Issue #109, opened 2014-11-24
rethinking the architecture


afeld:
For those not familiar with [the CAP project](https://18f.gsa.gov/dashboard/project/C2/), we are managing approval workflows for different groups in the executive branch, at the moment focusing on [purchase card](https://smartpay.gsa.gov/about-gsa-smartpay/tax-information/purchase-card) holders getting approvals from supervisors.

We have multiple clients needing approval workflows, but all are sending different information through different templates. One group, for example, is "forwarding" [GSA Advantage](https://www.gsaadvantage.gov/) carts (which copies the order information into the email), whereas another is (in the near term) going to be sending a work order as a spreadsheet. With the current architecture, the size of this codebase will grow linearly with the number of different groups (specifically, the number of types of workflows) we need to support, in terms of parsers, templates, etc.

We need to start splitting the generalized approval functionality from the client-specific functionality. More specifically, let's separate the email templates from the approval logic, so that C2 (or whatever we rename this central piece to) knows nothing about Navigator, carts, etc). The central piece would essentially be a state machine that

* Knows about approval groups/flows
* Sends notification emails
* Handles approvals/rejections
* Keeps an audit trail of activity

Options:

## API

Many of 18F's other projects are service/API-oriented, having a central server which our clients can communicate with through REST.

### Pros

* Improvements effective for everyone immediately
* Theoretically minimal setup burden for new clients
* Centralized place for reporting

### Cons

* We have to be more careful about breaking changes
* Single point of failure
* Handling of attachments more difficult

### Open questions

* How are the templates specified? We could:
    * Have an admin UI where they provide [Liquid](http://docs.shopify.com/themes/liquid-documentation/basics)/[Mustache](http://mustache.github.io/) templates, then have them POST the values to be filled in for each thing to be approved
    * Have them send the HTML to be included in the email in the POST

## Gem/[Engine](http://guides.rubyonrails.org/engines.html)

In this setup, each client would create a new Rails/Rack app that includes the common functionality as a gem.

### Pros

* They have full control over their app
* The information is stored in their database, meaning they can debug/query the data more easily
* The shared component only upgrades when they explicitly tell it to

### Cons

* Additional setup per client
* They have to manage their own infrastructure, including
    * Database
    * Email sending/handling
* The have to use Ruby
* They have to upgrade the shared component manually
* Engines/Railties aren't widely used...there may be a good reason

---

Questions? Thoughts? Alternative ideas?

afeld:
Closing as stale.

PR #127, opened 2014-12-15
[#89168592] create a Proposal model

afeld:
I have been thinking about [how to generalize](https://github.com/18F/C2/issues/109) the app, and short of actually splitting the application into multiple pieces, this PR splits a Proposal model out from Cart. The medium-term goal is to separate any GSA Advantage-specific logic from that dealing with the overall workflow and approvals. After this, I plan to create [a new API](https://www.pivotaltracker.com/story/show/77296838) for Proposals (not Carts), which can be used by Navigator, WHSC, etc.


app/models/proposal.rb
```diff
@@ -0,0 +1,13 @@
+class Proposal < ActiveRecord::Base
+  has_one :cart
```
cmc333333: Is there an equivalent that'd allow zero or one?

afeld: Yep, `has_one` does that, though probably worth noting that it won't always be present in an inline comment. Good call!

app/controllers/communicarts_controller.rb
```diff
@@ -24,7 +24,10 @@ def send_cart
   end
 
   def approval_reply_received
-    cart = Cart.where(external_id: params['cartNumber'].to_i).where(status: 'pending').first
+    cart = Cart.joins(:proposal).where(
+      external_id: params['cartNumber'].to_i,
```
cmc333333: Should the external id be on the proposal, eventually?

afeld: Not every Proposal will have one, but I could see it being useful for enough use cases that we should keep it in there. On the other hand, it might be more clear if it's on the use-case-specific tables, e.g. the `cl_number` (?) for NCR, or the `concur_voucher_id` if we were handling travel requests. Thoughts?

afeld:
Had a couple train rides worth of downtime – updated!

cmc333333:
Seems like a good step towards your goal. Let me know if I can help integrating this with the workflow piece.

db/migrate/20141214220050_create_proposals.rb
```diff
@@ -0,0 +1,37 @@
...
+      t.timestamps
+    end
+    add_reference :carts, :proposal
+
+    reversible do |dir|
+      Cart.reset_column_information
+
+      dir.up do
+        Cart.find_each do |cart|
+          cart.create_proposal!(
```
afeld: Note that it's keeping the same `id` on the carts, so none of the other references need to change.

afeld:
Phew, ok, updated. Feels like this was a lot of work for what it was, but I am still _fairly_ confident about the direction of having a Proposal model and then distinct models for each use case. Someone mind taking another look-see?

cmc333333:
:+1: from me -- I'd like to see this merged soon so it doesn't drift too far from master. I don't feel confident enough to merge myself, though.

@phirefly - when you get a chance, review and merge?

phirefly:
We'll probably need a backfill script. Any in the works?

db/migrate/20141214220050_create_proposals.rb
```diff
@@ -0,0 +1,37 @@
+class CreateProposals < ActiveRecord::Migration
+  def change
+    create_table :proposals do |t|
+      t.string :status
+      t.string :flow
+      t.timestamps
+    end
+    add_reference :carts, :proposal
```
phirefly: Will NCR be of type Cart? Would it be worthwhile to name that appropriately at this point? I'm also happy with capturing that in a separate story.

db/migrate/20141214220050_create_proposals.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    reversible do |dir|
```
phirefly: Scratch my `backfill` comment. Here it is. ;)

db/migrate/20141214220050_create_proposals.rb
afeld: For the moment, yes. My plan was to deal with splitting the Cart into use-case-specific models separately.

afeld:
Updated! Not sure if I succeeded in *simplifying* anything though :confused: To minimize the number of changes, I left all the state-related calls to the Cart (`approve!`, `rejected?`, etc.), but this required jumping through some hoops to get the delegation to work. Moving the Approval-related stuff (the relationships and the methods) to the Proposal later will help.

phirefly:
:shipit:

PR #131, opened 2014-12-17
81308238 portal page

sashax:


sashax:
OK, @afeld changes made. Plus extra page! Also more tests.

app/controllers/application_controller.rb
```diff
@@ -5,7 +5,8 @@ class ApplicationController < ActionController::Base
   private
 
   def current_user
-    User.find_by(email_address: session[:user]['email']) if session[:user].present?
+    # User.find_by(email_address: session[:user]['email']) if session[:user].present?
+    @current_user ||= User.find_or_create_by(email_address: session[:user]['email']) if session[:user] && session[:user]['email']
```
afeld: Whoa, we weren't doing this before?? Derp.

app/controllers/carts_controller.rb
```diff
@@ -7,4 +8,24 @@ def show
     @show_comments = true
   end
 
+  def index
+    role = 'requester'
+    my_carts = Cart.joins(:approvals).where(:approvals => {:role => role, :user_id => @current_user[:id]})
+    if my_carts.empty?
+      role = 'approver'
+      my_carts = Cart.joins(:approvals).where(:approvals => {:role => role, :user_id => @current_user[:id]})
+    end
```
afeld: I think this could all be simplified to

```ruby
@current_user.carts
```

app/controllers/carts_controller.rb
```diff
@@ -7,4 +8,24 @@ def show
...
+    @closed_carts = my_carts.closed
+    @open_carts = my_carts.open
+    render :template => "carts/index", :locals => {:open_carts => @open_carts,
+                                                   :closed_carts => @closed_carts,
+                                                   :user => @current_user, :role => role,
+                                                   limit: CLOSED_CART_LIMIT}
```
afeld: The idiomatic Rails thing would be to leave the `render` implicit and passing all of the models via instance variables. If you prefer setting the local variables explicitly, I would make them regular variables within the controller action.

app/controllers/application_controller.rb
```diff
@@ -5,7 +5,8 @@ class ApplicationController < ActionController::Base
   private
 
   def current_user
-    User.find_by(email_address: session[:user]['email']) if session[:user].present?
+    # User.find_by(email_address: session[:user]['email']) if session[:user].present?
```
afeld: Mind removing this?

app/controllers/carts_controller.rb
```diff
@@ -7,4 +8,24 @@ def show
...
+  end
+
+  def archive
+    role = params[:role] || 'requester'
+    @closed_cart_full_list = Cart.joins(:approvals).where(:approvals => {:role => role, :user_id => @current_user[:id]}).closed
```
afeld: Similar to above:

```ruby
@carts = current_user.carts.closed
```

Do you specifically want the ability to filter by role?

app/controllers/carts_controller.rb
```diff
@@ -7,4 +8,24 @@ def show
...
+    render :template => "carts/archive", :locals => {:closed_carts => @closed_cart_full_list, :user => @current_user, :role => role}
```
afeld: Same as above: use implicit rendering, or don't use instance variables.

app/helpers/home_helper.rb
```diff
@@ -0,0 +1,5 @@
+module HomeHelper
+  def signed_in?
+    session[:user] && !session[:user].empty?
+  end
+end
```
afeld: Not necessary, since it's a [`helper_method`](http://api.rubyonrails.org/classes/AbstractController/Helpers/ClassMethods.html#method-i-helper_method) from the `ApplicationController`.

app/views/home/index.html.erb
```diff
@@ -1,32 +1,29 @@
...
   <div class="home-main-body">
+
       <div class="row">
         <div class="col-sm-12">
           <h2 id="project_tagline">&quot;P-Card holders deserve easy, modern, automated management of their shopping carts.&quot;</h2>
-        </div>
-      </div>
-      <div class="row">
-        <div class="col-sm-12">
+          <p></p>
```
afeld: Can we remove this?

app/views/shared/_cart_list.html.erb
```diff
@@ -0,0 +1,27 @@
...
+<%#-- "limit" is the maximum noumber of carts to show. --%>
+<table class="cart-list">
+  <tr class="header">
+    <th class="first" scope="col"><h5>Request</h5></th>
+    <th class="sixth" scope="col"><h5>Total Price</h5></th>
+    <th class="sixth" scope="col"><h5><%= if role=='requester' then 'To' else 'From' end%> </h5></th>
+    <th class="sixth" scope="col"><h5>Date</h5></th>
+  </tr>
+  <%- carts.each_with_index do |cart, index| %>
+    <%- if defined?(limit) && index >= limit then break end %>
```
afeld: Can we use [`limit`](http://guides.rubyonrails.org/active_record_querying.html#limit-and-offset) on the query instead?

app/views/shared/_cart_list.html.erb
```diff
@@ -0,0 +1,27 @@
...
+    <tr>
+      <td class="first"><a href="/carts/<%= cart.id %>"><%= cart.name %></a></td>
+      <td class="sixth"><%= number_to_currency(cart.decorate.total_price) %></td>
+      <td class="sixth">
+        <%- cart.approvals.each do |approval| %>
+          <%- if approval.role != role && (closed || approval.status == "pending") %>
+            <p><%= User.find(approval.user_id).email_address %></p>
```
afeld: This can be `approval.user.email_address`.

app/views/shared/_cart_list.html.erb
```diff
@@ -0,0 +1,27 @@
...
+    <th class="sixth" scope="col"><h5>Date</h5></th>
+  </tr>
+  <%- carts.each_with_index do |cart, index| %>
+    <%- if defined?(limit) && index >= limit then break end %>
+    <tr>
+      <td class="first"><a href="/carts/<%= cart.id %>"><%= cart.name %></a></td>
+      <td class="sixth"><%= number_to_currency(cart.decorate.total_price) %></td>
+      <td class="sixth">
+        <%- cart.approvals.each do |approval| %>
+          <%- if approval.role != role && (closed || approval.status == "pending") %>
```
afeld: Not quite sure what this is checking... maybe make it a helper method?

spec/controllers/carts_controller_spec.rb
```diff
@@ -0,0 +1,68 @@
+describe CartsController do
+  let(:user1) { FactoryGirl.create(:user, email_address: 'user1@some-dot-gov.gov') }
+  let(:user2) { FactoryGirl.create(:user, email_address: 'user2@some-dot-gov.gov') }
+  let(:user3) { FactoryGirl.create(:user, email_address: 'user3@some-dot-gov.gov') }
+  let(:user4) { FactoryGirl.create(:user, email_address: 'user4@some-dot-gov.gov') }
+  let(:approval_group1) { FactoryGirl.create(:approval_group, name: 'test-approval-group1') }
+  let(:approval_group2) { FactoryGirl.create(:approval_group, name: 'test-approval-group2') }
+  @cart1
```
afeld: Leftover?

spec/controllers/carts_controller_spec.rb
```diff
@@ -0,0 +1,68 @@
+describe CartsController do
+  let(:user1) { FactoryGirl.create(:user, email_address: 'user1@some-dot-gov.gov') }
+  let(:user2) { FactoryGirl.create(:user, email_address: 'user2@some-dot-gov.gov') }
+  let(:user3) { FactoryGirl.create(:user, email_address: 'user3@some-dot-gov.gov') }
+  let(:user4) { FactoryGirl.create(:user, email_address: 'user4@some-dot-gov.gov') }
```
afeld: No need to pass in emails here.

spec/controllers/carts_controller_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  let(:approval_group1) { FactoryGirl.create(:approval_group, name: 'test-approval-group1') }
+  let(:approval_group2) { FactoryGirl.create(:approval_group, name: 'test-approval-group2') }
+  @cart1
+
+
+  before do
+    UserRole.create!(user_id: user1.id, approval_group_id: approval_group1.id, role: 'approver')
+    UserRole.create!(user_id: user2.id, approval_group_id: approval_group1.id, role: 'requester')
+    UserRole.create!(user_id: user3.id, approval_group_id: approval_group1.id, role: 'approver')
+    #why do I have to do it this way? I dunno, but I do. :/
```
afeld: Could you elaborate?

spec/controllers/carts_controller_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    p = {}
+    p['approvalGroup'] =  'test-approval-group1'
+    p['cartName'] = 'cart1'
+    @cart1 = Commands::Approval::InitiateCartApproval.new.perform(p)
+    session[:user] = {}
```
afeld: What are you trying to do in these five lines?

app/controllers/carts_controller.rb
sashax: Yes.

app/views/home/index.html.erb
sashax: not if we want it to layout as specified

afeld: Can we use a `margin-bottom` on the `h2` or something rather than an empty tag?

sashax: Meh, It looks ok without it. I just took it out.

app/controllers/carts_controller.rb
```diff
@@ -7,4 +8,19 @@ def show
     @show_comments = true
   end
 
+  def index
+    @role = 'requester'
+    my_carts = Cart.joins(:approvals).where(:approvals => {:role => @role, :user_id => @current_user[:id]})
+    if my_carts.empty?
+      @role = 'approver'
+      my_carts = Cart.joins(:approvals).where(:approvals => {:role => @role, :user_id => @current_user[:id]})
+    end
```
afeld: This seems over-complicated code and UX-wise... can we just do this?

```ruby
my_carts = current_user.carts.order('created_at DESC')
```

We can [do a fancier query](http://stackoverflow.com/questions/6332043/sql-order-by-multiple-values-in-specific-order) if we want to show carts where the user is a requester before those where they're an approver, but

* I don't think there will be a lot of overlap between requesters and approvers, at least for WHSC
* My instinct is that the user would want to see all carts they're involved with in one list anyway

sashax: Well, I agree that there's not likely to be overlap. There are (minor)
differences between how carts are represented depending on whether one is
an approver or a requester.
 We could simplify the query by cheating and just picking the role of the
first cart.
I believe that it's useful to distinguish between open and closed carts.

app/controllers/carts_controller.rb
```diff
@@ -7,4 +8,19 @@ def show
...
+    @closed_carts = my_carts.closed
+    @open_carts = my_carts.open
+  end
+
+  def archive
+    @role = params[:role] || 'requester'
+    @closed_cart_full_list = Cart.joins(:approvals).where(:approvals => {:role => @role, :user_id => @current_user[:id]}).closed
```
afeld: This can be simplified to

```ruby
@closed_cart_full_list = current_user.carts.where(approvals: {role: @role}).closed
```

sashax: oooh, nice!

app/controllers/carts_controller.rb
afeld: Ohhh I see the issue.. the `@role` wouldn't be set. I think the logic makes more sense on a cart-by-cart basis:

```
for each cart
  show the cart summary
  if user is an approver on the cart
    show action buttons
```

rather than one role for the overall page. If you agree, feel free to add an inline `TODO` and we can worry about it later. Otherwise, no worries.

sashax: Sure, I'm fine with that. I'll add the comment.

afeld:
Cool, just the one note above about limiting the number of carts displayed via the query. Otherwise, :+1:

sashax:
Is there a straightforward way to limit the number of records in the query
and also know if there are more records than the limit? We have been using
if (count > limit) to control the visibility of the View the Archive link.
Unless I'm missing something, that behavior will be awkward to replicate in
this case.

afeld:
What about something like this? (simplified example)

```erb
<% carts.limit(LIMIT).each do |cart| %> 
  ...
<% end %>

<% if carts.size > LIMIT %>
  <%= link_to "View the archive" %>
<% end %>
```

sashax:
oh, yeah, right., I was limiting it in the Wrong Place. Change submitted,
although I think it ends up being only a marginal improvement. I did catch
a typo, though!

PR #139, opened 2014-12-23
do all configuration through environment variables

afeld:
Per @dlapiduz's comment in https://github.com/18F/C2/pull/137#issuecomment-67983337, this PR includes a handful of things:

* Everything that needs to be configurable can be done so through environment variables
* Better naming of environment variables
* Switch to supporting a `.env` file instead of `config/environment_variables.yml` and `config/feature_flags.yml`
* Document optional environment variables
* Remove unused environment variables

TODOs before merging/deploying:

* [x] Corresponding changes in cloud-cutter
* [ ] Update configs in S3
* [ ] Verify that assets work in emails


monfresh:
I came across this while browsing on govcode.org, and wanted to let @afeld know about the excellent [Figaro](https://github.com/laserlemon/figaro) gem, which is my preferred way of managing config vars, and it already comes with a way to define required keys.

afeld:
> I came across this while browsing on govcode.org

Helping 18F Find 18F Since 2014 :wink: 

> the excellent Figaro gem, which is my preferred way of managing config vars

Looked into it a *bit* at the time, but probably worth seeing what it can simplify. Thanks!

https://www.pivotaltracker.com/story/show/87317434

PR #149, opened 2014-12-31
add Capistrano

afeld:
Currently a simple `uptime` script as a proof-of-concept, but I want to add more of the things listed in https://github.com/18F/C2/pull/148#issue-53166287.

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -1,5 +1,6 @@
 # Load DSL and set up stages
 require 'capistrano/setup'
+require_relative 'lib/capistrano/server_with_nat'
 
 # Include tasks from other gems included in your Gemfile
 #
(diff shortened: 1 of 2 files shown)


dlapiduz:
@afeld I am a bit late but can you tell me a bit of what sort of tasks you need to do with ssh?

afeld:
Commands I was planning to add, in order of developer usefulness:

1. Open a Rails console
1. Pull (a scrubbed copy of?) the production database to the local machine
1. Open an bash console (for troubleshooting asset compilations, etc.)

Read-only for all of these things is totally fine. Happy to consider alternate approaches!

phirefly:
Can we invoke this from the cap script? https://www.pivotaltracker.com/story/show/83707776

afeld:
Totes :smiley: Or at the very least 2. would allow us to do it locally.

PR #150, opened 2015-01-02
81528646 whsc web approval status

phirefly:
Added approval status into web layout view


app/views/shared/_linear_approval_status.html.erb
```diff
@@ -0,0 +1,46 @@
...
+                <%- last_status = approval.status %>
+                <%- end %>
+              </table>
+            </td>
+          </tr>
+        </table>
+      </td>
+    </tr>
+  </table>
+</div>
```
afeld: Can we share the existing [`_email_status`](https://github.com/18F/C2/blob/ce06e1558bafd0a4da2483a546692385677c1208/app/views/communicart_mailer/_email_status.html.erb) template?

app/views/shared/_approval_status.html.erb
```diff
@@ -0,0 +1,9 @@
+<%- if @cart.flow == 'linear' %>
+  <%= render partial: "shared/linear_approval_status" %>
+<%- else %>
+  <div class="col-md-4  col-xs-12 communicart_description">
+    <div class="purchase-status">
+      <h5 class="<%= @cart.status %>"><%= @cart.display_status %></h5>
+    </div>
+  </div>
+<%- end %>
```
afeld: Created a story with an idea to avoid this split: https://www.pivotaltracker.com/story/show/85405198

app/views/shared/_linear_approval_status.html.erb
phirefly: The email_status template used to be shared but we split them up into web and email versions. The rationale was that we can refactor the web version to a table-free version that would more easily work with responsive markup. @sashax, additional thoughts?

sashax:
I'm confused about whether the proposed file: _linear_approval_status.html.erb is supposed to be an email or web partial. It *looks* as if the partial's being used in a web view, in which case we may have problems with responsiveness, since it's using tables. 

I *believe* @afeld is advocating we combine linear and parallel approval views, not (necessarily) that we combine mail and web templates. I have no real opinion on the former, other than it may be premature optimization (since as I understand it we have no actual parallel-approval-process-using customers), but I believe we probably have to continue to provide both email and responsive layouts for each partial (unless we make the decision to punt on responsiveness, of course). This said, it may be that we can get away with using tables for layout in this specific case.

phirefly:
Yes, _linear_approval_status.html.erb is a web partial. My understanding was that we were keeping mailer partials under `views/communicart_mailer`. Correct me if I'm wrong. 

Not sure if I'm misunderstanding your statement about not having a parallel-approval-process-using customer, but the Navigator team is using parallel approvals.

A couple of things I would be in favor of:
* Combining linear and parallel views into one but having separate views for the web and email markup. I am currently splitting the linear/parallel views within the web layout but am happy to merge the two.
* Keeping tables in the web views (approval_status and linear_approval_status in app/views/shared) for this story but creating one specifically focused on responsive layouts for the email status. That said, I'm happy to just piggyback that effort onto this, too. :)

Additional thoughts?

afeld:
I don't feel strongly about whether to share the mailer+web templates or not. The styles are shared at this point, right? In other words, the templates would need to stay in sync anyway. If this is the case, I say we keep them shared (for now), but it would probably make sense to split them in the long run.

> advocating we combine linear and parallel approval views... may be premature optimization

:+1: A separate task, at least.

phirefly:
Updated with suggestions. Let me know if any stories need to be added (I think we go them all) or any other feedback. Otherwise. I'm good to merge.

PR #151, opened 2015-01-05
get rid of views/shared/

afeld:
In response to https://github.com/18F/C2/pull/150, it was a bit confusing what the partials were being "shared" by, so moved them into more appropriate directories.


phirefly:
:+1: 
Will update https://github.com/18F/C2/pull/150 accordingly

PR #154, opened 2015-01-05
[#85360084] add fake budget approver(s) to WHSC requests

afeld:


phirefly:
To your earlier point, should the email be in the repo? Can we pull the emails out into a client config of sorts?

afeld:
Refactored, and made the emails optionally configurable – defaults should work across our [deploys](http://12factor.net/codebase), but are overridable. Does that work for you, or would you prefer that they weren't in the repo at all?

phirefly:
:ship: it

PR #158, opened 2015-01-07
Conditionally display the portal link

phirefly:
While testing, I realized the 'Back to main portal' link is showing up on the homepage as well as on the very page it links to. Added a quick way to handle this.

### DIFF
--- a/app/views/layouts/_communicart_header.html.erb
+++ b/app/views/layouts/_communicart_header.html.erb
@@ -32,6 +32,8 @@
   </div>
 </div>
 
-<ul id='breadcrumb-nav'>
-  <li><%= link_to "Back to main portal", carts_path %></li>
-</ul>
+<%- unless excluded_portal_link %>
+  <ul id='breadcrumb-nav'>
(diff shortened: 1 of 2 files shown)


app/views/layouts/_communicart_header.html.erb
```diff
@@ -32,6 +32,8 @@
   </div>
 </div>
 
-<ul id='breadcrumb-nav'>
-  <li><%= link_to "Back to main portal", carts_path %></li>
-</ul>
+<%- unless excluded_portal_link %>
```
afeld: How about `unless current_path == carts_path`?

phirefly: :+1:

phirefly: I still want to exclude the home page so would need to keep this within the excluded_portal_link method

phirefly: `current_path` may have been deprecated. I found `current_page?(carts_path)` [here](http://apidock.com/rails/ActionView/Helpers/UrlHelper/current_page%3F)

PR #161, opened 2015-01-13
Reject button

sashax:


app/helpers/communicart_mailer_helper.rb
```diff
@@ -14,12 +14,12 @@ def generate_bookend_class(index, count)
     end
   end
 
-  def approve_all_url(approval)
+  def approve_all_url(approval, approve = true)
```
afeld: Can we make a `reject_cart_url` method instead?

afeld:
Just the one comment – otherwise :shipit:!

Tangential question, but is there an ability to leave a comment after approval/rejection? If not, we should support that through the web flow too.

phirefly:
Yep, comments can be created through the web on the carts show page.

app/views/communicart_mailer/cart_notification_email.html.erb
```diff
@@ -31,7 +31,9 @@
     <tr id="approval-actions">
       <td colspan=2 align="center">
         <p>
-          <%= link_to "Approve all items with one click", approve_all_url(@approval), class: 'form-button' %>
+          <%= link_to "Approve all items with one click", approve_all_url(@approval, true), class: 'form-button' %>
+          <%= link_to "Reject all items with one click", approve_all_url(@approval, false), class: 'form-button' %>
```
phirefly: I would be in favor of these two updates:
* de-emphasizing the rejection action a bit by having a normal link. The easy way to do this may be to just remove the class, so that it isn't a button. Sorry, I did add the comment to make this a link in story but it mentions a button in the description.

* Just updating the text to say 'Reject all items', again to simplify it.

Outside of these two comments, :shipit: 
Thanks @sashax!

afeld: Could the buttons just say "approve" and "reject"? Pretty sure we don't have item-level rejection at this point anyway.

app/helpers/communicart_mailer_helper.rb
sashax: I had it that way originally. Problem is that we end up with two methods with 75% identical code, or another layer of helper indirection. I don't feel immensely strongly about it, but I'm not sure either is an improvement over what's there.

app/views/communicart_mailer/cart_notification_email.html.erb
sashax: I will include both y'alls suggestions.

app/helpers/communicart_mailer_helper.rb
afeld: How about `def approval_action_url(approval, action = 'approve')`?

app/helpers/communicart_mailer_helper.rb
```diff
@@ -14,12 +14,12 @@ def generate_bookend_class(index, count)
     end
   end
 
-  def approve_all_url(approval)
+  def approve_all_url(approval, action = 'approve')
```
afeld: Last thing! Can we update the method name, since it's not always for approvals?

PR #166, opened 2015-01-14
Sort carts by date so most recentest is on top.

sashax:


sashax:
https://www.pivotaltracker.com/story/show/86073814

PR #168, opened 2015-01-14
85857380 fix total amount display

phirefly:
• Updated #total_price to consider a cart's `origin` property
• Remove custom currency helpers in favor of Rails `number_to_currency`

### DIFF
--- a/app/decorators/cart_decorator.rb
+++ b/app/decorators/cart_decorator.rb
@@ -2,8 +2,13 @@ class CartDecorator < Draper::Decorator
   delegate_all
 
   def total_price
-    price = object.cart_items.reduce(0) do |sum,citem| sum + citem.quantity * citem.price end
-    Float("%0.02f" % price)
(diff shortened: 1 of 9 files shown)


app/decorators/cart_decorator.rb
```diff
@@ -2,8 +2,11 @@ class CartDecorator < Draper::Decorator
   delegate_all
 
   def total_price
-    price = object.cart_items.reduce(0) do |sum,citem| sum + citem.quantity * citem.price end
-    Float("%0.02f" % price)
+    if self.getProp('origin') == 'ncr'
+      self.getProp('amount').to_f
+    else
+      object.cart_items.reduce(0) { |sum,citem| sum + citem.quantity * citem.price }
+    end
```
afeld: This is a case where having separate decorator classes would be useful for the different origins, but we can tackle that in the Proposal refactoring.

PR #171, opened 2015-01-26
86223830 fix placeholder approvals

phirefly:
Fixed bug that displays placeholder text instead of actual approval information

### DIFF
--- a/app/views/communicarts/approval_response.html.haml
+++ b/app/views/communicarts/approval_response.html.haml
@@ -1,7 +1,0 @@
-!!!
-%head
-  %title= "Communicart: Approval"
-%html
-
-  %body.communicart-notification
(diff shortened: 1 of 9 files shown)


app/views/communicarts/approval_response.html.haml
```diff
@@ -1,7 +0,0 @@
-!!!
-%head
-  %title= "Communicart: Approval"
-%html
-
-  %body.communicart-notification
-= render partial: "carts/cart", locals: {cart: @cart, title: "Cart #{@cart.id}"}
```
afeld: Are we trying to move away from HAML?

phirefly: I prefer haml, but the general consensus from other has been to kill it. I went ahead and converted since I was in the template.

PR #174, opened 2015-01-27
fix .css.scss extension deprecation

afeld:
https://github.com/rails/sass-rails/pull/291


afeld:
Ran this:

```bash
# http://stackoverflow.com/a/5394379/358804
for file in **/*.css.scss
do
  git mv "$file" "${file/css./}"
done
```

phirefly:
:zap:

PR #176, opened 2015-01-28
display cart properties on show page

afeld:
Thinking they should probably have less spacing, but was fighting with some conflicting styles a bit, so didn't want to spend much time on it before getting a second opition. @ericadeahl, thoughts on a quick fix for that middle section?

c2

### DIFF
--- a/app/assets/stylesheets/common/cart_properties.scss
+++ b/app/assets/stylesheets/common/cart_properties.scss
@@ -1,5 +1,18 @@
+@import "bootstrap/variables";
+
 .cart_properties {
+  $sideMargin: $grid-gutter-width / 2;
+
+  td:first-child {
+    width: 25%;
+  }
+
   .header td {
-    padding: 10px;
+    padding: $sideMargin;
+  }
+
+  // force specificity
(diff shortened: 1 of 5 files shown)


ericadeahl:
@afeld Middle section looks ok to me. I would just:
- reduce the spacing so each row is about half the height it is now
- left align the labels (left column of text) with the rules
- move the right column text over closer to the labels (so the labels column is 1/4 of the width, right column is 3/4 of the width) 

Is that what you need? I'm lacking some context info :)

afeld:
@ericadeahl Exactly what I was looking for, thanks!

@phirefly Hold off on reviewing/merging this... going to split up into more focused PRs and do the layout changes suggested above.

afeld:
Updated!

screen shot 2015-01-29 at 1 45 51 am

phirefly:
@afeld: Does this PR include the desired splitting up of commits or should we still hold off on this?

afeld:
Yep! Rebased and ready for review.

phirefly:
Beauteous. :shipit:

PR #178, opened 2015-01-28
split up cart templates

afeld:
Just a refactoring.


phirefly:
:+1: Increased beauteousness

PR #179, opened 2015-01-28
start to organize stylesheet files by corresponding view

afeld:
A humble proposal:

I'm starting to feel the pain of managing a decent number of styles in web+email contexts. This is a proposal to add some convention around CSS class names and corresponding files. Thoughts?

@unendin was the master of this... do you remember the convention that we used at Jux?


app/assets/stylesheets/view_specific/README.md
```diff
@@ -0,0 +1,11 @@
...
+For example:
+
+```scss
+// carts/cart_items.scss
+// corresponds to
+// carts/_cart_items.html.erb
+.view-carts-cart_items {
+  // ...
+}
+```
```
afeld: Down the road, I could see having a gem to enforce this.

app/assets/stylesheets/common/communicarts.scss
```diff
@@ -1,6 +1,6 @@
 @import "constants";
 @import "webflow";
-@import "../communicart_mailer/email_status";
+@import "../view_specific/shared/email_status";
```
msecret: The name "view_specific" feels a little awkward to me. Whats the reason you don't want to just call it "view"?

app/assets/stylesheets/view_specific/shared/email_status.scss
```diff
@@ -1,4 +1,4 @@
-.email-status {
+.view-shared-email_status {
```
msecret: My concern here is that this will end up as a long class name. I like the idea of a single letter prefix. Some options:

- v-email_status - this would be dropping the shared prefix, which might be ok.
- vs-email_status 
- v-s-email_status

app/assets/stylesheets/common/communicarts.scss
afeld: This felt more descriptive, but I don't feel super strongly about it.

app/assets/stylesheets/view_specific/shared/email_status.scss
afeld: Changing the `view-` prefix to `v-` sounds good, but I want to keep the path in there to avoid naming conflicts (e.g. `v-index`).

afeld: ...and to make the corresponding views easier to find.

afeld:
Updated!

phirefly:
:ship: it

PR #183, opened 2015-01-29
approval_response fixes

afeld:
See commits.


app/views/communicarts/approval_response.html.erb
```diff
@@ -1,11 +1 @@
-<!DOCTYPE html>
```
phirefly: Doh! Good catch!

phirefly:
:shipit:

PR #186, opened 2015-01-29
[#86662010] display the relevant name in the email From field

afeld:
For new carts, it's the requester, for new comments, it's the commenter, etc.

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -1,4 +1,5 @@
 class CommentsController < ApplicationController
+  before_filter :authenticate_user!
 
   def index
     @commentable = find_commentable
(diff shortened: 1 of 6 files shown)


app/controllers/comments_controller.rb
```diff
@@ -1,4 +1,5 @@
 class CommentsController < ApplicationController
+  before_filter :authenticate_user!
```
afeld: Is there a reason we *wouldn't* want to do this?

phirefly:
Good catch! Thankfully, this is just a show page.

phirefly:
:rocket:

afeld:
Actually, it included the creation of comments, too. I've been meaning to run [Brakeman](http://brakemanscanner.org/) on our app to find more things like this.

PR #188, opened 2015-01-29
[#86662046] display cart name on show page

afeld:
Debating about which layout I like best:

a)

screen shot 2015-01-29 at 4 09 36 pm

b)

screen shot 2015-01-29 at 4 07 04 pm

c) (in this commit)

screen shot 2015-01-29 at 4 06 34 pm

d)

screen shot 2015-01-29 at 4 16 36 pm

I don't have a sense of how long these descriptions will be on average, which I think would help us decide. The ones I saw were pretty short (not even a full sentence). @ericadeahl Thoughts?

### DIFF
--- a/app/views/carts/_cart.html.erb
+++ b/app/views/carts/_cart.html.erb
@@ -1,12 +1,15 @@
 <div class="inset">
   <div class="row">
 
-    <div class="col-md-8 col-xs-12">
+    <div class="col-md-12 col-xs-12">
       <h1 class="communicart_header">
-        Purchase Request <%= cart.id %>
+        <%= cart.name %>
       </h1>
 
       <div class="communicart_description">
+        <p>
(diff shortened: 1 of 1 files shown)


phirefly:
C seems reasonable from a placement standpoint. The integration of the number into the description in b) seems to overemphasize it. In d) it causes its placement to vary based on the description size.

@ericadeahl, along these lines, we will need to do a design audit to do fix existing designs as well as visit some design issues we have encountered during the buildout of the templates -- mainly due to the differences between the 'parallel' and 'linear' flows. I will be discussing with Russ about resource, so it may not be you, but just wanted to give you the heads up. :)

afeld:
In that case, merging now since it fixes the issue, but happy to do another to address any design feedback. Thanks!

PR #189, opened 2015-01-29
[#86660976] Removed rejection reply tip

phirefly:


afeld:
Do we still want to leave this bit in at the top?

```
------------------ 
Please reply to this email with the response 'APPROVE' or 'REJECT' on its own line. 
Additional comments:
```

If yes, :shipit:

aminedak15:
lf yes amine yo ablo arbic

aminedak15:
Do we still want to leave this bit in at the top?

phirefly:
Yep, we'll keep this for now. Thanks!

PR #190, opened 2015-01-30
display detailed status on cart index page

afeld:
In [a previous commit](https://github.com/18F/C2/commit/ce8d72294607d27eb72bf1457f22c7bb9dd753a6#commitcomment-9500132), @phirefly mentioned that we didn't have a spec for the helper method that gathered the email addresses to be displayed for each cart on the index page. Instead of verifying the existing functionality, this PR:

* Displays all carts the current_user is involved with, whether they are an approver, a requester, or an observer
* Lists the approver(s) who are being waited on for the cart to progress

This screenshot shows all the different states: 

screen shot 2015-01-30 at 2 04 48 am

### DIFF
--- a/app/controllers/carts_controller.rb
+++ b/app/controllers/carts_controller.rb
@@ -12,7 +12,7 @@ class CartsController < ApplicationController
     # TODO: handle role on a cart-by-cart basis
     # (including action buttons in approver/open case)
     @role = requester_or_approver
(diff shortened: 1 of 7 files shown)


phirefly:
:shipit: 
Our next round of design reviews/updates may tweak this, but very nice improvements nonetheless. Lots of nice new things to put in front of the users. Thanks for the spec!

PR #191, opened 2015-01-30
Allow DB config during development

cmc333333:
Also updates the README.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -2,13 +2,12 @@ default: &default
   adapter: postgresql
   encoding: unicode
   database: <%= ENV['DB_NAME'] || "c2_#{Rails.env}" %>
-  pool: 5
-
-development: *default
-test: *default
-production:
-  <<: *default
   username: <%= ENV['DB_USERNAME'] %>
   password: <%= ENV['DB_PASSWORD'] %>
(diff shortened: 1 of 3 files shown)


config/database.yml
```diff
@@ -2,13 +2,12 @@ default: &default
...
-  pool: 5
-
-development: *default
-test: *default
-production:
-  <<: *default
   username: <%= ENV['DB_USERNAME'] %>
   password: <%= ENV['DB_PASSWORD'] %>
   host: <%= ENV['DB_HOST'] %>
   port: <%= ENV['DB_PORT'] %>
```
afeld: If the keys are set, do we have to pass explicit defaults? For example, do we need something like this?

```erb
<%= ENV['DB_USERNAME'] || root %>
```

phirefly: I would prefer something like that

cmc333333: An empty host and port definitely equate to the system defaults for postgres (localhost and.... 5321 -- something like that). I assume the username and password would be the same -- i.e. setting a blank `username` is equivalent to not setting one.

My setup is different than yours, would you mind confirming that a default is needed?

afeld: Works On My Machine™

PR #195, opened 2015-02-02
Add "offices" field to NCR form

cmc333333:
* The list of offices currently only includes one dummy example - in the future, it'd make sense for this to be dynamic
* Also adds `letter_opener` and `letter_opener_web` to view emails which *would* be sent during development

screen shot 2015-02-02 at 5 04 22 pm
screen shot 2015-02-02 at 5 12 57 pm


app/views/ncr/proposals/new.html.erb
```diff
@@ -32,6 +32,15 @@
       <% end %>
     </div>
     <div class="form-group">
+      <%= f.label :office %>
+      <%= f.select(:office, [], {}, {class: 'form-control', multiple: false}) do %>
+        <%= content_tag(:option, '', value: '') %>
+        <% Ncr::ProposalForm::OFFICES.each do |office| %>
+          <%= content_tag(:option, office, value: office) %>
+        <% end %>
+      <% end %>
```
afeld: I *think* this could all be simplified to

```erb
<%= f.collection_select :office, Ncr::ProposalForm::OFFICES, :to_s, :to_s, {include_blank: true}, {class: 'form-control'} %>
```

afeld:
Looks good, though I'm somewhat hesitant to merge it without real options in there. Maybe worth feature flagging for now?

phirefly:
I will ping Carol at the budget office for the list and CC you guys on it. Maybe we hold off for today but feature flag if we don't hear anything by tomorrow?

cmc333333:
Updated to include the true office data, use of JQuery `chosen` and the template rewrite @afeld recommended. The number of entries clutter the model quite a bit; more impetuous to put these in the DB.

afeld:
How about moving them to `config/data/ncr_offices.yml`?

cmc333333:
I like it

cmc333333:
Updated to use YAML for the options. What do you all think?

afeld:
Much better, thanks!

PR #196, opened 2015-02-03
[#87250330] sign the user in via token

afeld:
When the approver clicks approve/reject from the notification email, 1019568 now logs that user in to the system via the token. With 3552747, they now get redirected to the normal cart show page, instead of rendering a different view at a different URL.

Checked w/ @cmc333333 and @NoahKunin [in #cybersec](https://18f.slack.com/archives/cybersec/p1422567486000148) to verify that this is ok, security-wise.

### DIFF
--- a/spec/controllers/communicarts_controller_spec.rb
+++ b/spec/controllers/communicarts_controller_spec.rb
@@ -74,9 +74,7 @@ describe CommunicartsController do
       end
     end
 
-    context 'no approval_group is indicated' do
-      #TODO: Write specs
-    end
+    skip 'no approval_group is indicated'
 
     context 'nonexistent approval_group is specified' do
       it 'should return 400 error' do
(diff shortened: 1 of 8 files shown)


spec/controllers/communicarts_controller_spec.rb
```diff
@@ -265,21 +263,23 @@
...
       end
 
       it 'successfully validates the user_id and cart_id with the token' do
-        approver
-        expect_any_instance_of(Approval).to receive(:update_attributes)
         expect { put 'approval_response', @json_approval_params_with_token }.not_to raise_error
       end
 
+      it "signs the user in" do
+        put 'approval_response', @json_approval_params_with_token
```
cmc333333: debugging code

spec/controllers/communicarts_controller_spec.rb
```diff
@@ -303,7 +303,17 @@
         expect { put 'approval_response', @json_approval_params_with_token }.to raise_error(AuthenticationError)
       end
 
-      skip 'marks a token as used'
+      it 'marks a token as used' do
+        expect(ApiToken).to receive(:find_by).with(access_token: "5a4b3c2d1ee1d2c3b4a5").and_return(token)
+        expect_any_instance_of(Approval).to receive(:update_attributes)
+        approver
+
+        put 'approval_response', @json_approval_params_with_token
```
cmc333333: debugging code

spec/controllers/communicarts_controller_spec.rb
afeld: Ha, no, `put` as in "HTTP PUT".

http://api.rubyonrails.org/classes/ActionController/TestCase/Behavior.html#method-i-put

cmc333333: My bad!

spec/controllers/communicarts_controller_spec.rb
```diff
@@ -303,7 +303,17 @@
...
+
+        expect(response).to redirect_to(cart_path(cart))
+        token.reload
+        expect(token.used_at).to_not eq(nil)
```
cmc333333: Might be worth extending this just a bit further to show that if a user tries to log in with this again, they cannot. I think that's implied by the token, but I'm not sure.

Related question - if a user finds the same email again, and they click one of the links, can they still get to the cart (after logging in)?

cmc333333:
I don't understand all the ruby/rails-isms but the token-based login makes sense to me

phirefly:
Would love to see a browser spec

phirefly:
Great! Just the one comment. Then :shipit:

PR #197, opened 2015-02-03
86055782 ncr send email after request

phirefly:
• Sends an email to the requester when submitting an NCR proposal. 
• Allows NCR-specific logic for sending an approval notice to the requester.
• Sends an email to the requester if an approval is rejected.


app/mailers/communicart_mailer.rb
```diff
@@ -17,12 +17,22 @@ def cart_observer_email(to_email, cart)
     send_cart_email(sender, to_email, cart)
   end
 
+  def sent_confirmation_email(cart)
```
phirefly: Open to better naming suggestions, perhaps something that indicates it is going to the requester?

lib/ncr_dispatcher.rb
```diff
@@ -0,0 +1,6 @@
+class NcrDispatcher < LinearDispatcher
```
phirefly: Open to doing away with this altogether and maybe injecting this logic somehow into `requires_approval_notice?` in LinearDispatcher.

app/mailers/communicart_mailer.rb
afeld: How about `send_proposal_confirmation`, or `proposal_created_confirmation`? Not too concerned about being wordy here.

lib/dispatcher.rb
```diff
@@ -10,21 +10,37 @@ def email_observers(cart)
     end
   end
 
+  def email_sent_confirmation(cart)
```
afeld: See suggestions above for related name.

lib/ncr_dispatcher.rb
```diff
@@ -0,0 +1,6 @@
...
+
+  def requires_approval_notice? approval
+    approval.cart_approvals.approvable.order('position ASC').last == approval
+  end
```
afeld: How about this:

```ruby
def requires_approval_notice?(approval)
  final_approval(approval.cart) == approval
end

def final_approval(cart)
  cart.ordered_approvals.last
end
```

lib/ncr_dispatcher.rb
afeld: Argh, yeah, really hoping to avoid more NCR-specific logic. The important thing here is that the requesters don't want as many emails? It sounded like they had mixed opinions on it, so I am reluctant to add the logic to give them less information until they are actually using the system.

spec/integration/cart_creation_without_approval_group_spec.rb
```diff
@@ -85,9 +85,9 @@
     expect(Approval.count).to eq 3
   end
 
-  it 'delivers emails to approver email addresses indicated in the toAddress field' do
+  it 'delivers emails to requester and approver email addresses indicated in the toAddress field' do
     post 'send_cart', @json_params_1
-    expect(ActionMailer::Base.deliveries.count).to eq 2
+    expect(ActionMailer::Base.deliveries.count).to eq 3
```
afeld: Maybe check the emails?

```ruby
expect(email_recipients).to eq(%w(...))
```

lib/ncr_dispatcher.rb
phirefly: :+1:

lib/ncr_dispatcher.rb
phirefly: Yes, Chris mentioned this in both of our meetings with him. I think we should have this in our next demo, whether it is in its current form or if the logic is passed into the existing `LinearDispatcher` method in a block. I don't mind quickly doing the latter, PR into the PR style. ;)

afeld: It's not the fact that there's a subclass, so much as I question whether that desire is actually NCR-specific/wide. Seems like it's either the case that it would be a requester-specific preference, or we should reduce the number of notifications for *all* requesters across the board. If this is just a hold-over until we know, though, totally fine.

afeld: Bump!

phirefly: I'm happy leaving this ncr-specific for now, although user-specific is the more realistic use case if/when we start managing user preferences. The main problem that this would need to solve for now is that Navigator folks are happy with sending an email to requesters every time an approval happens, while the NCR folks are not.

lib/dispatcher.rb
```diff
@@ -10,21 +10,37 @@ def email_observers(cart)
...
+    true
+  end
+
   def self.initialize_dispatcher(cart)
     case cart.flow
     when 'parallel'
       ParallelDispatcher.new
     when 'linear'
-      LinearDispatcher.new
+      if cart.getProp('origin') == 'ncr'
```
afeld: Use `cart.ncr?`.

lib/ncr_dispatcher.rb
afeld: Maybe just add a comment saying this is a placeholder for doing per-user preferences, and linking to the story?

phirefly:
Great feedback. Thanks.

phirefly:
Updated!

PR #199, opened 2015-02-04
Removed NcrLinearDispatcher and moved logic into base Dispatcher

phirefly:
An approach for removing the NcrDispatcher


lib/dispatcher.rb
```diff
@@ -28,7 +28,17 @@ def on_approval_status_change(approval)
   end
 
   def requires_approval_notice?(approval)
-    true
+    if approval.cart.getProp('origin') == 'ncr'
+      self.with_approval_notice_requirements { approval.is_final_approval? }
```
phirefly: This could turn into a rules block that we pull out of here, potentially covering all of the different clients and addressing other client-specific functions

cmc333333: +1 separating this out. Ideally, the code should never explicitly check "am I ncr?". Calling could be based on method overriding or configurations. For the latter, I'd imagine an api like this:

```ruby
if approval.cart.getProp('origin').needs_notice
```

afeld:
The proc passing is strange to me... I think I prefer the approach in https://github.com/18F/C2/pull/197/. Feel free to reopen if you disagree!

phirefly:
I'm cool with it. This may be relevant later but the [other way](https://github.com/18F/C2/pull/197) seems like an appropriate small step, easy to grok.

PR #200, opened 2015-02-06
86218872: Add guard when sending emails in linear flow

cmc333333:
If a cart gets approved or rejected, don't send emails to the "next" approver. I'm not restricting to "rejected" in case there's ever a way to approve the whole cart outside of the normal work flow.

### DIFF
--- a/lib/linear_dispatcher.rb
+++ b/lib/linear_dispatcher.rb
@@ -2,7 +2,9 @@ class LinearDispatcher < Dispatcher
   def next_pending_approval(cart)
     # we don't care how the cart was approved/rejected
     if cart.pending?
-      return cart.ordered_awaiting_approvals.first
+      cart.ordered_awaiting_approvals.first
+    else
+      nil
     end
   end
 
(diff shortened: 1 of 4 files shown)


lib/linear_dispatcher.rb
```diff
@@ -1,21 +1,24 @@
 class LinearDispatcher < Dispatcher
-  def next_approval(cart)
-    cart.ordered_awaiting_approvals.first
+  def next_pending_approval(cart)
+    # we don't care how the cart was approved/rejected
+    if cart.pending?
+      return cart.ordered_awaiting_approvals.first
+    end
```
afeld: I'm laughing to myself as I type this, but mind making the return value more explicit, but *excluding* the explicit `return`?

```ruby
def next_pending_approval(cart)
  # we don't care how the cart was approved/rejected
  if cart.pending?
    cart.ordered_awaiting_approvals.first
  else
    nil
  end
end
```

#rubyisms

cmc333333: Fair enough. Is the `else` relevant or does in an implicit nil okay?

afeld: I prefer to be more explicit about it, but don't feel super strongly.

PR #201, opened 2015-02-06
[#87244764] add Spring to speed up loading Rails in development

afeld:


afeld:
FYI, this took starting RSpec went from seven seconds to three on my machine :rocket:

phirefly:
Needs a merge with master and then :ship: it!

PR #202, opened 2015-02-06
remove sdoc gem

afeld:
We don't use this, and if we did want API docs, might as well point people to http://www.rubydoc.info/github/18F/C2/master.


phirefly:
:shipit:

PR #204, opened 2015-02-06
consolidate spec directories

afeld:
`integration` specs are the same as `request` specs, as far as RSpec-Rails is concerned.


phirefly:
:shipit:

PR #206, opened 2015-02-10
87947534: Use i18n to override field names

cmc333333:
Specifically, the rwa field.


cmc333333:
Updated to account for a flaky test -- enforce an order to parallel approvers.

app/models/cart.rb
```diff
@@ -69,7 +69,8 @@ def ordered_awaiting_approvals
   # users with outstanding cart_notification_emails
   def currently_awaiting_approvers
```
phirefly: We can use a spec for this

cmc333333:
Specs have been added

phirefly:
:ship: it

PR #207, opened 2015-02-10
[#87139776] Update icon image helper for emails to account for linear

phirefly:
In anticipation of us reconciling the two treatments for linear/parallel status displays, I kept this simple by just removing the vertical line between approvals.

Before:
screen shot 2015-02-10 at 10 55 12 am

After:
screen shot 2015-02-10 at 10 54 48 am


cmc333333:
:+1:

PR #208, opened 2015-02-13
Workflow on approvals

cmc333333:
This dips a toe in workflows by using one to replace approval status. On the plus side, a step or two of indirection has been removed. I wasn't sure where to put the workflow, so I moved it into a "helper" -- let me know if there's a more logical place.

This provides no new tests as it should be logically equivalent to the existing code.


app/models/approval.rb
```diff
@@ -66,10 +59,15 @@ def approved_at
...
 
-  private
-
-  def set_default_status
-    self.status ||= 'pending'
+  # Used by the state machine
+  def on_approved_entry(new_state, event)
+    self.cart.update_approval_status
+    Dispatcher.on_approval_status_change(self)  # todo - move this out
   end
```
afeld: I would normally say we should pull the contents of these two identical methods into a common function, but I don't think they *should* be the same, or that they will stay the same for very long. Adding the `ThreeStateWorkflow` to the `Cart` will probably take care of this.

afeld:
So glad this is getting done – thanks!

PR #209, opened 2015-02-13
87736992 approve reject cart show

phirefly:
WIP: Feedback would be great, especially around bypassing token validation through the use of a queryparam. I'm starting to feel the pain of managing role, so that work will likely clean up the checks for cart status in the carts helper.

### DIFF
--- a/app/controllers/communicarts_controller.rb
+++ b/app/controllers/communicarts_controller.rb
@@ -56,7 +56,7 @@ private
 
   def validate_access
     request_from_email = params['email_delivery'].presence || true
-    return unless request_from_email == 'true'
+    return unless request_from_email == true
 
     @token = ApiToken.find_by(access_token: params[:cch])
     if !@token
--- a/app/controllers/communicarts_controller.rb
+++ b/app/controllers/communicarts_controller.rb
@@ -42,7 +42,7 @@ class CommunicartsController < ApplicationController
     approval = cart.approvals.find_by(user_id: params[:user_id])
 
     Commands::Approval::UpdateFromApprovalResponse.new.perform(approval, approval_response_status)
-    @token.update_attribute(:used_at, Time.now)
+    @token.update_attribute(:used_at, Time.now) if @token
 
     flash[:success] = "You have #{approval_response_status} Cart #{cart.public_identifier}."
     redirect_to cart_path(cart)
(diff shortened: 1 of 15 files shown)


app/controllers/communicarts_controller.rb
```diff
@@ -52,6 +55,9 @@ def approval_response
 private
 
   def validate_access
+    request_from_email = params['email_delivery'].presence || true
+    return unless request_from_email == true
```
cmc333333: This logic is kind of confusing. If `params['email_delivery']` is not truthy, we'll get `true` (due of the second half of the or); we'll then return. If `params['email_delivery']` *is* truthy, we won't return. It doesn't look like it's ever truthy, though?

spec/steps/approval_steps.rb
```diff
@@ -12,13 +16,24 @@ module ApprovalSteps
...
+    @cart.approvals << FactoryGirl.create(:approval, role: 'approver', user_id: approver.id, position: position)
   end
 
   step "feature flag :flag_name is :value" do |flag, value|
     ENV[flag] = value
   end
 
+  step 'the cart has been approved by the logged in user' do
+    approval = @cart.approvals.where(user_id: @current_user.id).first
+    approval.update_attributes(status: 'approved')
```
cmc333333: Consider `approval.approve!` for consistency.

app/controllers/communicarts_controller.rb
phirefly: Yeah, it is confusing. From the web requests, I'm using a queryparam `email_delivery=false` which bypasses the token validation.

phirefly: I can probably simplify to `return if params['email_delivery'] == false`

phirefly: So, to answer your previous question, the current code sets `request_from_email` to true unless we explicitly say it's false, which is what I'm doing in the request. If we don't pass the param, the default value is true. This is the case when we click Approve from one of the emails.

spec/steps/approval_steps.rb
phirefly: I don't see that method. But it should exist!

phirefly: Perhaps it's avilable after the merge? ;)

cmc333333: Yep; after you merge master, you'll have access to the workflow methods, including `approve!`

app/controllers/communicarts_controller.rb
cmc333333: `return if params['email_delivery'] == false` is more readable to my untrained eye.

Bigger question: why is whether or not a user came from email the deciding factor? I'd expect the token to be ignored (though, perhaps still expired) if the user was already logged in, regardless of how the user reached this page.

phirefly:
Squashing and submitting another PR

PR #210, opened 2015-02-13
ThreeStateWorkflow on cart

cmc333333:
Added a "partial_approve" event to the three state workflow and updated the Cart to use it.

Replace many chunks of code with their equivalent using a workflow.


afeld:
Very slick!

PR #213, opened 2015-02-18
use scopes for queries throughout the app

afeld:
Fewer hard-coded strings!


app/models/cart.rb
```diff
@@ -64,11 +64,11 @@ def currently_awaiting_approvers
   end
 
   def approved_approvals
-    self.approver_approvals.where(status: 'approved')
+    self.approver_approvals.approved
   end
 
   def all_approvals_received?
-    self.approver_approvals.where('status != ?', 'approved').empty?
+    self.approver_approvals.pending.empty?
```
cmc333333: What about rejected approvals?

afeld: Hmmmmm, good catch. Maybe this method needs to be renamed, because I'm seeing it used as both "all approvals got a response", as well as "everything was approved".

afeld: A test you added later caught this failure. Thanks from The Future @cmc333333!

afeld:
Allrighty, ready for another look!

app/models/user_role.rb
```diff
@@ -11,4 +11,8 @@ class UserRole < ActiveRecord::Base
   validates_presence_of :approval_group_id
   validates :role, presence: true, inclusion: {in: ROLES}
   # TODO Limit requester to only one at this time
+
+  ROLES.each do |role|
+    scope role.pluralize.to_sym, -> { where(role: role) }
```
cmc333333: Cute

app/models/cart.rb
```diff
@@ -18,18 +18,19 @@ class Cart < ActiveRecord::Base
...
+  workflow_spec.states.keys.each do |state|
+    scope state, -> { where(status: state) }
+  end
+  scope :closed, -> { where(status: ['approved', 'rejected']) }
 
   ORIGINS = %w(navigator ncr)
 
   def rejections
-    self.approvals.where(status: 'rejected')
+    self.approver_approvals.rejected
```
cmc333333: Good call -- this is no doubt the intention

PR #214, opened 2015-02-18
87736992 approve reject cart show 2

phirefly:


app/controllers/communicarts_controller.rb
```diff
@@ -62,6 +65,8 @@ def approval_response
 private
 
   def validate_access
+    return if signed_in?
```
afeld: We should probably mark their token as used even if they're signed in.

app/views/carts/_approval_actions.html.erb
```diff
@@ -0,0 +1,19 @@
+<% if display_response_actions?(cart, current_user) %>
```
afeld: Minor: can we put the conditional outside of the partial?

app/views/carts/_approval_actions.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  <div id='reply-section'>
+    <p>
+      <%= link_to "Approve",
+          approval_response_path(approver_action: 'approve',
+                                 cart_id: cart.id,
+                                 scheme: 'https',
```
afeld: Is this needed? Should just be a path.

PR #216, opened 2015-02-19
88355056: First steps of restarting a proposal

cmc333333:
* Adds a restart event, which moves from rejected/pending into a pending state
* When this event is called, app approvals are set to "pending", tokens are expired, and emails are sent out again
* Modified the NCR proposals controller to have an end point for editing an existing proposal - when successful, this will trigger the restart event

No way to access this page without the url at the moment, but I wanted to push this up for feedback first


app/views/ncr/proposals/new.html.erb
```diff
@@ -2,7 +2,7 @@
   <h4>National Capital Region</h4>
   <h3>Purchase Card Approval Request</h3>
 
-  <%= form_for @proposal_form do |f| %>
+  <%= form_for @proposal_form, url: @form_url, method: @form_method do |f| %>
```
afeld: Pretty sure the additions aren't needed. From http://guides.rubyonrails.org/form_helpers.html#relying-on-record-identification:

> you can just pass the model instance and have Rails figure out model name and the rest... Record identification is smart enough to figure out if the record is new by asking record.new_record?. It also selects the correct path to submit to and the name based on the class of the object.

app/controllers/ncr/proposals_controller.rb
```diff
@@ -29,6 +32,46 @@ def create
...
+      if cart.requester != current_user
+        redirect_to new_ncr_proposal_path, :alert => 'You cannot restart that proposal'
+      end
+    end
+
+    def edit
+      cart = Cart.find(params[:id])
+      @proposal_form = Ncr::ProposalForm.from_cart(cart)
+      @form_url, @form_method = {action: "update"}, "put"
+      render 'new'
```
afeld: Can we rename the view to `form.html.erb`? This is the Rails-y way of indicating that it's used for both `new` and `edit`.

app/controllers/ncr/proposals_controller.rb
```diff
@@ -29,6 +32,46 @@ def create
...
+      if cart.approved?
+        redirect_to new_ncr_proposal_path, :alert => "That proposal's already approved. New proposal?"
+      end
+    end
+    def cart_owner
+      cart = Cart.find(params[:id])
+      if cart.requester != current_user
+        redirect_to new_ncr_proposal_path, :alert => 'You cannot restart that proposal'
+      end
+    end
```
afeld: How about

* Combining these methods and calling it `redirect_if_cart_cant_be_edited` or something?
* Moving them/it to the bottom of the controller?

app/controllers/ncr/proposals_controller.rb
```diff
@@ -29,6 +32,46 @@ def create
       end
     end
 
+    def not_approved
+      cart = Cart.find(params[:id])
```
afeld: Mind making a shared method so there's only a single call to the database between all of the methods?

```ruby
def cart
  @cart ||= Cart.find(params[:id])
end
```

app/controllers/ncr/proposals_controller.rb
```diff
@@ -29,6 +32,46 @@ def create
...
+      if cart.approved?
+        redirect_to new_ncr_proposal_path, :alert => "That proposal's already approved. New proposal?"
+      end
+    end
+    def cart_owner
+      cart = Cart.find(params[:id])
+      if cart.requester != current_user
+        redirect_to new_ncr_proposal_path, :alert => 'You cannot restart that proposal'
```
afeld: This is a good start, but we should really have a more comprehensive authorization layer sooner or later.

https://www.ruby-toolbox.com/categories/rails_authorization

app/controllers/ncr/proposals_controller.rb
```diff
@@ -29,6 +32,46 @@ def create
...
+    end
+
+    def update
+      cart = Cart.find(params[:id])
+      @proposal_form = Ncr::ProposalForm.new(params[:ncr_proposal])
+      @form_url, @form_method = {action: "update"}, "put"
+      @proposal_form.requester = current_user
+      if @proposal_form.valid?
+        @proposal_form.update_cart(cart)
+        if cart.persisted?
```
afeld: Go with [`cart.changed?`](http://api.rubyonrails.org/classes/ActiveModel/Dirty.html#method-i-changed-3F) or `cart.errors.any?` instead.

app/models/cart.rb
```diff
@@ -250,4 +250,10 @@ def on_rejected_entry(prev_state, event)
       Dispatcher.on_cart_rejected(self)
     end
   end
+
+  def restart
+    self.api_tokens.update_all(expires_at: Time.now)
+    self.approver_approvals.update_all(status: 'pending')
```
afeld: What about

```ruby
self.approver_approvals.each do |approval|
  unless approval.pending?
    approval.restart!
  end
end
```

dr evil

app/models/ncr/proposal_form.rb
```diff
@@ -59,23 +59,49 @@ def create_cart
...
+        amount: self.amount,
+        expense_type: self.expense_type,
+        vendor: self.vendor,
+        not_to_exceed: self.not_to_exceed,
+        building_number: self.building_number,
+        rwa_number: self.rwa_number,
+        office: self.office
+      )
+      cart.set_requester(self.requester)
+    end
```
afeld: Niiiiiiice.

app/models/ncr/proposal_form.rb
```diff
@@ -59,23 +59,49 @@ def create_cart
...
       end
 
       cart
     end
+
+    def update_cart(cart)
+      cart.name = self.description
+      if cart.save
+        cart.approver_approvals.destroy_all
+        cart.clear_props!
```
afeld: Hmmmmm.... are we sure we want to do this? Not aware of any issues at the moment, but seems like this could lead to a bug.

app/models/ncr/proposal_form.rb
```diff
@@ -59,23 +59,49 @@ def create_cart
...
+      )
+      cart.set_requester(self.requester)
+    end
+
+    def self.from_cart(cart)
+      form = self.new(cart.deserialized_properties)
+      form.approver_email = cart.ordered_approvals.first.user.email_address
+      form.description = cart.name
+      form
+    end
```
afeld: This method could use tests – easy to accidentally leave a property unassigned.

app/views/ncr/proposals/new.html.erb
```diff
@@ -24,12 +24,9 @@
...
       <%= f.label :building_number %>
-      <%= f.select(:building_number, [], {}, {class: 'form-control', multiple: false}) do %>
-        <%= content_tag(:option, '', value: '') %>
-        <% Ncr::ProposalForm::BUILDING_NUMBERS.each do |building| %>
-          <%= content_tag(:option, building, value: building) %>
-        <% end %>
-      <% end %>
+      <%= f.collection_select(
+            :building_number, Ncr::ProposalForm::BUILDING_NUMBERS, :to_s, :to_s,
+            {include_blank: true}, {class: 'form-control'}) %>
```
afeld: :clap:

spec/features/ncr_proposal_spec.rb
```diff
@@ -87,5 +87,49 @@
       expect(current_path).to eq("/carts/#{Cart.last.id}")
       expect(page).to have_content("RWA Number")
     end
+
+    it "can be restarted if pending" do
+      proposal = FactoryGirl.build(:proposal_form)
+      proposal.requester = requester
+      cart = proposal.create_cart
```
afeld: Not sure that you actually need the Proposal, right?

```ruby
cart = FactoryGirl.create(:cart)
cart.set_requester(requester)
```

spec/features/ncr_proposal_spec.rb
```diff
@@ -87,5 +87,49 @@
...
+
+      visit "/ncr/proposals/#{cart.id}/edit"
+      expect(find_field("ncr_proposal_building_number").value).to eq(
+        'Entire WH Complex')
+      fill_in 'Vendor', with: 'ACME'
+      click_on 'Submit for approval'
+      expect(current_path).to eq("/carts/#{cart.id}")
+      expect(page).to have_content("ACME")
+      expect(page).to have_content("resubmitted")
```
afeld: You might also want to check that the value was persisted:

```ruby
cart.reload
expect(cart.getProp('vendor')).to eq('ACME')
```

spec/models/cart_spec.rb
```diff
@@ -136,5 +136,27 @@
...
+  describe '#restart' do
+    it 'resets approval states when rejected' do
+      cart = FactoryGirl.create(:cart_with_approvals)
+      Dispatcher.deliver_new_cart_emails(cart)
+      expect(cart.api_tokens.length).to eq(2)
+
+      cart.approver_approvals.first.approve!
+      cart.approver_approvals.last.reject!
+      expect(cart.approvals.where(status: 'approved').length).to eq(1)
+      expect(cart.approvals.where(status: 'rejected').length).to eq(1)
```
afeld: Hmm, wonder if we can move some of this setup into a factory or something. Not a blocker.

spec/models/cart_spec.rb
```diff
@@ -136,5 +136,27 @@
...
+      expect(cart.rejected?).to eq(true)
+
+      cart.restart!
+
+      expect(cart.pending?).to eq(true)
+      expect(cart.api_tokens.where("expires_at < ?", Time.now).length).to eq(2)
+      expect(cart.api_tokens.where("expires_at > ?", Time.now).length).to eq(2)
```
afeld: Use `cart.unexpired`.

spec/models/cart_spec.rb
```diff
@@ -136,5 +136,27 @@
...
+      cart.approver_approvals.first.approve!
+      cart.approver_approvals.last.reject!
+      expect(cart.approvals.where(status: 'approved').length).to eq(1)
+      expect(cart.approvals.where(status: 'rejected').length).to eq(1)
+      expect(cart.rejected?).to eq(true)
+
+      cart.restart!
+
+      expect(cart.pending?).to eq(true)
+      expect(cart.api_tokens.where("expires_at < ?", Time.now).length).to eq(2)
```
afeld: Mind adding a `Cart.expired` scope?

afeld:
Hope you don't mind all the nitpicks! I think users will really appreciate this feature.

app/models/cart.rb
afeld: Also, we should add an inline note/TODO that the history isn't being kept.

afeld:
> No way to access this page without the url at the moment, but I wanted to push this up for feedback first

app/views/ncr/proposals/new.html.erb
cmc333333: The problem is that a proposal form is never actually saved anywhere, so regardless of if we are "editing" or not, form_for will think we're creating a new instance. If there was some flag on the model itself that I could say "hey, pretend live you're going to be edited", that'd be much cleaner. Does that exist?

app/controllers/ncr/proposals_controller.rb
cmc333333: Absolutely

app/controllers/ncr/proposals_controller.rb
cmc333333: Sounds good.

app/controllers/ncr/proposals_controller.rb
cmc333333: Sure

app/controllers/ncr/proposals_controller.rb
cmc333333: Oof, yeah that makes sense.

app/models/cart.rb
cmc333333: Delightfully evil

app/models/ncr/proposal_form.rb
cmc333333: This is meant to remove any weird versioning issues -- e.g. fields that were present in a previous version of the form but are not anymore. It would also remove other properties set on the cart, if we began collecting more meta data, which would probably be bad.

Maybe a list of fields (which only grows) to remove could account for both types of bugs?

app/models/ncr/proposal_form.rb
cmc333333: Good point

spec/features/ncr_proposal_spec.rb
cmc333333: I wanted the properties from the factorygirl proposal form. If I just used `:cart`, I wouldn't get those, I don't think.

spec/features/ncr_proposal_spec.rb
cmc333333: Good idea!

app/views/ncr/proposals/new.html.erb
afeld: Aaaah, good question. Try this:

http://stackoverflow.com/a/5969673/358804

Happy to take a stab at it in a separate PR if you like.

app/models/ncr/proposal_form.rb
afeld: @phirefly Any gut feeling here? I'm fine with leaving it with a note that "we may not actually want to do this".

spec/models/cart_spec.rb
```diff
@@ -136,5 +136,27 @@
       expect(email_recipients).to eq(['requester@some-dot-gov.gov'])
     end
   end
+  describe '#restart' do
```
afeld: You might also add tests to see that the properties are removed.

spec/features/ncr_proposal_spec.rb
afeld: Fair enough!

app/views/ncr/proposals/new.html.erb
cmc333333: I've played with this a bit more and had no luck. I'm going to leave it for future refactoring.

app/controllers/ncr/proposals_controller.rb
cmc333333: https://www.pivotaltracker.com/story/show/89017762

spec/models/cart_spec.rb
cmc333333: Should restarting a cart always remove properties? Right now it's up to the caller to update the properties.

cmc333333:
Updated the PR with several of @afeld 's suggestions.

spec/models/cart_spec.rb
afeld: Oh, right, that makes sense.

app/views/ncr/proposals/new.html.erb
afeld: I'm gonna take a crack at this in the morning – will time box to an hour and submit a PR on top of this one. If I can't get it, will go ahead and merge.

afeld: Don't want to block this – will send another PR if I can figure it out.

PR #217, opened 2015-02-24
87736992 fix parallel approval display

phirefly:
Hides approval approval actions for requesters per @cmc333333. Also included some previous feedback from @afeld.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -22,7 +22,7 @@ class User < ActiveRecord::Base
     self.carts.where(approvals: {role: 'requester'})
   end
 
-  def is_approver?(cart)
+  def approver_of?(cart)
     cart.approvers.include? self
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -22,6 +22,10 @@ class User < ActiveRecord::Base
     self.carts.where(approvals: {role: 'requester'})
   end
 
+  def is_approver?(cart)
+    cart.approvers.include? self
+  end
+
   def last_requested_cart
     self.requested_carts.order('carts.created_at DESC').first
   end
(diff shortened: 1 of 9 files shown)


app/models/user.rb
```diff
@@ -22,6 +22,10 @@ def requested_carts
     self.carts.where(approvals: {role: 'requester'})
   end
 
+  def is_approver?(cart)
```
afeld: Minor: I'd call it `approver?` or `approver_of?`.

cmc333333:
If I read this right, it'll prevent both the requester and observers from seeing the approve/reject buttons (yay). To test the boundaries a bit more, would an approver who has already rejected/approved continue to see the buttons?

afeld:
@cmc333333 *Good. Eye.* :clap: 

/cc https://www.pivotaltracker.com/story/show/89017762

app/models/user.rb
phirefly: I like `approver_of?` Thx!

phirefly:
+1
For now, we are handling that here:
Parallel approvals have to be in [pending status](https://github.com/18F/C2/blob/master/app/helpers/carts_helper.rb#L26)
Linear approvals have to be pending, since we call `ordered_awaiting_approvals` [here](https://github.com/18F/C2/blob/master/app/helpers/carts_helper.rb#L34)

On that last point, I may add the guard `return false unless cart.linear?` instead of using `cart.linear?` for consistency with the previous method. Let me know what you think.

phirefly:
So, in a nutshell, the buttons would not be visible if an approver has already rejected/approved.

afeld:
Can we add some specs to check that the buttons don't appear in those scenarios?

phirefly:
Updated with specs `Displaying approval actions for a parallel cart` and `Displaying approval actions for a linear cart`

cmc333333:
I'm ready to merge unless you want to make that name change, @phirefly

PR #218, opened 2015-02-26
create a .states method in ThreeStateWorkflow

afeld:
Also, validate `Cart#status`.


afeld:
Going to merge so I need to update #127 *one less time*, but happy to address any feedback in a follow-up PR.

cmc333333:
:+1:

PR #220, opened 2015-02-26
[#87565664] redirect back to cart when comment is added

afeld:


app/controllers/comments_controller.rb
```diff
@@ -7,12 +7,16 @@ def index
...
+    commentable = find_commentable
+    comment = commentable.comments.build(comment_params)
+    comment.user = current_user
+    if comment.save
+      flash[:notice] = "You successfully added a comment for #{commentable.class.name} #{commentable.id}"
+      if commentable.respond_to?(:cart)
+        redirect_to commentable.cart
+      else
+        redirect_to commentable
+      end
```
afeld: This is the relevant part.

phirefly:
:shipit: 

It might be worth a team discussion on possibly merging the spec/features and spec/acceptance tests.

afeld:
Agreed. I am personally not a fan of Gherkin/Cucumber/Turnip (happy to elaborate if you like), so would be happy with going all-RSpec.

PR #221, opened 2015-02-26
[#87051300] remove unused styles

afeld:
The gems mentioned in the story didn't work out of the box unfortunately, so I just started removing some big chunks by hand.


phirefly:
:ship: Thanks for the upgrades too.

PR #222, opened 2015-03-02
fix Proposal migration to not rely on models

afeld:
Before this change, I was getting an exception during the `up` part of the migration that `delegate` needs the associated object to be present... in other words, a :chicken: and :egg: problem of the Proposal needing to exist for every Cart in order for the associated Proposals to be created.

This solution is a bit of a hack, but it's certainly easier than writing raw SQL. @phirefly Mind checking out this branch locally and giving it a try?

### DIFF
--- a/db/migrate/20141214220050_create_proposals.rb
+++ b/db/migrate/20141214220050_create_proposals.rb
@@ -1,4 +1,16 @@
 class CreateProposals < ActiveRecord::Migration
+  ## http://makandracards.com/makandra/15575-how-to-write-complex-migrations-in-rails ##
+  class TempCart < ActiveRecord::Base
+    self.table_name = 'carts'
+    belongs_to :temp_proposal, foreign_key: 'proposal_id'
+  end
+
+  class TempProposal < ActiveRecord::Base
+    self.table_name = 'proposals'
+    has_many :temp_carts
+  end
+  ######################################################################################
+
   def change
     create_table :proposals do |t|
       t.string :status
@@ -8,11 +20,12 @@ class CreateProposals < ActiveRecord::Migration
     add_reference :carts, :proposal
 
     reversible do |dir|
-      Cart.reset_column_information
+      TempCart.reset_column_information
+      TempProposal.reset_column_information
 
       dir.up do
-        Cart.find_each do |cart|
-          cart.create_proposal!(
+        TempCart.find_each do |cart|
+          cart.create_temp_proposal!(
             status: cart.status,
             flow: cart.flow,
             created_at: cart.created_at,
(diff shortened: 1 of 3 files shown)


afeld:
Also, starting to think that the `remove_column` (the statements that delete data) should be moved to a separate migration so they can be run separately... thoughts?

monfresh:
What is the scenario where the original migration fails? What kind of content does the database have to contain for this to fail? I see that the migration was created back in December. Did it fail then too? If not, why now?

arowla:
This seems like an OK approach, though I wonder if it could also be done by reserving the addition of the foreign key reference for last. And actually, like you mentioned, the removal of columns would ideally be done in a separate step, at least for a production app, to avoid breaking backwards compatibility with code while the app is running-- so perhaps it makes sense to break the whole thing into 2 to 3 steps.

phirefly:
migrate works, but rollback [fails](https://gist.github.com/phirefly/d024947f103159006f55)

cpapazian:
not quite sure what is going on here, but it looks like you are mixing data and schema migrations? if you didn't have to worry about migrating the data, would you have had a problem here? my advice would be to move the data migration to a separate script. whether that script is considered a "migration" or just a script is up to you ... there are downsides to either.

afeld:
I am never using one of the `app/models/` in a migration again :smirk: 

> migrate works, but rollback fails

Fixed now. Mind trying the migration in both directions locally again?

> What is the scenario where the original migration fails?

```
git:(2a3db7b) $ bin/rake db:migrate
== 20141214220050 CreateProposals: migrating ==================================
-- create_table(:proposals)
   -> 0.0053s
-- add_reference(:carts, :proposal)
   -> 0.0008s
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

Cart#status delegated to proposal.status, but proposal is nil: #<Cart id: 5, name: "Test Cart needing approval", created_at: "2015-01-28 17:15:04", updated_at: "2015-03-03 05:21:35", external_id: nil, status: "pending", flow: "linear", proposal_id: nil>
/Users/aidanfeldman/dev/18F/c2/app/models/concerns/proposal_delegate.rb:25:in `rescue in status'
/Users/aidanfeldman/dev/18F/c2/app/models/concerns/proposal_delegate.rb:25:in `status'
/Users/aidanfeldman/dev/18F/c2/db/migrate/20141214220050_create_proposals.rb:16:in `block (3 levels) in change'
```

Aaaah, looking back at it again, it was because [I was using `cart.status`](https://github.com/18F/C2/blob/2a3db7b8b00eaa45bd7677d8f7a8f901cdb46649/db/migrate/20141214220050_create_proposals.rb#L16) (which is delegated), rather than `cart[:status]`. Anyway, still probably better to not touch the real models.

> the removal of columns would ideally be done in a separate step

Good call – will re-remove the columns in a separate PR.

phirefly:
:ship: it

PR #223, opened 2015-03-02
Ignore node modules now Cartmaker was removed in #84957974

phirefly:


afeld:
No need – just delete the directory!

PR #225, opened 2015-03-03
[#84953008] remove support for Mario

afeld:
Builds on https://github.com/18F/C2/pull/224. Depends on https://github.com/18F/cloud-cutter/pull/37.

*Oh man* deleting large swaths of code is so cathartic :sweat_smile:


spec/requests/cart_linear_approval_spec.rb
```diff
@@ -1,145 +0,0 @@
...
-      json_approval_params["fromAddress"] = "approver3@some-dot-gov.gov"
-      approve
-
-      expect(cart.approved?).to eq true
-      expect(cart.approvals.approved.count).to eq 3
-      expect(email_recipients).to eq(['test-requester@some-dot-gov.gov'])
-    end
-  end
-
-  context 'NCR mailing behavior' do
```
cmc333333: Seems like this shouldn't be deleted?

cmc333333:
This pulls out all of the JSON end points, right? Aren't those used by a client?

spec/requests/cart_linear_approval_spec.rb
afeld: I removed these test files "whole hog" because they were all done through around the endpoint we don't need anymore, though I'll go through and make sure tests are in place for all the underlying functionality this was covering. At the very least, I'll fix the drops in code coverage.

screen shot 2015-03-03 at 2 52 37 pm

afeld:
Oops, didn't mean to delete the branch.

> This pulls out all of the JSON end points, right?

Not all... just `/approval_reply_received`.

> Aren't those used by a client?

I believe Navigator is only using `/send_cart`, and [I'm not seeing it anywhere in their code](https://github.com/search?utf8=%E2%9C%93&q=approval_reply_received+user%3Agsa&type=Code&ref=searchresults). @phirefly :+1: :-1:?

spec/requests/cart_linear_approval_spec.rb
afeld: ...but good catch!

afeld:
Updated!

cmc333333:
I'm fine with removing this, but would like confirmation from @phirefly that removing this end point won't be an issue.

phirefly:
Removing this endpoint is ok, since it was used entirely by Mario. 

One note: While our walkthroughs appear to be confirming the approvers like to click the Approval button, It is becoming clear that the ability to comment (not approve/reject) through the normal email reply is important as well. This resulted in [this story](https://www.pivotaltracker.com/story/show/89824332). This could mean that a Mario-ish solution could still have a place if we need to parse emails for comments. The Mandrill / Mailgun route is probably a better approach, but just want to give additional context.

Unless we want to continue the discussion, :shipit:

afeld:
> the ability to comment...through the normal email reply is important

Yeah, I've been feeling the same way.

> The Mandrill / Mailgun route is probably a better approach

:+1:

PR #226, opened 2015-03-03
Add a link to restart to NCR cart page

cmc333333:
Also generalizes a CSS selector.

I don't like the ncr-specific check in the ruby code. It'd be much nicer if there was a way to say something like:

```
<%= render partial: cart.origin + '/restart_link', locals: {cart: cart} %>
```

and [account for errors](http://stackoverflow.com/questions/3559419/is-there-any-rails-function-to-check-if-a-partial-exists). Yea Nay? 

Looks like:

screen shot 2015-03-03 at 10 50 22 am


app/helpers/carts_helper.rb
```diff
@@ -22,6 +22,11 @@ def display_response_actions?(cart, user)
     current_linear_approval?(cart, user)
   end
 
+  # Todo: Move this to an NCR specific template?
+  def display_restart?(cart, user)
```
afeld: Minor: you can access `current_user` in here directly – don't need to pass it in. Unless you prefer passing it explicitly?

spec/features/ncr_proposal_spec.rb
```diff
@@ -136,5 +136,36 @@
...
+      cart.setProp('origin', 'somewhereElse')
+      visit "/carts/#{cart.id}"
+      expect(page).not_to have_content('Restart this Cart?')
+
+      cart.setProp('origin', 'ncr')
+      req = cart.approvals.where(role: 'requester').first
+      req.update_attribute(:user, FactoryGirl.create(:user))
+      visit "/carts/#{cart.id}"
+      expect(page).not_to have_content('Restart this Cart?')
+    end
```
afeld: Mind breaking these up into five separate tests?

afeld:
> I don't like the ncr-specific check in the ruby code.

I think we will learn more as we build out the different use cases, but what you suggested sounds reasonable to me.

app/helpers/carts_helper.rb
cmc333333: Booo global state. I'll rewrite it, though ;)

afeld: _Technically_ it's another method (that has internal state), but I hear ya. Whatever you prefer!

cmc333333:
Created a `client_partial` function, which will include a partial template if the template is present in the client/origin directory.

Also cleaned up some redundancy in the tests related to restarting for ncr.

app/models/cart.rb
```diff
@@ -203,7 +203,9 @@ def import_cart_items(cart_items_params)
   end
 
   def origin
-    self.getProp('origin')
+    # In practice, carts should always have an origin. Account for test cases
+    # and old data with this "or"
+    self.getProp('origin') || ''
```
afeld: TODO backfill `origin`s on old Carts.

PR #230, opened 2015-03-04
add link to create a new NCR proposal

afeld:
screen shot 2015-03-04 at 6 34 13 pm

### DIFF
--- a/app/views/layouts/_communicart_header.html.erb
+++ b/app/views/layouts/_communicart_header.html.erb
@@ -25,7 +25,9 @@
     <%- if signed_in? %>
       <div class="mycart-link">
         <ul class="header-nav">
-            <li><%= link_to("My Requests", carts_path) %> <%= image_tag "img-cart.png", width: '18px', height:'16px' %></li>
+          <li><%= link_to "My Requests", carts_path %> <%= image_tag "img-cart.png", width: '18px', height:'16px' %></li>
+          <%# TODO only show to NCR people %>
+          <li><%= link_to "New Request", new_ncr_proposal_path %></li>
         </ul>
       </div>
     <%- end %>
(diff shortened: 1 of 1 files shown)


app/views/layouts/_communicart_header.html.erb
```diff
@@ -25,7 +25,9 @@
     <%- if signed_in? %>
       <div class="mycart-link">
         <ul class="header-nav">
-            <li><%= link_to("My Requests", carts_path) %> <%= image_tag "img-cart.png", width: '18px', height:'16px' %></li>
+          <li><%= link_to "My Requests", carts_path %> <%= image_tag "img-cart.png", width: '18px', height:'16px' %></li>
+          <%# TODO only show to NCR people %>
+          <li><%= link_to "New Request", new_ncr_proposal_path %></li>
```
afeld: This brings up the question... do we need to start assigning users to the different use cases in some way? We don't want this link to show up for Ric, for example.

cmc333333: That's a good question. It certainly seems like users need to be restricted to a domain/client/use case (or perhaps multiple).

When this was implemented at a previous employer, we had different schemas for each client. Basically, all requests included a header which provided the client id; that client id then determined which db connection was used. This acted as a form of context throughout the rest of the request. That might be overkill here, but it'd certainly be a good idea to separate which models are associated with a single client (carts, approvals), which could have multiple clients (perhaps users), and which are system wide (perhaps none, right now?)

phirefly:
I'd like to merge with these two considerations:
1) To @cmc333333's point about having users be on multiple clients, I would be in favor of updating the label to say `New NCR Request`, just to minimize confusion while we're dealing with multiple clients (18F and Navigator) and to remind us that it's conceivable for a user to create different types of requests.

2) Get something in place that would allow us to identify each user with client(s). Not necessary for this to be merged, but @nickbristow's efforts on the 18F request form may give us reason to explore this user/client link sooner rather than later. Until we have that working, it may be appropriate to feature flag this to hide from Navigator and (soon-to-be) 18F folks in our production environment.

Let me know what you guys think.

afeld:
> I would be in favor of updating the label to say "New NCR Request"

Done!

> identify each user with client(s)

Good call – made https://www.pivotaltracker.com/story/show/89841186.

PR #231, opened 2015-03-06
18f purchase form

nickbristow:
Work In progress, 

Please hold off merging.


afeld:
Screenshot or it didn't happen :wink:

nickbristow:
c2
@afeld

app/models/gsa18f/proposal_form.rb
```diff
@@ -0,0 +1,124 @@
+#module Ncr
+module Gsa18f
+  DATA = YAML.load_file("#{Rails.root}/config/data/18f.yaml")
+
+  class ProposalForm
+    include SimpleFormObject
+
+    EXPENSE_TYPES = %w(BA61 BA80)
```
phirefly: These are NCR specific and can be removed

app/models/gsa18f/proposal_form.rb
```diff
@@ -0,0 +1,124 @@
...
+
+    URGENCY = DATA['URGENCY']
+    OFFICES = DATA['OFFICES']
+    attribute :origin, :string
```
phirefly: Similarly, we can remove NCR-specific attributes

app/models/gsa18f/proposal_form.rb
```diff
@@ -0,0 +1,124 @@
...
+
+    def finance_approver_email
+      ENV['NCR_FINANCE_APPROVER_EMAIL'] || 'communicart.ofm.approver@gmail.com'
+    end
+
+    def requires_finance_approval?
+      self.expense_type == 'BA61'
+    end
+
+    def approver_emails
```
phirefly: In the 18F case, I'm thinking we can just make Ric the approver_email and not even require the requester to type in his email address. This means that 1) removal from the form, and 2) updating `approver_emails` to be `['rics@email_address']`

phirefly: Another option is to not hard code his email and keep it in the yaml. Thoughts?

cmc333333: Ultimately, I'd put it in the db. In the meanwhile, I'd just add an environment variable for 18F_APPROVER_EMAIL

phirefly:
I'm good with that

PR #232, opened 2015-03-06
[#87818416] expire session after two weeks

afeld:
Duration choice was completely arbitrary.


phirefly:
Sounds like a reasonable place to start. :shipit:

PR #233, opened 2015-03-07
[Delivers #89129918] don't pass user_id to approval_response

afeld:


cmc333333:
Thank you! @phirefly I think this'll break your presentation flow, as you'll need to change user accounts. If there's a demo coming up, let's hold off until that's updated. Otherwise, let's merge

phirefly:
Let's discuss in IPM today. We have a demo with Ric/18F as well as NCR user testing later this week.

phirefly:
Noted. And I will change my demo flow. Thank you! 

:shipit:

PR #234, opened 2015-03-08
[Delivers #89771296] remove CartItems and CartItemTraits

afeld:


cmc333333:
This makes sense to me and I didn't see any gotchas in the patch. @phirefly if you can confirm that items are no longer used, let's merge.

phirefly:
Good to delete. Per our conversation earlier this week, let's keep the related Cart data as is in the db.

:ship:

PR #236, opened 2015-03-10
Added check for double approval

nickbristow:
Needs tests and review

### DIFF
--- a/app/controllers/communicarts_controller.rb
+++ b/app/controllers/communicarts_controller.rb
@@ -26,10 +26,8 @@ class CommunicartsController < ApplicationController
     when 'approve'
       current_status = approval.status  
       approval.approve!
-      if approval.status == 'rejected'
-        flash[:success] = "You have already rejected Cart #{cart.public_identifier}."
-      elsif current_status == approval.status
-        flash[:success] = "You have already #{params[:approver_action]}d Cart #{cart.public_identifier}."
+      if current_status == approval.status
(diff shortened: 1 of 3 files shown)


afeld:
@nickbristow My instinct is that we want to handle this at the model level. @cmc333333 What do you think about adding an no-op `:approve` action for the `:approved` state in the `ThreeStateWorkflow`. Also `:reject` for `:rejected`, etc... basically ignoring repeat actions.

https://github.com/18F/C2/blob/master/app/models/concerns/three_state_workflow.rb

nickbristow:
@afeld that was the other way i thought about doing it, but should the user get a different message then?

afeld:
Maybe we can handle it in a more generic way, saying that "if the previous state and new state are the same, just say 'you already `#{action}`d this proposal'".

afeld:
..which would also cover rejection, etc.

nickbristow:
@afeld ok, one sec.

cmc333333:
@afeld There's one use case that is kinda tricky, though: what if a rejected approval gets an "approve" message or vice versa?

I'm more inclined to say "you've already chosen. you can choose no more"

afeld:
:+1:

phirefly:
Per @cmc333333's comment, does this mean we need an `unless approval.status == 'pending'` to cover both rejected and approved?

cmc333333:
:+1:

PR #237, opened 2015-03-10
[Delivers #79483250] document the architecture

afeld:
...among other things. Also made the app ignore the SMTP settings in development mode, since they aren't used.


afeld:
Ready for review!

dlapiduz:
:+1: :+1: :+1: :+1: :+1: :+1: :+1:

doc/overview.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Use cases
+
+This application contains code for several independent but similar use cases. Users will generally be segmented into one use case or another in terms of how the Proposals are initiated, though the approval workflow is (largely) the same.
+
+### [National Capitol Region (NCR) service centers](http://www.gsa.gov/portal/category/21528)
+
+The NCR use case was built around GSA service centers (paint shops, landscapers, etc.) needing approvals for their superiors and various budget officials for credit card purchases. They use the "linear" workflow described [above](#proposal-flows):
+
+1. The requester logs in via MyUSA.
```
phirefly: Do we need to mention that users currently need to be whitelisted?

afeld: Not sure what you mean... I don't know of a user whitelist on either the MyUSA or C2 side.

afeld: Ah ok, you're right. Where is the whitelist specified?

cmc333333:
:+1:

afeld:
Added a note about the whitelisting. Cool if we merge sooner than later, with the assumption that the docs will always need improvement?

PR #239, opened 2015-03-11
[Delivers #90162132] associate ApiTokens with Approvals

afeld:
Makes things easier for some forthcoming refactoring.


spec/lib/parallel_dispatcher_spec.rb
```diff
@@ -13,9 +13,12 @@
     end
 
     it 'creates a new token for each approver' do
```
phirefly: Minor: The name of this specs seems a little off now

app/models/api_token.rb
```diff
@@ -1,17 +1,26 @@
...
+  has_one :user, through: :approval
+
+  validates_presence_of :approval_id
 
   scope :unexpired, -> { where('expires_at >= ?', Time.now) }
   scope :expired, -> { where('expires_at < ?', Time.now) }
   scope :unused, -> { where(used_at: nil) }
   scope :fresh, -> { unused.unexpired }
 
+  def cart_id
```
phirefly: We can delegate `cart_id` to `approval` like we do in [approval](https://github.com/18F/C2/blob/master/app/models/approval.rb#L10)

app/models/api_token.rb
```diff
@@ -1,17 +1,26 @@
...
+    self.approval.cart_id
+  end
+
   def generate_token
```
phirefly: I just noticed we don't have any specs for ApiToken. This method looks spec worthy.

app/models/approval.rb
```diff
@@ -5,7 +5,9 @@ class Approval < ActiveRecord::Base
 
   belongs_to :cart
   belongs_to :user
+  has_one :api_token, -> { fresh }
```
phirefly: Neato

spec/lib/parallel_dispatcher_spec.rb
afeld: I don't think it's inaccurate.

afeld:
Updated!

afeld:
Totally confused about why Timecop doesn't seem to be working in [the failing test](https://circleci.com/gh/18F/C2/578). Need to investigate.

afeld:
Phew, fixed. Stupid floating point comparison! Anyway, ready for another look.

cmc333333:
Splitting things up, one step at a time!

PR #241, opened 2015-03-12
Poltergeist for testing + emergency field

cmc333333:
* Added a generic JS script for filtering fields based based on a declarative syntax
* Emergency field sets all would-be approvers as observers, and starts the cart as approved
* Turbolinks fix

@phirefly @afeld @nickbristow if this gets merged, you'll need to have phantomjs to run tests

Looks like
screen shot 2015-03-12 at 12 00 28 pm
screen shot 2015-03-17 at 12 14 23 am
screen shot 2015-03-17 at 12 14 31 am


cmc333333:
There are some ugly hacks in here to account for SimpleFormObject turning everything into a string. I'll work on separating that model out next

README.md
```diff
@@ -17,6 +17,7 @@ To get the database and tests running:
 
 1. Start PostgreSQL. You will need to be able to create databases; set the
    `DB_`* variables in your environment if needed.
+1. Install [PhantomJS](http://phantomjs.org/download.html) and have it in your PATH
```
afeld: Worth adding to [the bootstrap script](https://github.com/18F/C2/blob/master/script/bootstrap) instead?

app/assets/javascripts/filter.js
```diff
@@ -0,0 +1,54 @@
...
+  //  Add controls
+  $.each(filters, function(key, filter) {
+    filter.hideAll();
+    filter.$('input:radio[data-filter-control=' + key + ']').each(
+        function(idx, control) {
+          filter.addRadio($(control));
+        }
+    );
+  });
+});
```
afeld: Fancy! Might be worth some JS unit tests sooner or later.

afeld: Also, this would be a lot more compact in CoffeeScript (which we have support for), unless you specifically prefer raw JS.

app/assets/stylesheets/common/static.scss
```diff
@@ -15,3 +15,7 @@ h3 {
   border: 1px solid #333399;
   width: 30em;
 }
+
+[aria-hidden=true] {
```
afeld: What's "aria"?

afeld: Derp, nevermind. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA

app/assets/javascripts/filter.js
```diff
@@ -0,0 +1,54 @@
...
+  });
+  //  Add controls
+  $.each(filters, function(key, filter) {
+    filter.hideAll();
+    filter.$('input:radio[data-filter-control=' + key + ']').each(
+        function(idx, control) {
+          filter.addRadio($(control));
+        }
+    );
+  });
```
afeld: How about moving these to named methods?

app/models/ncr/proposal_form.rb
```diff
@@ -92,12 +89,33 @@ def set_props_on(cart)
         vendor: self.vendor,
         not_to_exceed: self.not_to_exceed,
         building_number: self.building_number,
-        rwa_number: self.rwa_number,
         office: self.office
       )
+      case self.expense_type
+        when 'BA61'
+          # Hack to account for SimpleFormObject bugs
+          cart.set_props(emergency: self.emergency != false && self.emergency != "0")
```
afeld: Is the intention to store the string, or an actual boolean? Anyway, I would pull this out to an `emergency_bool` method (or something).

http://stackoverflow.com/questions/8119970/string-true-and-false-to-boolean#comment32650264_21042466

app/models/ncr/proposal_form.rb
```diff
@@ -92,12 +89,33 @@ def set_props_on(cart)
         vendor: self.vendor,
         not_to_exceed: self.not_to_exceed,
         building_number: self.building_number,
-        rwa_number: self.rwa_number,
         office: self.office
       )
+      case self.expense_type
+        when 'BA61'
+          # Hack to account for SimpleFormObject bugs
```
afeld: Can you elaborate (in the comment)?

app/models/ncr/proposal_form.rb
```diff
@@ -92,12 +89,33 @@ def set_props_on(cart)
...
+          cart.set_props(emergency: self.emergency != false && self.emergency != "0")
+        when 'BA80'
+          cart.set_props(rwa_number: self.rwa_number)
+      end
       cart.set_requester(self.requester)
     end
 
+    def add_approvals_on(cart)
+      # Hack to account for SimpleFormObject bugs
+      if self.emergency == true || self.emergency == "1"
```
afeld: Ditto comments above.

spec/features/ncr_proposal_spec.rb
```diff
@@ -93,6 +92,18 @@
...
+    it "hides fields based on expense", :js => true do
+      visit '/ncr/proposals/new'
+      expect(find_field("RWA Number", visible: false).visible?).to be false
+      expect(find_field("emergency", visible: false).visible?).to be false
+      choose 'BA61'
+      expect(find_field("RWA Number", visible: false).visible?).to be false
+      expect(find_field("emergency", visible: false).visible?).to be true
+      choose 'BA80'
+      expect(find_field("RWA Number", visible: false).visible?).to be true
+      expect(find_field("emergency", visible: false).visible?).to be false
```
afeld: Minor: instead of being tied to the implementation (the fields being hidden rather than removed), thoughts about checking whether it's present at all or not?

```ruby
visit '/ncr/proposals/new'
expect(find_field("RWA Number")).to be_nil
expect(find_field("emergency")).to be_visible
# etc.
```

app/views/ncr/proposals/form.html.erb
```diff
@@ -9,21 +9,37 @@
...
+            <%= f.radio_button :expense_type, "BA61",
+                               :"data-filter-control" => "expense-type" %>
+            BA61
+          <% end %>
+        </div>
+        <div class="form-group" data-filter-key="expense-type"
+                                data-filter-value="BA61">
+          <%= f.label :emergency do %>
+            <%= f.check_box :emergency %>
+            I received a verbal NTP to address this emergency
```
afeld: They will know what an "NTP" is?

spec/rails_helper.rb
```diff
@@ -70,6 +70,11 @@
   end
 
   config.before(:each) do
+    if Capybara.current_driver == :rack_test
+      DatabaseCleaner.strategy = :transaction
+    else
+      DatabaseCleaner.strategy = :truncation
+    end
```
afeld: Mind linking to the relevant documentation?

spec/features/ncr_proposal_spec.rb
```diff
@@ -169,5 +180,29 @@
...
+      }.to change { Cart.count }.from(0).to(1)
+
+      expect(page).to have_content("Proposal submitted")
+      expect(current_path).to eq("/carts/#{Cart.last.id}")
+      expect(page).to have_content("0 of 0 approved")
+
+      cart = Cart.last
+      expect(cart.getProp(:emergency)).to eq(true)
+      expect(cart.approved?).to eq(true)
+    end
```
afeld: This could be a rabbit hole (of negative assertion testing), but might be worth adding a test to verify that "if they select BA61, mark it as an 'emergency', then select BA80, the Cart should not be marked as emergency, and the Proposal should be 'pending'."

afeld:
Sorry for the slow turnaround on this.

> Emergency field sets all would-be approvers as observers

I know we talked about it a bit before, but this is preferable to setting them as approvers in the `approved` state? Seems like the latter might be preferable for reporting purposes.

> Turbolinks fix

For the record, I am fine with ditching Turbolinks as soon as they cause more trouble than they are worth.

---

screen shot 2015-03-16 at 2 32 13 pm

Thinking the new field should appear *below* the BA80 radio button, or at least be indented or something so that the visual tie between the radio buttons isn't broken. @nickbristow Curious to hear your thoughts.

app/models/ncr/proposal_form.rb
cmc333333: Good call

spec/features/ncr_proposal_spec.rb
cmc333333: Makes sense to me

README.md
cmc333333: I think they are pretty different instructions for installing with different OSes. Do you think it's worth the extra logic?

app/assets/javascripts/filter.js
cmc333333: For the time being, this is tested indirectly through poltergeist, but I agree.

I've not used coffeescript, but I'll try it out.

app/assets/javascripts/filter.js
cmc333333: Looks like a copy-paste error, too!

app/views/ncr/proposals/form.html.erb
cmc333333: This is the text from the client -- I have no idea what NTP means ;)

spec/features/ncr_proposal_spec.rb
cmc333333: That's a great test!

cmc333333:
> I know we talked about it a bit before, but this is preferable to setting them as approvers in the approved state? Seems like the latter might be preferable for reporting purposes.

I think having the system say that they approved the request is misleading, so I went with the simpler, observer approach. Until we have evidence that that's incorrect, I don't think it should change? We'll understand the needs more when we get to reporting, I think.

> For the record, I am fine with ditching Turbolinks as soon as they cause more trouble than they are worth.

Doesn't matter to me; if we keep them, we just need to implement them correctly.

> Thinking the new field should appear below the BA80 radio button, or at least be indented or something so that the visual tie between the radio buttons isn't broken.

The field placement was a direct result of this mockup: 

Indentation seems like a good idea

phirefly:
:+1: observer treatment

README.md
afeld: No worries then.

app/assets/javascripts/filter.js
afeld: Oh, did you copy this from somewhere? Link to the original source, please!

app/assets/javascripts/filter.js
afeld: Not a blocker! but check out http://coffeescript.org/#classes specifically.

afeld:
@brethauer Thoughts about the checkbox placement?

app/assets/javascripts/filter.js
cmc333333: Nah, the last chunks of code are just duplicated. JS has silently ignored them...

cmc333333:
I believe all the comments have been addressed; also updated screen shots.

app/assets/javascripts/.jshintrc
```diff
@@ -0,0 +1,27 @@
...
+  "validthis": true,
+  "boss": true,
+  "browser": true,
+  "globalstrict": true,
+  "sub": true,
+
+  "globals": {
+    "$": true
+  }
+}
```
afeld: TODO: hook up JSHint in our Travis build.

afeld: ...slash [coffee-jshint](https://github.com/Clever/coffee-jshint) or [coffeelint](http://www.coffeelint.org/).

app/assets/javascripts/filter.js.coffee
```diff
@@ -0,0 +1,37 @@
...
+        filters[key] = new Filter($scope, key)
+    filters
+
+$ ->
+  #  @todo - better scope
+  $scope = $(document)
+  filters = Filter.generateIn($scope)
+  for key, filter of filters
+    filter.hideAll()
+    filter.addRadios()
```
afeld: Looks good! What did you think of CoffeeScript?

cmc333333: I like that it's more concise, but I'd want a bit more experience before evaluating it.

spec/rails_helper.rb
afeld: Bump!

afeld:
Just the one open comment – otherwise, :ship:

cmc333333:
Documentation for the capybara config is a few lines above -- check out the latest diff rather than the inline version (which is outdated)

PR #242, opened 2015-03-12
18f purchase form

nickbristow:
Adds 18f form at gsa18f/proposals/new


app/controllers/Gsa18f/proposals_controller.rb
```diff
@@ -0,0 +1,86 @@
+module Gsa18f
+  class ProposalsController < ApplicationController
+    before_filter :authenticate_user!
+    before_filter :redirect_if_cart_cant_be_edited, only: [:edit, :update]
+
+    def new
+      @proposal_form = Gsa18f::ProposalForm.new
+      @form_url, @form_method_18f = {action: "create"}, "post"
```
afeld: Minor style note: can you set these two variables on separate lines?

afeld:
Replaces https://github.com/18F/C2/pull/231. Will continue inline review tonight or in the morning, but:

* [ ] Make sure all new code is tested – opening `coverage/index.html` after running the test suite will help

app/controllers/Gsa18f/proposals_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+      current_user.last_requested_cart
+    end
+
+    def last_approvers
+      last_cart.try(:approvers)
+    end
+
+    def suggested_approver
+      last_approvers.try(:first)
+    end
```
afeld: These two methods aren't needed.

app/decorators/cart_decorator.rb
```diff
@@ -4,6 +4,8 @@ class CartDecorator < Draper::Decorator
   def total_price
     if self.getProp('origin') == 'ncr'
       self.getProp('amount').to_f
+    elsif self.getProp('origin') == 'gsa18f'
+      self.getProp('cost_per_unit').to_f * self.getProp('quantity').to_f
     else
```
afeld: Mind switching to a `case` statement?

http://ruby-doc.org/docs/keywords/1.9/Object.html#method-i-case

app/models/gsa18f/proposal_form.rb
```diff
@@ -0,0 +1,105 @@
+#module Ncr
```
afeld: Leftover comment.

app/models/gsa18f/proposal_form.rb
```diff
@@ -0,0 +1,105 @@
...
+  DATA = YAML.load_file("#{Rails.root}/config/data/18f.yaml")
+
+  class ProposalForm
+    include SimpleFormObject
+
+    URGENCY = DATA['URGENCY']
+    OFFICES = DATA['OFFICES']
+    
+    #18f additions
+    #attribute :approver_email, :text
```
afeld: Extra comments.

app/models/gsa18f/proposal_form.rb
```diff
@@ -0,0 +1,105 @@
...
+    attribute :requester, :user
+    attribute :office, :string
+    attribute :justification, :text
+    attribute :link_to_product, :string
+    attribute :quantity, :decimal
+    attribute :date_requested, :string
+    attribute :urgency, :string
```
afeld: Thinking we should store the index (integer) instead...

app/models/gsa18f/proposal_form.rb
```diff
@@ -0,0 +1,105 @@
...
+    def create_cart
+      cart = Cart.new(
+        name: self.product_name_and_description,
+        proposal_attributes: {flow: 'linear'}
+      )
+      if cart.save
+        self.set_props_on(cart)
+        self.approver_emails.each do |email|
+          cart.add_approver(email)
+        end
```
afeld: Can we simplify this to `cart.add_approver('Richard.L.Miller@gsa.gov')`?

afeld: Also, maybe move that into a method so it can be shared with `update_cart`.

app/models/gsa18f/proposal_form.rb
```diff
@@ -0,0 +1,105 @@
...
+        Dispatcher.deliver_new_cart_emails(cart)
+      end
+
+      cart
+    end
+
+    def update_cart(cart)
+      cart.name = self.product_name_and_description
+      if cart.save
+        cart.approver_approvals.destroy_all
```
afeld: Is there a benefit to destroying rather than `restart!`ing them?

/cc @cmc333333

app/models/gsa18f/proposal_form.rb
```diff
@@ -0,0 +1,105 @@
...
+        # @todo: do we actually want to clear all properties?
+        cart.clear_props!
+        self.set_props_on(cart)
+        self.approver_emails.each do |email|
+          cart.add_approver(email)
+        end
+        cart.restart!
```
afeld: Wondering how much of the logic from above we could move into the `restart` action on the model...

/cc @cmc333333

app/assets/stylesheets/webapp/ncr.scss
```diff
@@ -7,3 +7,13 @@
     margin-bottom: 20px;
   }
 }
+
+.controller-gsa18f-proposals {
+  .content.container {
```
cmc333333: Do we think these two (`controller-ncr-proposals` and `controller-gsa18f-proposals`) are going to diverge much? If not, let's just use a single `proposal-form` or similar class

app/models/gsa18f/proposal_form.rb
```diff
@@ -0,0 +1,105 @@
...
+    #   ENV['NCR_FINANCE_APPROVER_EMAIL'] || 'communicart.ofm.approver@gmail.com'
+    # end
+
+    # def requires_finance_approval?
+    #   self.expense_type == 'BA61'
+    # end
+
+    def approver_emails
+      emails = ['Richard.L.Miller@gsa.gov']
+      emails
```
cmc333333: Best to include real addies in the code base. Take the `ENV` approach above?

app/models/gsa18f/proposal_form.rb
cmc333333: On NCR it's required as the user could select different addies when restarting the form. That's less a concern here (as the approvers are constant)

app/models/gsa18f/proposal_form.rb
cmc333333: I don't think much, I'm afraid, at least not until `restart` logic is delegated to client-specific models.

app/models/gsa18f/proposal_form.rb
```diff
@@ -0,0 +1,105 @@
...
+    end
+
+    def self.from_cart(cart)
+      relevant_properties = self._attributes.map(&:name).map(&:to_s)
+      properties = cart.deserialized_properties.select{
+        |key, val| relevant_properties.include? key
+      }
+      form = self.new(properties)
+      form.approver_email = cart.ordered_approvals.first.user.email_address
+      form.description = cart.name
```
cmc333333: These two fields don't exist on the 18f model. I think you want to get rid of the first and use `product_name_and_description` for the second?

app/models/gsa18f/proposal_form.rb
```diff
@@ -0,0 +1,87 @@
...
+      greater_than_or_equal_to: 0,
+      less_than_or_equal_to: 3000
+    }
+    validates :quantity, numericality: {
+      greater_than_or_equal_to: 1
+    }
+    validates :product_name_and_description, presence: true
+
+    def set_approver_on(cart)
+      cart.add_approver(ENV['GSA18F_APPROVER_EMAIL'] || 'Richard.L.Miller@gsa.gov')
```
cmc333333: Let's not have real email addies in the code base; instead, make this a dummy like `18fapprover@gsa.gov`

app/controllers/gsa18f/proposals_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+        flash[:error] = @proposal_form.errors.full_messages
+        render 'form'
+      end
+    end
+
+    protected
+
+    def last_cart
+      current_user.last_requested_cart
+    end
```
afeld: Can be removed.

app/controllers/gsa18f/proposals_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      @cart ||= Cart.find(params[:id])
+    end
+
+    def redirect_if_cart_cant_be_edited
+      if self.cart.approved?
+        redirect_to new_gsa18f_proposal_path, :alert => "That proposal's already approved. New proposal?"
+      elsif self.cart.requester != current_user
+        redirect_to new_gsa18f_proposal_path, :alert => 'You cannot restart that proposal'
+      end
+    end
```
afeld: TODO: do something more generic around Proposals across all the controllers.

app/models/gsa18f/proposal_form.rb
```diff
@@ -0,0 +1,87 @@
...
+      greater_than_or_equal_to: 0,
+      less_than_or_equal_to: 3000
+    }
+    validates :quantity, numericality: {
+      greater_than_or_equal_to: 1
+    }
+    validates :product_name_and_description, presence: true
+
+    def set_approver_on(cart)
+      cart.add_approver(ENV['GSA18F_APPROVER_EMAIL'] || '18fapprover@gsa.gov')
```
afeld: Minor: mind moving the email part to an `#approver_email` method?

app/models/gsa18f/proposal_form.rb
afeld: That might be something we want to revisit in the future, but ok for now.

app/views/gsa18f/proposals/form.html.erb
```diff
@@ -0,0 +1,75 @@
...
+    </div>
+    <!-- Justification -Text Area- -->
+    <div class="form-group">
+      <%= f.label :justification %>
+      <%= f.text_area :justification, class: 'form-control' %>
+    </div>
+    <!-- Link to Product -->
+    <div class="form-group">
+      <%= f.label :link_to_product %>
+      <p> Search GSA Advantage first; then Amazon; then other vendor. Paste link here.</p>
```
afeld: Mind linking to [GSA Advantage](https://www.gsaadvantage.gov/)?

app/models/gsa18f/proposal_form.rb
```diff
@@ -0,0 +1,87 @@
...
+    URGENCY = DATA['URGENCY']
+    OFFICES = DATA['OFFICES']
+    
+    attribute :requester, :user
+    attribute :office, :string
+    attribute :justification, :text
+    attribute :link_to_product, :string
+    attribute :quantity, :integer
+    attribute :date_requested, :datetime
+    attribute :urgency, :string
```
afeld: I thought I left a comment about it before, but I think we should store the urgency as an int.

app/views/gsa18f/proposals/form.html.erb
```diff
@@ -0,0 +1,75 @@
...
+      <%= f.label :urgency %>
+      <%= f.collection_select(
+            :urgency, Gsa18f::ProposalForm::URGENCY, :to_s, :to_s,
+            {include_blank: true}, {class: 'form-control'}) %>
+    </div>
+    <!-- Additional Info -->
+    <div class="form-group">
+      <%= f.label :additional_info, 'Please include any additional information you feel we need to correctly execute the purchase' %>
+      <%= f.text_area :additional_info, class: 'form-control' %>
+    </div>
```
afeld: Not your fault, but kind of irks me that there is so much duplication between the markup of each field. We should consider making a helper method, or switching to something like [simple_form](https://github.com/plataformatec/simple_form) down the road.

spec/models/gsa18f/proposal_form_spec.rb
```diff
@@ -0,0 +1,19 @@
+describe Gsa18f::ProposalForm do
+  describe '#create_cart' do
+    def approver_emails(form)
+      cart = form.create_cart
+      approvals = cart.ordered_approvals
+      approvals.map {|a| a.user.email_address }
+    end
+
+    it "adds Ric Miller as approver email" do
```
afeld: As @cmc333333 suggested, we should take out the specific name.

app/views/gsa18f/proposals/form.html.erb
```diff
@@ -0,0 +1,75 @@
...
+    <div class="form-group">
+      <%= f.hidden_field :origin, value: 'gsa18f' %>
+    </div>
```
afeld: Shouldn't be necessary...we should assign the `origin` explicitly in the ProposalForm.

app/models/gsa18f/proposal_form.rb
nickbristow: Ah, i thought you were talking about something else with that comment, my bad @afeld

cmc333333:
:+1: will merge in a few hours to give Aidan time to review

app/models/gsa18f/proposal_form.rb
```diff
@@ -0,0 +1,100 @@
...
+      less_than_or_equal_to: 3000
+    }
+    validates :quantity, numericality: {
+      greater_than_or_equal_to: 1
+    }
+    validates :product_name_and_description, presence: true
+
+    def ordered_urgencies
+      URGENCY.each_with_index().to_a
+    end
```
afeld: Minor: this should be a class method.

afeld:
Just the one minor note – otherwise, :shipit:!

app/models/gsa18f/proposal_form.rb
cmc333333: This can probably be deleted, hey?

PR #243, opened 2015-03-13
[Delivers #90311894] associate Approvals w/ Proposals

afeld:
Includes #239, so here's the effective diff:

https://github.com/18F/C2/compare/api-token-association...90311894_proposal_approvals


db/migrate/20150311080458_associate_approvals_with_proposals.rb
```diff
@@ -0,0 +1,33 @@
...
+      # http://stackoverflow.com/a/6257679/358804
+
+      dir.up do
+        execute <<-SQL
+          UPDATE approvals
+          SET proposal_id = proposals.id
+          FROM carts
+          -- get associated proposal through the cart
+          LEFT OUTER JOIN proposals ON carts.proposal_id = proposals.id
+          WHERE approvals.cart_id = carts.id;
```
cmc333333: I think this would have been a bit simpler, but not worth changing now:
```sql
UPDATE approvals
SET proposal_id = carts.proposal_id
FROM carts
WHERE approvals.cart_id = carts.id;
```

Could also do a subselect (the RDMS should make the query more efficient than it appears)
```sql
UPDATE approvals
SET proposal_id = (SELECT proposal_id FROM carts WHERE cart_id = carts.id);
```

cmc333333:
The diff between branches made sense to me. Mind merging upstream? That'll give us a bit more time if someone else wants to take a look, too.

afeld:
Merged `master`, so should be easier to look at now.

cmc333333:
One small step...

db/migrate/20150311080458_associate_approvals_with_proposals.rb
afeld: I spent waaay too long getting this to work – will definitely bug you for advice on the next one :wink:

PR #244, opened 2015-03-16
[#89129918] Removed user_id from approval request

phirefly:


afeld:
Wow, I left this out of https://github.com/18F/C2/pull/233? :laughing:

PR #245, opened 2015-03-17
Model for NCR Work Orders + Property list

cmc333333:
This is the start of replacing the NCR proposal form with a model. Ultimately, this effort should result in standardized interfaces for client-specific additions (including models, logic, templates, etc.) This PR takes a step in that direction

* Adds a polymorphic field on the Proposals model
* Adds a model for NCR Work Orders
* Starts to define an interface for these client-specific data types; adds it to `Cart` for the time being
* Update templates to display fields based on Proposals rather than Carts


app/models/cart.rb
```diff
@@ -226,4 +226,19 @@ def parallel?
...
+    # Some fields aren't meant for the clients' eyes
+    exclusions = ['origin', 'contractingVehicle', 'location', 'configType']
+    props = self.deserialized_properties.to_a
+    props = props.select {|k,v| !exclusions.include? k }
+    props = props.map {|key,value|
+      [I18n.t('helpers.label.cart.' + key, default: key.underscore.humanize),
+       value]
+    }
+    props
+  end
```
afeld: Mind splitting up into a handful of constants/methods?

app/models/ncr/work_order.rb
```diff
@@ -0,0 +1,42 @@
+module Ncr
+  # Make sure all table names use 'ncr_XXX'
+  def self.table_name_prefix
+    'ncr_'
+  end
```
afeld: Is this used?

app/models/ncr/work_order.rb
```diff
@@ -0,0 +1,42 @@
...
+      attributes = self.attribute_names
+      case self.expense_type
+      when "BA61"
+        exclusions = exclusions.push("rwa_number")
+      when "BA80"
+        exclusions = exclusions.push("emergency")
+      end
+      attributes = attributes - exclusions
+      attributes.map{|key| [WorkOrder.human_attribute_name(key), self[key]]}
+    end
```
afeld: Minor: mind splitting to smaller methods?

app/models/proposal.rb
```diff
@@ -4,6 +4,7 @@ class Proposal < ActiveRecord::Base
   workflow_column :status
 
   has_one :cart
+  belongs_to :clientdata, polymorphic: true
```
afeld: Thinking this should have an underscore.

spec/models/ncr/work_order_spec.rb
```diff
@@ -0,0 +1,38 @@
+describe Ncr::WorkOrder do
+  describe 'fields_for_display' do
+    it "shows BA61 fields" do
+      wo = Ncr::WorkOrder.new(
+        amount: 1000, expense_type: "BA61", vendor: "Some Vend", 
+        not_to_exceed: false, emergency: true, rwa_number: "RWWAAA #",
+        building_number: Ncr::WorkOrder::BUILDING_NUMBERS[0],
+        office: Ncr::WorkOrder::OFFICES[0])
```
afeld: Mind making a factory?

https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md

spec/models/ncr/work_order_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      fields = wo.fields_for_display
+      expect(fields).to include(["Amount", 1000])
+      expect(fields).to include(["Expense type", "BA61"])
+      expect(fields).to include(["Vendor", "Some Vend"])
+      expect(fields).to include(["Not to exceed", false])
+      expect(fields).to include(["Emergency", true])
+      expect(fields).to include(["Building number",
+                                 Ncr::WorkOrder::BUILDING_NUMBERS[0]])
+      expect(fields).not_to include(["RWA Number", "RWWAAA #"])
+      expect(fields).to include(["Office", Ncr::WorkOrder::OFFICES[0]])
```
afeld: Maybe do something like

```ruby
expect(fields.sort). include([
  ["Amount", 1000],
  ["Building number", Ncr::WorkOrder::BUILDING_NUMBERS[0]],
  # ...
])
```

to also ensure there's no extra properties being included. Ditto below.

app/models/cart.rb
cmc333333: None of the steps in between are relevant outside of this method. How about more descriptive variable names?

spec/models/ncr/work_order_spec.rb
cmc333333: There's a factory, but I chose not to use it here since all of the fields are being tested. I can swap it out, though.

spec/models/ncr/work_order_spec.rb
cmc333333: Makes sense

app/models/proposal.rb
cmc333333: Can do

app/models/ncr/work_order.rb
cmc333333: Will give it another shot

app/models/cart.rb
afeld: Even so, I think it's worthwhile for grouping/abstracting/decluttering the functionality. The method as written isn't _bad_ – just seems like it would be easy to split up.

spec/models/ncr/work_order_spec.rb
afeld: Oh, fair enough. No, this is fine then.

app/models/ncr/work_order.rb
cmc333333: Yep. Without it, active record would look at the `work_order` table. Adding a prefix prevents collisions, but we need to notify active record that that's what's happening.

afeld: :+1:

cmc333333:
Updated to address your comments @afeld , thanks!

PR #248, opened 2015-03-20
Front end and final hookups for NCR models

cmc333333:
* New controller, view, and url for NCR `work_order` (instead of `proposal`)
* Move `public_identifier` to client data indirection
* Moved constants into the Ncr namespace
* Cleaned up the ignored_fields concept: instead focuses on what fields are visible/editable
* Updates tests to correspond to test the new versions of all of the above
* Added proposal to several templates
* Removed several "are you an ncr cart?" checks
* Migration for exisiting data (and reverting)


cmc333333:
Sorry for the large diff - a good chunk of it is just renaming and moving things. In terms of logic, the controller and model act slightly differently and the migration would be worth reviewing.

app/controllers/ncr/work_orders_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+        errors = errors << "Description is required"
+      end
+      if !@approver_email.is_a?(String) || @approver_email.strip().empty?
+        errors = errors << "Approver email is required"
+      end
+      if !@work_order.valid?
+        errors = errors + @work_order.errors.full_messages
+      end
+      errors
+    end
```
afeld: Mind making this a protected method? Just to distinguish that it's a helper method rather than a controller action.

app/controllers/ncr/work_orders_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+      @work_order = Ncr::WorkOrder.new
+      @approver_email = self.suggested_approver_email
+      @description = ""
+      render 'form'
+    end
+
+    def errors
+      # @TODO we can use a nested model once we get rid of the Cart requirement
+      errors = []
+      if !@description.is_a?(String) || @description.strip().empty?
```
afeld: Check out the [`blank?`](http://api.rubyonrails.org/classes/Object.html#method-i-blank-3F) method.

app/controllers/ncr/work_orders_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+
+    def new
+      @work_order = Ncr::WorkOrder.new
+      @approver_email = self.suggested_approver_email
+      @description = ""
+      render 'form'
+    end
+
+    def errors
+      # @TODO we can use a nested model once we get rid of the Cart requirement
```
afeld: Would be nice to move this logic into the `WorkOrder` model even before that... check out

http://guides.rubyonrails.org/active_record_validations.html#custom-methods

and I'm happy to talk through it or pair if you like.

afeld: e.g.

```ruby
  validate :description_must_be_present

  # ...

  def description_must_be_present
    if self.description.blank?
      self.errors.add(:base, "must have a description")
    end
  end
end
```

app/controllers/ncr/work_orders_controller.rb
```diff
@@ -0,0 +1,99 @@
+module Ncr
+  class WorkOrdersController < ApplicationController
+    before_filter :authenticate_user!
+    before_filter :redirect_if_cart_cant_be_edited, only: [:edit, :update]
+
+    def new
+      @work_order = Ncr::WorkOrder.new
+      @approver_email = self.suggested_approver_email
+      @description = ""
```
afeld: Don't think this line is necessary...?

app/models/ncr/work_order.rb
```diff
@@ -6,15 +6,17 @@ def self.table_name_prefix
...
   DATA = YAML.load_file("#{Rails.root}/config/data/ncr.yaml")
 
+  EXPENSE_TYPES = %w(BA61 BA80)
+
+  BUILDING_NUMBERS = DATA['BUILDING_NUMBERS']
+  OFFICES = DATA['OFFICES']
+
   class WorkOrder < ActiveRecord::Base
     # In practice, each work order only has one proposal
     has_many :proposals, as: :client_data
```
afeld: Wait, what? Why is this a `has_many`? Sorry if I missed this before.

app/views/ncr/work_orders/form.html.erb
```diff
@@ -2,10 +2,10 @@
   <h4>National Capital Region</h4>
   <h3>Purchase Card Approval Request</h3>
 
-  <%= form_for @proposal_form, url: @form_url, method: @form_method do |f| %>
+  <%= form_for @work_order do |f| %>
     <div class="form-group">
-      <%= f.label :description %>
-      <%= f.text_area :description, class: 'form-control' %>
+      <%= label_tag :description %>
+      <%= text_area_tag :description, @description, class: 'form-control' %>
```
afeld: How about making a `WorkOrder#description` method? The fewer instance variables that need to be passed around, the better.

app/models/ncr/work_order.rb
cmc333333: Proposal has a `client_data_type` and a `client_data_id`, via the `belongs_to` relationship. The `has_many` here is just the other side of that. Looks like I can do `has_one` instead? Does that make sense?

app/models/cart.rb
```diff
@@ -238,4 +220,25 @@ def fields_for_display
...
       [label, value] }
   end
+  def client
+    # TODO set the origin
+    if self.origin.blank?
+      "gsa_advantage"
+    else
+      self.origin
+    end
+  end
```
afeld: Can we combine this with the `#origin` method?

app/models/cart.rb
```diff
@@ -238,4 +220,25 @@ def fields_for_display
...
+  # @todo - the method name (e.g. :external_id) should live on a "client"
+  # model
+  def public_identifier
+    self.external_id
+  end
+  def total_price
+    case self.getProp('origin')
+    when 'gsa18f'
```
afeld: Mind changing to `if self.origin == 'gsa18f'`? Maybe better yet, add a `#gsa18f?` method.

app/models/ncr/work_order.rb
afeld: Yep, `has_one` would be the way to go. Thanks!

app/models/ncr/work_order.rb
```diff
@@ -24,21 +26,81 @@ class WorkOrder < ActiveRecord::Base
...
-    def ignore_fields
-      case self.expense_type
+    def set_defaults
+      self.not_to_exceed ||= false
+      self.emergency ||= false
+    end
+
+    # @todo - this is an awkward dance due to the lingering Cart model. Remove
+    # that dependence
+    def create_cart(approver_email, description, requester)
```
afeld: `#create_ASSOCIATION_NAME` is reserved in ActiveRecord, so might be confusing, even if we're not using the other form at the moment. Mind changing the name?

app/models/ncr/work_order.rb
```diff
@@ -24,21 +26,81 @@ class WorkOrder < ActiveRecord::Base
...
+      cart.restart!
+      cart
+    end
+
+    def system_approvers
+      emails = [ENV['NCR_BUDGET_APPROVER_EMAIL'] ||
+                'communicart.budget.approver@gmail.com']
+      if self.expense_type == 'BA61'
+        emails << (ENV['NCR_FINANCE_APPROVER_EMAIL'] ||
+                   'communicart.ofm.approver@gmail.com')
```
afeld: Mind making `#budget_approver` and `#finance_approver` methods? Just to make this a bit easier to understand.

app/models/ncr/work_order.rb
afeld: I'd also add a `has_one :cart, through: :proposal`.

app/models/ncr/work_order.rb
```diff
@@ -24,21 +26,81 @@ class WorkOrder < ActiveRecord::Base
...
+      else
+        emails
+      end
+    end
+
+    def add_approvals_on(cart, approver_email)
```
afeld: The WorkOrder already has the association, so shouldn't need the Cart to be passed in.

app/models/ncr/work_order.rb
```diff
@@ -24,21 +26,81 @@ class WorkOrder < ActiveRecord::Base
...
       when "BA80"
-        ["emergency"]
+        fields + [:rwa_number]
       end
     end
 
     # Methods for Client Data interface
     def fields_for_display
-      attributes = self.attribute_names - ["id"] - self.ignore_fields
+      attributes = Ncr::WorkOrder.relevant_fields(self.expense_type)
```
afeld: Minor: mind moving this into a `#relevant_fields` instance method that calls the class method?

app/models/ncr/work_order.rb
```diff
@@ -24,21 +26,81 @@ class WorkOrder < ActiveRecord::Base
...
       attributes.map{|key| [WorkOrder.human_attribute_name(key), self[key]]}
     end
+    def client
+      "ncr"
+    end
+    # @todo - this is pretty ugly
+    def public_identifier
+      self.proposals[0].cart.id
+    end
```
afeld: Thinking that we should use the `Proposal#id`, actually. That might be a separate story though.

config/routes.rb
```diff
@@ -20,7 +20,7 @@
   end
 
   namespace :ncr do
-    resources :proposals
+    resources :work_orders
   end
```
afeld: Should we worry about redirecting from the old URLs? My guess is no, but wanted to bring it up.

/cc @phirefly

db/migrate/20150319212424_migrate_ncr_to_models.rb
```diff
@@ -0,0 +1,69 @@
...
+      Hash[self.properties.collect{|p| [p.property, YAML::load(p.value)]}]
+    end
+  end
+  class Proposal < ActiveRecord::Base
+    has_one :cart
+  end
+  class NcrWorkOrder < ActiveRecord::Base
+    has_many :proposals, -> { where client_data_type: 'Ncr::WorkOrder' },
+              foreign_key: "client_data_id"
+  end
```
afeld: Minor: I've been prefixing the migration-only models w/ `Temp` so that I don't accidentally reference the normal ones.

lib/dispatcher.rb
```diff
@@ -43,7 +43,8 @@ def self.initialize_dispatcher(cart)
     when 'parallel'
       ParallelDispatcher.new
     when 'linear'
-      if cart.ncr?
+      # @todo: dynamic dispatch for selection
+      if cart.proposal.client_data_legacy.client == "ncr"
```
afeld: Why this instead of the old way?

afeld:
This PR is a beast :sweat_smile: 

Have you done some manual testing on this? Just want to make sure there isn't some thing we're missing in the tests, e.g. the notification emails being sent without showing any properties.

Also, general Ruby formatting note: mind leaving a blank newline between method definitions?

Kudos for all the effort on this!

lib/dispatcher.rb
cmc333333: Ideally, there'd be zero "are you an instance of ncr?" checks. These should all be implicit based on dynamic dispatch, file existence, etc.

app/controllers/ncr/work_orders_controller.rb
cmc333333: Sure

app/controllers/ncr/work_orders_controller.rb
cmc333333: Sounds good, thanks!

app/controllers/ncr/work_orders_controller.rb
cmc333333: Sure. There's some ugly gymnastics around getting to the cart, but should be possible.

app/controllers/ncr/work_orders_controller.rb
cmc333333: Probably not; will remove

app/controllers/ncr/work_orders_controller.rb
cmc333333: I take that back. The cart won't exist by the time the work order is validated. Are you suggesting NCR descriptions shouldn't rely on the cart?

app/models/ncr/work_order.rb
cmc333333: The reliance on cart is going away, right? Making it painful to connect a work order to a cart is a way to speed up that process.

app/models/cart.rb
cmc333333: Can do

app/models/cart.rb
cmc333333: I can do the first. I think the gsa18f proposal model will go away before the latter would be helpful.

app/models/ncr/work_order.rb
cmc333333: Sure thing

app/models/ncr/work_order.rb
cmc333333: Don't you think that's redundant given the env variable names?

app/models/ncr/work_order.rb
cmc333333: Sure

app/models/ncr/work_order.rb
cmc333333: Sure

app/models/ncr/work_order.rb
cmc333333: Let's delay that. I'm not super confident that the public identifier is *used* outside of gsa advantage.

config/routes.rb
cmc333333: Can also keep the old urls if that's desirable. It was easier to have a separate system so both could be available during development, but we could switch them back now

db/migrate/20150319212424_migrate_ncr_to_models.rb
cmc333333: Makes sense

cmc333333:
Thanks for looking at it! Yep, I tested manually, but I can take some screen shots in case there's something subtle.

app/controllers/ncr/work_orders_controller.rb
cmc333333: Updated in the PR (though it's not showing in the inline)

app/views/ncr/work_orders/form.html.erb
cmc333333: This got swallowed in an outdated diff:

Do you think description should not depend on cart? I can definitely add it as its own field and add an indirection layer as with `total_amount`

afeld:
No worries, I trust ya :smile:

app/controllers/ncr/work_orders_controller.rb
afeld: Ah oh, maybe more trouble than it's worth then.

app/models/ncr/work_order.rb
afeld: Haha true. I love the paradox of intentionally laying obstacles to force us to make the code better later.

app/views/ncr/work_orders/form.html.erb
afeld: I was thinking of a method that delegates to the Cart at least, which could then be swapped out for an attribute later.

app/models/ncr/work_order.rb
afeld: I was thinking

```ruby
def finance_approver_email
  ENV['NCR_FINANCE_APPROVER_EMAIL'] || 'communicart.ofm.approver@gmail.com'
end
```

Not a huge deal though.

app/models/ncr/work_order.rb
afeld: It's not used by the system in terms of parsing or anything at the moment, but Chris and others expressed interest in having a canonical request identifier that they can refer to.

config/routes.rb
afeld: Nah, don't worry about it.

app/views/ncr/work_orders/form.html.erb
afeld: e.g.

```ruby
  # ...
  def description
    self.cart.try(:description) || ''
  end
end
```

app/models/ncr/work_order.rb
cmc333333: Sure. Adding as protected methods

cmc333333:
I think everything's been addressed one way or another now. Also fixed a few references to external_id, which I don't think ever worked for NCR.

app/models/cart.rb
```diff
@@ -238,4 +217,22 @@ def fields_for_display
       EXCLUDE_FIELDS_FROM_DISPLAY.include? key}.map{ |key,value,label|
       [label, value] }
   end
+
+  def client
+    self.origin
+  end
```
afeld: Why do we need this _and_ `#origin`?

app/models/ncr/work_order.rb
```diff
@@ -24,21 +25,97 @@ class WorkOrder < ActiveRecord::Base
...
+      "ncr"
+    end
+
+    # @todo - this is pretty ugly
+    def public_identifier
+      self.proposal.cart.id
+    end
+
+    def total_price
+      self.amount or 0.0
```
afeld: Minor: use `||` for consistency.

afeld:
Just a couple minor things!

app/models/cart.rb
cmc333333: Origin doesn't exist on the ncr model. Are you saying you want the interface method renamed to `origin`?

afeld: Oh, I see. Nvm!

afeld: Actually, maybe just get rid of the `#origin` method and use `#client` instead.

cmc333333: Smart, can do

cmc333333:
Updated to remove the origin method and use `||``

PR #249, opened 2015-03-20
Use proposal_id in this migration so that it can be ran in order

cmc333333:
Two branches were developed in parallel. When applied sequentially, migrations was breaking. This patch fixes that, at least from a schema perspective. I've relatively confident that the logic is maintained, but if you could verify it's doing what you intended @afeld, that'd be wonderful.


afeld:
I appreciate it!

PR #250, opened 2015-03-20
have Cloud Foundry ignore files ignored by Git

afeld:
https://trello.com/c/MIRccDiT/22-have-cf-automatically-ignore-files-from-the-gitignore-by-default

/cc @dlapiduz @ozzyjohnson

### DIFF
--- a/.cfignore
+++ b/.cfignore
@@ -1,0 +1,1 @@
+.gitignore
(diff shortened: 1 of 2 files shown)


.cfignore
```diff
@@ -0,0 +1 @@
+.gitignore
```
afeld: This is a symlink (weird that it shows up like a normal file).

PR #251, opened 2015-03-23
Work order code

cmc333333:
Builds on #248 . The only new bit is in 9713196

Looks like:
screen shot 2015-03-23 at 10 37 27 am
screen shot 2015-03-23 at 10 36 56 am


afeld:
:shipit:

PR #254, opened 2015-03-24
88560300 expire token 1

phirefly:
This test was accidentally added to master and didn't fail because there was an undefined step. This commit will expire tokens when approving through the web app instead of an email link.

### DIFF
--- a/app/controllers/communicarts_controller.rb
+++ b/app/controllers/communicarts_controller.rb
@@ -21,8 +21,9 @@ class CommunicartsController < ApplicationController
   def approval_response
     cart = Cart.find(params[:cart_id]).decorate
     client_data = cart.proposal.client_data_legacy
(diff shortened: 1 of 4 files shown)


app/controllers/communicarts_controller.rb
```diff
@@ -21,8 +21,9 @@ def send_cart
   def approval_response
     cart = Cart.find(params[:cart_id]).decorate
     client_data = cart.proposal.client_data_legacy
-    approval = cart.approvals.find_by(user_id: user_id) 
-    
+    approval = cart.approvals.find_by(user_id: user_id)
+    @token ||= ApiToken.find_by(approval_id: approval.id)
```
afeld: Why is this necessary?

cmc333333:
Makes sense to me and patches up a small security hole

PR #255, opened 2015-03-24
[Delivers #90634128] Error message when approving your own cart

nickbristow:
Addresses https://www.pivotaltracker.com/story/show/90634128

### DIFF
--- a/app/controllers/communicarts_controller.rb
+++ b/app/controllers/communicarts_controller.rb
@@ -21,13 +21,18 @@ class CommunicartsController < ApplicationController
   def approval_response
     cart = Cart.find(params[:cart_id]).decorate
     client_data = cart.proposal.client_data_legacy
-    approval = cart.approvals.find_by(user_id: user_id)
+    approval = cart.approvals.approvable.find_by(user_id: user_id)
+    
+    if !approval
+      flash[:error] = "Sorry. You are not allowed to approve your own request."
+      redirect_to cart_path(cart)
+      return 
+    end 
+    
     @token ||= ApiToken.find_by(approval_id: approval.id)
     
     if !approval.pending?
       flash[:error] = "You have already logged a response for Cart #{client_data.public_identifier}"
-    elsif !approval.approvable?
-      flash[:error] = "Sorry. You are not allowed to approve your own request."
     else
       case params[:approver_action]
       when 'approve'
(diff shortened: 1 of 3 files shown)


app/controllers/communicarts_controller.rb
```diff
@@ -21,10 +21,12 @@ def send_cart
   def approval_response
     cart = Cart.find(params[:cart_id]).decorate
     client_data = cart.proposal.client_data_legacy
-    approval = cart.approvals.find_by(user_id: user_id) 
+    approval = cart.approvals.find_by(user_id: user_id)
```
afeld: Let's use `cart.approvals.approvable.find_by(user_id: user_id)` instead of the separate check below.

afeld: Though I suppose that wouldn't allow the error message for that specific case to be shown...though not sure that needs to handled separately from a "you aren't allowed to approve this" message.

cmc333333:
The so-called "raphy" case. Mind merging master? I defer to @afeld re `is_approvable?` vs `approvable?`

afeld:
> `is_approvable?` vs `approvable?`

The latter is more Ruby-ish.

app/controllers/communicarts_controller.rb
```diff
@@ -21,9 +21,16 @@ def send_cart
   def approval_response
     cart = Cart.find(params[:cart_id]).decorate
     client_data = cart.proposal.client_data_legacy
-    approval = cart.approvals.find_by(user_id: user_id)
+    approval = cart.approvals.approvable.find_by(user_id: user_id)
+    
+    if !approval
+      flash[:error] = "Sorry. You are not allowed to approve your own request."
```
afeld: Trying to think if there are other cases where the `approval` could be `nil`...if we want to show this message specifically, might be better to go back to the way you were handling it before. Sorry!

app/controllers/communicarts_controller.rb
```diff
@@ -21,9 +21,16 @@ def send_cart
...
+      redirect_to cart_path(cart)
+      return
```
afeld: I'm personally allergic to mid-method `return`s...mind using an `if`/`else` instead? @cmc333333 Curious to hear your thoughts.

PR #256, opened 2015-03-24
[Delivers 90512374]: Require comment, send email on comment

cmc333333:
Changes:
* Makes comment text a required field
* Implements comment emails via an event system

If the event system described holds water, we can use it to untangle the various "did X just happen? then send an email" systems. Instead, we can emit specific events and hook up observers which will send email as appropriate. I could imagine a per-instance configuration that only adds observers relevant to a specific client, but there are other complications with that.

I'm not a fan of the initializers system, but I think it's satisfactory for now.

I reused existing (but seemingly unused) comment emailing methods. Resulting email looks like:
screen shot 2015-03-24 at 5 03 15 pm


app/models/concerns/observable_model.rb
```diff
@@ -0,0 +1,13 @@
...
+  included do
+    after_create :send_create_event
+    # Add more as needed
+  end
+
+  def send_create_event
+    ActiveSupport::Notifications.instrument("#{self.class.model_name}.create",
+                                            self)
+  end
+end
```
afeld: Iiiinteresting. Do you like this better than https://github.com/rails/rails-observers?

app/views/communicart_mailer/comment_added_email.html.erb
```diff
@@ -1,16 +1,18 @@
...
+        Action need on Request 
+        #<%= cart.proposal.client_data_legacy.public_identifier %>
+        (<%= cart.name %>):
+      </p>
+    </td></tr>
+    <tr><td class="w-container html-email-message">
+      <%= @comment.comment_text %>
+    </td></tr>
+    <tr><td class="w-container html-email-message">
+      <%= link_to "View / Update Request", cart_url(cart, anchor: 'cart-comments'), class: 'form-button' %>
```
afeld: The requester is the only person who can "update" the request at the moment, right? Not a huge deal – we can iterate.

PR #257, opened 2015-03-25
[Delivers #89169198] split observers and requester from Approval model

afeld:
`Approval`s no longer have a `role` – there is an equivalent `Observation` model, and a `requester_id` on the `Proposal` directly. These were the minimum required changes...I'll do some other cleanup in a follow-up PR.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,7 @@
 source 'https://rubygems.org'
 
 gem 'acts_as_list'
+gem 'ar_outer_joins'
 gem 'autoprefixer-rails'
 gem 'awesome_print'
 gem 'bootstrap-sass', '~> 3.3.0'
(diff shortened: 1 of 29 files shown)


cmc333333:
Looks good to me. Relevant: master now include another instance of `observers` in `lib/dispatcher.rb`

Otherwise, I'd be happy to merge

afeld:
> master now include another instance of `observers` in `lib/dispatcher.rb`

Sorry, I don't follow? The model is called `Observation`, so there shouldn't be any naming conflict.

cmc333333:
Sorry, I meant `observing`. https://github.com/18F/C2/blob/master/lib/dispatcher.rb#L93 was added after you made your last commit, I think, so the tests aren't failing, but that line refers to a scope that's deleted in this PR.

afeld:
Updated!

cmc333333:
Makes sense to me, as does the Rspec tweak. @phirefly this may affect your PR. Will merge in the morning unless you see something else

afeld:
can haz merge?

cmc333333:
/me is defeated by the cute kitty

PR #260, opened 2015-03-25
89820292 resubmit without reapproving [do not merge]

phirefly:
Thought I'd post the WIP while working on tests in case anyone has feedback about the approach. 

Also have a question that I will post.

### DIFF
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -50,7 +50,8 @@ class Proposal < ActiveRecord::Base
   def restart
     # Note that none of the state machine's history is stored
     self.cart.api_tokens.update_all(expires_at: Time.now)
-    self.cart.approver_approvals.where.not(status: 'approved').each do |approval|
+    # self.cart.approver_approvals.where.not(status: 'approved').each do |approval|
+    self.cart.approver_approvals.each do |approval|
       approval.restart!
     end
     Dispatcher.deliver_new_cart_emails(self.cart)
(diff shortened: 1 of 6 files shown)


lib/dispatcher.rb
```diff
@@ -10,11 +10,18 @@ def email_observers(cart)
...
+    cart.approved_approvals.each do |approval|
+      CommunicartMailer.cart_notification_email(approval.user_email_address, approval, false).deliver
+    end
+  end
+
   def email_sent_confirmation(cart)
     CommunicartMailer.proposal_created_confirmation(cart).deliver
   end
 
   def deliver_new_cart_emails(cart)
```
phirefly: Question: This method doesn't appear to be sending new emails to approvers. With this newest change, it will send to 1) approvers that have already approved, 2) observers, and 3) the requester.

Am I missing something? Perhaps the sending of emails to the rest of approvers is happening elsewhere.

cmc333333: It's currently happening in the linear/parallel dispatchers. https://www.pivotaltracker.com/story/show/90215472 is an effort to make this more consistent

app/models/proposal.rb
```diff
@@ -47,7 +47,7 @@ def on_rejected_entry(prev_state, event)
   def restart
     # Note that none of the state machine's history is stored
     self.cart.api_tokens.update_all(expires_at: Time.now)
-    self.cart.approver_approvals.each do |approval|
+    self.cart.approver_approvals.where.not(status: 'approved').each do |approval|
```
cmc333333: I'd avoid "restart"ing altogether if you want to keep it in and renaming this completely if not. Perhaps you want a "edited" message that causes the proposal to stay in the pending state, but triggers emails. Certainly don't want to `deliver_new_cart_emails` as this no longer resembles a new cart

PR #264, opened 2015-03-26
random cleanup of TODOs

afeld:
They were starting to get on my nerves :smirk:


cmc333333:
Is there a ruby linter/checker you prefer? Hopefully one of "reek", "rubocop", or "rubylint" (which all integrate with [syntastic](https://github.com/scrooloose/syntastic))

afeld:
I think I've only used Rubocop, but I the settings are a lot more important than which linter is used. I'm open to adding it to our tests, though very strongly that non-egregious issues should only be fixed as we go, rather than at once in bulk. @monfresh and I ran into this issue on save-ferris, and I know he's been doing other work w/ linting since.

afeld:
Also I think I subconsciously avoid adding them to projects to avoid bikeshedding over style preferences :wink:

afeld:
...though concede that it might be worthwhile to just have the discussion once and pull that bandaid off, rather than having little debates on an ongoing basis.

arowla:
It might be worth it to establish standards for each language, so that projects are started with a coding standard from the outset. I think that as a group we need to start looking towards the longer term, and standards are important for maintainability, including keeping code history functional and informative, with a minimum of formatting-only changes.

PR #265, opened 2015-03-26
Added description instruction as per Story

nickbristow:
https://www.pivotaltracker.com/story/show/90113244


nickbristow:
Heres a screenshot. 
c2

PR #266, opened 2015-03-26
Updated to be more robust

nickbristow:
It was still producing an error in some cases where the user couldn't be found. 

Updated to check and make sure it has an approval before running through the test cases

### DIFF
--- a/app/controllers/communicarts_controller.rb
+++ b/app/controllers/communicarts_controller.rb
@@ -27,8 +27,6 @@ class CommunicartsController < ApplicationController
       @token ||= ApiToken.find_by(approval_id: approval.id)
       if !approval.pending?
         flash[:error] = "You have already logged a response for Cart #{proposal.public_identifier}"
-      elsif !approval.approvable?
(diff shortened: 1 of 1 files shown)


cmc333333:
Looks like this conflicted with #264. Fight, fight, fight

app/controllers/communicarts_controller.rb
```diff
@@ -22,26 +22,29 @@ def approval_response
...
+          approval.reject!
+          flash[:success] = "You have rejected Cart #{proposal.public_identifier}."
+        end
+      end
+      
+      if @token && !@token.used?
+        @token.use!
       end
+    else
+      flash[:error] = "Sorry. You are not allowed to approve your own request."
```
cmc333333: `Sorry, you're not an approver on #{proposal.public_identifier}.` is a bit more accurate

cmc333333:
Would it be simpler to just move `@token ||= ApiToken.find_by(approval_id: approval.id)` into the `else` clause of the existing check? This diff makes the code more "pointy" / [arrow-y?](http://blog.codinghorror.com/flattening-arrow-code/)

PR #268, opened 2015-03-27
90728562 modify request button

phirefly:
Updated the `Edit and Restart?` button to `Modify Request`. Made it more prominent, moving it just below the request to be approved.

Changed the 18F form to reflect this change


phirefly:
Screenshot

screen shot 2015-03-27 at 6 18 28 am

app/views/communicart_mailer/_gsa18f_cart.html.erb
```diff
@@ -1,4 +1,4 @@
-<div id="ncr-layout">
+<div id="gsa18f-layout">
```
cmc333333: Just verifying: this change doesn't break the 18f layout, right?

cmc333333: @phirefly says it looks the same

PR #269, opened 2015-03-27
[Delivers 90103790]: Adds title and description field to ncr work orders

cmc333333:
* Adds the mentioned fields
* Saves/edits the *work order*'s name field, rather than the cart's

These changes were originally made when I thought NCR's field display would need to be different than other clients. The mechanics are in this PR, but turned out to not be required.
* Moves `client_partial` into its own helper
* Modified client partial to look for a `default` client
* Moves proposal_properties to this default client

Looks like:
screen shot 2015-03-27 at 2 48 57 pm
screen shot 2015-03-27 at 12 14 41 pm


app/controllers/ncr/work_orders_controller.rb
```diff
@@ -77,16 +73,12 @@ def redirect_if_cart_cant_be_edited
     def permitted_params
       fields = Ncr::WorkOrder.relevant_fields(
         params[:ncr_work_order][:expense_type])
-      params.require(:ncr_work_order).permit(*fields)
+      params.require(:ncr_work_order).permit(:name, *fields)
```
afeld: We don't want the `:name` in the `relevant_fields`?

app/helpers/client_helper.rb
```diff
@@ -0,0 +1,16 @@
+# Contains functions which make client selection/branching easier
```
afeld: :+1:

afeld:
Hooray! I tweaked some Cart-related stuff in the WorkOrder model (and merged myself before yours – sorry!)...mind updating there, and also making the `description` available in the API?

> Saves/edits the work order's name field, rather than the cart's

Ooh, good step in moving towards removing the dependence on Carts!

app/controllers/ncr/work_orders_controller.rb
cmc333333: Name is special; it's never displayed with the other fields.

cmc333333:
Updated, thanks

PR #270, opened 2015-03-27
[WIP] [Delivers #91304582] add very basic work order listing API

afeld:


cmc333333:
So far so good. Probably want to add requester, approvers, and dates

spec/requests/api/ncr/work_orders_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+          'id' => work_order.id,
+          'not_to_exceed' => work_order.not_to_exceed,
+          'office' => work_order.office,
+          'proposal' => {
+            'created_at' => proposal.created_at.iso8601(3),
+            'flow' => proposal.flow,
+            'id' => proposal.id,
+            'status' => 'pending',
+            'updated_at' => proposal.updated_at.iso8601(3)
+          },
```
afeld: Can't decide if I like the nested format here, or if we should just flatten the attributes from the Proposal up to the top level.

phirefly:
What do you think about just keeping this in a `hackathon` branch that people can pull down and use for now?

afeld:
@phirefly I was thinking we deploy it and then they don't even need to set up the app locally at all.

afeld:
Added a feature flag that I enabled on the dev server, so these changes are innocuous where not enabled. Going to merge now so that a) the hackathon participants don't have to worry about using a special branch, and b) I/we can do more focused PRs. Feedback still more than welcome, though!

PR #272, opened 2015-03-28
make populate scripts work in production

afeld:


afeld:
Verified by @phirefly on dev.

PR #273, opened 2015-03-28
various work order API improvements

afeld:
See commits.


afeld:
Going to merge and deploy for the hackathon folks, but would still love feedback!

PR #276, opened 2015-03-30
[Delivers #90730418] "Request has changed" warning

cmc333333:
This simple implementation tacks on the `client_data`'s updated_at time, as an integer. We can easily replace that with real version numbers should the need arise.

1. Create request
2. Modify version number slightly in email link
3. See:
screen shot 2015-03-30 at 12 19 59 pm


cmc333333:
Another way to test:
1. Make cart
2. Have supervisor log in, but not approve cart; waiting on the page
3. Edit/Restart cart
4. Supervisor clicks approve button -- they should see the above warning.

app/models/cart.rb
```diff
@@ -203,4 +203,9 @@ def total_price
       0.0
     end
   end
+
+  # may be replaced with paper-trail or similar at some point
```
afeld: Since this may be the case, wondering if we should call the param `updated_at_i` or something so that we don't run into a problem distinguishing them down the road?

db/migrate/20150327220945_add_timestamps_to_work_order.rb
```diff
@@ -0,0 +1,5 @@
+class AddTimestampsToWorkOrder < ActiveRecord::Migration
+  def change
+    add_timestamps :ncr_work_orders
```
afeld: Wondering if we should bother populating old ones from the Proposals/Carts...

spec/controllers/communicarts_controller_spec.rb
```diff
@@ -157,6 +157,30 @@
...
+        params = approval_params_with_token
+        params[:version] = cart.proposal.version.to_s
+        put 'approval_response', params
+        approval.reload
+        expect(approval).to be_approved
+        expect(flash[:error]).not_to be_present
+      end
+      it "doesn't work if versions don't match" do
+        params = approval_params_with_token
+        params[:version] = (cart.proposal.version + 1).to_s
```
afeld: Technically, wouldn't the param value only ever be less than that of the Proposal? Could see myself being confused by this later...mind changing to `cart.proposal.version - 1`?

app/models/proposal.rb
```diff
@@ -13,7 +13,7 @@ class Proposal < ActiveRecord::Base
 
   # The following list also servers as an interface spec for client_datas
   delegate :fields_for_display, :client, :public_identifier, :total_price,
-           :name, to: :client_data_legacy
+           :name, :version, to: :client_data_legacy
```
afeld: What do you think about changing it to this?

```ruby
def version
  [
    self.updated_at,
    self.client_data_legacy.updated_at
  ].max.to_i
end
```

There might be unintended consequences I'm not thinking of...

spec/controllers/communicarts_controller_spec.rb
```diff
@@ -157,6 +157,30 @@
...
+    context "version number" do
+      it "doesn't need to be present" do
+        put 'approval_response', approval_params_with_token
+        approval.reload
+        expect(approval).to be_approved
+        expect(flash[:error]).not_to be_present
+      end
+      it "works if version matches" do
+        params = approval_params_with_token
+        params[:version] = cart.proposal.version.to_s
```
afeld: Minor: if you still like the idea of [removing the Cart model](https://www.pivotaltracker.com/story/show/91211506), might be good for our specs to not reference them if they don't need. Mind putting a `let(:proposal) { cart.proposal }` above and changing this and the following specs?

afeld:
:clap: for the elegant solution! This will be a nice touch for users.

> Another way to test

Think turning that into a feature spec would be useful? If you think it would be redundant with the other tests you added, no worries.

app/models/cart.rb
cmc333333: I'm not sure I follow. If we add paper-trail or something similar, we'd keep the `version` name, but back it with the paper-trail data (rather than updated_at data). Maybe `version` is a keyword with paper-trail, though, and the interface should be `version_name` or something?

db/migrate/20150327220945_add_timestamps_to_work_order.rb
cmc333333: Makes sense, and shouldn't be too hard to implement.

spec/controllers/communicarts_controller_spec.rb
cmc333333: Gotcha, yep, makes sense.

app/models/proposal.rb
cmc333333: That ties it down to comparable version identifiers, which is okay for the time being. Will implement.

spec/controllers/communicarts_controller_spec.rb
cmc333333: Sure thing

app/models/cart.rb
afeld: All I'm saying is that we shouldn't change what underlying value the `version` corresponds to, and `updated_at_i` was a way to state that more explicitly. I think we're on the same page?

cmc333333: From an architectural stand point, I like that the version identifier could be swapped out easily (and per client, if needed). There is a concern that we'd switch to a new version system and effectively expire all pending approvals, so I'll add a comment to `Proposal#version` which calls out the backwards compatibility concerns.

cmc333333:
Updated to address your comments. Thanks!

PR #279, opened 2015-03-31
[#89017762] First step to adding consistent authorization

cmc333333:
Opening this PR to see what we think of this direction. Using Pundit, policy checking for restarting a cart is now centralized.

Added a mechanism for combining policy checks, as a single "can this user edit?" could fail due for multiple reasons.


app/controllers/application_controller.rb
```diff
@@ -1,10 +1,34 @@
...
+    else
+      queries = [query]
+    end
+
+    queries.each do |q|
+      super(record, q)  # default processing
+    end
+
+    true
+  end
```
afeld: Mind adding some inline comments to explain what's going on in more detail?

app/policies/proposal_policy.rb
```diff
@@ -0,0 +1,32 @@
...
+    }
+  end
+
+  def initialize(user, proposal)
+    @user = user
+    @proposal = proposal
+  end
+
+  def author?
+    @proposal.requester_id == @user.id
```
afeld: Minor: if you're using `attr_reader`, access the attributes with `self.ATTRNAME`. If you spell instance variables wrong, they give back `nil`, which is really annoying, so I tend to use them as little as possible.

app/policies/tree_policy.rb
```diff
@@ -0,0 +1,28 @@
...
+    else    # deferring to another query/permission
+      self.flatten_tree(value)
+    end
+  end
+
+  def test_all(key)
+    values = self.flatten_tree(key).map{ |k| self.send(k) }
+    values.all?
+  end
+end
```
afeld: This seems like it could make a great wrapper gem for pundit! Or maybe a PR to the original.

afeld:
:heart_eyes: 

If this works, I say :shipit: Happy to deal with authorization for different parts of the application in different pieces.

app/policies/proposal_policy.rb
cmc333333: The above diff is misleading; `attr_reader`'s been removed now

app/policies/tree_policy.rb
cmc333333: Mayhaps. Let's use pundit a bit more -- we very well might be using it "wrong"

cmc333333:
Updated; I think it can be merged before all of the policies are in place.

PR #283, opened 2015-04-02
[#89017762] Another incremental step regarding authorization and proposals

cmc333333:
* Adds an authorization check for whether or not a user can see the approve/reject buttons (though nothing was added to the corresponding handler)
* That check allowed many decorator methods to be removed
* Streamlined 7 or so related but ever so slighty different ways of accessing approv(ers/als) into 4 entry points: all approv* and "current" approv*
* Defaulted approvals to ordering by the position field. This only needs to be overridden in one place, but removes a dozen "ordered" calls
* Moved `linear?` and `parallel?` to proposal


app/decorators/cart_decorator.rb
```diff
@@ -10,7 +10,8 @@ def total_approvers
   end
 
   def approvals_by_status
-    object.approvals.order(
+    # Override default scope
+    object.approvals.reorder(
```
afeld: Had no idea that method existed!

app/policies/proposal_policy.rb
```diff
@@ -21,6 +21,11 @@ def not_approved?
     !@proposal.approved?
   end
 
+  def approve_reject?
```
afeld: What do you think about prefixing the Policy method names with `can_`?

afeld:
:clap:

app/policies/proposal_policy.rb
cmc333333: Makes sense, though the pundit people might take offense ;)

PR #284, opened 2015-04-03
[Delivers #91153332] fix rendering of error pages

afeld:
screen shot 2015-04-02 at 9 46 18 pm


app/controllers/home_controller.rb
```diff
@@ -1,6 +1,7 @@
 class HomeController < ApplicationController
   before_filter :setup_mygov_client
   before_filter :setup_mygov_access_token
+  before_action :authenticate_user!, only: :error
```
afeld: Just so we don't get spam.

config/application.rb
```diff
@@ -41,7 +41,6 @@ class Application < Rails::Application
     }
     config.roadie.url_options = config.action_mailer.default_url_options
 
-    config.exceptions_app = self.routes
```
afeld: This is the actual fix.

app/controllers/communicarts_controller.rb
```diff
@@ -78,7 +78,7 @@ def user_id
 
   def authentication_error(e)
     flash[:error] = e.message
-    redirect_to "/498.html"
+    render 'authentication_error', status: 498
```
cmc333333: 498? That looks to be specific to ESRI. Should this be a 403?

config/routes.rb
```diff
@@ -3,6 +3,7 @@
   match 'approval_response', to: 'communicarts#approval_response', via: [:get, :put]
   root :to => 'home#index'
   match "/auth/:provider/callback" => "home#oauth_callback", via: [:get]
+  get '/error' => 'home#error'
```
cmc333333: Is this only supposed to be accessible when in dev mode?

cmc333333:
There's a lot of parts I don't understand here; can you explain why this fixes?

app/controllers/communicarts_controller.rb
afeld: Heh, I didn't even question it. Good catch.

config/routes.rb
afeld: No, I think it's useful in various environments for testing things like New Relic and logging.

afeld:
> There's a lot of parts I don't understand here

Take a look at the commits separately – might make more sense?

config/routes.rb
afeld: ...and that the 500 page shows up properly :smile:

cmc333333:
Should have looked up `config.exceptions_app`, now I think I understand, thanks.

PR #285, opened 2015-04-03
[#91629492] add environment variable to force HTTPS

afeld:
* [ ] Set `FORCE_HTTPS=true` on production


afeld:
Updated to use [a middleware](https://github.com/tobmatth/rack-ssl-enforcer) (which does stuff like force secure cookies).

/cc @konklone

konklone:
:+1: :us:

.env.example
```diff
@@ -15,6 +15,7 @@ MYUSA_SECRET: consumer_secret_key
 # DEFAULT_URL_PORT
 # DEFAULT_URL_SCHEME
 # FLAG_ITEM_COMMENT_UI
+# FORCE_HTTPS=true
```
cmc333333: Just checking - will this need to be overridden on dev boxes?

afeld: You see that it's commented out? HTTPS-only is off by default.

cmc333333: All the lines in `app/assets/javascripts/application.js` are commented, too, but things aren't always what they appear in rails land ;)

afeld: Lol, touché.

PR #287, opened 2015-04-03
add an API for users

afeld:
Will be used by forthcoming [auto-complete](https://www.pivotaltracker.com/story/show/91734988) of approver emails. Once we add a notion of a `client` per user, we can add that param to filter them.


spec/requests/api/users_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+
+    it "can be `offset`" do
+      users = 3.times.map do
+        FactoryGirl.create(:user)
+      end
+
+      json = get_json('/api/v1/users.json?offset=1')
+
+      ids = json.map {|user| user['id'] }
+      expect(ids).to eq(users.map(&:id)[1..-1])
```
cmc333333: It doesn't look like a sort is involved here? Does offset make sense without a sort?

PR #288, opened 2015-04-03
[Delivers #91734988] auto-complete approver email

afeld:
Builds on #287...see https://github.com/18F/C2/compare/user-api...91734988_auto_complete_email.

demo

@msecret @shawnbot @konklone @wylieconlon Thanks for the suggestions – feedback appreciated!


app/assets/javascripts/forms.js.coffee
```diff
@@ -0,0 +1,24 @@
...
+      opts.searchField = ['email_address', 'first_name', 'last_name']
+      opts.valueField = 'email_address'
+
+    $el.selectize(opts)
+
+    if src
+      # load options from server
+      selectize = $el[0].selectize
+      $.getJSON src, (data) ->
+        selectize.addOption(data)
```
afeld: This file is the important part.

afeld:
A bit baffled as to why the tests are failing, since nothing changed on the backend/specs compared to the other branch, which passed :unamused: Will investigate.

cmc333333:
Mind merging master to make the diff easier to view? Haven't looked closely, but are the test errors because the API is off by default, yet is being called in the existing NCR tests?

cmc333333:
Note that if we include vendor files, we should update the `LICENSES` file, a la https://github.com/18F/foia-hub/blob/master/LICENSE.md . Though, feel free to make that more pretty ;)

afeld:
> Mind merging master

Done!

> are the test errors because the API is off by default, yet is being called in the existing NCR tests?

I'm [enabling it in the spec](https://github.com/18F/C2/blob/bbdbeea98a93790f2e91be4e0ac1c4617f5ab2f4/spec/requests/api/ncr/work_orders_spec.rb#L15)...was passing on `master` and locally, which is totally confusing. Shouldn't be the kind of thing where spec order would matter, either :rage1: 

> if we include vendor files, we should update the `LICENSES` file

Hmm, good point. Asked for guidance in https://github.com/18F/open-source-policy/issues/36.

afeld:
* [ ] Handle when `API_ENABLED` isn't set

afeld:
Updated! Should now work in all environments.

cmc333333:
Vendor issue shouldn't hold up this PR. Very nice!

PR #289, opened 2015-04-03
fix setting and cleanup of API_ENABLED in specs

afeld:
...hopefully. Easier to understand [ignoring whitespace](https://github.com/18F/C2/pull/289/files?w=1).


cmc333333:
`?w=1` is amazing

PR #291, opened 2015-04-03
refactor form JS

afeld:
Builds on #290.

https://github.com/18F/C2/compare/91100514_vendor_auto_complete...form-js-refactor

### DIFF
--- a/app/assets/javascripts/forms.js.coffee
+++ b/app/assets/javascripts/forms.js.coffee
@@ -41,7 +41,6 @@ class Selectizer
     # TODO make sorting smarter, e.g. approvers/vendors they have used before
     $.ajax(
       url: @src()
-      data: {limit: 100}
       dataType: 'json'
       cache: true
       context: @
--- a/app/assets/javascripts/forms.js.coffee
+++ b/app/assets/javascripts/forms.js.coffee
@@ -10,10 +10,10 @@ $ ->
       opts.maxItems = 1
 
     if src
-      # TODO make configurable
-      opts.labelField = 'email_address'
-      opts.searchField = ['email_address', 'first_name', 'last_name']
-      opts.valueField = 'email_address'
+      attr = $el.attr('data-attr')
+      opts.labelField = attr
+      opts.searchField = [attr]
+      opts.valueField = attr
 
     $el.selectize(opts)
 
--- a/app/assets/javascripts/forms.js.coffee
+++ b/app/assets/javascripts/forms.js.coffee
@@ -1,31 +1,58 @@
 $ ->
-  $('.js-selectize').each (i, el) ->
-    $el = $(el)
-    src = $el.attr('data-src')
-    opts = {}
+  class Selectizer
+    constructor: (el) ->
+      @$el = $(el)
 
-    if $el.is('input')
-      # allow free-form input
-      opts.create = true
-      opts.maxItems = 1
+    isFreeForm: ->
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/forms.js.coffee
```diff
@@ -1,24 +1,58 @@
...
+    onOptionsLoaded: (data) ->
+      selectize = @selectizeObj()
+      selectize.addOption(data)
+
+    # load options from server
+    loadOptions: ->
+      # TODO make sorting smarter, e.g. approvers/vendors they have used before
+      $.ajax(
+        url: @src()
+        data: {limit: 100}
```
afeld: Not sure if we should just keep the limit off until we have backend filtering by email/vendor...

app/assets/javascripts/forms.js.coffee
```diff
@@ -1,24 +1,60 @@
...
+
+    onOptionsLoaded: (data) ->
+      selectize = @selectizeObj()
+      selectize.addOption(data)
+
+    loadRemoteOptions: ->
+      # TODO make sorting smarter, e.g. approvers/vendors they have used before
+      $.ajax(
+        url: @src()
+        data: {limit: 100}
```
cmc333333: The API doesn't filter based on search term, that's all happening client side, right? If so, adding a limit wouldn't be a good idea, lest we only search through the first X results.

app/assets/javascripts/forms.js.coffee
```diff
@@ -1,24 +1,60 @@
 $ ->
+  class Selectizer
```
cmc333333: This whole class probably belongs outside the `$ ->`, yes?

app/assets/javascripts/forms.js.coffee
```diff
@@ -1,24 +1,60 @@
...
+    constructor: (el) ->
+      @$el = $(el)
```
cmc333333: Do you actually need to carry around the dom element, or can you pull out the relevant attributes and keep the class light?

app/assets/javascripts/forms.js.coffee
```diff
@@ -1,24 +1,60 @@
...
-    $el.selectize(opts)
-
-    if src
-      # load options from server
-      selectize = $el[0].selectize
-      $.getJSON src, (data) ->
-        selectize.addOption(data)
+    selectizer = new Selectizer(el)
+    selectizer.enable()
+    selectizer.loadOptionsIfRemote()
```
cmc333333: Why not `enable()` and `loadOptionsIfRemote()` in the initialization?

app/assets/javascripts/forms.js.coffee
afeld: Works for me. It'll force us to fix it when we have a large number of records and the page gets super slow :smirk:

app/assets/javascripts/forms.js.coffee
afeld: We do, since we need to call `selectize()` on it.

app/assets/javascripts/forms.js.coffee
afeld: I'm personally a fan of keeping constructors innocuous, but could go either way if you like it better.

afeld:
Updated – thanks for the feedback!

cmc333333:
Excited to see it

PR #293, opened 2015-04-06
Policies for Auth Token and Approve/Reject

cmc333333:
Also separates our token login handling into a separate method.

Probably the most interesting bit here is that, when validating an auth token, we don't have access to a `User` (by definition). We can, however, treat the `params` hash as if it were a user, due to Pundit's flexible permissions model.


app/policies/proposal_policy.rb
```diff
@@ -13,24 +14,32 @@ def initialize(user, proposal)
...
+  def is_not_approved?
     !@proposal.approved?
   end
 
-  def approve_reject?
+  def is_approver?
+    !@proposal.approvals.find_by(user: @user).nil?
+  end
+
+  def is_pending_approver?
```
afeld: Typically, methods that return booleans in Ruby aren't prefixed with `is_*`.

spec/controllers/communicarts_controller_spec.rb
```diff
@@ -130,24 +130,6 @@
...
-        token.update_attributes(expires_at: 8.days.ago)
-        bypass_rescue
-        expect { put 'approval_response', approval_params_with_token }.to raise_error(AuthenticationError)
-      end
-
-      it 'fails when the token has already been used once' do
-        token.update_attributes(used_at: 1.hour.ago)
-        bypass_rescue
-        expect { put 'approval_response', approval_params_with_token }.to raise_error(AuthenticationError)
-      end
```
afeld: I think it still might be useful to have a smoke test at the controller/request level, even if it just uses [mocks](https://relishapp.com/rspec/rspec-mocks/docs) to ensure that the appropriate policy is being called.

app/controllers/communicarts_controller.rb
```diff
@@ -19,69 +15,89 @@ def send_cart
...
-  def authentication_error(e)
-    flash[:error] = e.message
-    render 'authentication_error', status: 403
+  def auth_errors(exception)
+    if exception.record == :api_token
+      self.token_validation_errors(exception.query)
+    else
+      self.user_access_errors(exception.query, exception.message)
+    end
   end
```
afeld: Would be nice to get this logic (basically all of the `protected` methods) out of the controller. Not a blocker, though.

cmc333333: I agree; the error messages are tied very tightly to the policy that failed, for example. I'll tinker with this a bit.

app/policies/proposal_policy.rb
cmc333333: Okidok. I was trying to match the can_XXX names, but I can definitely strip off the is'es

spec/controllers/communicarts_controller_spec.rb
cmc333333: Can do.

cmc333333:
@afeld : updated to address your comments, thanks! Removed the TreePolicy as it was too confusing. Replaced it with an exception-based system. Policies are now written to raise exceptions (with specific error messages -- this binds error messages much closer to their policy). When the policies are called with a trailing `?`, that exception is caught, if raised, and `false` is returned instead.

Also removed all of the `is_` prefixes and added back the authorization smoke tests.

app/controllers/application_controller.rb
```diff
@@ -10,24 +10,20 @@ class ApplicationController < ActionController::Base
...
-    end
+  def authorize(record, query=nil, user=nil)
+    user ||= @current_user
+    policy = Pundit.policy(user, record)
 
-    queries.each do |q|
-      super(record, q)  # have Pundit process each permission
+    # use the action as a default permission
+    query ||= ("can_" + params[:action].to_s + "!").to_sym
+    unless policy.public_send(query)
```
afeld: Passing user input to a `#send` is dangerous...think it's worth checking against a whitelist?

app/policies/api_token_policy.rb
```diff
@@ -0,0 +1,37 @@
...
+    @api_token = ApiToken.find_by(access_token: params[:cch])
+  end
+
+  def exists!
+    check(!@api_token.nil?,
+          "Something went wrong with the token (nonexistent)")
+  end
+
+  def not_expired!
+    exists! && check(!@api_token.expires_at || @api_token.expires_at > Time.now,
```
afeld: Is checking for `!@api_token.expires_at` correct? Don't we want the opposite?

afeld: Also, maybe move the latter check into an `ApiToken#expired?` method.

app/controllers/application_controller.rb
afeld: ...though maybe it's not as big of a deal in this case since it's prefixing (and suffixing) the method name.

app/policies/exception_policy.rb
```diff
@@ -0,0 +1,35 @@
...
+    if !guard
+      # will need to replace this when a new version of pundit arrives
+      ex = Pundit::NotAuthorizedError.new(message)
+      caller_name = caller_locations(1, 1)[0].label
+      ex.query, ex.record, ex.policy = caller_name.to_sym, @record, self
+      raise ex
+    end
+    guard
+  end
+end
```
afeld: This class is complex enough that it could probably use some dedicated unit tests.

afeld:
New approach works for me!

app/controllers/application_controller.rb
cmc333333: I was under the assumption that `params[:action]` was a whitelist already?

All queries should be coming from code rather than a user. If we make a mistake and allow user input to get in, however, that user would only have access to the policy object, which is intentionally limited.

app/policies/api_token_policy.rb
cmc333333: This duplicates existing logic: if there is no `expires_at` set on the token, we proceed. I assume this was a backwards compatibility thing.

Can definitely move that into `#expired?`

app/policies/exception_policy.rb
cmc333333: Can do.

app/controllers/application_controller.rb
afeld: Yeah, just tested...it gets overridden, even if that query param is present. Probably ok for now, but maybe worth flagging for a future security review.

app/policies/api_token_policy.rb
afeld: > if there is no `expires_at` set on the token, we proceed.

That seems like a mistake, or a hand-wave-y thing we did to make our tests work. It *should* be present for all tokens. Storied:

https://www.pivotaltracker.com/story/show/92351222

cmc333333:
Updated to use an `expired?` method and add specs specifically for the ExceptionPolicy mixin

PR #294, opened 2015-04-07
Authorization added to cart views

cmc333333:
Builds on #293: https://github.com/18F/C2/compare/authorization...cart-auth?expand=1

Adds an authorization check to cart `show` handler, as well as a scoping check for the `list` and `archive` handlers. Modifies these controllers/views to use `proposal` instead of `cart`


app/policies/proposal_policy.rb
```diff
@@ -13,24 +14,60 @@ def initialize(user, proposal)
...
     !@proposal.approved?
   end
 
-  def approve_reject?
+  def is_approver?
+    !@proposal.approvals.find_by(user: @user).nil?
+  end
+
+  def is_observer?
+    @proposal.observers.include? @user
```
phirefly: Minor: you can use Activerecord's helper `proposal.approvals.exists?(user: @user)`

app/policies/proposal_policy.rb
```diff
@@ -13,24 +14,60 @@ def initialize(user, proposal)
...
+
+  # equivalent of can_show?
+  class Scope
+    def initialize(user, scope)
+      @user = user
+      @scope = scope
+    end
+
+    def resolve
+      # Use subselects instead of left joins to avoid duplication removal
```
afeld: Out of curiosity, why would we want it to return duplicates?

app/policies/proposal_policy.rb
```diff
@@ -13,24 +14,60 @@ def initialize(user, proposal)
...
+  def is_not_approved?
     !@proposal.approved?
   end
 
-  def approve_reject?
+  def is_approver?
+    !@proposal.approvals.find_by(user: @user).nil?
+  end
+
+  def is_observer?
```
afeld: Leave off the `is_*` prefix.

afeld:
Just the couple of minor notes!

app/policies/proposal_policy.rb
cmc333333: Can do.

app/policies/proposal_policy.rb
cmc333333: Nice.

app/policies/proposal_policy.rb
cmc333333: A better comment may be 
```ruby
# use subselects instead of left joins to avoid an explicit duplication-removal step
```

If we were to use an inner join, we'd be skipping over carts with no observers or no approvers (as they'd be filtered out). If we instead used a left join, we'd get one row for every cart x cart's observer x cart's approver tuple. The duplicates can be removed with a `distinct`, but that makes the query even more complicated, and may make it play less well with further filters.

Instead, this uses subselects, which require explicit sql, but should make working with the results much easier. Postgres should also be smart enough to make these subselects efficient.

cmc333333:
Updated to include the exception style proposed in #293. Also removed the `is_` prefixes, modified the comment, and used `exists?`

app/policies/proposal_policy.rb
```diff
@@ -35,4 +40,30 @@ def can_edit!
   end
 
   alias_method :can_update!, :can_edit!
+
+  def can_show!
+    check(author? || approver? || observer?,
```
afeld: Wondering if we should simplify this to `@proposal.users.include?(user)`, so that it (eventually) will be one SQL call instead of three.

afeld:
Just the one minor suggestion. Otherwise, :shipit:!

app/policies/proposal_policy.rb
cmc333333: Sure thing

cmc333333:
Updated to use proposal.users

PR #295, opened 2015-04-07
License for third parties. Sort auto complete

cmc333333:
In the case of a score tie (e.g. when the page first loads), sort by the search field.


afeld:
Thank you!

PR #296, opened 2015-04-13
[Delivers #91070178] Comment Authorization and connection to Proposal

cmc333333:
Builds on #294: https://github.com/18F/C2/compare/cart-auth...comment-auth?expand=1

* Adds a proposal field for comments, migrates existing comments to use that field, and removes the existing polymorphic field.
* Removes all references to said polymorphic field
* Adds authorization when creating comments
* Specs to test that authorization

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -1,5 +1,7 @@
 class CommentsController < ApplicationController
   before_filter :authenticate_user!
+  before_filter ->{authorize self.proposal, :can_show!}
+  rescue_from Pundit::NotAuthorizedError, with: :auth_errors
 
   def index
     @proposal = self.proposal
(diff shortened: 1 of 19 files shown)


app/controllers/comments_controller.rb
```diff
@@ -1,40 +1,35 @@
 class CommentsController < ApplicationController
   before_filter :authenticate_user!
+  before_filter ->{authorize self.proposal, :can_show!}
```
afeld: Does `can_show!` still make sense for the `create` action?

cmc333333: `can_show!` is a policy on proposals, not comments. So, to create/view/etc. comments, we are testing the ability to see the associated proposal.

We could add a new method which performs the same action (`can_create_comment`?) but that seems to obscure the test.

PR #297, opened 2015-04-13
[Delivers #89834540] Client field on User

cmc333333:
* Adds a client field to user
* Uses it when determining whether or not to show the NCR link

Note, this performs no data migrations.


afeld:
Some test failures, but otherwise :+1:

PR #298, opened 2015-04-15
[#91211506] Proposals Controller

cmc333333:
The bulk of this PR is switching all of the `/carts` urls to be `/proposals`. Associated changes:

* Carts urls still exist, redirecting the user to the appropriate proposal url
* Replaced `cart_template_name` and `prefix_template_name` with `client_partial` equivalents
* Removed `CartsHelper` as its presentation logic was only used in one template and is not terribly complicated
* Cart decorator is now a Proposal decorator
* Compiled a few, single-use partials inline


app/controllers/carts_controller.rb
```diff
@@ -1,32 +1,16 @@
-# @todo Rename
+# @todo Eventually, delete
```
afeld: I think we can delete this controller now and do the redirection in `routes.rb`:

http://guides.rubyonrails.org/routing.html#redirection

app/views/proposals/_proposal_list.html.erb
```diff
@@ -17,11 +17,20 @@
...
-      <td class="sixth"><a href="/carts/<%= proposal.cart.id %>"><%= proposal.cart.id %></a></td>
-      <td class="first"><a href="/carts/<%= proposal.cart.id %>"><%= proposal.name %></a></td>
+      <td class="sixth"><a href="<%= proposal_url(proposal) %>"><%= proposal.public_identifier %></a></td>
+      <td class="first"><a href="<%= proposal_url(proposal) %>"><%= proposal.name %></a></td>
       <td class="sixth"><%= number_to_currency(proposal.total_price) %></td>
       <td class="sixth">
-        <p><%= display_status(proposal.cart, current_user) %></p>
+        <% approvers = proposal.currently_awaiting_approvers %>
+        <% if proposal.pending? && approvers.include?(current_user) %>
+          <%# This isn't used anywhere %>
```
afeld: What do you mean? That condition would be `true` for an outstanding approval, right?

app/views/proposals/show.html.erb
```diff
@@ -0,0 +1 @@
+<%= render partial: "proposal", locals: { proposal: @proposal}  %>
```
afeld: Hmm, is that partial used in multiple places? Maybe we can just move that file in here.

afeld:
This is great, thanks for doing it!

app/views/proposals/_proposal_list.html.erb
cmc333333: The condition would be true if the user were an approver on an outstanding approval, but we only list carts by the requester. As the requester shouldn't be an approver on his/her own cart, I don't think this text is ever shown.

app/controllers/carts_controller.rb
cmc333333: Nice; can do.

app/views/proposals/_proposal_list.html.erb
afeld: > we only list carts by the requester

Aaaah, that didn't used to be the case...I missed that change. [Storied](https://www.pivotaltracker.com/story/show/92792448).

app/views/proposals/show.html.erb
cmc333333: Good call

cmc333333:
Updated to remove the cart controller and embed the partial, as it was only used in one template.

afeld:
Needs merge from `master`, but otherwise :shipit:!

cmc333333:
Merged master

PR #299, opened 2015-04-17
[Delivers #89433964] File upload

cmc333333:
Adds paperclip and an attachments model to allow file uploading. This will use S3 if the appropriate environmental variables are set (namely a bucket name, the private, and public key).

Also generalized some styles to work with file uploads as well as comments.

Looks like:
screen shot 2015-04-15 at 4 46 29 pm


Gemfile
```diff
@@ -5,6 +5,7 @@ gem 'acts_as_list'
 gem 'ar_outer_joins'
 gem 'autoprefixer-rails'
 gem 'awesome_print'
+gem 'aws-sdk-v1'
```
afeld: Why `v1`? `aws-sdk` is the newer one, which it seems like Paperclip supports.

config/routes.rb
```diff
@@ -22,6 +22,7 @@
     end
 
     resources :comments, only: [:index, :create]
+    resources :attachments, only: [:create]
```
afeld: Might as well put this under `/proposals/`.

Gemfile
```diff
@@ -16,6 +17,7 @@ gem 'jquery-rails'
 gem 'jquery-turbolinks'
 gem 'newrelic_rpm'
 gem 'omniauth-myusa', git: 'https://github.com/18F/omniauth-myusa.git'
+gem "paperclip", "~> 4.2"
```
afeld: There are differing opinions on this, but I would leave the version restriction off unless there's a specific reason to lock the range.

afeld: At the very least, for consistency.

Gemfile
cmc333333: Not quite yet: https://github.com/thoughtbot/paperclip/issues/1764

app/controllers/attachments_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  def proposal
+    @cached_proposal ||= Cart.find(params[:cart_id]).proposal
+  end
+
+  def attachments_params
+    params.require(:attachment).permit(:file)
+  end
+
+  def auth_errors(exception)
+    redirect_to carts_path, :alert => "You are not allowed to see that cart"
```
afeld: Maybe "you aren't allowed to add an attachment to that proposal"?

app/views/carts/_cart.html.erb
```diff
@@ -75,10 +75,10 @@
 
 
 
-<div class="cart-comments-container">
-  <div id="cart-comments">
-    <h3>Comments on this purchase request</h3>
-    <%- if @show_comments %>
+<%- if @include_comments_files %>
```
afeld: I'm not sure that we need this anymore, but that can be a separate discussion.

app/views/carts/_cart.html.erb
```diff
@@ -120,9 +120,50 @@
...
+      <%= form_for [cart, Attachment.new] do |f| %>
+        <div class="line-item">
+          <div class="row">
+            <%= f.file_field :file %>
+          </div>
+        </div>
+        <div class='row text-area-info-container'>
+          <div class='col-xs-7 col-sm-6 text-area-info-web'>
+            <p>
+              Attaching a file will associate it with this proposal. Files cannot be removed
```
afeld: Not sure that the last part is necessary.

spec/controllers/attachments_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  describe 'permission checking' do
+    let (:proposal) { FactoryGirl.create(:proposal, :with_approvers,
+                                         :with_observers, :with_requester,
+                                         :with_cart) }
+    let (:params) {{
+      cart_id: proposal.cart.id, 
+      attachment: { file: fixture_file_upload('icon-user.png', 'image/png') }
+    }}
+                     
+    it "allows the requester to comment" do
```
afeld: I think each of these should me "...to add an attachment"?

spec/controllers/attachments_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      login_as(proposal.requester)
+      post :create, params
+      expect(flash[:success]).to be_present
+      expect(flash[:error]).not_to be_present
+      expect(response).to redirect_to(proposal.cart)
```
afeld: We should probably check that the file was properly attached to the model.

Gemfile
afeld: Far enough. Mind adding an inline comment linking to that issue?

app/controllers/attachments_controller.rb
cmc333333: Can do

config/routes.rb
cmc333333: Just waiting for that shoe to drop :)

app/views/carts/_cart.html.erb
cmc333333: Let's loop in @phirefly ; I think he added this initially.

app/views/carts/_cart.html.erb
cmc333333: Will delete

spec/controllers/attachments_controller_spec.rb
cmc333333: Copypasta

spec/controllers/attachments_controller_spec.rb
cmc333333: Good call

config/routes.rb
afeld: Let's not wait.

app/views/carts/_cart.html.erb
```diff
@@ -120,9 +120,50 @@
...
+      <%= form_for [cart, Attachment.new] do |f| %>
+        <div class="line-item">
+          <div class="row">
+            <%= f.file_field :file %>
+          </div>
+        </div>
+        <div class='row text-area-info-container'>
+          <div class='col-xs-7 col-sm-6 text-area-info-web'>
+            <p>
+              Attaching a file will associate it with this proposal
```
afeld: Minor: seems like this would be obvious...no?

app/views/carts/_cart.html.erb
```diff
@@ -120,9 +120,50 @@
...
+          <p class='col-xs-5 col-sm-6 text-area-button'>
+            <%= submit_tag "Attach a File", id: :add_a_file %>
+          </p>
+        </div>
+      <%- end %>
+      <% cart.proposal.attachments.each do |attachment| %>
+        <div class="line-item">
+          <div class="row">
+            <p class="col-sm-6 col-xs-12">
+              <a href="<%= attachment.file.expiring_url(10*60) %>"><%= attachment.file_file_name %></a>
```
afeld: What is the magic number here? Maybe move to a helper or decorator or something?

config/application.rb
```diff
@@ -46,5 +46,18 @@ class Application < Rails::Application
     config.autoload_paths << Rails.root.join('lib')
 
     config.assets.precompile << 'common/communicarts.css'
+
+    # Paperclip's attachment settings are determined by S3 env vars
```
afeld: Maybe mention that it will be saved to the filesystem otherwise?

afeld:
A few suggestions, but no blockers. Branch-deploying.

app/views/carts/_cart.html.erb
cmc333333: Changing to "Attach a file (e.g. receipts, contract documents, etc.)"

app/views/carts/_cart.html.erb
cmc333333: Can do. I'll just replace override file_url.

Magic number is for an arbitrary timeout: 10 minutes.

config/application.rb
cmc333333: Can do

spec/controllers/attachments_controller_spec.rb
cmc333333: This was addressed just below the diff.

afeld:
* [x] Create bucket for testing
* [ ] Create bucket for production
* [x] Set environment variables on `c2-dev`
* [ ] Set environment variables on `c2-staging`
* [ ] Set environment variables on `c2-prod`

PR #300, opened 2015-04-17
Feature/18f recurring charges

nickbristow:
This PR converts 18f to a proper model and adds the recurring fields feature. 

c2
c2
c2


app/models/gsa18f/procurement.rb
```diff
@@ -0,0 +1,111 @@
...
+      greater_than_or_equal_to: 1
+    }
+    validates :product_name_and_description, presence: true
+
+    def set_defaults
+      # self.not_to_exceed ||= false
+      # self.emergency ||= false
+    end
+
+    def init_and_save_cart(approver_email, requester)
```
afeld: Why does the `approver_email` need to be passed in explicitly? Can't this method just call that method directly?

app/controllers/gsa18f/procurements_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+      render 'form'
+    end
+
+    def update
+      @procurement = self.procurement
+      @procurement.update(permitted_params)
+      @approver_email = @procurement.approver_email;
+      if self.errors.empty?
+        cart = self.cart
+        @procurement.save
```
afeld: I don't think you need to `save` again.

app/controllers/gsa18f/procurements_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+        flash[:error] = errors
+        render 'form'
+      end
+    end
+
+    def permitted_params
+      fields = Gsa18f::Procurement.relevant_fields(
+        params[:gsa18f_procurement][:recurring])
+      params.require(:gsa18f_procurement).permit(:name, *fields)
+    end
```
afeld: Mind making this a `protected` method?

app/controllers/gsa18f/procurements_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+      @procurement ||= Gsa18f::Procurement.find(params[:id])
+    end
+
+    def errors
+      errors = []
+      if !@procurement.valid?
+        errors += @procurement.errors.full_messages
+      end
+      errors
+    end
```
afeld: I think this could be simplified to

```ruby
def errors
  @procurement.validate
  @procurement.errors.full_messages
end
```

app/models/gsa18f/procurement.rb
```diff
@@ -0,0 +1,111 @@
...
+    }
+    validates :quantity, numericality: {
+      greater_than_or_equal_to: 1
+    }
+    validates :product_name_and_description, presence: true
+
+    def set_defaults
+      # self.not_to_exceed ||= false
+      # self.emergency ||= false
+    end
```
afeld: Just remove this if it isn't necessary.

app/models/gsa18f/procurement.rb
```diff
@@ -0,0 +1,111 @@
...
+      "gsa18f"
+    end
+
+    # @todo - this is pretty ugly
+    def public_identifier
+      self.cart.id
+    end
+
+    def total_price
+      self.cost_per_unit * self.quantity || 0.0
```
afeld: Mind adding some parentheses here to make the grouping more obvious?

app/models/gsa18f/procurement.rb
```diff
@@ -0,0 +1,111 @@
...
+    end
+
+    # may be replaced with paper-trail or similar at some point
+    def version
+      self.updated_at.to_i
+    end
+
+    def approver_email
+      ENV['GSA18F_APPROVER_EMAIL'] || '18fapprover@gsa.gov'
+    end
```
afeld: Minor: I think this would make more sense as a class method.

app/views/default/_proposal_properties.html.erb
```diff
@@ -10,7 +10,11 @@
         <strong><%= name %>
       </td>
       <td>
-        <%= value %>
+        <% if value.class == BigDecimal %>
+          <%= number_to_currency(value) %>
```
afeld: Hmm, I wonder if we can safely assume all decimals should be displayed as $$...ok for now I suppose.

app/views/gsa18f/procurements/form.html.erb
```diff
@@ -2,12 +2,12 @@
   <h3>GSA 18f</h3>
   <h4>Purchase Request - Supplies</h4>
   <p>Eg; ergonomic keyboards, foamcore boards, ergonomic mice, monitor stands, post-its, sharpies, etc. </p>
-  <%= form_for @proposal_form, url: @form_url, method: @form_method do |f| %>
+  <%= form_for @procurement do |f| %>
```
afeld: Yay!

db/migrate/20150410163833_add_urgency_to_gsa18f_procurement.rb
```diff
@@ -0,0 +1,5 @@
+class AddUrgencyToGsa18fProcurement < ActiveRecord::Migration
+  def change
+  	add_column :gsa18f_procurements, :urgency, :string
```
afeld: Careful with indentation!

db/migrate/20150414194609_add_time_stamps_to_gsa18f_procurements.rb
```diff
@@ -0,0 +1,7 @@
+class AddTimeStampsToGsa18fProcurements < ActiveRecord::Migration
+  def change
+  	change_table :gsa18f_procurements do |t|
+  		t.timestamps
+  	end
```
afeld: Ditto.

spec/features/18f_procurement_spec.rb
```diff
@@ -1,17 +1,17 @@
...
-    let(:gsa18f) {
-      proposal = FactoryGirl.build(:gsa18f_proposal_form)
-      proposal.requester = requester
-      proposal.create_cart
-    }
+    # let(:gsa18f) {
+    #   proposal = FactoryGirl.build(:gsa18f_procurement)
+    #   proposal.requester = requester
+    #   proposal.create_cart
+    # }
```
afeld: Please remove if not needed!

spec/features/18f_procurement_spec.rb
```diff
@@ -81,94 +82,100 @@
...
       # keeps the form values
     end
+    let(:gsa18f) {
+      pr = FactoryGirl.create(:gsa18f_procurement)
+      pr.init_and_save_cart('approver@email.com', requester)
+      pr
+    }
+    let(:gsa18f_cart) {
+      gsa18f.proposal.cart
+    }
```
afeld: Rather than having `let` statements in the middle of a spec file, mind moving them to the top (if they're used by all specs in the file), or surrounding this and the subsequent specs with a `context` block?

app/models/gsa18f/procurement.rb
```diff
@@ -0,0 +1,104 @@
...
+    def name
+      self.product_name_and_description
+    end
+
+    def self.approver_email
+      ENV['GSA18F_APPROVER_EMAIL'] || '18fapprover@gsa.gov'
+    end
+
+    protected
+    def system_approvers
```
cmc333333: Doesn't look like this is ever used

config/data/18f.yaml
```diff
@@ -7,4 +7,8 @@ OFFICES:
   - 'Chicago'
   - 'Dayton'
   - 'San Francisco'
-  - 'Me! (Remote Worker)'
\ No newline at end of file
+  - 'Me! (Remote Worker)'
+RECURRENCE:
+  - 'Daily'
+  - 'Monthly'
+  - 'Yearly'
```
cmc333333: This is okay for now, but I assume we'll want these values to have programatic meaning at some point, so we'll likely want them within the main code rather than in a configuration file.

db/schema.rb
```diff
@@ -47,6 +47,28 @@
...
+    t.string   "url"
+    t.text     "notes"
+    t.integer  "quantity"
+    t.text     "details"
+    t.string   "part_number"
+    t.float    "price"
+    t.integer  "cart_id"
+    t.datetime "created_at"
+    t.datetime "updated_at"
+  end
```
cmc333333: Try merging master and migrating the db. I don't think `cart_items` and `cart_items_traits` are still used.

cmc333333:
This is good; only thing that seems critical is the schema issue

cmc333333:
Merge upstream. Ran through a request locally with no trouble. What do you think, @afeld ?

Note that it needs the demo server to be migrated.

PR #301, opened 2015-04-19
[Delivers #92448720] add notion of approval delegates

afeld:
A User can have zero or more ApprovalDelegates, which are other users who can approve requests instead of them. Not loving the names I used for new models/columns/associations...very open to suggestions for alternatives. Open questions:

* If an approver has delegates, should they still be allowed to approve?

TODOs:

* [x] Document how to add delegates
* [x] Ensure that the delegate can view the proposals
* [x] Ensure that the delegate can comment on the proposals
* [x] Don't auto-login from notification emails that have delegates

### DIFF
--- a/spec/controllers/comments_controller_spec.rb
+++ b/spec/controllers/comments_controller_spec.rb
@@ -40,6 +40,7 @@ describe CommentsController do
       expect {
         post :create, params
       }.to change{ proposal.comments.count }.from(0).to(1)
+      expect(Comment.last.user).to eq(delegate)
     end
 
     it "does not allow others to comment" do
--- a/spec/controllers/comments_controller_spec.rb
+++ b/spec/controllers/comments_controller_spec.rb
@@ -33,7 +33,7 @@ describe CommentsController do
     it "allows a delegate to comment" do
       approver = proposal.approvals.first.user
       delegate = FactoryGirl.create(:user)
-      approver.outgoing_delegates.create!(assignee: delegate)
+      approver.add_delegate(delegate)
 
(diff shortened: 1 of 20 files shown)


cmc333333:
This looks good. Note that, as hinted in your last test, this does not categorize the shared inbox user as anything special, so someone clicking from the shared inbox who had not been logged in, would sign in as the shared inbox user.

afeld:
I'm thinking that's ok for now...happy to take care of that as a separate story if we want to add the restriction of users with delegates not being able to approve, but I could see the delegates being used for things like vacation, so would probably want it as a separate flag anyway.

afeld:
> Not loving the names I used for new models/columns/associations

In particular, should the join model be called a `Delegation` (or `ApprovalDelegation`), which has a `delegate` and a `delegator`?

afeld:
Allrighty, ready for review!

afeld:
Updated to account for mailbox users forwarding notification emails to their delegates.

spec/controllers/comments_controller_spec.rb
```diff
@@ -30,6 +30,18 @@
...
+    it "allows a delegate to comment" do
+      approver = proposal.approvals.first.user
+      delegate = FactoryGirl.create(:user)
+      approver.add_delegate(delegate)
+
+      login_as(delegate)
+
+      expect {
+        post :create, params
+      }.to change{ proposal.comments.count }.from(0).to(1)
```
cmc333333: Would you mind adding a check that the comment was "made" by the delegated user?

PR #304, opened 2015-04-21
[Delivers #92876748] delegate notification emails shouldn't use URL to automatically login

afeld:
Just use the simple cart URL.


cmc333333:
Simple!

PR #305, opened 2015-04-21
prompt delegate to log in when approving

afeld:
A little janky, and not convinced about the approach, but it should work for now. The new flow, when logged out:

1. Delegate sees notification email
1. They click "Approve" from the email
1. If signed in:
    * Yes: Approved.
    * No:
        1. They are redirected to the homepage, with a flash message saying they should log in
        1. They click "Sign in with MyUSA"
        1. After completing the OAuth flow, Approved.

Could use some manual testing as well if someone has a moment to try it.

### DIFF
--- a/app/controllers/communicarts_controller.rb
+++ b/app/controllers/communicarts_controller.rb
@@ -57,10 +57,16 @@ class CommunicartsController < ApplicationController
   end
 
   def auth_errors(exception)
-    flash[:error] = exception.message
     if exception.record == :api_token
-      render 'authentication_error', status: 403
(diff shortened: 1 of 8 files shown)


afeld:
The actual title of this should be "if there's a problem with the token, just tell them to sign in".

cmc333333:
This makes sense; will test locally first, then merge.

afeld:
Updated w/ fixed link for delegates.

cmc333333:
Tested with both a logged in delegated user and a non-logged in delegated user.

The latter route is rough around the edges, but an improvement!

PR #306, opened 2015-04-22
[Delivers #92981132] move flash message below the header

afeld:
Curious to hear @jehlers's opinion on it, but I think it's an improvement in terms of where there eyes will be on the page.

screen shot 2015-04-21 at 11 52 15 pm


jehlers:
Looks good to me! Would it be helpful to add a close or hide X on the right side? Or is this not the kind of message that needs that?

Also maybe make the dotted stroke a red/pink so that it blends a little more into the background?

PR #309, opened 2015-04-22
[Delivers #92876870] fix display of approver after a delegate approves

afeld:
Fixes two issues:

* The subsequent notification email was showing the delegator as having approved, not the delegate.
* The delegate wasn't being assigned in the case of rejection.

### DIFF
--- a/app/controllers/communicarts_controller.rb
+++ b/app/controllers/communicarts_controller.rb
@@ -17,15 +17,14 @@ class CommunicartsController < ApplicationController
   def approval_response
     proposal = self.cart.proposal
     approval = proposal.approval_for(current_user)
+    if approval.user.delegates_to?(current_user)
+      # assign them to the approval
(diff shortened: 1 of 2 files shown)


app/controllers/communicarts_controller.rb
```diff
@@ -17,15 +17,14 @@ def send_cart
   def approval_response
     proposal = self.cart.proposal
     approval = proposal.approval_for(current_user)
+    if approval.user.delegates_to?(current_user)
```
phirefly: Wouldn't this just need to move above line 27, where we send out the email?

cmc333333: This is saying "if *any* action is taken", which seems like an okay solution.

afeld: @phirefly No, because 'Approval#approve!' and 'reject!' trigger emails.

PR #310, opened 2015-04-22
ensure that delegates can only be assigned to an approver once

afeld:
Just noticed that we have some duplicates in our database. Cleaned them up, but this will prevent new ones.


phirefly:
Ship it

PR #314, opened 2015-04-23
[Delivers #92792448] list all proposals that the user is involved in

afeld:
...on `/carts`. Includes Proposals for which the users is a:

* Requester
* Approver
* Delegate
* Observer


app/policies/proposal_policy.rb
```diff
@@ -79,9 +79,19 @@ def resolve
...
+        -- approver
         OR EXISTS (SELECT id FROM approvals
                    WHERE proposal_id = proposals.id AND user_id = :user_id)
+        -- delegate
+        OR EXISTS (
+          SELECT approvals.id FROM approvals
+          LEFT OUTER JOIN approval_delegates
+          ON approval_delegates.assigner_id = approvals.user_id
+          WHERE approval_delegates.assignee_id = :user_id
+        )
```
afeld: _Somehow_ got this right on the first try. `</humblebrag>`

skeletons dance

cmc333333: It looks like there's not restriction by proposal? I think all that needs to change is to add:
```sql
AND approvals.proposal_id = proposals.id
```

afeld: Way to burst my bubble, @cmc333333.

afeld:
Updated! Fixed the SQL condition issue...not sure how/if to write a test for it though.

PR #315, opened 2015-04-24
[Delivers #89820292] editing a request

cmc333333:
This avoids the `restart` work flow while editing a request. Instead, approvers which had already approved the request are _notified_ of the change, but do not need to re-approve.

It also accounts for a corner case where the requester edits the proposal *after* the first approver has approved. Here, the UI makes that field non-editable ("frozen"). If the first approver has _not_ already approved, then his/her approval is deleted (to avoid any lingering token issues) and a new approval put in its place.


afeld:
> approvers which had already approved the request are notified of the change, but do not need to re-approve.

I got the impression that they wouldn't care, assuming these are usually small/administrative changes. @phirefly @brethauer?

> If the first approver has not already approved, then his/her approval is deleted

Maybe we should notify them? Making a story to take care of that is fine though.

phirefly:
@afeld It has been unanimous that a previously approved request should not be required to be approved again. That said, one or two people have mentioned the desire to get an email telling them that things have changed -- but if it was already approved, they don't want to be required to approve it again.

afeld:
> one or two people have mentioned the desire to get an email telling them that things have changed

Yeah, this was my question specifically.

phirefly:
I see. Let's keep this as the default for now: sending them an email. I've added a story to handle this in preference: https://www.pivotaltracker.com/story/show/93245548

phirefly:
Also, @cmc333333 I'd be in favor of squashing all my wips into your commits before merging

afeld:
Meh, not worth it.

PR #316, opened 2015-04-27
[Delivers #93334670] Notify more approvers on comment creation

cmc333333:
Now, the requester, actionable approvers, already-approved approvers (the new bit), and observers get notified.

This also removes the `:with_requester` option in the Proposal factory. We assume that a requester is present.

Also removed a stubbed method in a mailer spec.


app/models/comment.rb
```diff
@@ -25,4 +27,16 @@ def self.attributes
...
     ]
   end
+
+  # All of the users who should be notified when a comment is created
+  def listeners
+    users_to_notify = []
+    users_to_notify += self.proposal.currently_awaiting_approvers
+    users_to_notify += self.proposal.approvals.approved.map(&:user)
+    users_to_notify += self.proposal.observers
+    users_to_notify << self.proposal.requester
```
afeld: What do you think about sharing some logic with `Proposal#users`? Also, there *shouldn't* be duplicates, but I'd make `users_to_notify` a [`Set`](http://ruby-doc.org/stdlib-2.2.2/libdoc/set/rdoc/Set.html) anyway.

app/models/comment.rb
```diff
@@ -25,4 +27,16 @@ def self.attributes
       'updated_at'
     ]
   end
+
+  # All of the users who should be notified when a comment is created
+  def listeners
+    users_to_notify = []
+    users_to_notify += self.proposal.currently_awaiting_approvers
+    users_to_notify += self.proposal.approvals.approved.map(&:user)
```
afeld: Wouldn't these two lines be the same as

```ruby
users_to_notify += self.proposal.approvers
```

?

afeld:
Needs merge, but :+1: overall!

app/models/comment.rb
cmc333333: No, `currently_awaiting_approvers` only includes approvers which can take action _right now_. That means any approvers later in the chain aren't included.

app/models/comment.rb
cmc333333: `Set` makes sense. I'm not sure there's a good way to use `Proposal#users` here. We could try to _subtract_ the approvers that shouldn't be notified, but then we run the risk of someone appearing twice in the chain and being removed.

app/models/comment.rb
afeld: Ah ok, totally missed that. Might be worth an inline comment.

afeld: ...or moving to a `past_and_current_approvers` method or something.

cmc333333:
Master merged, tweaked based on comments

PR #317, opened 2015-04-28
[Delivers #93440210] Friendly error message when attachment not present

cmc333333:
Looks like:
screen shot 2015-04-28 at 10 53 17 am


afeld:
Would like to take another pass at the uploading UI to prevent this problem from happening, but good to have the validation regardless. The test failure doesn't really make sense to me here...

cmc333333:
Rerunning with a cleared cache resolved

afeld:
:hankey:

PR #318, opened 2015-04-28
[Delivers #89433078] Add Fiscal Year to Work Order Identifier

cmc333333:
Along the way, be more consistent about including the cart identifier in all communications. This won't lead to proper threading, yet, but gets one step closer.

Note that the identifiers are also keyed by the work order rather than the cart (one more step towards removing Carts).

Looks like:
screen shot 2015-04-28 at 11 27 58 am
screen shot 2015-04-28 at 11 26 54 am
screen shot 2015-04-28 at 11 57 46 am
screen shot 2015-04-28 at 11 57 36 am


app/models/ncr/work_order.rb
```diff
@@ -92,9 +92,13 @@ def client
...
     end
 
-    # @todo - this is pretty ugly
     def public_identifier
-      self.cart.id
+      fiscal_year = self.created_at.year
+      if self.created_at.month >= 10
+        fiscal_year += 1
+      end
+      fiscal_year = fiscal_year % 100   # convert to two-digit
```
afeld: Minor: maybe move the above to a `#fiscal_year` method?

app/models/ncr/work_order.rb
```diff
@@ -92,9 +92,13 @@ def client
...
+      "FY" + fiscal_year.to_s.rjust(2, "0") + " ##{self.id}"
```
afeld: Thinking leaving out the space will make it more easily searchable, e.g. `FY15-26`

afeld:
Any potential pitfalls of using this identification scheme?

/cc @phirefly @brethauer

afeld:
Otherwise, looks good!

phirefly:
Agree with @afeld on format `FY15-26` instead of having the space and the pound.

afeld:
Needs merge from `master`, but otherwise :shipit:

cmc333333:
Master merged, format updated

PR #320, opened 2015-04-30
[#91211506] make dispatchers/mailer Proposal-centric

afeld:
...in an effort to remove the dependence on the Cart model for clients that don't need it. Should be zero change in functionality.


spec/mailers/communicart_mailer_spec.rb
```diff
@@ -153,13 +155,13 @@ def expect_csvs_to_be_exported
...
-    let(:cart) { FactoryGirl.create(:cart) }
-    let(:comment) { FactoryGirl.create(:comment, proposal: cart.proposal) }
+    let(:proposal) { FactoryGirl.create(:proposal, :with_cart) }
+    let(:comment) { FactoryGirl.create(:comment, proposal: proposal) }
     let(:email) { "commenter@some-dot-gov.gov" }
     let(:mail) { CommunicartMailer.comment_added_email(comment, email) }
 
     it 'renders the subject' do
-      expect(mail.subject).to eq("A comment has been added to Cart ##{cart.id}")
+      expect(mail.subject).to eq("A comment has been added to Cart #13579")
```
cmc333333: Probably shouldn't be hard coded?

cmc333333:
Only a minor comment, but overall great, so I'll merge before all our trees get too far out of date.

PR #323, opened 2015-05-01
Copy-pasta failure

cmc333333:
Due to #321 
screen shot 2015-05-01 at 5 08 00 pm
screen shot 2015-05-01 at 5 09 56 pm


phirefly:
Ship it

PR #324, opened 2015-05-01
Update/name field

nickbristow:
As per https://www.pivotaltracker.com/story/show/93638458

this PR requires a DB migration

c2

### DIFF
--- a/app/models/ncr/work_order.rb
+++ b/app/models/ncr/work_order.rb
@@ -110,6 +110,9 @@ module Ncr
       self.updated_at.to_i
     end
 
+    def name
+      self.project_title
+    end
 
     protected
 
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -65,28 +65,6 @@ ActiveRecord::Schema.define(version: 20150501212215) do
     t.datetime "updated_at"
   end
 
-  create_table "cart_item_traits", force: true do |t|
-    t.text     "name"
-    t.text     "value"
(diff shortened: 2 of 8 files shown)


db/schema.rb
```diff
@@ -65,6 +65,28 @@
...
+    t.text     "notes"
+    t.integer  "quantity"
+    t.text     "details"
+    t.string   "part_number"
+    t.float    "price"
+    t.integer  "cart_id"
+    t.datetime "created_at"
+    t.datetime "updated_at"
+  end
+
```
afeld: Not sure what happened here...I can fix it if you need.

afeld: Your DB might be in a weird state.

app/models/ncr/work_order.rb
```diff
@@ -110,6 +110,9 @@ def version
       self.updated_at.to_i
     end
 
+    def name
+      self.project_title
+    end
```
afeld: Can we get rid of this?

afeld:
Nice work! Just a couple minor things.

app/models/ncr/work_order.rb
cmc333333: Not quite -- `name` is part of the `client_data` interface; it's used in most listing and in a few email templates, iirc.

afeld: Ah ok, no worries then.

PR #325, opened 2015-05-01
[#89626204] replace /approval_response with /proposals/:id/approve

afeld:


afeld:
Thoughts about `/proposal/:id/approve` vs . `/approval/:id/approve`? The latter is *technically* more accurate, though from the user's perspective, the former makes more sense. Not that they will need to care about these URLs, though... Preference?

cmc333333:
I don't see a good reason to surface approval objects as separate entities at this point, so I'd vote for the former. If we ever needed to address an individual approval, I'd expect it to always have the context of either a specific proposal or a specific user, i.e. `/proposal/:proposal_id/approval/:approval_id` and/or `/user/:user_id/approval/:approval_id`

afeld:
Ready for some :eyes:

app/views/proposals/show.html.erb
```diff
@@ -185,11 +185,7 @@
 <% if policy(@proposal).can_approve_or_reject? %>
   <div class="centered">
     <p>
-      <%= link_to "Approve",
-          approval_response_path(approver_action: 'approve',
-                                 # @todo: get rid of cart
-                                 cart_id: @proposal.cart.id, version: @proposal.version
-                                 ), class: 'form-button' %>
+      <%= link_to "Approve", approve_proposal_url(@proposal, version: @proposal.version), class: 'form-button' %>
```
cmc333333: `version` isn't need anymore because it's built into `approve_proposal_url`, yes?

spec/helpers/communicart_mailer_helper_spec.rb
```diff
@@ -1,32 +1,34 @@
 describe CommunicartMailerHelper do
   describe '#approval_action_url' do
     it "returns a URL" do
-      approval = FactoryGirl.create(:approval, :with_cart, :with_user)
+      approval = FactoryGirl.create(:approval, :with_proposal, :with_user)
       token = approval.create_api_token!
+      proposal = approval.proposal
+
+      expect(proposal).to receive(:version).and_return(123)
```
cmc333333: Why mock this out? Wouldn't using the actual version generated be cleaner?

spec/requests/proposals_spec.rb
```diff
@@ -13,5 +13,104 @@
...
+      proposal.reload
+      proposal.approvals.each do |approval|
+        expect(approval.status).to eq(status)
+      end
+      expect(proposal.status).to eq(status)
+    end
+
+    before do
+      allow_any_instance_of(Proposal).to receive(:client).and_return('ACME')
+      allow_any_instance_of(Proposal).to receive(:public_identifier).and_return('123')
```
cmc333333: Why not se these when calling FactoryGirl?

config/routes.rb
```diff
@@ -25,6 +24,11 @@
   }
 
   resources :proposals, only: [:index, :show] do
+    member do
+      get 'approve'
```
cmc333333: Not a result of your changes, but we really should get rid of the `get` version of this url. It'd be pretty easy to send a follow up email to all of my approvers that included an image which, when loaded, approved my request.

app/views/proposals/show.html.erb
afeld: It's added in `approval_action_url` within the mailer helper...not sure if we should combine them, since that needs a URL not a path, and this one shouldn't have a token. Thoughts?

app/helpers/communicart_mailer_helper.rb
```diff
@@ -19,17 +19,14 @@ def auto_login?(user)
...
 
-  def approval_action_url(approval, action = 'approve')
-    opts = {
-      approver_action: action,
-      cart_id: approval.cart_id,
-      version: approval.proposal.version
-
-    }
+  # approve-only at the moment
+  def approval_action_url(approval)
```
afeld: Would love a different name from this method to distinguish it from the built-in route helpers. Prepending `custom_` feels like cheating, but would do the trick.

spec/requests/proposals_spec.rb
afeld: For better or worse, I didn't want to have to create the associated model just to have these two delegated attributes available. Maybe I should just bite the bullet, short of doing https://www.pivotaltracker.com/story/show/93936068.

spec/helpers/communicart_mailer_helper_spec.rb
afeld: Ditto below.

config/routes.rb
afeld: Per our [Slack convo](https://18f.slack.com/archives/cap/p1430921797004017), not going to worry about this for now, since form support in HTML emails isn't available everywhere.

app/helpers/communicart_mailer_helper.rb
cmc333333: Ah, yes I was confused. Maybe `generate_approve_url`?

spec/requests/proposals_spec.rb
cmc333333: Oof. That's no good. Flagging it for being too blunt, but we can merge anyway.

afeld:
Updated!

cmc333333:
A lot of work went into this; great job!

PR #327, opened 2015-05-04
[Delivers #86061736] MVP Dashboard

cmc333333:
1. Creates a minimal, NCR-specific dashboard based on amount spent and raw count
1. Creates a minimal "query" page for proposals. This can be extended later, but currently allows for customizable start/end dates for searching. This is used as a drill-down from the dashboard
1. Adds a secure mechanism for the "back to XXX" link at the top of most pages to be modified on demand. This is then used in the query page (which is generic) to link back to the dashboard page. Probably went overboard here.

Looks like:
screen shot 2015-05-04 at 5 48 55 pm

screen shot 2015-05-04 at 5 49 09 pm

screen shot 2015-05-04 at 5 49 17 pm


cmc333333:
Also, I can rewrite the templates in erb if that's preferred. Haml was just particularly useful with the table structure.

app/controllers/application_controller.rb
```diff
@@ -26,6 +27,25 @@ def authorize(record, query=nil, user=nil)
...
+  def return_to
+    if params[:return_to]
+      return_to = params.require(:return_to)
+      proper_sig = self.make_return_to(return_to.require(:name),
+                                       return_to.require(:path))[:sig]
+      if return_to.require(:sig) == proper_sig
+        return_to.permit([:path, :name])
+      end
+    end
+  end
```
afeld: Are we just trying to replicate the Back button? If so, I say we get rid of the whole "return to X" link.

app/controllers/application_controller.rb
```diff
@@ -26,6 +27,25 @@ def authorize(record, query=nil, user=nil)
     end
   end
 
+  def param_date(sym)
+    begin
+      Date.strptime(params[sym].to_s)
+    rescue
+      nil
+    end
```
afeld: Mind checking for the possible error scenarios, rather than a catch-all `rescue`?

app/controllers/ncr/dashboard_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+        start_date = Date.new(row["year"].to_i, row["month"].to_i, 1)
+        end_date = start_date + 1.month
+        {path: query_proposals_path(start_date: start_date.strftime(),
+                                    end_date: end_date.strftime(),
+                                    return_to: return_params),
+         month: I18n.t("date.abbr_month_names")[start_date.month],
+         year: start_date.year,
+         count: row["count"].to_i,
+         cost: row["cost"].to_f}
+      }
```
afeld: Maybe move these queries into a dedicated class?

app/views/proposals/query.html.haml
```diff
@@ -0,0 +1,9 @@
+.inset
+  %h1
+    Proposals participated in
+    -# A month of time. This can be expanded to include other intervals later
+    - if @start_date && @end_date && @end_date == @start_date + 1.month
+      = "(" + I18n.t("date.abbr_month_names")[@start_date.month]
+      = @start_date.year.to_s + ")"
```
afeld: Umm, maybe these lines would make more sense as named methods in a helper?

afeld:
> I can rewrite the templates in erb if that's preferred

Nah, HAML's good!

app/controllers/application_controller.rb
cmc333333: I would generally agree, but this was part of the story's spec. @phirefly what do you think?

app/controllers/application_controller.rb
cmc333333: Sure

app/controllers/ncr/dashboard_controller.rb
cmc333333: This is a single query, and more or less the only thing that this controller does. I can split it into query and output, though?

app/views/proposals/query.html.haml
cmc333333: Sure; something better at least

phirefly:
+1 using haml

app/controllers/ncr/dashboard_controller.rb
cmc333333: Moved in a later commit

app/controllers/application_controller.rb
phirefly: I'd like to leave it + test it. I do remember one of our testers not having any clue how to get back. This is a pretty lightweight way of offering that. That said, we may find that longer lists will require that this link be placed at the bottom, but placement is good as the first pass.

cmc333333:
Updated per comments

app/controllers/application_controller.rb
afeld: Can we combine this with how `session[:return_to]` is being used in other parts of the app?

app/helpers/return_to_helper.rb
```diff
@@ -0,0 +1,10 @@
+module ReturnToHelper
+  def make_return_to(name, path)
+    sig = OpenSSL::HMAC.digest(
+      OpenSSL::Digest::SHA256.new,
+      Rails.application.secrets.secret_key_base,
+      name + "$" + path
+    )
+    {name: name, path: path, sig: Base64.urlsafe_encode64(sig)}
+  end
+end
```
afeld: This feels like it could be moved out to your first gem!

app/controllers/proposals_controller.rb
```diff
@@ -28,4 +42,17 @@ def auth_errors(exception)
...
+    if @start_date && @end_date
+      # month span
+      if @start_date.mday == 1 && @end_date == @start_date + 1.month
+        month_name = I18n.t('date.abbr_month_names')[@start_date.month]
+        "(#{month_name} #{@start_date.year})"
+      else
+        "(#{@start_date.iso8601} - #{@end_date.iso8601})"
+      end
+    end
+  end
```
afeld: Since it's only used in the view, can we put it in `app/helpers/proposals_helper.rb`?

afeld: Also, I'd take the `start_date` and `end_date` as arguments to make it not rely on external state, and to be more testable.

app/views/proposals/query.html.haml
```diff
@@ -0,0 +1,6 @@
+.inset
+  %h1
+    Proposals participated in
+    = datespan_header
+
+  = client_partial (current_user.client_slug || 'ncr'), "proposal_list", locals: {proposals: @proposals}
```
afeld: Maybe we should move that default into the `client_partial` helper for now?

app/controllers/ncr/dashboard_controller.rb
afeld: Not a blocker, but probably worth making a dedicated DashboardResult class.

spec/features/return_to_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    expect(page).not_to have_content('other')
+  end
+
+  it 'validates sig (path)' do
+    different_path = return_to.merge(path: 'other')
+    visit query_proposals_path(return_to: different_path)
+    expect(page).to have_content('Back to main portal')
+    expect(page).not_to have_content(return_to[:name])
+  end
+end
```
afeld: Does this cover a failure case, passing in an invalid `return_to`? Sorry if I'm missing it.

cmc333333: Yes, "Back to main portal" implies the signature is invalid.

app/views/proposals/query.html.haml
cmc333333: That feels like a step in the wrong direction (towards more indirection). How about defaulting users to having that slug upon creation and migrating existing users? That'd make it painfully obvious what needs to change.

app/controllers/proposals_controller.rb
cmc333333: Sure.

app/controllers/ncr/dashboard_controller.rb
cmc333333: We can make it reusable the second time we need it. I'm pretty sure this will all be wiped out when we get more feedback.

app/controllers/application_controller.rb
cmc333333: `session` variables hang around until explicitly removed, which is not desirable here. We can definitely replace `session[:return_to]` with this GET-based version, though.

spec/features/return_to_spec.rb
afeld: You might consider making the spec descriptions something like "sends them back to the main portal if the path signature is invalid".

app/helpers/return_to_helper.rb
cmc333333: May need some hand holding

app/views/proposals/query.html.haml
afeld: > defaulting users to having that slug 

That doesn't feel right either...even if we have them select their `client` when setting up their account. We would probably want to save their `User` before they make that selection. Suppose they close that page and come back...how would we know whether they are actually NCR or not?

Not any great options here yet, unfortunately :confused:

app/controllers/ncr/dashboard_controller.rb
afeld: Just for readability, I mean.

app/views/proposals/query.html.haml
cmc333333: Going to revert the change which set the default as it causes lots of errors during tests. Notably, listing pages use the NCR listing, which includes fields not present on a non-ncr model.

cmc333333:
Updated per feedback

afeld:
:shipit:

PR #329, opened 2015-05-05
[Delivers #89820292] More emails when editing

cmc333333:
Currently, already-approver approvers are notified when a work order is edited. If a new approver is added, they are also notified, but the logic for that lives inside a model.

This improves on that by moving all of the logic into the NCR dispatcher, and adding a new email to all actionable approvers.


app/mailers/communicart_mailer.rb
```diff
@@ -7,9 +7,11 @@ class CommunicartMailer < ActionMailer::Base
   add_template_helper ClientHelper
 
 
-  def proposal_notification_email(to_email, approval, show_approval_actions=true)
+  def proposal_notification_email(to_email, approval,
+                                  show_approval_actions=true, alert_partial=nil)
```
afeld: Maybe we should use [keyword arguments](https://robots.thoughtbot.com/ruby-2-keyword-arguments)? Would make the calling more readable.

afeld: Come to think of it, the complex method signature might be a hint that we should split the pure notification email up from the one that includes the approval actions, even if they share most of the same logic/partials. Thoughts?

cmc333333: I've split them into two methods (one calling the other) but kept a single template. If we find significant divergence, we can separate the templates further.

lib/ncr_dispatcher.rb
```diff
@@ -13,11 +13,18 @@ def final_approval(proposal)
...
-  # modified
+  # modified. Also notify current approvers that the proposal has been updated
   def on_proposal_update(proposal)
     proposal.approvals.approved.each{|approval|
-      CommunicartMailer.proposal_notification_email(
-        approval.user_email_address, approval, false).deliver
+      CommunicartMailer.notification_for_approver(approval.user_email_address, approval, "already_approved").deliver
+    }
+    proposal.currently_awaiting_approvals.each{|approval|
+      if approval.api_token
```
afeld: What are you *actually* checking for here? Maybe an `approval.notified?` method or something?

cmc333333: Added a comment.

PR #330, opened 2015-05-05
updated verbage for emails and proposal view

nickbristow:
Here is the updated language
### 18f
letteropenerweb

letteropenerweb

letteropenerweb

letteropenerweb
This used to say "Cart"

### NCR

letteropenerweb

letteropenerweb

### DIFF
--- a/app/models/gsa18f/procurement.rb
+++ b/app/models/gsa18f/procurement.rb
@@ -76,7 +76,7 @@ module Gsa18f
 
     # @todo - this is pretty ugly
     def public_identifier
-      self.cart.id
+      "##{self.cart.id}"
     end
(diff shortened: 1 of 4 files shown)


app/models/gsa18f/procurement.rb
```diff
@@ -76,7 +76,7 @@ def client
 
     # @todo - this is pretty ugly
     def public_identifier
-      self.cart.id
+      "##{self.cart.id}"
```
cmc333333: What do you think about using `self.id` ?

nickbristow: No reason not to! @cmc333333

PR #332, opened 2015-05-06
[Delivers #93639720] Validate RWA Number format

cmc333333:
Per the spec. Only thing I'm not too clear on is how rails handles empty values -- should I just be using `allow_nil` (and not testing the empty string)?

### DIFF
--- a/app/models/ncr/work_order.rb
+++ b/app/models/ncr/work_order.rb
@@ -29,6 +29,10 @@ module Ncr
     validates :expense_type, inclusion: {in: EXPENSE_TYPES}, presence: true
     validates :vendor, presence: true
     validates :building_number, presence: true
+    validates :rwa_number, format: {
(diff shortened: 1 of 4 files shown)


app/models/ncr/work_order.rb
```diff
@@ -29,6 +29,10 @@ class WorkOrder < ActiveRecord::Base
     validates :expense_type, inclusion: {in: EXPENSE_TYPES}, presence: true
     validates :vendor, presence: true
     validates :building_number, presence: true
+    validates :rwa_number, format: {
+      with: /[a-zA-Z][0-9]{7}/,
```
afeld: @phirefly Can we look at a list of RWAs, or ask someone to double-check that this format is correct?

cmc333333: Format is per spec from Anthony

PR #335, opened 2015-05-06
[#93698442] Script for importing 18f users

cmc333333:
Call like:
```sh
rake import_users:team_yaml DIR=/path/to/data-private/team/
```

This did not include a CSV importer, but you can imagine how similar it'd be


afeld:
Clever! Didn't occur to me to use the `data-private`.

PR #340, opened 2015-05-07
[Finished #93639564] Display the time for comments

phirefly:
Displaying time for comments. Decided to display in ET since our first group is in that timezone. Also displaying the month, day, year in a more readable format since we have the space.

screen shot 2015-05-07 at 5 05 10 pm


cmc333333:
I think I like @afeld 's idea of "X time ago" a bit better, but this is a step in the right direction.

PR #344, opened 2015-05-09
Disable/Enable comment button if text field is empty

ssing128:
Disable comment button if text field is empty.
Enable comment button if text field has value.


app/assets/javascripts/forms.js.coffee
```diff
@@ -53,8 +53,18 @@ class Selectizer
...
     selectizer.loadOptionsIfRemote()
+
+  #Disable/Enable button if textbox is empty
+  $('#add_a_comment').attr 'disabled', true
+  $('#comment_comment_text').keyup ->
+    disable = false
+    if $(this).val() == ''
+      disable = true
+    $('#add_a_comment').prop 'disabled', disable
+    return
```
afeld: Do you need the explicit return?

app/assets/javascripts/forms.js.coffee
```diff
@@ -53,8 +53,18 @@ class Selectizer
...
     selectizer.enable()
     selectizer.loadOptionsIfRemote()
+
+  #Disable/Enable button if textbox is empty
+  $('#add_a_comment').attr 'disabled', true
+  $('#comment_comment_text').keyup ->
+    disable = false
+    if $(this).val() == ''
+      disable = true
+    $('#add_a_comment').prop 'disabled', disable
```
afeld: I think you want to use `attr()` again (though I find the distinction somewhat confusing).

afeld: Sorry, I'm wrong...it should use `prop()` both places.

> The `.prop()` method should be used to set disabled and checked instead of the `.attr()` method

http://api.jquery.com/prop/#prop2

afeld:
Sorry for the slow response – thanks for doing this! Just a couple small things, and also needs a merge from `master`.

```bash
git pull https://github.com/18F/C2.git master
```

afeld:
Looks good to go! Still needs a merge from `master` – see instructions above.

afeld:
:tada: :rocket: :guitar:

PR #346, opened 2015-05-10
Use inheritance and a JSON field rather than separate tables

cmc333333:
Rather than having two distinct objects (proposal and work order/procurement), this makes work order/procurements instances of proposals. That allows all of the indirection, both forward (from proposal to `client_data`) and backwards (from work order/procurement to `proposal`) to vanish. It also opens the possibility of dynamically-specified proposal fields.

As active record takes hints from the database, conversions between certain types (e.g. numbers, which rails encodes as strings, and dates) must be more explicit.

Todo:
* [ ] ~~Remove Cart completely~~
* [ ] ~~Update gherkin test to not specify public identifier, as that's not core proposal functionality~~
* [x] Update to allow for the lingering Cart model
* [x] Update API docs
* [x] General cleanup
* [ ] Migrate existing data & remove old fields/tables


app/models/ncr/work_order.rb
```diff
@@ -33,31 +23,31 @@ class WorkOrder < ActiveRecord::Base
...
-    # @todo - this is an awkward dance due to the lingering Cart model. Remove
-    # that dependence
-    def init_and_save_cart(approver_email, requester)
-      cart = Cart.create(
-        proposal_attributes: {flow: 'linear', client_data: self,
-                              requester: requester}
-      )
-      self.add_approvals(approver_email)
-      Dispatcher.deliver_new_proposal_emails(proposal)
-      cart
```
afeld: :fire: :clap:

db/migrate/20150510022120_add_client_fields_to_proposal.rb
```diff
@@ -0,0 +1,6 @@
+class AddClientFieldsToProposal < ActiveRecord::Migration
+  def change
+    add_column :proposals, :client_fields, :json
+    add_column :proposals, :subclass, :string
```
afeld: I'd use a `type` column to have Rails handle the subclass conversion automatically.

http://api.rubyonrails.org/classes/ActiveRecord/Inheritance.html

spec/features/version_check_spec.rb
```diff
@@ -1,12 +1,12 @@
 describe "Version check" do
-  let(:proposal) { FactoryGirl.create(:proposal, :with_approvers, :with_cart) }
+  let(:proposal) { FactoryGirl.create(:proposal, :with_approvers) }
 
   it "occurs if the cart is modified in after seeing the profile page" do
     login_as(proposal.approvals.first.user)
     visit "/proposals/#{proposal.id}"
 
     sleep 1.second  # wait to get a new update time
```
afeld: TODO use Timecop.

afeld:
There's some really good refactoring in here – would you mind if we pull the things that don't require schema changes into a separate PR, so we can keep this one focused? I know that's a bit of a pain – I'm happy to do the grunt work if you like.

> That allows all of the indirection, both forward (from proposal to client_data) and backwards (from work order/procurement to proposal) to vanish

I think a lot of that can be made transparent via delegation...see [`ProposalDelegate`](https://github.com/18F/C2/blob/master/app/models/concerns/proposal_delegate.rb), which hasn't been `include`d in the use case models yet.

---

I'm still pretty reluctant about storing the properties as JSON. Below are the concerns I have with not using real columns:

* Gem compatibility
* Migrations become harder
    * No equivalent of `rename_column`
* Do the properties have full query support?
    * e.g. equivalent of `Ncr::WorkOrder.where('amount > ?', 1000)`
* Our RDS databases would need to be upgraded to 9.4 to take advantage of `jsonb`, which is more performant than `json` post-insertion.
    * https://twitter.com/schneems/status/598143711680180224
* Can ActiveRecord validations be used for the `json` properties?

Since I haven't used `json` in Postgres before, I don't really know what the tradeoffs are/aren't. Would love some thoughts from the broader team.

/cc @catherinedevlin @arowla 

---

Useful links:

* https://www.compose.io/articles/is-postgresql-your-next-json-database/
* https://twitter.com/aidanfeldman/status/598139771391594496
* https://18f.slack.com/archives/dev/p1431444650000121 ff.

cmc333333:
> Gem compatibility

It could go either way. ActiveRecord sure likes relying on a DB schema, but that doesn't mean that's a good solution.
 
> Migrations become harder

Kind of. Migrations certainly become different, but if we don't think of client-specific fields as we do database schemas, it's not that radical

> No equivalent of rename_column

Right, data would need to be versioned. If we allow the client to change the schema at will, though, rename_column wouldn't do much good

> Do the properties have full query support?

Define "full". Fields can be [treated](http://www.postgresql.org/docs/9.3/static/functions-json.html) as integers, strings, and I believe booleans

> Our RDS databases would need to be upgraded to 9.4 to take advantage of jsonb, which is more performant than json post-insertion.

That's not really a criticism, that's an optimization.

> Can ActiveRecord validations be used for the json properties?

Yes; they're being used in the PR.

cmc333333:
> I think a lot of that can be made transparent via delegation...see ProposalDelegate, which hasn't been included in the use case models yet.

That's magic, which is the opposite of transparency. Delegation is also what's led to numerous explosions in the app, as the controller/template/etc. believes that a particular type of `client_data` is present, when it is not.

afeld:
> ActiveRecord sure likes relying on a DB schema, but that doesn't mean that's a good solution.

Regardless, it's what we've got and what's well supported in Rails, so moving off of it would not be a good idea or worthwhile IMHO.

> Migrations certainly become different

It means we're going to need to start doing migrations by looping through records, which is a lot more fragile than single `ALTER TABLE` statements.

> data would need to be versioned

This sounds like a pain...

> If we allow the client to change the schema at will

I think the major disagreement here is that I don't see this as a goal in the forseeable future, and trying to future-proof for it adds unnecessary burden (or at least is a zero-sum trade).

> > Our RDS databases would need to be upgraded to 9.4
>
> That's not really a criticism

I just mean that it's extra work, for what I don't see as a lot of gain.

> the opposite of transparency

You're right, "transparency" wasn't the right word. What I meant is "if you're not within one of the models or the use-case-specific templates, you shouldn't have to care."

> Delegation is also what's led to numerous explosions in the app, as the controller/template/etc. believes that a particular type of client_data is present, when it is not.

I don't see how using the embedded `json` is any different. It's currently expecting a relation, this PR makes it expect an embedded object with certain properties.

cmc333333:
> Regardless, it's what we've got and what's well supported in Rails, so moving off of it would not be a good idea or worthwhile IMHO.

If our use case required something that Rails doesn't support well, we need to work around Rails.

> It means we're going to need to start doing migrations by looping through records, which is a lot more fragile than single ALTER TABLE statements.

Only if we need to change the schema of client data. I'm saying: don't view client data as a static schema and that's no longer a big problem.

> This sounds like a pain...

Sure. There's always cost to design decisions. I'd argue it's painful now to have two disparate objects acting as if they were one. It's also a pain to set up new clients and to run `Proposal` tests in isolation.

> I think the major disagreement here is that I don't see this as a goal in the forseeable future, and trying to future-proof for it adds unnecessary burden (or at least is a zero-sum trade).

Right. I don't think making a new app per client use case is feasible. It's slow, expensive, and unnecessarily constrains our users. I'd rather build the plumbing than build several dozen one-off apps.

> I just mean that it's extra work, for what I don't see as a lot of gain.

That'd be like adding a database index. It's extra work, but may be worth the effort.

> You're right, "transparency" wasn't the right word. What I meant is "if you're not within one of the models or the use-case-specific templates, you shouldn't have to care."
> I don't see how using the embedded json is any different. It's currently expecting a relation, this PR makes it expect an embedded object with certain properties.

The errors _are_ in use-case-specific templates. The underlying assumption that the templates made was wrong. The data they were presenting wasn't what they expected. We should force that expectation of flexible fields throughout the app so that these types of errors don't pop up.

afeld:
I don't feel like we're coming any closer to agreement on this...not sure how to proceed :confused: At the very least, are you ok with [splitting the PR](https://github.com/18F/C2/pull/346#issuecomment-101361658)?

cmc333333:
I'm not sure how much can be gleaned without the underlying inheritance change, but go for it.

cmc333333:
Closing as I haven't kept this branch up to date. Will reopen a related issue soon

PR #347, opened 2015-05-11
Added target "C2" to email links so they open in same tab

nickbristow:


app/views/ncr/_proposal_mail.html.erb
```diff
@@ -3,7 +3,7 @@
     <tr>
       <td>
         <h1 class="communicart_header"> 
-          <%= link_to proposal.name, proposal_url(proposal),  {target: '_blank', 'aria-label' => 'opens in a new window'} %>
+          <%= link_to proposal.name, proposal_url(proposal),  {target: 'C2', 'aria-label' => 'opens in a new window'} %>
```
cmc333333: "Opens in a new window" might not be accurate now, right?

PR #348, opened 2015-05-11
Added "Discard Changes" to edit request pages"

nickbristow:
ncr_request_-_c2

This only appears when editing a request. The link takes the user to the proposals show page.


app/views/gsa18f/procurements/form.html.erb
```diff
@@ -89,5 +89,8 @@
       <%= f.text_area :additional_info, class: 'form-control' %>
     </div>
     <%= f.submit "Submit for approval", class: 'btn btn-default' %>
+    <% if !@procurement.id.nil? %>
```
cmc333333: Might be cleaner with [`persisted?`](http://apidock.com/rails/v4.2.1/ActiveRecord/Persistence/persisted%3F)

PR #349, opened 2015-05-12
[Delivers #94399790] Account for proposal lists including mixes of proposals

cmc333333:
While this wouldn't affect normal users so much, devs often have proposals of various types in their proposal list. Don't let that lead to an error.


afeld:
Code looks good, but wondering about the best approach. Should we be handling this in the template, or make the query in the controller only return the appropriate Proposals?

cmc333333:
This is a bandaid solution until we determine a longer-term strategy for dealing with users with multiple proposal types. I'm not sure that limiting which proposals they see is a good solution.

PR #351, opened 2015-05-13
[Delivers #93639788] update building list

afeld:


cmc333333:
Glad we've got instructions now. Maybe we'll add a cron script to keep the list up-to-date at some point.

afeld:
I was gonna write a rake task or something, but you need to a) be on the GSA network and b) enter a CAPTCHA, so *meh*.

afeld:
Also I can't imagine the list changes terribly frequently.

PR #352, opened 2015-05-13
Fixed discard changes bug, add tests

nickbristow:
Noticed on dev the discard changes was using work order id, when it needed proposal. 

Fixed that and added a test.


cmc333333:
Another example of the `client_data` being separate from `proposal` biting us ;)

PR #353, opened 2015-05-14
[Delivers #94019990] Be more careful when approving via GET

cmc333333:
Using a GET-able URL when performing an action is problematic due to the confused deputy problem. Imagine a malicious requester crafted an email with an `img` with the src pointing to the approval url. If an approver loaded this "image", their browser would be acting on their behalf to approve the request.

This limits the surface area of this attack a bit by generally forcing a POST to occur. The exception is that the "approve" button in an email cannot be a POST and work correctly in most mail readers. Since this had to be a GET, we single out this exception by adding a special check that a token is present.

Continues to look like:
screen shot 2015-05-13 at 9 01 57 pm

### DIFF
--- a/app/assets/stylesheets/common/communicarts.scss
+++ b/app/assets/stylesheets/common/communicarts.scss
@@ -281,7 +281,7 @@ p {
   display: inline-block;
   margin-top: 15px;
   padding: 15px 25px;
-  border-bottom: 2px solid #095ba7;
+  border: 2px solid #095ba7;
   border-radius: 5px;
   background-color: #0d66bf;
   color: black;
--- a/app/controllers/concerns/token_auth.rb
+++ b/app/controllers/concerns/token_auth.rb
@@ -17,6 +17,13 @@ module TokenAuth
     rescue_from Pundit::NotAuthorizedError, with: :auth_errors
   end
 
+  # As a security precaution, certain actions must be POSTed to.
+  def needs_token_on_get
+    if request.get?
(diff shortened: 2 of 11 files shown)


cmc333333:
Hold off on merging this -- need to think through the delegation case

cmc333333:
Updated to account for the delegation case. Instead of checking for delegation when sending the email, the check is performed when trying to act on the token.

afeld:
Thinking through if there are downsides of including the token when there are delegates...only thing I can come up is
 
1. Approver forwards the proposal notification (with token) to their delegate
1. Approver removes that delegate
1. Delegate clicks approval email
1. Delegate is signed in as the approver
 
I suppose that's no worse than any other email with a token getting forwarded, regardless of whether there are delegates or not.

Do you prefer this to the POST-via-JS or confirmation button approach? Overall seems fine to me, but this is probably a flow that could use some manual QA.

cmc333333:
That's definitely a hole stemming from treating login credentials and "you-came-from-an-email" checks as the same thing. 

The JS-POST technique is not generally a good option as it leads both to user confusion (as they see a flash of a page) and breaks a RESTful contract (a GET is taking an action). If our use case didn't demand action-from-email, I'd say that a confirmation screen would have been the best option.

I'll take another look at this and see if there's not a way to separate out authentication from "I come from email" checks.

afeld:
Meh, we can iterate. Tally-ho!

PR #354, opened 2015-05-14
[Delivers #93639592] Disable submit when nothing is present

cmc333333:
This extends #344 by
* changing the style of disabled buttons so that it's obvious they're disabled
* abstracting the disable-checking code into a declarative style
* using that style to re-use the code for file submission
* adding tests

Also fixes a reference to a non-existent image. Looks like:

screen shot 2015-05-14 at 12 27 55 pm

screen shot 2015-05-14 at 12 28 38 pm


phirefly:
Nice. :shipit:

PR #355, opened 2015-05-14
[Delivers #93638088] allow users to search for Proposals

afeld:
Search options are documented here:

https://docs.google.com/document/d/1q3dsdkqAKjK_XDohjTtBdiaxduaEnm3KLHrKMtg94rI/edit#

Ended up going with the simplest possible Postgres Full Text Search.

### DIFF
--- a/lib/proposal_search.rb
+++ b/lib/proposal_search.rb
@@ -22,7 +22,7 @@ class ProposalSearch
         SELECT
           proposals.id AS pid,
           (
-            setweight(to_tsvector(to_char(proposals.id, '999')), 'A') ||
+            setweight(to_tsvector(proposals.id::text), 'A') ||
             setweight(to_tsvector(coalesce(ncr_work_orders.project_title, '')), 'B')
             -- TODO handle additional properties
           ) AS document
--- a/lib/proposal_search.rb
+++ b/lib/proposal_search.rb
@@ -38,7 +38,6 @@ class ProposalSearch
         LEFT OUTER JOIN ncr_work_orders ON
           ncr_work_orders.id = proposals.client_data_id AND
           proposals.client_data_type = 'Ncr::WorkOrder'
-        GROUP BY proposals.id, ncr_work_orders.id, gsa18f_procurements.id
       ) p_search
       ON proposals.id = p_search.pid
     SQL
--- a/lib/proposal_search.rb
+++ b/lib/proposal_search.rb
@@ -1,4 +1,12 @@
-# http://blog.codeclimate.com/blog/2012/10/17/7-ways-to-decompose-fat-activerecord-models/
+# Query logic modified from
+#
+#   http://blog.lostpropertyhq.com/postgres-full-text-search-is-good-enough/#ranking
+#
+# Note that all search computation is done at runtime, so may need to use one or more VIEWs or INDEXes down the road to make it more performant. Query object based on
+#
+#   http://blog.codeclimate.com/blog/2012/10/17/7-ways-to-decompose-fat-activerecord-models/
+#
+# TODO sanitize query
 class ProposalSearch
   attr_reader :relation
 
@@ -6,15 +14,7 @@ class ProposalSearch
     @relation = proposals
   end
 
-  def execute(query)
-    # Modified from
-    #
-    #   http://blog.lostpropertyhq.com/postgres-full-text-search-is-good-enough/#ranking
-    #
-    # Note that all search computation is done at runtime, so may need to use one or more VIEWs or INDEXes down the road to make it more performant.
-
-    # TODO sanitize query
-
+  def joined
     join = <<-SQL
       INNER JOIN (
         -- TODO handle other use case models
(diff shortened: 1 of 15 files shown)


cmc333333:
This is a good start! Some suggestions:

* This might operate better as a manipulation of an existing queryset -- i.e. you want to filter and order an existing query rather than trying to merge the two queries later
* This currently conflates filtering with ordering. Using an `in` clause doesn't inherently affect the order of the results, so the sort is misleading
* Generating all of the term vectors at query time will be very taxing -- this transformation has to be performed on all relevant columns, on all rows for each query. It might be better to precompute that by adding a term vector column. Of course, if this column exists, it can/should have an index.
* Weighing certain columns over others will be a useful feature in the future, but I wonder if it's only hiding the logic in the query right now
* Doing all of the joins and listing the individual fields will become very painful as the number of clients expands (both to implement and at run time). If you go the precomputed column route, you could have each subtype insert its own values into that term vector column. Responsibility, then, falls on the subtype rather than the querier
* It looks like there are some existing gems, e.g. [pg_search](https://github.com/Casecommons/pg_search), which might make this much easier for you

I'll posit that this is another argument for flexible fields rather than distinct tables. The term vector _couldn't_ be specified statically, so we'd be forced to design a solution which worked for everyone out of the box.

cmc333333:
Obviously not important for the first release, but comment text is another blob to search over. Maaaaybe file name, too, for attachments.

afeld:
Great feedback, thanks.

> Using an `in` clause doesn't inherently affect the order of the results, so the sort is misleading

D'oh! Good catch.

> Generating all of the term vectors at query time will be very taxing

Totally. Doing a red-green-refactor process here, and doing the vectors at runtime means I don't need to redo a migration every time I tweak.

> Weighing certain columns over others will be a useful feature in the future, but I wonder if it's only hiding the logic in the query right now

I mean, if I search "12", I would definitely want Proposal #12 to appear before one with a description of "12-foot pipe".

> you could have each subtype insert its own values into that term vector column

Yeah, definitely leaning that direction, since common attributes like the `name` are all defined in the model.

> It looks like there are some existing gems

Yeah, I've been looking at that and [texticle](http://texticle.github.io/texticle/)...just want to understand how Full Text Search works before knowing which gem/strategy would be preferable.

## TODOs

* [ ] Add search UI
* [ ] Fix search ranking
* [ ] Decide what fields should be searchable
* [ ] Investigate using a VIEW vs. separately inserted records

afeld:
Allrighty, I think this is ready for consideration! The search currently requires a table scan, so I looked into a few options to make it more performant:

* `gin` indexes
* Materialized views
* Triggers
* `after_save` callbacks to insert from the model

We may want to switch to some combination of these at some point, but searching over 1000 records on my local machine only took 14ms, so I'm not worried about search performance (at the expense of extra bookkeeping code) in the immediate future. Happy to address in a separate PR in the next sprint if it should be a priority.

Here's the UI, which leverages the existing "query" page:

screen shot 2015-05-18 at 2 02 30 am

I've added the search box to the proposal `show` and `index` pages, too. Thanks!

lib/proposal_search.rb
```diff
@@ -0,0 +1,65 @@
...
+            setweight(to_tsvector(coalesce(gsa18f_procurements.additional_info, '')), 'C')
+          ) AS document
+        FROM proposals
+        LEFT OUTER JOIN gsa18f_procurements ON
+          gsa18f_procurements.id = proposals.client_data_id AND
+          proposals.client_data_type = 'Gsa18f::Procurement'
+        LEFT OUTER JOIN ncr_work_orders ON
+          ncr_work_orders.id = proposals.client_data_id AND
+          proposals.client_data_type = 'Ncr::WorkOrder'
+        GROUP BY proposals.id, ncr_work_orders.id, gsa18f_procurements.id
```
cmc333333: Is this `group by` needed? I don't see any aggregations

lib/proposal_search.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def filtered(query)
+    self.joined.where('p_search.document @@ plainto_tsquery(?)', query)
+  end
+
+  def ordered(query)
+    sanitized_query = ActiveRecord::Base::sanitize(query)
+    ordering = <<-SQL
+      ts_rank(p_search.document, plainto_tsquery(#{sanitized_query})) DESC
+    SQL
```
cmc333333: This works, but is a little messy. Something to consider for later: I bet activerecord has a way to surface the rank through a `SELECT`. Something like
```ruby
query = self.filtered(query)
query = query.add_select('ts_rank(p_search.document, plainto_tsquery(?)) AS rank', query)
query.order('rank DESC')
```

afeld: Ah, good call. I had tried this before and it complained, but I think that's because I was trying to use the computed value in a `WHERE` clause instead of doing that `JOIN`.

http://stackoverflow.com/a/10099006/358804

lib/proposal_search.rb
afeld: Good catch – I think it was left over from a previous version.

afeld:
Updated!

cmc333333:
Very nice

PR #356, opened 2015-05-14
Error message is focused when it appears

nickbristow:


cmc333333:
My one worry here is that this overrides any focus the user _would_ have had. I'm not sure if that's actually a problem or not...

nickbristow:
This should only run when the page loads, so it really just moves focus
from the top of the page to the error.

On Thursday, May 14, 2015, CM Lubinski <notifications@github.com> wrote:

> My one worry here is that this overrides any focus the user *would* have
> had. I'm not sure if that's actually a problem or not...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/18F/C2/pull/356#issuecomment-102239106>.
>


-- 
Nick Bristow
18f <http://18f.gsa.gov> Innovation Specialist
Nick.Bristow@gsa.gov

cmc333333:
Right, I'm wondering about the cases where the page would have loaded and jumped to a particular div (are there other cases where the focus would be set on page load?). Again, I'm not sure it's a problem, just thinking aloud.

cmc333333:
We'll test with live users ;)

PR #358, opened 2015-05-15
[Delivers #93997522] Don't rely on rails magic to POST/DELETE

cmc333333:
Logging a user out used to require Javascript to convert a link into a POST request. When the home page was replaced, that magic disappeared, so logging out became broken.

Instead of fixing the Javascript, allow everyone visiting `/logout` to be logged out.


cmc333333:
Hold off on merging - it looks like `button_to` could be used and the POST kept.

cmc333333:
Updated to use `button_to` for both the logout action (POST) and the attachment deletion action (DELETE).

This required a little styling love to work well, particularly with the heavily-styled home page.

PR #359, opened 2015-05-18
[#93640022] Add rake task for importing one user

cmc333333:
Not a big departure from rails console, but a bit easier to script if needed.

### DIFF
--- a/lib/tasks/import_users.rake
+++ b/lib/tasks/import_users.rake
@@ -18,9 +18,9 @@ namespace :import_users do
   task team_yaml: :environment do
     dir = ENV['DIR']
     if !dir
-      puts "DIR must be specified. e.g. rake import_users:team_yaml DIR=/path/to/data-private/team/"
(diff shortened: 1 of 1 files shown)


lib/tasks/import_users.rake
```diff
@@ -30,4 +30,27 @@ namespace :import_users do
       Dir.glob(dir + "private" + File::SEPARATOR + "*.yml").each(&method(:gsa18f_yaml_file))
     end
   end
+
+  task one: :environment do
+    email = ENV['EMAIL']
+    if !email
+      puts 'EMAIL must be specified. e.g. rake import_users:one EMAIL=anna.smith@some.gov FIRST=Anna LAST=Smith CLIENT=gsa18f'
```
afeld: Minor: maybe `raise` so that the command fails?

afeld:
Not sure that I totally understand the need here, but the code looks just fine!

cmc333333:
Updated. This should make setting individual user names/creating individuals a bit easier.

PR #361, opened 2015-05-20
[Delivers #91211506 #93936068] remove Cart dependency from client models

afeld:
I started this branch a week or so ago but put it aside for a while...this PR removes the Cart dependence from the WorkOrder and Procurement models. There's some other light refactoring, mostly taking advantage of the `ProposalDelegate` mixin and some better variable names. Sorry it's so monstrous :grin:


spec/factories/gsa18f/procurements.rb
```diff
@@ -6,13 +6,7 @@
...
     office Gsa18f::Procurement::OFFICES[0]
     urgency Gsa18f::Procurement::URGENCY[0]
-    trait :with_proposal do
-      proposal
-    end
-
-    trait :with_cart do
-      association :proposal, :with_cart
-    end
+    proposal
```
cmc333333: flow should be 'linear', hey?

cmc333333:
Glad to see `Cart` be removed from more places.

Something to consider, with all of the nil wrappers for `client_data` in `Proposal`, you've created a pseudo base class. Part of my argument in #346 is that it's simpler to use language-supported inheritance.

spec/factories/gsa18f/procurements.rb
afeld: Good catch!

afeld:
Updated!

> you've created a pseudo base class...it's simpler to use language-supported inheritance

That's fair. I still anticipate [other tradeoffs](https://github.com/18F/C2/pull/346#issuecomment-101361658), but this PR hopefully narrows the gap so that code outside of the model doesn't have to care as much about how the data is persisted.

PR #363, opened 2015-05-20
[Delivers #91726006] remove exporters

afeld:
a.k.a. the CSV attachments on the final emails.


cmc333333:
Makes sense, and would be super easy to recreate if needed.

PR #364, opened 2015-05-20
[Delivers #94394824] Limit Proposals Approvers Can See

cmc333333:
Previously, approvers (/delegates) could see any proposal they were associated with, regardless of whether or nor they could take action on the proposal. This patch limits the proposals an approver sees to just those they have acted on before or can current act on.

It does this by adding an "actionable" state to the approval state machine. Approvals start as "pending", but move to "actionable" as soon as the approver can take action. From there, they move to approved/rejected. This state can be queried directly.

Parallel carts have all approvals start in "actionable" state; linear carts have only the first. As approvals come in, they trigger the "next" approval to move into the "actionable" state.

This change also allowed the removal of `ParallelDispatcher`, as its logic became redundant.


app/models/approval.rb
```diff
@@ -1,7 +1,16 @@
...
+    state :pending do
+      event :make_actionable, transitions_to: :actionable
+    end
+    state :actionable do
+      event :approve, transitions_to: :approved
+      event :reject, transitions_to: :rejected
+    end
+    state :approved
+    state :rejected
+  end
```
afeld: Everything's so simple (a `ThreeStateWorkflow`) until it isn't :smirk:

app/models/proposal.rb
```diff
@@ -80,7 +97,11 @@ def users
   # returns the Approval
   def add_approver(email)
     user = User.for_email(email)
-    self.approvals.create!(user_id: user.id)
+    if self.parallel? || self.approvals.empty?
+      self.approvals.create!(user_id: user.id, status: 'actionable')
+    else
+      self.approvals.create!(user_id: user.id)
+    end
```
afeld: I don't think we want to do this distinction inside of this method...`actionable` should be set on the first Approval for a Proposal based on the `position`, not the order in which the Approval is added.

afeld: ...and that way the logic can be shared with `#restart`.

app/models/proposal.rb
```diff
@@ -160,12 +169,30 @@ def on_rejected_entry(prev_state, event)
 
   def restart
     # Note that none of the state machine's history is stored
+    # Further note: this isn't used anywhere
```
afeld: What do you mean? `expires_at` is checked by `#expired?`, as well as the `expired`/`unexpired`/`fresh` scopes on the ApiToken.

app/policies/proposal_policy.rb
```diff
@@ -33,19 +33,12 @@ def observer!
...
-    self.actionable_approvers.include?(@user)
+    @proposal.currently_awaiting_approvers.include?(@user)
   end
 
   def pending_delegate?
-    # TODO convert to SQL
-    self.actionable_approvers.any? do |approver|
-      approver.outgoing_delegates.exists?(assignee_id: @user.id)
-    end
+    ApprovalDelegate.where(assigner_id: @proposal.currently_awaiting_approvers, assignee: @user).exists?
```
afeld: :clap:

app/models/proposal.rb
cmc333333: I don't follow

app/models/proposal.rb
cmc333333: There is no interface for restarting a proposal

app/policies/proposal_policy.rb
```diff
@@ -81,19 +74,13 @@ def resolve
...
-          ON approval_delegates.assigner_id = approvals.user_id
-          -- ...on the proposal...
-          WHERE approvals.proposal_id = proposals.id
-          -- ...who has the specifified user as a delegate
-          AND approval_delegates.assignee_id = :user_id
+          SELECT * FROM approvals
+          LEFT JOIN approval_delegates ON (assigner_id = user_id)
+          WHERE proposal_id = proposals.id
+            AND status <> 'pending'
+            AND (user_id = :user_id OR assignee_id = :user_id)
```
afeld: Surprised you don't need to qualify the column names here. Nice refactor!

app/views/shared/_email_status.html.erb
```diff
@@ -16,16 +16,17 @@
           <table class="approval_details data_container">
             <%- last_status = nil %>
             <%- @proposal.approvals_in_list_order.each_with_index do |approval,index| %>
+              <%- display_status = approval.actionable? ? 'pending' : approval.status %>
```
afeld: Minor: maybe worth moving to a helper/decorator?

lib/linear_dispatcher.rb
```diff
@@ -8,19 +8,10 @@ def next_pending_approval(proposal)
...
-  def deliver_new_proposal_emails(proposal)
-    self.email_next_pending_approver(proposal)
-    super
-  end
-
   def on_approval_approved(approval)
-    self.email_next_pending_approver(approval.proposal)
+    if next_approval = self.next_pending_approval(approval.proposal)
+      self.email_approver(next_approval)
+    end
```
afeld: I kind of liked the descriptive method name...

spec/lib/linear_dispatcher_spec.rb
```diff
@@ -17,21 +17,21 @@
     end
 
     it "returns the first pending approval by position" do
-      proposal.approvals.create!(position: 6)
-      last_approval = proposal.approvals.create!(position: 5)
+      proposal.approvals.create!(position: 6, status: 'actionable')
+      last_approval = proposal.approvals.create!(position: 5, status: 'actionable')
```
afeld: They shouldn't both be `actionable`, right? Also the description needs to be updated.

app/models/proposal.rb
afeld: I mean that the `restart` method below is basically doing the same thing as these new lines, so we should move it out to a `set_actionable_approvals` method or something.

app/models/proposal.rb
afeld: https://github.com/18F/C2/blob/ec6c27ea30755e52b1e26b1f617c6d7fb9e01b5d/app/controllers/gsa18f/procurements_controller.rb#L37

cmc333333: My bad. I'll remove the comment and add a ticket to evaluate - I'm guessing I'm not the only one who thought NCR and 18f operated the same way when editing.

app/models/proposal.rb
cmc333333: Can do

lib/linear_dispatcher.rb
cmc333333: I don't think `self.email_approver(next_approval)` is less descriptive than `self.email_next_pending_approver`. I can change the variable name to `next_pending_approval` if you really want?

Removing this indirection function reduces the surface area of this class, which improves encapsulation. When refactoring in the future, we won't need to see "is anyone calling this method" before changing it.

spec/lib/linear_dispatcher_spec.rb
cmc333333: I can update the description. This is specifically testing the order, not the `actionable` flag, so approvals are created unusually.

app/views/shared/_email_status.html.erb
cmc333333: Makes sense

cmc333333:
Updated. @afeld let me know what you think of `initialize_approvals`. It feels too heavy to me, like it's asking too much of the caller.

afeld:
Sorry for taking so long to re-review! Just the one question...otherwise, good to merge!

> let me know what you think of `initialize_approvals`. It feels too heavy to me, like it's asking too much of the caller.

I hear ya, though I don't have a better idea. We can iterate.

app/models/proposal.rb
```diff
@@ -80,7 +97,17 @@ def users
   # returns the Approval
   def add_approver(email)
     user = User.for_email(email)
-    self.approvals.create!(user_id: user.id)
+    approval = self.approvals.create!(user_id: user.id)
+    approval
+  end
+
+  def initialize_approvals()
+    if self.linear? && self.approvals
```
cmc333333: I thought they'd evaluate the same, but yep, I meant `self.approvals.any?` Will update.

This check is necessary as we only want the _first_ linear approval to be marked as `actionable` -- the others will enter this state as approvals come in.

PR #366, opened 2015-05-20
Feature/changes tracked

nickbristow:
A special comment is added to the proposal to track changes.


phirefly:
@nickbristow Can you add a screengrab, please? Thx!

afeld:
Did you and @cmc333333 end up pairing on this the other day? Let me know when it's ready for :eyes:!

nickbristow:
@afeld yeah, we paired, refactoring

nickbristow:
I had to change the method for saving the comment to a manual approach to get it to have the right content in the email

app/models/proposal.rb
```diff
@@ -167,5 +167,11 @@ def restart
     Dispatcher.deliver_new_proposal_emails(self)
   end
 
+  def changed_fields comment_text
+    Comment.new({comment_text: comment_text,
```
nickbristow: I had to change the method for saving the comment to a manual approach to get it to have the right content in the email

nickbristow:
paint_the_blue_walls_-_c2

afeld:
Cool, just leave a comment here when you'd like feedback!

nickbristow:
go for it @afeld would love feedback

app/controllers/ncr/work_orders_controller.rb
```diff
@@ -39,9 +39,12 @@ def update
...
       @work_order.assign_attributes(permitted_params)   # don't hit db yet
       @approver_email = params[:approver_email]
-
       if self.errors.empty?
-        @work_order.save
+        if @work_order.changed?
+          comment_text = @work_order.changed_attributes_comment
+          @work_order.proposal.changed_fields(comment_text)
+          @work_order.save
+        end
```
afeld: Can we do this in an `after_update` callback on the model? Just so there's less logic needed in the controller.

http://api.rubyonrails.org/classes/ActiveRecord/Callbacks.html

app/models/ncr/work_order.rb
```diff
@@ -13,6 +13,9 @@ def self.table_name_prefix
 
   class WorkOrder < ActiveRecord::Base
     include ObservableModel
+    include ActiveModel::Dirty
```
afeld: I think this is included with `ActiveRecord::Base`.

app/models/ncr/work_order.rb
```diff
@@ -13,6 +13,9 @@ def self.table_name_prefix
...
+    include ValueHelper
+    # TODO include ProposalDelegate
```
afeld: Mind removing this? Included below.

app/models/proposal.rb
```diff
@@ -167,5 +167,11 @@ def restart
     Dispatcher.deliver_new_proposal_emails(self)
   end
 
+  def changed_fields comment_text
+    Comment.new({comment_text: comment_text, 
+                 update_comment: true, 
+                 user_id: self.requester_id, 
+                 proposal_id: self.id}).save
```
afeld: FYI that

```ruby
SomeModel.new(...).save
```

is equivalent to

```ruby
SomeModel.create(...)
```

afeld: Also, I would use `self.comments.create(...)` so that it takes care of the association.

app/views/communicart_mailer/_comments.html.erb
```diff
@@ -6,22 +6,41 @@
...
+          <td valign='top' class='comment-detail'>
+            <p>
+              <% unless c.user.nil? %>
+                <strong><%= c.user_full_name %></strong><br/>
+              <% end %>
+              <%= c.comment_text %>
+            </p>
+          </td>
+        </tr>
+      <% end %>
```
afeld: I think this can be DRYed up by changing it to

```erb
<tr <% if c.update_comment %>class="comment-update"<% end %>>
  <td valign='top' class='comment-detail first'>
    <p>
      <%= l(c.created_at.to_date) %>
    </p>
  </td>

  <td valign='top' class='comment-detail'>
    <p>
      <% unless c.user.nil? %>
        <strong>
          <% if c.update_comment %>
            Request modified by
          <% end %>
          <%= c.user_full_name %>
        </strong><br/>
      <% end %>
      <%= markdown(c.comment_text) %>
    </p>
  </td>
</tr>
```

Might be a little easier to read in HAML, frankly.

afeld: Also, we validate that Comments need to have an associated User, so we can remove the `c.user.nil?` check(s).

app/views/proposals/show.html.erb
```diff
@@ -109,25 +109,46 @@
...
-                <%= c.comment_text %>
-              </p>
+              <div class='row'>
+                <p class='comment-text col-sm-6 col-xs-12'>
+                  <%= c.comment_text %>
+                </p>
+              </div>
             </div>
-          </div>
+          <% end %>
```
afeld: Ditto.

afeld:
Nice work! I totally understand the MVP of storing things in a Comment, though it still makes me squeamish :blush: @cmc333333 Still ok with storing it this way, knowing that we'll need to write a migration to parse them out sooner or later?

nickbristow:
Hey @afeld I updated based on your feedback. Thanks for the Feedback!!

Also, your not about ``` self.comments.create```
when i did that, it wasn't adding the user or the created_at info. Thought that was a little strange. Not sure what to do about that, but I changed it from new to create

nickbristow:
Oh, another note, I used ```before_update``` because i want the comment before the update is made so that the notification email includes the update comment.

app/models/ncr/work_order.rb
```diff
@@ -13,11 +13,15 @@ def self.table_name_prefix
 
   class WorkOrder < ActiveRecord::Base
     include ObservableModel
+    include ValueHelper
+    # TODO include ProposalDelegate
```
afeld: Mind removing this comment?

app/models/proposal.rb
```diff
@@ -167,5 +167,11 @@ def restart
     Dispatcher.deliver_new_proposal_emails(self)
   end
 
+  def changed_fields comment_text
+    Comment.create({comment_text: comment_text, 
+                 update_comment: true, 
+                 user_id: self.requester_id, 
+                 proposal_id: self.id})
+  end
```
afeld: Try

```ruby
self.comments.create(
  comment_text: comment_text, 
  update_comment: true, 
  user_id: self.requester_id
)
```

app/models/ncr/work_order.rb
```diff
@@ -13,11 +13,15 @@ def self.table_name_prefix
...
 
     has_one :proposal, as: :client_data
     include ProposalDelegate
 
     after_initialize :set_defaults
+    #creates a comment containing the changes
+    before_update :changed_attributes_comment
```
afeld: Maybe make the comment obsolete by naming the callback `:record_changes` or something?

afeld:
Just a couple small things – looks good!

Sorry for totally missing this before, but can we add a couple specs to make sure the updates are tracked correctly? Doing them at the model level is fine. Happy to talk through this if you like.

cmc333333:
@afeld yep, I'm okay with prototyping this way. We'll probably eventually move to something else, but this'll do for now

nickbristow:
@afeld thanks for the comments, updated

afeld:
:shipit:

Mind just adding a couple tests checking that the update comments are created, and/or the list of changes displays properly?

afeld:
We can do that in a separate PR if you want.

nickbristow:
@afeld yeah, working on that now

afeld:
...though we wouldn't if we were being dogmatic about test coverage :wink:

PR #368, opened 2015-05-21
[Delivers #95058962] add flag to hide BA61 option

afeld:
If the environment variable is set, hides the radio buttons and defaults to BA80. Easier to see [ignoring whitespace](https://github.com/18F/C2/pull/368/files?w=1).


cmc333333:
Nice and clean

PR #370, opened 2015-05-22
Added style to change color of login link

nickbristow:
Made log in/log out link whte

### DIFF
--- a/app/assets/stylesheets/webapp/webviews.scss
+++ b/app/assets/stylesheets/webapp/webviews.scss
@@ -315,9 +315,6 @@ body.webapp {
     text-decoration: underline;
     color: #0C75E1;
   }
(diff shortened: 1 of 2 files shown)


nickbristow:
my_requests_-_c2

app/assets/stylesheets/webapp/webviews.scss
```diff
@@ -315,6 +315,11 @@ body.webapp {
     text-decoration: underline;
     color: #0C75E1;
   }
+
+  .button_to input.login-link{
+    color: #ffffff; 
+  }
+
```
cmc333333: This placement is by no means bad, but would it make more sense in the header-specific styles? Also, should we use `@white`?

PR #372, opened 2015-05-29
[Delivers #94604016] create common controller for different use cases

afeld:


app/models/gsa18f/procurement.rb
```diff
@@ -24,6 +24,8 @@ class Procurement < ActiveRecord::Base
     }
     validates :product_name_and_description, presence: true
 
+    after_create :add_approvals
```
cmc333333: This is nice and clean

app/controllers/use_case_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  def model_instance_variable_name
+    self.model_class.name.demodulize.underscore
+  end
+
+  def assign_model_instance_variables(val)
+    @model_instance ||= val
+
+    var_name = "@#{self.model_instance_variable_name}".to_sym
+    instance_variable_set(var_name, val)
```
cmc333333: Would it make more sense to change the templates to reference `@model_instance`? It seems more straight forward

afeld: Yeah, you're right. Was trying to minimize the surface area of the changes, but I agree it would make things easier to follow.

app/controllers/use_case_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  def errors
+    @model_instance.valid? # force validation
```
cmc333333: Would `@model_instance.validate` remove the need for this comment?

spec/factories/users.rb
```diff
@@ -1,6 +1,14 @@
 FactoryGirl.define do
   factory :user do
-    sequence(:email_address, User.count) {|n| "liono#{n}@some-cartoon-show.com" }
+    # don't freak out if the database isn't created yet
+    # http://stackoverflow.com/a/27745684/358804
```
cmc333333: This seems a little hacky. If it's needed, should it be present in the `rails_helper.rb` instead?

afeld: Not sure what you mean?

afeld:
Updated!

spec/factories/users.rb
cmc333333: I figured you were trying to trigger something which would create the database. If this is just to provide a default, we should figure out what's really causing the error and resolve it.

app/controllers/use_case_controller.rb
```diff
@@ -0,0 +1,74 @@
+# Abstract controller – requires the following methods on the subclass:
+# * model_class
+# * permitted_params
+class UseCaseController < ApplicationController
+  before_filter :authenticate_user!
+  before_filter ->{authorize self.proposal}, only: [:edit, :update]
+  rescue_from Pundit::NotAuthorizedError, with: :auth_errors
+  before_filter :find_model_instance, only: [:edit, :update]
```
cmc333333: I think this is redundant with line 6 (both would set `@model_instance`). Should they get combined?

afeld: It happens to be using the same method, but I wouldn't rely on `self.proposal` setting the instance variable. I actually don't mind the redundancy in this case, unless you would prefer a `find_model_instance_and_authorize` callback?

cmc333333: Nah, keeping consistent authorization commands across controllers seems more important -- we can probably even pull those out into the base controller/trait.

spec/factories/users.rb
afeld: The problem is that it's trying to `SELECT COUNT(*) FROM users;` when loading the `test` environment, even if the database/table doesn't exist yet.

```
$ RAILS_ENV=test rake db:drop
$ RAILS_ENV=test rake db:setup
rake aborted!
ActiveRecord::StatementInvalid: PG::UndefinedTable: ERROR:  relation "users" does not exist
LINE 5:                WHERE a.attrelid = '"users"'::regclass
                                          ^
:               SELECT a.attname, format_type(a.atttypid, a.atttypmod),
                     pg_get_expr(d.adbin, d.adrelid), a.attnotnull, a.atttypid, a.atttypmod
                FROM pg_attribute a LEFT JOIN pg_attrdef d
                  ON a.attrelid = d.adrelid AND a.attnum = d.adnum
               WHERE a.attrelid = '"users"'::regclass
                 AND a.attnum > 0 AND NOT a.attisdropped
               ORDER BY a.attnum
...
/Users/aidanfeldman/dev/18F/c2/spec/factories/users.rb:3:in `block (2 levels) in <top (required)>'
...
```

Weirdly, `RAILS_ENV=test rake db:schema:load` works, which is what CircleCI uses, and thus why the tests are passing. Not sure of the best way to handle this.

cmc333333: Startlingly similar to https://github.com/18F/tock/pull/161 . The inclusion of a file probably shouldn't force a DB call. Does this have the same semantics?

```patch
diff --git a/spec/factories/users.rb b/spec/factories/users.rb
index 7916d8a..5c4d0d5 100644
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :user do
-    sequence(:email_address, User.count) {|n| "liono#{n}@some-cartoon-show.com" }
+    sequence(:email_address) {|n| "liono#{n}@some-cartoon-show.com" }
     first_name "Liono"
     last_name "Thunder"
```

afeld: Pretty sure I added that "starting value" because I had run `rake populate:*` before, but when I re-ran it, it tried to start the sequence from `0`, which threw a validation error because that email was already taken.

cmc333333: How about

```diff
diff --git a/spec/factories/users.rb b/spec/factories/users.rb
index 7916d8a..c9b3cce 100644
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :user do
-    sequence(:email_address, User.count) {|n| "liono#{n}@some-cartoon-show.com" }
+    sequence(:email_address) {|n| "liono#{User.count}@some-cartoon-show.com" }
     first_name "Liono"
     last_name "Thunder"
```

afeld: Cool, that works. https://github.com/18F/C2/pull/376/

afeld: Also, had no idea you could do `diff` formatting in Markdown. Sweet.

afeld:
Updated!

cmc333333:
Yay!

PR #373, opened 2015-05-31
93639582 full approval

phirefly:
Displays a more explicit message of all approvals being completed. Allows ability to have custom messages per client.


app/mailers/communicart_mailer.rb
```diff
@@ -42,8 +42,8 @@ def proposal_created_confirmation(proposal)
   def approval_reply_received_email(approval)
     @approval = approval
     @proposal = approval.proposal.decorate
+    @alert_partial = 'approvals_complete' if @proposal.approved?
```
cmc333333: Good call on the consistency with `notification_for_approver`

cmc333333:
Straightforward solution, nice

PR #376, opened 2015-06-03
don't call to the database when the test environment is loaded

afeld:
See https://github.com/18F/C2/pull/372/files#r31368285.

/cc @annalee


afeld:
Merging to unblock @annalee.

annalee:
Thanks a ton, @afeld and @cm! (there really was something wrong with postgres, so this was not my main blocker--but thanks for fixing it so fast!)

PR #377, opened 2015-06-03
[#94861538] Simple CSV user import

cmc333333:
Give it a CSV file and a client slug and this script will replace/create appropriate users. It'll guess at the appropriate column names (which can be overridden). It'll trim the email address and title case the name fields.


lib/tasks/import_users.rake
```diff
@@ -52,4 +52,35 @@ namespace :import_users do
...
+    if !first_col || !last_col || !email_col
+      raise "Couldn't determine one or more of first name, last name, email: #{headers}"
+    end
+
+    data.each do |row|
+      email = row[email_col]
+      if !email
+        warn "Email is empty: " + row.inspect
+      else
+        User.for_email(email.strip).update(
```
afeld: Wonder if we should downcase as well?

lib/tasks/import_users.rake
```diff
@@ -52,4 +52,35 @@ namespace :import_users do
...
+          first_name: (row[first_col] || "").titleize,
+          last_name: (row[last_col] || "").titleize,
```
afeld: I would leave them `nil` if not present.

afeld: ...a la


```ruby
row[last_col].presence.try(:titleize)
```

afeld:
What's your thinking for how the CSVs will make it onto Cloud Foundry? Wonder if it should (alternatively?) accept a URL.

Not necessarily a blocker, but I'd suggest moving the bulk of the Rake task under a class/module under `lib` and put some tests around it.

Lastly, mind documenting why and how to use it?

cmc333333:
> What's your thinking for how the CSVs will make it onto Cloud Foundry? 
We can just include the file when pushing. That's how the 18f contracts were loaded

> Wonder if it should (alternatively?) accept a URL.
That's a good idea; it'd require proper authentication at the external site, though, which makes me think it's not worth it unless there's a specific use case

> Not necessarily a blocker, but I'd suggest moving the bulk of the Rake task under a class/module under lib and put some tests around it.
Sure thing. It has more re-use potential than the 18f import

> Lastly, mind documenting why and how to use it?
Can do

lib/tasks/import_users.rake
cmc333333: Good call

lib/tasks/import_users.rake
cmc333333: Makes sense -- I think there might even be an existing library somewhere in our app

afeld:
> We can just include the file when pushing.

Mind including that in a comment or a Markdown file or something?

afeld:
> it'd require proper authentication at the external site

Yeah, no worries.

cmc333333:
Updated with better docs and moving the bulk of CSV processing to a separate, tested file.

lib/csv_user_importer.rb
```diff
@@ -0,0 +1,31 @@
...
+    @csv = CSV.read(csv_path, headers: true)
+    @client = client
+    headers = @csv.headers()
+    @first_name_col = headers.find {|h| h.downcase.include? "first"}
+    @last_name_col = headers.find {|h| h.downcase.include? "last"}
+    @email_col = headers.find {|h| h.downcase.include? "email"}
+  end
+
+  def process_rows
+    if !@first_name_col || !@last_name_col || !@email_col
```
afeld: Minor: as long as you have the accessors, I would assign/read using `self.*`.

afeld: Goes for all instance variables in this file, really.

cmc333333:
Updated to use `self` when referring to `attr_accessor` attributes.

PR #379, opened 2015-06-04
[Finishes #96227784][Finishes #96109016] Bug fixes

cmc333333:
While tracking down an [issue](https://www.pivotaltracker.com/story/show/96227784) where the first approver was not receiving an email alert, I found that a useful controller spec was not being run (due to file name). Renaming it caused the tests to run, but with error due to an [issue](https://www.pivotaltracker.com/story/show/96109016) while "editing" without modifying the work order.

This resolves the second issue (by checking if the "update" text would be empty), and adds a hook to the `create` method in `use_case_controller`. The Work Order controller uses that hook to set its approvers before the email goes out.


app/models/proposal.rb
```diff
@@ -194,10 +194,12 @@ def partial_approve
   end
 
   def changed_fields comment_text
-    self.comments.create(
-      comment_text: comment_text, 
-      update_comment: true, 
-      user_id: self.requester_id
-    )
+    if !comment_text.blank?
```
afeld: FYI that Rails also provides [`#present?`](http://api.rubyonrails.org/classes/Object.html#method-i-present-3F).

afeld:
> a useful controller spec was not being run

:facepalm: Good catch!

PR #380, opened 2015-06-05
[Delivers #95573418] Allow attachments at work order creation

cmc333333:
Adds file fields to the NCR intake form. Non-JS users will get 10 fields, JS users will get 1 with the ability to add up to 9 additional. First approver and requester will get emails including links to the attachments.

Looks like:
screen shot 2015-06-05 at 9 07 20 am
screen shot 2015-06-05 at 9 07 39 am
screen shot 2015-06-05 at 9 11 48 am


spec/features/ncr_work_orders_spec.rb
```diff
@@ -131,6 +131,29 @@
...
 
+    it "allows attachments to be added during intake without JS" do
+      visit '/ncr/work_orders/new'
+      expect(page).to have_content("Attachments")
+      expect(page).not_to have_selector(".js-am-minus")
+      expect(page).not_to have_selector(".js-am-plus")
+      expect(page).to have_selector("input[type=file]", count: 10)
+    end
+
+    it "allows attachments to be added during intake with JS", :js => true do
```
afeld: Nice.

app/views/ncr/work_orders/form.html.erb
```diff
@@ -98,6 +98,17 @@
                           disabled: approver_email_frozen?,
                           data: {attr: 'email_address', src: '/api/v1/users.json'} %>
     </div>
+    <% if @model_instance.new_record? %>
```
afeld: Would prefer to not have two UIs for uploading, for usability and maintainability purposes. We should consider doing it all through the form, and getting rid of the Upload button on the `show` page.

afeld:
Code looks good, though I would still like to look into [a fancier uploader](https://www.pivotaltracker.com/story/show/94044148) sooner or later. Thanks!

PR #381, opened 2015-06-05
[#94312822] add 18F purchaser as an observer

phirefly:
Add a purchaser as an observer when submitting and supply request.


app/models/gsa18f/procurement.rb
```diff
@@ -76,7 +80,13 @@ def name
     end
 
     def self.approver_email
-      ENV['GSA18F_APPROVER_EMAIL'] || '18fapprover@gsa.gov'
+      raise 'No 18F approver was set' unless ENV['GSA18F_APPROVER_EMAIL']
+      ENV['GSA18F_APPROVER_EMAIL']
+    end
+
+    def self.purchaser_email
+      raise 'No 18F purchaser was set' unless ENV['GSA18F_PURCHASER_EMAIL']
```
phirefly: Thoughts on pulling out this email as an environment variable like we do for the approver? Also noticed that the default email for approver was a bogus email so what if we simply raise an error if the env var isn't set?

spec/lib/proposal_search_spec.rb
```diff
@@ -35,6 +35,16 @@
     end
 
     context Gsa18f::Procurement do
+      around(:each) do |example|
```
phirefly: I'll need to pull this out of here since we're sharing this in other tests.

app/models/gsa18f/procurement.rb
```diff
@@ -76,7 +80,13 @@ def name
     end
 
     def self.approver_email
-      ENV['GSA18F_APPROVER_EMAIL'] || '18fapprover@gsa.gov'
+      raise 'No 18F approver was set' unless ENV['GSA18F_APPROVER_EMAIL']
+      ENV['GSA18F_APPROVER_EMAIL']
```
afeld: How about `ENV.fetch('GSA18F_APPROVER_EMAIL')`?

http://ruby-doc.org/core-2.2.2/ENV.html#method-c-fetch

app/models/gsa18f/procurement.rb
afeld: :+1: on both.

spec/support/integration_spec_helper.rb
```diff
@@ -19,4 +19,12 @@ def setup_mock_auth(service_name=:myusa, user=FactoryGirl.create(:user))
       }
     )
   end
+
+  def with_18f_env_variables(setup_vars=nil)
```
afeld: How about `with_18f_procurement_env_variables`?

spec/support/integration_spec_helper.rb
```diff
@@ -19,4 +19,12 @@ def setup_mock_auth(service_name=:myusa, user=FactoryGirl.create(:user))
...
+    ENV['GSA18F_APPROVER_EMAIL'] = 'test_approver@some-dot-gov.gov'
+    ENV['GSA18F_PURCHASER_EMAIL'] = 'test_purchaser@some-dot-gov.gov'
+    yield
+    ENV['GSA18F_APPROVER_EMAIL'] = nil
+    ENV['GSA18F_PURCHASER_EMAIL'] = nil
```
afeld: Minor nit, but mind setting them back to whatever they were before?

afeld: ...slash use the existing `with_env_var` helper.

spec/models/gsa18f/procurement_spec.rb
```diff
@@ -1,4 +1,13 @@
 describe Gsa18f::Procurement do
+  around(:each) do |example|
+    ENV['GSA18F_APPROVER_EMAIL'] = 'test_approver@some-dot-gov.gov'
+    ENV['GSA18F_PURCHASER_EMAIL'] = 'test_purchaser@some-dot-gov.gov'
+    example.run
+
+    ENV['GSA18F_APPROVER_EMAIL'] = nil
+    ENV['GSA18F_PURCHASER_EMAIL'] = nil
+  end
+
```
afeld: Can we re-use the same helper method?

app/models/gsa18f/procurement.rb
phirefly: This would set the value to nil if it isn't set in the env vars. Is this what we want?

afeld: How so?

> If the given name does not exist...an IndexError is raised

phirefly: Got it. I was totally looking at the wrong documentation. :)

phirefly:
@afeld updated!

app/models/proposal.rb
```diff
@@ -196,6 +196,14 @@ def restart
     Dispatcher.deliver_new_proposal_emails(self)
   end
 
+  def changed_fields comment_text
+    self.comments.create(
+      comment_text: comment_text,
+      update_comment: true,
+      user_id: self.requester_id
+    )
+  end
```
afeld: Is this being used anywhere?

spec/support/integration_spec_helper.rb
```diff
@@ -19,4 +19,15 @@ def setup_mock_auth(service_name=:myusa, user=FactoryGirl.create(:user))
...
+  def with_18f_procurement_env_variables(setup_vars=nil)
+    old_approver_email = ENV['GSA18F_APPROVER_EMAIL']
+    old_purchaser_email = ENV['GSA18F_PURCHASER_EMAIL']
+
+    ENV['GSA18F_APPROVER_EMAIL'] = 'test_approver@some-dot-gov.gov'
+    ENV['GSA18F_PURCHASER_EMAIL'] = 'test_purchaser@some-dot-gov.gov'
+    yield
+    ENV['GSA18F_APPROVER_EMAIL'] = old_approver_email
+    ENV['GSA18F_PURCHASER_EMAIL'] = old_purchaser_email
+  end
```
afeld: Can we put this in a different helper?

phirefly: Recommendations?

afeld: Maybe a new one called `EnvironmentHelper`?

app/models/proposal.rb
phirefly: I thought this was from a master merge but I don't see the code in master now... will clean this up.

phirefly:
Updated

spec/rails_helper.rb
```diff
@@ -60,8 +60,9 @@
   # Add modules for helpers
   config.include ControllerSpecHelper, type: :controller
   config.include RequestSpecHelper, type: :request
-  [:feature, :request].each do |type|
+  [:feature, :request, :model].each do |type|
     config.include IntegrationSpecHelper, type: type
+    config.include EnvironmentSpecHelper, type: type
```
afeld: We don't want to include the `IntegrationSpecHelper` for models...mind moving this line out of the loop to be `config.include EnvironmentSpecHelper, type: :model`?

PR #382, opened 2015-06-05
upgrade to Rails 4.2

afeld:
https://github.com/rails/rails/issues/20077 may be a blocker :confused:


afeld:
Huzzah, it works!

db/schema.rb
```diff
@@ -16,8 +16,8 @@
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
-  create_table "api_tokens", force: true do |t|
-    t.string   "access_token"
+  create_table "api_tokens", force: :cascade do |t|
+    t.string   "access_token", limit: 255
```
cmc333333: Where are these limits coming from? Do we need a corresponding migration?

More info: https://github.com/rails/rails/pull/14579

afeld: Seems to be accurate. On production:

```
> pp ActiveRecord::Base.connection.columns('api_tokens').find{|col| col.name == 'access_token' }
#<ActiveRecord::ConnectionAdapters::PostgreSQLColumn:0x007fd669756f60
 @array=false,
 @coder=nil,
 @default=nil,
 @default_function=nil,
 @limit=255,
 @name="access_token",
 @null=true,
 @oid_type=
  #<ActiveRecord::ConnectionAdapters::PostgreSQLAdapter::OID::Identity:0x007fd6671a6220>,
 @precision=nil,
 @primary=nil,
 @scale=nil,
 @sql_type="character varying(255)",
 @type=:string>
```

Open to doing a migration to remove the limits, but don't think it matters that much for now.

PR #385, opened 2015-06-09
[Delivers #95888854] Allow Work Orders to be edited post approval

cmc333333:
This allows approved work orders to be edited, adds a notice when editing such a work order, and tags edit comments as occurring post approval.

To do that:
* Added a mechanism for Work Orders (or other client-specific models) to override proposal's policies. In this case, this allowed Work Orders to be edited post approval while other client-specific models cannot be
* Adds a line to the "edited" comment noting that this edit took place after the proposal was approved
* Moved "edited" comment logic into Ncr::WorkOrder, as it's quite coupled at the moment

Looks like:
__Edit form__
screen shot 2015-06-09 at 12 30 17 pm

__Email header__
screen shot 2015-06-09 at 1 03 23 pm

__Web interface__
screen shot 2015-06-09 at 1 02 55 pm

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,10 +10,10 @@ class ApplicationController < ActionController::Base
   helper_method :current_user, :signed_in?, :return_to
 
   protected
-  # We are overriding this method to account for permission trees. See
-  # TreePolicy
+  # We are overriding this method to account for ExceptionPolicies
   def authorize(record, query=nil, user=nil)
     user ||= @current_user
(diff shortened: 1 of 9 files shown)


cmc333333:
@phirefly there was some discussion about what needed to implemented in https://www.pivotaltracker.com/story/show/95888854 -- would you mind verifying that this is an acceptable solution?

afeld:
Looks good to me!

phirefly:
:shipit:

PR #386, opened 2015-06-10
[Delivers #93637304] improve email grouping

afeld:
The goal is to get all emails related to a request to show up in a single thread in GMail and other mail clients.

### DIFF
--- a/app/decorators/proposal_decorator.rb
+++ b/app/decorators/proposal_decorator.rb
@@ -55,4 +55,8 @@ class ProposalDecorator < Draper::Decorator
   def progress_status_message
     "#{number_approved} of #{total_approvers} approved."
   end
+
+  def email_subject
+    "Request #{proposal.public_identifier}"
+  end
 end
--- a/app/decorators/proposal_decorator.rb
+++ b/app/decorators/proposal_decorator.rb
@@ -59,4 +59,8 @@ class ProposalDecorator < Draper::Decorator
   def email_subject
     "Request #{proposal.public_identifier}"
   end
+
+  def email_msg_id
+    "<proposal-#{self.id}@#{DEFAULT_URL_HOST}>"
+  end
 end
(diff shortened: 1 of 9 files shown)


afeld:
:zap: 

screen shot 2015-06-10 at 1 47 04 am

app/decorators/proposal_decorator.rb
```diff
@@ -55,4 +55,12 @@ def completed_status_message
   def progress_status_message
     "#{number_approved} of #{total_approvers} approved."
   end
+
+  def email_subject
+    "Request #{proposal.public_identifier}"
```
afeld: Open to suggestions about this. The format suggested in [the story](https://www.pivotaltracker.com/story/show/93637304) is more likely to change during the lifetime of a request and thus more likely to break the threading...not sure what the right balance is. Some ideas, from most to least volatile/safe:

1. `Request #<ID>`
1. `Request #<ID> – <Proposal name>`
1. `Request #<ID>, <Org Code>, <Building code>, from <Requester>`

spec/mailers/communicart_mailer_spec.rb
```diff
@@ -209,16 +215,14 @@ def sender_names(mail)
...
+    it_behaves_like "a Proposal email"
 
     it 'renders the receiver email' do
       expect(mail.to).to eq([proposal.requester.email_address])
     end
 
-    it 'renders the sender email' do
-      expect(mail.from).to eq(["reply@stub.gov"])
+    it "sets the sender name" do
+      expect(sender_names(mail)).to eq([nil])
```
cmc333333: I'm not sure what this and the `it "sets the sender name"` above are actually testing. Seems like `it_behaves_like "a Proposal email"` covers the sender?

cmc333333:
Big improvement! Only one small suggestion

afeld:
Updated!

PR #387, opened 2015-06-10
add Peek

afeld:
...to help annoy us into fixing slow pages. Note that it will be hidden by default, but can be shown by pressing the backtick key.

screen shot 2015-06-10 at 2 41 06 am

More info:

https://github.com/peek/peek


cmc333333:
It worries me a bit that would be enabled and accessible to all users in a production environment. Generally, you don't want to advertise these types of metrics to potentially malicious users.

Can we add a user flag?

afeld:
We could, though I favor transparency around this. Started a conversation in https://github.com/peek/peek/issues/71.

afeld:
Ok, updated to only display for "admins" outside of development.

* [ ] Set `ADMIN_EMAILS` on servers

PR #388, opened 2015-06-10
[Delivers #96626124] eager load the associated models for the WorkOrder API

afeld:
Was noticing a *ton* of SQL queries in the log when loading the NCR form...turns out it was doing a separate database call for every association :grin: This brings it down to a single query. On my local machine, loading http://localhost:3000/api/v1/ncr/work_orders.json with 841 WorkOrders:

**Before**

```
Completed 200 OK in 6391ms (Views: 5853.2ms | ActiveRecord: 535.2ms)
```

**After**

```
Completed 200 OK in 1274ms (Views: 1258.1ms | ActiveRecord: 13.5ms)
```

### DIFF
--- a/app/controllers/api/ncr/work_orders_controller.rb
+++ b/app/controllers/api/ncr/work_orders_controller.rb
@@ -3,7 +3,9 @@ module Api
     class WorkOrdersController < BaseController
       def index
         # TODO use a scope for ordering
-        orders = ::Ncr::WorkOrder.joins(:proposal).order('proposals.created_at DESC')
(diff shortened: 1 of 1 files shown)


cmc333333:
Nice improvement. I'm pretty sure that all of the listing pages suffer this.

Unfortunately, n+1 style bugs are notorious for arising during regressions. Would you mind adding a test for this? A pattern I've seen work well is to create, say 10 records, load a listing page and assert that the total query count was less than some constant, say 5.

cmc333333:
Ah, #389 does this with a gem! Very slick

PR #389, opened 2015-06-10
add Bullet gem to find optimizable queries

afeld:


afeld:
@cmc333333 Thoughts about making Bullet throw errors opt-in? Figure we only benefit from getting the n+1 fixes in sooner than later, and we can pick off more and more as we go.

cmc333333:
:+1: good idea

afeld:
Updated!

PR #390, opened 2015-06-11
[Delivers #93381468] Add an endpoint for feedback, along with its form

cmc333333:
Also tweaks the logic around js filters to allow for logic when an option is _not_ selected. There are currently no links to this form.

## initial form
screen shot 2015-06-11 at 5 39 57 pm

## is a bug
screen shot 2015-06-11 at 5 40 08 pm

## submitted (not logged in)
screen shot 2015-06-11 at 5 40 29 pm

## submitted (logged in)
screen shot 2015-06-11 at 5 41 00 pm

## email
screen shot 2015-06-11 at 5 41 37 pm

### DIFF
--- a/app/controllers/feedback_controller.rb
+++ b/app/controllers/feedback_controller.rb
@@ -1,5 +1,7 @@
 class FeedbackController < ApplicationController
-  # note that index is rendered implicitly as it's just a template
+  def index
+    @skip_footer = true
+  end
 
   def create
     fields = [:bug, :context, :expected, :actually, :comments, :satisfaction, :referral]
--- a/app/controllers/feedback_controller.rb
+++ b/app/controllers/feedback_controller.rb
@@ -2,18 +2,11 @@ class FeedbackController < ApplicationController
   # note that index is rendered implicitly as it's just a template
 
   def create
-    message = []
-    [:bug, :context, :expected, :actually, :comments, :satisfaction, :referral].each do |key|
-      if !params[key].blank?
-        message << "#{key}: #{params[key]}"
-      end
-    end
-    message = message.join("\n")
-    if !message.blank?
-      if current_user
-        message += "\nuser: #{current_user.email_address}"
-      end
-      CommunicartMailer.feedback(message).deliver
+    fields = [:bug, :context, :expected, :actually, :comments, :satisfaction, :referral]
+    fields = fields.select {|key| !params[key].blank?}
+    form_values = fields.map {|key| [key, params[key]]}
+    unless form_values.empty?
+      CommunicartMailer.feedback(current_user, form_values).deliver
     end
   end
 end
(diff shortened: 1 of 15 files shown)


app/controllers/feedback_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+        message << "#{key}: #{params[key]}"
+      end
+    end
+    message = message.join("\n")
+    if !message.blank?
+      if current_user
+        message += "\nuser: #{current_user.email_address}"
+      end
+      CommunicartMailer.feedback(message).deliver
+    end
```
afeld: We might want to move this logic to a Plain Old Ruby Object down the road.

app/mailers/communicart_mailer.rb
```diff
@@ -66,6 +66,18 @@ def comment_added_email(comment, to_email)
     end
   end
 
+  def feedback(message)
+    mail(
+      to: CommunicartMailer.support_email,
+      subject: 'Feedback submission',
+      from: default_sender_email,
+      body: message
+    )
```
afeld: How about adding them as the `reply_to`? Just to save us a step.

afeld:
Looks good! Kind of bummed that we need to implement/maintain this ourselves and can't use something off-the-shelf, though :confused:

app/controllers/feedback_controller.rb
cmc333333: I'll move some of the logic into the mailer as a compromise

app/mailers/communicart_mailer.rb
cmc333333: Good idea

cmc333333:
Updated to simplify the logic. Added the user as a CC (so that they will receive a receipt)

cmc333333:
Updated to link to the feedback form on standard pages, the landing page, and 400s. The 500 template includes a mailto link instead. Added a little logic so that the feedback link does not appear on the feedback form.

Looks like:
### Landing Page
screen shot 2015-06-15 at 8 40 06 pm
### Normal Pages
screen shot 2015-06-15 at 8 40 22 pm
### 404s
screen shot 2015-06-15 at 8 39 45 pm
### 500
screen shot 2015-06-15 at 8 22 50 pm
### 500 email
screen shot 2015-06-15 at 8 22 56 pm

app/controllers/feedback_controller.rb
```diff
@@ -0,0 +1,15 @@
+class FeedbackController < ApplicationController
+  def index
+    @skip_footer = true
+  end
+
+  def create
+    fields = [:bug, :context, :expected, :actually, :comments, :satisfaction, :referral]
+    fields = fields.select {|key| !params[key].blank?}
+    form_values = fields.map {|key| [key, params[key]]}
```
afeld: Not sure that it makes a *huge* difference, but this is basically what [strong parameters](http://edgeguides.rubyonrails.org/action_controller_overview.html#strong-parameters) are for.

PR #391, opened 2015-06-12
[Delivers #96691408] make #name present for all Proposals

afeld:
Also make `WorkOrder#project_title` required. I know it wasn't in this sprint, but I had some time on the train and it was the easiest story to do offline :smile_cat:


cmc333333:
Good stuff. Using the default if `blank?` might lead to some odd corner cases, but we can tweak it if the need ever arises.

PR #392, opened 2015-06-12
formated error messages for amount

nickbristow:


cmc333333:
Much more approachable

PR #393, opened 2015-06-15
[Delivers #96974108] only have single budget approver for BA61

afeld:
@phirefly Not positive I interpreted the story correctly...mind confirming?

* [ ] Switch `NCR_BA61_TIER2_BUDGET_MAILBOX` to `NCR_BA61_BUDGET_MAILBOX` on servers


afeld:
Specifically:

* Are there any cases where they *would* want the Tier 1 *and* Tier 2 approvers for BA61?
* Should we still refer to the latter as "Tier 2", or just "budget"?

afeld:
From @phirefly:

If it's BA61 and it has org code `P1122021 – (192X,192M) WHITE HOUSE DISTRICT`, then only use the Tier 2 mailbox. Otherwise use both Tier 1 and Tier 2.

cmc333333:
Lots of special cases there. Is it time for a rules engine?

cmc333333:
To be clear - there are additional changes needed w/ the logic before this should be merged, right?

afeld:
Hopefully just what I listed above, yeah.

afeld:
> Is it time for a rules engine?

Not sure that it's *that* complicated yet, but would be curious to see what you have in mind.

afeld:
Replaced by #397.

PR #394, opened 2015-06-15
Fixed error message, said amount amount

nickbristow:


afeld:
Actually, did you test this? Rails sometimes automatically prepends the name of the field, depending on how you retrieve the errors (which is kind of annoying).

PR #395, opened 2015-06-16
[Delivers #97070284] Add shim for IE8

cmc333333:
This was all that was required to fix selectize in IE8 (and 9)


afeld:
Looking for a document that gives more context, but I believe we are supposed to avoid using shared CDNs. Mind loading the script from a relative path?

https://github.com/es-shims/es5-shim

afeld:
https://github.com/18F/blog-drafts/issues/199

afeld:
Mind deploying the branch to c2-dev and verifying that you can access the JS file at the generated path? My guess is that you will need to add it to `config.assets.precompile`.

http://guides.rubyonrails.org/asset_pipeline.html#precompiling-assets

cmc333333:
Looks like I didn't need to add anything to the precompile list; works in IE on dev

PR #397, opened 2015-06-17
[Delivers #96974108] skip Tier 1 for WHSC

afeld:


app/models/ncr/work_order.rb
```diff
@@ -114,7 +115,35 @@ def name
       self.project_title
     end
 
+    def system_approvers
+      results = []
+      if self.expense_type == 'BA61'
+        unless self.organization.whsc?
+          results << self.class.ba61_tier1_budget_mailbox
+        end
```
afeld: This is the important part...the rest is simply refactoring to support it.

app/models/ncr/work_order.rb
```diff
@@ -97,6 +92,12 @@ def client
       "ncr"
     end
 
+    def organization
+      # TODO reference by `code` rather than storing the whole thing
+      code = self.org_code.split(' ', 2)[0]
```
cmc333333: Is it worth accounting for a nil `org_code` here?

lib/ncr/organization.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+  end
+end
+
+rows = CSV.read("#{Rails.root}/config/data/ncr/org_codes_2015-05-18.csv", headers: true)
+
+Ncr::Organization::BY_CODE = {}
+rows.each do |row|
+  org = Ncr::Organization.new(row)
+  Ncr::Organization::BY_CODE[org.code] = org
```
cmc333333: This is a solution to our duplicate org code problem, too

afeld:
Updated!

app/models/ncr/work_order.rb
```diff
@@ -114,7 +115,35 @@ def name
       self.project_title
     end
 
+    def system_approvers
+      results = []
+      if self.expense_type == 'BA61'
+        unless self.organization.whsc?
```
cmc333333: To account for nils, maybe a `try` here?

afeld: Good catch!

afeld:
Updated!

cmc333333:
Thanks!

PR #398, opened 2015-06-17
remove comment index page

afeld:
Doesn't seem to be used anymore.


cmc333333:
I don't think so either; @phirefly can you be the triple check?

afeld:
It isn't referenced anywhere...I think it's safe.

phirefly:
Yes, this was the target page after adding a comment when we initially allowed comments to be added. :shipit:

PR #399, opened 2015-06-17
clean up styles for heading elements

afeld:
Clicked through all(?) pages in the web app, as well as the `approval_reply_received_email`, just to make sure they all look ok.


cmc333333:
:+1:, some screenshots in #400

PR #402, opened 2015-06-18
[WIP] reference buildings by `number`

afeld:
...rather than storing the entire string.


cmc333333:
@afeld this is pretty stale. Mind if we close for now?

PR #405, opened 2015-06-18
Feature/search by whole

nickbristow:
Search by:
ID number with Fiscal year, case insensitive, 

also allows searching by fiscal year, followed by any search term. 
Returns results from that fiscal year that matches search terms.


cmc333333:
@nickbristow good stuff. Do you mind adding some tests? Particularly around the search.

The data migration is a little fragile in that it relies on a class/method which may change in the future. Let me try to whip up something more stable

cmc333333:
```sql
UPDATE proposals
SET public_id = '#' || id;
```
```sql
UPDATE proposals
SET public_id = 'FY15-' || id
WHERE client_data_type = 'Ncr::WorkOrder';
```

cmc333333:
Rockin. We'll figure out the Gemlock file elsewheres.

PR #407, opened 2015-06-19
[#95897486] split help pages

afeld:
Just so the questions are easier to scan. Thanks to Anthony for the idea!

**Listing page**

screen shot 2015-06-19 at 4 51 38 pm

---

**Sub-page**

screen shot 2015-06-19 at 4 51 51 pm


app/views/help/index.html.haml
```diff
@@ -0,0 +1,12 @@
...
+    - # this is a hack to capture the :title from each page: render the template, without actually outputting it
+    - render template: "help/#{page}"
+    %li
+      = link_to help_path(page) do
+        - if content_for?(:title)
+          = yield(:title)
+        - else
+          = page
+- # this needs to be at the bottom of the template
+- title "FAQ"
```
afeld: This file is where the magic is happening :tophat: :sparkles: :rabbit2:

app/views/help/index.html.haml
```diff
@@ -0,0 +1,12 @@
+%ul
+  - @pages.each do |page|
+    - # this is a hack to capture the :title from each page: render the template, without actually outputting it
+    - render template: "help/#{page}"
```
cmc333333: What do you think about doing this pre-rendering step inside the controller instead of the view? Then `@pages` could be a list of `(path, title)` pairs? Ideally the controller would be able to fetch the meta data without rendering, but I don't see a simple way to do that either.

app/controllers/help_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  def index
+    dir = Rails.root.join('app', 'views', 'help', '*.md')
+    files = Dir.glob(dir)
+    @pages = files.map{|file| File.basename(file, '.md') }.sort
+  end
+
+  def show
+    page = params[:id]
+    # TODO make sure this isn't a security hole
```
cmc333333: Probably best to use a whitelist. Maybe pull out the listing logic in `index` into a shared function (or class var) to get such a list. Only render if `page` is a member of that list (404 otherwise)?

afeld: Good call!

app/views/help/index.html.haml
afeld: Have tried doing the rendering in a helper and the controller...can't figure out how to do it. Mind if we punt on it for now?

PR #408, opened 2015-06-20
[Delivers #97070798] add budget-related fields to work order form

afeld:
Easier to understand the spec changes [ignoring whitespace](https://github.com/18F/C2/pull/408/files?w=1).

---

screen shot 2015-06-19 at 9 34 10 pm

### DIFF
--- a/app/views/ncr/work_orders/form.html.erb
+++ b/app/views/ncr/work_orders/form.html.erb
@@ -116,7 +116,7 @@
       </div>
     <% else %>
       <div class="form-group">
-        <%= f.label :cl_number, 'CL number' %>
+        <%= f.label :cl_number %>
         <%= f.text_field :cl_number, class: 'form-control' %>
       </div>
       <div class="form-group">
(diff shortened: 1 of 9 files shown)


app/views/ncr/work_orders/form.html.erb
```diff
@@ -114,6 +114,19 @@
           </ul>
         <%- end %>
       </div>
+    <% elsif @model_instance.approved? %>
```
afeld: Is this the desired behavior?

afeld: @phirefly ^

app/views/ncr/work_orders/form.html.erb
```diff
@@ -114,6 +114,19 @@
...
+      <div class="form-group">
+        <%= f.label :cl_number, 'CL number' %>
+        <%= f.text_field :cl_number, class: 'form-control' %>
+      </div>
+      <div class="form-group">
+        <%= f.label :function_code %>
+        <%= f.text_field :function_code, class: 'form-control' %>
+      </div>
+      <div class="form-group">
+        <%= f.label :soc_code, 'SOC code' %>
```
cmc333333: You'll want to set these labels in the translation file instead so that the caps are correct on the detail page, etc.

app/views/ncr/work_orders/form.html.erb
phirefly: We can expose these even if it isn't approved.

phirefly:
Per our feedback from Budget folks, I will add another story to capture validations.

afeld:
Updated!

phirefly:
:shipit:

PR #409, opened 2015-06-20
[Delivers #97035202] improve the form submit button

afeld:
screen shot 2015-06-19 at 9 45 26 pm

---

screen shot 2015-06-19 at 9 45 15 pm

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -32,6 +32,9 @@ en:
         rwa_number: "RWA Number"
       cart:
         rwa_number: "RWA Number"
+    submit:
+      create: "Submit for approval"
+      update: "Update"
   activerecord:
     attributes:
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -32,6 +32,9 @@ en:
         rwa_number: "RWA Number"
       cart:
         rwa_number: "RWA Number"
+    submit:
+      create: "Submit for approval"
```
cmc333333: Should this get scoped more tightly? I can imagine we'll be adding an interface for creating users, etc. where "Submit for approval" isn't a good default text for the submit button.

phirefly:
@cmc333333 Agreed, although I am happy with this as our solution for launch. :ship:ping this.

PR #410, opened 2015-06-22
Added check for changes upon updating proposal

nickbristow:
If no model changes are made, it doesn't hit the DB and just redirects to the proposals show page.


app/controllers/use_case_controller.rb
```diff
@@ -41,7 +41,9 @@ def edit
   def update
     @model_instance.assign_attributes(self.permitted_params)   # don't hit db yet
 
-    if self.errors.empty?
+    if @model_instance.changed_attributes.blank? && self.errors.empty?
```
cmc333333: Does this still account for a situation where 1) the fields don't change but 2) a different approver is selected?

Unrelated: probably a good idea to add a flash notice

nickbristow:
@cmc333333 I changed how this works slightly. 

It goes through as normal if you change the email, I also added the flash response as noted in @phirefly 's comment.

app/controllers/use_case_controller.rb
```diff
@@ -54,6 +57,19 @@ def update
...
+  end
+
+  def approver_email_changed?
+    email = params[:approver_email]
+    if !email.nil?
+      @model_instance.approval_changed?(email)
+    else
+      false
+    end
+  end
```
cmc333333: I think this logic is specific to NCR, right? How about, in `UseCaseController`:
```ruby
def attribute_changes?
    !@model_instance.changed_attributes.blank?
end
```

Then, in the `Ncr::WorkOrderController`
```ruby
def attribute_changes?
    super || @model_instance.approval_changed?(@approver_email)
end
```

app/models/ncr/work_order.rb
```diff
@@ -54,6 +54,11 @@ def update_approver(approver_email)
       # no need to call initialize_approvals as they have already been set up
     end
 
+    def approval_changed? approval_email
```
cmc333333: Style thing: parens are generally included when defining a method

cmc333333: Also, suggestion on naming: `def approver_changed?(new_email)`

app/controllers/use_case_controller.rb
```diff
@@ -54,6 +57,19 @@ def update
 
   protected
 
+  def attribute_changes?
+    !@model_instance.changed_attributes.blank?
+  end
+
+  def approver_email_changed?
```
cmc333333: This method's no longer ever called, hey?

spec/features/ncr_work_orders_spec.rb
```diff
@@ -207,6 +207,16 @@
...
     end
 
+    it "does not resave unchanged requests" do 
+      visit "/ncr/work_orders/#{work_order.id}/edit"
+      click_on 'Submit for approval'
+      visit "/ncr/work_orders/#{work_order.id}/edit"
+      click_on 'Submit for approval'
+
+      expect(current_path).to eq("/proposals/#{work_order.proposal.id}")
+      expect(page).to have_content("No changes were made to the request")
```
cmc333333: Might be good to test that no emails were sent out, either. You should be able to use `deliveries` for this

cmc333333:
Sorry to do it @nickbristow , but would you mind upgrading bundle to fix the `Gemfile.lock` file. Should just be `gem update bundler`, I think

cmc333333:
@nickbristow agreed with pundit tweaks, so merging

PR #411, opened 2015-06-22
Delivers #90026260: a dashboard for 18F procurements.

annalee:
As an 18F admin, I can view a dashboard status of all accounted expenses at /gsa18f/dashboard.


cmc333333:
Copy-paste works for now (particularly to ship this client), but it'd be great if we spent some time thinking about how to make this more reusable.

PR #413, opened 2015-06-22
Fix master failure 20150622

phirefly:
Updated to fix test failures on master. 

Oddly, I noticed that only the `controller` type is needed to be explicitly mentioned. model and feature specs appear to be loading EnvironmentSpecHelper just fine.


phirefly:
Updated with note to pull out client-specific required env vars in the future

cmc333333:
I don't think the `around` is needed anymore, but it isn't hurting anyone

PR #416, opened 2015-06-24
[Delivers #94880126] ability to add observers

afeld:
This is a first pass of the functionality to add observers to Proposals. From the proposal show page, the user can click the "Add observers" link:

screen shot 2015-06-24 at 1 08 22 am

This takes them to a page that lists the observers, which they can add to inline:

screen shot 2015-06-24 at 1 08 30 am

which includes a confirmation once they've done so.

screen shot 2015-06-24 at 1 12 51 am

Still some TODOs noted in the PR, but thoughts overall?


afeld:
@phirefly @ericronne Bump!

phirefly:
Couple of comments:
• Deleting observers would be nice to add, if not now, in the next sprint.
• Placement of the `add observers` link will likely be moved but I'm okay with it for this first go around

Question: Can we notify the new observer at the time they are added? The original intent of this feature is to notify someone at the time the requester wants them to know about the request. We can discuss details in the tracker if need be.

afeld:
Updated!

> Deleting observers would be nice to add

Have a TODO in the controller, but yeah, let's take care of that separately.

> Can we notify the new observer at the time they are added?

Done in f7ea4b1.

app/controllers/observations_controller.rb
```diff
@@ -0,0 +1,30 @@
+class ObservationsController < ApplicationController
+  before_action :authenticate_user!
+  before_action :find_proposal
+  before_action ->{authorize @proposal, :can_edit!}
+
```
cmc333333: I think this is missing a `rescue_from` for authorization errors? It's part of a mixin in the ProposalsController

cmc333333:
Good stuff

app/controllers/observations_controller.rb
afeld: Good catch!

PR #417, opened 2015-06-24
[Delivers 94714266] link text to requests in emails

annalee:
Paired with @afeld 

When a requester modifies a request that has already been approved, the approver that has already taken action gets an email. This email does not require action (i.e. we do not display an 'Approve' button), but we still display `Or Send a Comment`.  Change this link to say `View this request.`


cmc333333:
Good stuff

PR #419, opened 2015-06-26
[Delivers #97164178] restrict creation of proposals to users with GSA emails

afeld:
...if the `RESTRICT_ACCESS` environment variable is set to `true`.


afeld:
I should add: we already have restriction on the viewing and editing of WorkOrders – this pull request optionally limits who can create them.

app/controllers/application_controller.rb
```diff
@@ -19,6 +19,7 @@ def authorize(record, query=nil, user=nil)
     user ||= @current_user
     record = self.authorizing_object(record)
     policy = Pundit.policy(user, record)
+    # TODO fall back to ProposalPolicy if there isn't a use-case-specific one
```
cmc333333: We do; see `authorizing_object`

app/controllers/use_case_controller.rb
```diff
@@ -3,6 +3,7 @@
 # * permitted_params
 class UseCaseController < ApplicationController
   before_filter :authenticate_user!
+  before_filter ->{authorize self.model_class.new}, only: [:new, :create]
```
cmc333333: I don't think you need to authorize `self.model_class.new` separately due to `authorizing_object`

spec/policies/gsa18f/procurement_policy_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  permissions :can_create? do
+    with_feature 'RESTRICT_ACCESS' do
+      it "doesn't allow someone with a non-GSA email to create" do
+        user = User.new(email_address: 'intruder@some.com')
+        procurement = Gsa18f::Procurement.new
+        expect(subject).not_to permit(user, procurement)
+      end
+    end
+  end
```
cmc333333: This covers the `RESTRICT_ACCESS=true` x non-gsa pair, but we probably want to test the three other permutations: (`RA=true`, gsa), (`RA=false`, non-gsa), (`RA=false`, gsa)

cmc333333:
This _would_ be more generic if GSA wasn't hard coded. What do you think about a regex env variable?

app/controllers/application_controller.rb
afeld: I think I had something wrong with my test that was causing a problem...*meh*.

app/controllers/use_case_controller.rb
afeld: Meaning it should just be `Proposal.new`?

afeld:
> This would be more generic if GSA wasn't hard coded.

I figure the restriction is going to vary by use case anyway (e.g. if we built one out for another agency), so not sure what benefit we'd get from making the details of the restriction configurable.

app/controllers/use_case_controller.rb
cmc333333: I thought you could collapse this into the following line, but the object doesn't exist yet, my mistake.

Authorizing on an empty object still seems suspicious. Would this work: `authorize self.model_class`?

cmc333333:
> I figure the restriction is going to vary by use case anyway
Sounds like a good candidate to be stored in the DB then ;) We can stick with this for now, but I'll throw another red flag here about hard coding values.

afeld:
Updated!

app/controllers/application_controller.rb
```diff
@@ -34,6 +34,11 @@ def authorize(record, query=nil, user=nil)
 
   # Proposals can have special authorization parameters in their client_data
   def authorizing_object(record)
+    if record.is_a?(Class)
+      # use an instance
+      record = record.new
+    end
```
cmc333333: I'm not sure this is required, but it doesn't hurt

cmc333333:
Thanks!

PR #420, opened 2015-06-26
97171464 fix email icons

phirefly:
Bypass the stripping of full domain url for css images that need to display within emails; temporary workaround pending updates to roady-rails and likely email design updates.

screen shot 2015-06-25 at 11 23 35 pm

### DIFF
--- a/app/assets/stylesheets/modules/shared/email_status.scss
+++ b/app/assets/stylesheets/modules/shared/email_status.scss
@@ -36,7 +36,7 @@
 
     &.last {
       img.approved.linear, img.pending.linear, img.rejected.linear {
-        background-image: none;
+        background-image: none !important;
       }
 
       img {
--- a/app/assets/stylesheets/modules/shared/email_status.scss
+++ b/app/assets/stylesheets/modules/shared/email_status.scss
@@ -42,7 +42,7 @@
 
     &.last {
       img.approved.linear, img.pending.linear, img.rejected.linear {
-        background-image: none !important;
+        background-image: none;
       }
 
       img {
@@ -60,22 +60,12 @@
 
     img {
       float: left;
-      padding-top: 2px;
-      &.approved.linear {
-        background-image: image-url('bg_approved_status.gif');
-      }
-
-      &.pending.linear {
-        background-image: image-url('bg_pending_status.gif');
-      }
-
-      &.rejected.linear {
-        background-image: image-url('bg_rejected_status.gif');
-      }
     }
 
     .approver {
       padding-left: 7px;
+      position: relative;
+      top: -5px;
     }
   }
 }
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/modules/shared/email_status.scss
```diff
@@ -36,7 +42,7 @@
 
     &.last {
       img.approved.linear, img.pending.linear, img.rejected.linear {
-        background-image: none;
+        background-image: none !important;
```
cmc333333: Is there a way to get what you need without the `!important`? Generally should be avoided

app/helpers/communicart_mailer_helper.rb
```diff
@@ -1,6 +1,12 @@
 module CommunicartMailerHelper
-  def status_icon_tag(status, linear=false)
-    image_tag("icon-#{status}.png", class: "status-icon #{status} #{'linear' if linear}")
+  def status_icon_tag(status, linear_display=false)
+    base_url = root_url.gsub(/\?.*$/,'').chomp("/")
+    bg_linear_image = base_url + image_path("bg_#{status}_status.gif")
+
+    image_tag(base_url + image_path("icon-#{status}.png"),
+              class: "status-icon #{status} #{'linear' if linear_display}",
+              style: ("background-image: url('#{bg_linear_image}');" if linear_display)
```
cmc333333: If this is an image tag, why does it need a `background-image`? Shouldn't the `src` be this image?

phirefly: We need to be able to have a variable vertical width to work with any font size. If we hard code the image, we aren't able to make the height variable.

app/assets/stylesheets/modules/shared/email_status.scss
phirefly: In order to bypass the roady-rails bug, I have to inline the styles directly in the markup. The inline styles are taking precedence over the css styles that remove the vertical line for the 'last' approver. Otherwise, it would show as follows:

screen shot 2015-06-26 at 1 42 09 pm

I can probably pull the `last` class into the img tag... will try that.

app/helpers/communicart_mailer_helper.rb
cmc333333: Gotcha -- the background image is the line that connects, my mistake.

app/assets/stylesheets/modules/shared/email_status.scss
cmc333333: :+1: I think that'd solve without requiring `!important`

afeld: Also, mind adding a comment about why it's needed?

cmc333333:
:+1:

PR #421, opened 2015-06-26
[Delivers #97017456] Adding BA60 expense type.

annalee:
Updates the NCR workorder form to include another `BA60` Expense type, and adds a feature flag to show/hide the BA60 radio button.

Paired with @afeld.


cmc333333:
Good stuff. Only question I have is around defaults. Hiding BA61 was the exception case (the system defaults to showing BA61). If I understand this, though, BA60 will default to being shown. Is that the case and is that desirable?

phirefly:
Let's default to hiding BA60.

This probably needs to be updated too:
https://github.com/18F/C2/blob/master/app/views/ncr/work_orders/form.html.erb#L5-L7

Would it help if we just changed the flags to be positive (e.g. `SHOW_BA60`, `SHOW_BA801, etc?) Perhaps it won't make a difference if we go live with all types next weeks but wanted to throw that out there.

phirefly:
Regarding defaulting to show, I would be ok with leaving as is but perhaps using the positive form of it in the future

phirefly:
Per Slack discussion, we'll leave https://github.com/18F/C2/blob/master/app/views/ncr/work_orders/form.html.erb#L5-L7 as is and plan to launch with 60/61 at the same time.

phirefly:
:shipit:

PR #422, opened 2015-06-26
more deprecation fixes

afeld:


cmc333333:
Any reason not to wait until there's a full release?

afeld:
Because the backtrace being printed by Guard on every test failure is really getting on my nerves :wink:

afeld:
Will do another PR with the released version as soon as it comes out though.

PR #423, opened 2015-06-26
[Delivers #97560640] add validation on budget fields

afeld:
Paired w/ @annalee. Spec changes are easier to see [without whitespace](https://github.com/18F/C2/pull/423/files?w=1).

### DIFF
--- a/app/models/ncr/work_order.rb
+++ b/app/models/ncr/work_order.rb
@@ -36,7 +36,7 @@ module Ncr
     validates :expense_type, inclusion: {in: EXPENSE_TYPES}, presence: true
     validates :function_code, format: {
       with: /\APG[A-Z0-9]{3}\z/,
-      message: "must start with 'PG', followed by letters or numbers"
+      message: "must start with 'PG', followed by three letters or numbers"
     }, allow_blank: true
     validates :project_title, presence: true
     validates :vendor, presence: true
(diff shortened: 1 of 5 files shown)


afeld:
See the (unusually long for me) commit message in 099ac45. We might consider making this the behavior for all (non-required?) string attributes: if they are `#blank?`, set them to `nil`. Especially important for the ones we'd actually care about querying over, I suppose.

app/models/ncr/work_order.rb
```diff
@@ -28,21 +29,56 @@ class WorkOrder < ActiveRecord::Base
...
       message: "must be greater than or equal to $0"
     }
+    validates :cl_number, format: {
+      with: /\ACL\d{7}\z/,
+      message: "must start with 'CL', followed by seven numbers"
+    }, allow_blank: true
     validates :expense_type, inclusion: {in: EXPENSE_TYPES}, presence: true
+    validates :function_code, format: {
+      with: /\APG[A-Z0-9]{3}\z/,
+      message: "must start with 'PG', followed by letters or numbers"
```
cmc333333: For consistency, should this say "3 letters or numbers"? Or "characters" to match `soc_code` below? "Characters" isn't exactly accurate, though.

app/models/ncr/work_order.rb
```diff
@@ -28,21 +29,56 @@ class WorkOrder < ActiveRecord::Base
...
+      if self.cl_number.present?
+        self.cl_number = self.cl_number.upcase
+        self.cl_number.prepend('CL') unless self.cl_number.start_with?('CL')
+      else
+        self.cl_number = nil
+      end
+
+      if self.function_code.present?
+        self.function_code.upcase!
+      else
```
cmc333333: Why not do the `prepend('PG')` thing?

afeld:
Updated!

cmc333333:
Thanks!

PR #424, opened 2015-06-27
[Delivers #97071484] [Delivers #94434346] No more api calls

cmc333333:
The Work Order edit/create page no longer accepts custom email addresses (the user must be from a pre-defined list) and no longer makes AJAX calls (fixing an issue around editing a pre-selected but custom building). This uses a a helper very similar to the one in #402


app/helpers/ncr/work_orders_helper.rb
```diff
@@ -0,0 +1,16 @@
+module Ncr
+  module WorkOrdersHelper
+    def approver_options
+      # @todo should this list be limited by client/something else?
+      # @todo is there a better order? maybe by current_user's use?
```
afeld: Maybe putting ones they've used before first would be good, but this is fine for now.

PR #425, opened 2015-06-28
[Delivers #94419078] Add scripts/manifest so that cloud foundry will run migrations on deploy

cmc333333:
App continues to work when deployed to dev and the output from `cf push` indicates that the migration script is ran. However, we'll want to wait until there's an actual migration to confirm.


phirefly:
:ship: it

afeld:
Does this change anything about how we deploy?

cmc333333:
Shouldn't, no. Only problem that might arise (as it did this morning) if pushing `master` followed by a previous commit (which doesn't have the `cf` rake command).

PR #426, opened 2015-06-28
[Finishes #97728212] Removes Observer/Event Queue Pattern

cmc333333:
This might make sense in the future, but right now the abstraction just makes logic harder to follow.


phirefly:
:shipit:

PR #427, opened 2015-06-29
95185144 cancel request

phirefly:
One more item in the task list to take care of but wanted to solicit feedback on the [Cancel request](https://www.pivotaltracker.com/story/show/95185144) story.

Do not merge.


app/controllers/proposals_controller.rb
```diff
@@ -23,6 +23,35 @@ def archive
...
 
+  def cancel_form
+    @proposal = self.proposal.decorate
+    unless policy(@proposal).can_cancel?
+     redirect_to proposals_path, alert: "You are not allowed to perform that action"
+    end
+
+    if @proposal.cancelled?
+     redirect_to proposal_path, id:@proposal.id, alert: "This request that has already been cancelled."
+    end
```
cmc333333: You shouldn't need an explicit check on `can_cancel?` -- that should be automatically checked when `authorize`ing. Further, `can_cancel?` should probably both check that the user is allowed to perform the action and that the proposal isn't cancelled already. See `ProposalPolicy#can_edit!` for an example.

You'll still need the redirect logic, however; see examples w/ `auth_errors`. You can check `exception.query` (which would be a symbol).

app/models/proposal.rb
```diff
@@ -9,15 +9,21 @@ class Proposal < ActiveRecord::Base
...
     end
     state :rejected do
       # partial approvals and rejections can't break out of this state
       event :partial_approve, :transitions_to => :rejected
       event :reject, :transitions_to => :rejected
       event :restart, :transitions_to => :pending
+      event :cancel, :transitions_to => :cancelled
+    end
+    state :cancelled do
+      event :partial_approve, :transitions_to => :cancelled
```
cmc333333: Can a canceled proposal be restarted?

Also, let's get rid of rejected while you're working on this

app/models/proposal.rb
```diff
@@ -48,6 +54,7 @@ class Proposal < ActiveRecord::Base
     scope status, -> { where(status: status) }
   end
   scope :closed, -> { where(status: ['approved', 'rejected']) }
+  scope :cancelled, -> { where(status: 'cancelled') }
```
cmc333333: Should `cancelled` be added to `:closed` as well?

app/views/proposals/show.html.erb
```diff
@@ -41,12 +41,18 @@
   </div>
 </div>
 
-<% if policy(@proposal).can_edit? %>
+<% if policy(@proposal).can_edit? && !@proposal.cancelled? %>
```
cmc333333: The `can_edit` policy should probably include a `cancelled` check, hey?

app/views/shared/_email_status.html.erb
```diff
@@ -1,45 +1,49 @@
 <tr class="m-shared-email_status">
   <td colspan="2" padding="10">
     <table id="status-container-detail" width="770" class="col-md-12">
-      <tr class="status-header-container">
+      <tr class="status-header-container cancelled">
```
cmc333333: Should the "cancelled" class be added regardless of the proposal's state?

app/models/proposal.rb
phirefly: At this point, we are not allowing them to restart a cancelled request, although it might be nice to look at that soon

app/models/proposal.rb
phirefly: Great thought

app/views/proposals/show.html.erb
phirefly: +1

app/views/shared/_email_status.html.erb
phirefly: Good catch. Thankya!

phirefly:
Updated per feedback. Screens attached:

screen shot 2015-07-16 at 12 32 12 pm

screen shot 2015-07-16 at 12 32 26 pm

screen shot 2015-07-16 at 12 32 41 pm

screen shot 2015-07-16 at 12 33 01 pm

app/controllers/proposals_controller.rb
```diff
@@ -71,4 +93,12 @@ def proposals
   def chronological_proposals
     self.proposals.order('created_at DESC')
   end
+
+  def auth_errors(exception)
```
phirefly: I'm still not sure how best to handle this. I would consider pulling this into TokenAuth, but the redirect to a different url adds a bit of complexity (knowledge about specific actions related to the different url). Thoughts?

app/controllers/proposals_controller.rb
```diff
@@ -2,11 +2,12 @@ class ProposalsController < ApplicationController
   include TokenAuth
 
   before_filter :authenticate_user!, except: :approve
-  before_filter ->{authorize self.proposal}, only: :show
+  before_filter ->{authorize self.proposal}, only: [:show, :cancel, :cancel_form]
   before_filter :needs_token_on_get, only: :approve
   before_filter :validate_access, only: :approve
   helper_method :display_status
   add_template_helper ProposalsHelper
+  rescue_from Pundit::NotAuthorizedError, with: :auth_errors
```
cmc333333: Not positive is this is needed, given the `TokenAuth` include

app/controllers/proposals_controller.rb
cmc333333: You raise a good point. We may want to move more out of `TokenAuth`, too

app/views/communicart_mailer/cancellation_email.html.erb
```diff
@@ -0,0 +1,12 @@
+<div>
+  <p>
+    Hello,<br/>
+    The request, <%= @proposal.name %> (<%= @proposal.public_id %>), has been cancelled.
+  </p>
```
cmc333333: May be useful to have the reason here, though clearly not pressing.

app/views/proposals/show.html.erb
```diff
@@ -42,12 +42,14 @@
...
 <% if policy(@proposal).can_edit? %>
-  <div class="centered">
-    <p>
-      <%= modify_client_button(@proposal) %>
-      <%= link_to "Add observers", proposal_observations_path(@proposal) %>
-    </p>
-  </div>
+  <%= client_partial(@proposal.client, 'restart_link',
+                     locals: {proposal: @proposal})
+  %>
```
cmc333333: I think this partial has been completely removed

cmc333333:
Comments are on a bunch of harmless little things. This is a big undertaking; nice job @phirefly

PR #432, opened 2015-06-30
rearrange NCR feature specs to distinguish creating vs. editing

afeld:
Just to make navigating the spec file a little easier.


phirefly:
:shipit:

PR #434, opened 2015-07-01
[Delivers #93031034] make the app production-ready

afeld:
This PR does a few things:

* Uses Puma instead of WEBrick as the server
* Enables garbage collection profiling in production (which is captured by New Relic [here](https://rpm.newrelic.com/accounts/921394/applications/5294993/ruby_vms))
* Sets the required Ruby version explicitly
* Gets rid of Cloud Foundry build warnings during `push`

Tested these changes locally and on `dev`.


config/puma.rb
```diff
@@ -0,0 +1,21 @@
+# https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server
+
+workers Integer(ENV['WEB_CONCURRENCY'] || 1)
```
afeld: Going any higher caused the app to crash due to low memory. CF gives `512M` by default...not sure if we should consider raising it.

afeld:
Also, if/when this is merged, everyone needs to run the following locally:

```bash
git fetch origin
git checkout origin/master
rvm get stable
rvm install 2.2.2
gem install bundler
bundle
```

cmc333333:
Makes sense to me, but I'd like a review by a ruby person

.ruby-version
```diff
@@ -1 +0,0 @@
-2.1.3
```
monfresh: This is actually used by local developers so that their RVM or rbenv knows which version is needed by the project. Specifying the Ruby version in the Gemfile has a separate purpose than specifying it in `.ruby-version`.

Procfile
```diff
@@ -0,0 +1 @@
+web: bundle exec rake cf:on_first_instance db:migrate && bundle exec puma
```
monfresh: To tell Puma to use the settings defined in `config/puma.rb`, you need to add `-C ./config/puma.rb` at the end of this line.

config/puma.rb
monfresh: That is strange. I have an instance of Ohana API running on 2 Heroku dynos with 2 Puma workers and 5 Puma threads, and memory usage rarely goes above 340MB.

.ruby-version
afeld: [Using the Gemfile works for RVM](https://rvm.io/workflow/projects#supported-files), but fair enough in case someone else might be using [rbenv](https://github.com/sstephenson/rbenv#choosing-the-ruby-version).

Procfile
afeld: It looks for `config/puma.rb` by default:

https://github.com/puma/puma#configuration-file

.ruby-version
monfresh: Yeah, I was wrong about that. However, if you only rely on the Gemfile for specifying the Ruby version, you will see this warning every time you `cd` into the project (unless you disable the warning per the instructions):
```
RVM used your Gemfile for selecting Ruby, it is all fine - Heroku does that too,
you can ignore these warnings with 'rvm rvmrc warning ignore /Users/moncefbelyamani/projects/18f/save-ferris/Gemfile'.
To ignore the warning for all files run 'rvm rvmrc warning ignore allGemfiles'.
```

config/puma.rb
afeld: Hmmmmmm...weird. Not sure that it would make a difference, but do you remember if you were using MRI?

Procfile
monfresh: Ahh, nice. I didn't realize that. I've always assumed you had to specify it.

config/puma.rb
```diff
@@ -0,0 +1,21 @@
...
+environment ENV['RACK_ENV'] || 'development'
+
+on_worker_boot do
+  # Worker specific setup for Rails 4.1+
+  # See: https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#on-worker-boot
+  ActiveRecord::Base.establish_connection
+
+  # https://github.com/Tomohiro/plate/commit/0a5a072c8d882f32a48d8bdbfa325e83719f58f8
+  require 'newrelic_rpm'
+  NewRelic::Agent.manual_start
```
afeld: @monfresh Did you ever find another work around for [this issue](https://github.com/puma/puma/issues/614#issuecomment-72401569)?

monfresh: I haven't seen that issue in a while. I didn't do anything specific to resolve it. I think removing the appsignal gem and/or updating to the latest newrelic_rpm and/or puma gems fixed it. Also, in ohana-api, I require newrelic_rpm all the time. I don't manually start it in [puma.rb](https://github.com/codeforamerica/ohana-api/blob/master/config/puma.rb).

monfresh: Another change I made when [switching from Unicorn to Puma](https://github.com/codeforamerica/ohana-api/commit/403fe4dc452e7be8a92e6271d95e5c44d559fc76) was to set the database pool to match the Puma `threads_count`. This was based on the Heroku documentation.

config/puma.rb
monfresh: Yes, I'm using Ruby 2.2.1 as provided by Heroku. Which reminds me to update to 2.2.2.

afeld: Doesn't seem like you have `newrelic_rpm` included in [the Gemfile](https://github.com/codeforamerica/ohana-api/blob/master/Gemfile)...am I missing something? Might be why you're not seeing the issue :wink:

config/puma.rb
afeld: Doesn't seem like you have `newrelic_rpm` included in [the ohana-api Gemfile](https://github.com/codeforamerica/ohana-api/blob/master/Gemfile)...am I missing something? Might be why you're not seeing the issue :wink:

config/database.yml
```diff
@@ -1,13 +1,17 @@
...
-  password: <%= ENV['DB_PASSWORD'] %>
-  host: <%= ENV['DB_HOST'] %>
-  port: <%= ENV['DB_PORT'] %>
-  pool: 5
 
-development: *default
-test: *default
+  # should match puma.rb
+  # https://devcenter.heroku.com/articles/concurrency-and-database-connections#threaded-servers
+  pool: <%= ENV['MAX_THREADS'] || 5 %>
```
afeld: No idea why this is causing [build failure](https://circleci.com/gh/18F/C2/1308) :rage:

config/puma.rb
monfresh: Sorry for the confusion. New Relic isn't included by default in the generic repo (but maybe it should be). [San Mateo County's instance](https://github.com/smcgov/ohana-api-smc/blob/master/Gemfile#L69) of the project has New Relic.

config/database.yml
monfresh: My guess is that this is due to your use of `.env`. Try defining `MAX_THREADS` in your `.env`.

config/puma.rb
jgrevich: Doesn't the mem usage depend on what libs you're including and your app code (storing large amounts of data in mem or mem leaks can cause for some bloated memory). It does seem a bit odd that it's using so much memory compared to what @monfresh is seeing with his app.

jgrevich: RE: raising mem, it sounds like increasing the available memory will allow for greater concurrency/threads which is a good thing assuming that the core resources (CPU, RAM (no swap), disk/net IO) on the host are not taxed.

I'm no expert but this is what I (think I) understand. If your code was 100% CPU bound then you want one thread per CPU (I think the ideal is two with hyper threading (HT) enabled chips). A 100% CPU bound app is usually never the case and you're going to have to do some experimenting to find out what amount of conc/threads works best for your setup/resource constraints (historically we'd be limited to physical HW, 2-8 core chips and max of ~32GB ram per host). To test, I think you want to map concurrency to the number of CPU cores. Then, start adding more threads and your performance should increase since the processor is most likely going to be waiting on disk/net IO and free to do other tasks (you'll want to be watching cpu, mem (swap), disk/net IO during these tests). I would keep adding threads until you see one of your core resources constrained (memory is swapping, cpu is > 100%, etc) or a steady degradation in performance in which case your now loosing to the excess context switching between threads, not the disk/net IO. If memory is flexible, just keep an eye on it and make sure there's enough to prevent it from having to swap.

Let us know how it goes.

config/database.yml
jgrevich: According to circleci `MAX_THREADS` isn't defined anywhere and it doesn't seem to like that. Does setting that constant fix the issue or just using an integer value for that attribute?

afeld:
Mmk fixed the build failure.

cmc333333:
I'm cool merging unless you wanted to tweak anything else per @monfresh @jgrevich 's comments

afeld:
we'll do it live

PR #449, opened 2015-07-05
[Delivers #98159348] remove Cart model

afeld:
Amazing how cathartic deleting code is for me :relaxed: 

* [x] Remove all references to "carts"
* [ ] Remove unused tables (will do this is a separate PR, to be pushed out later)
    * `carts`
    * `approval_groups`
    * `properties`


cmc333333:
Good stuff! Logic continues to make sense to me, though I didn't dive super deep.

There are a few specs which probably need to be copy-pasted + tweaked (rather than just deleted) if their equivalent doesn't already exist. From what I saw, these include:
* `describe 'partial approvals'`
* `context 'scopes'`
* `describe '#on_rejected_entry'`
* `describe '#restart'`

afeld:
:sweat_smile: Got through it! Still need to look into @cmc333333's suggestion above, but otherwise should be pretty close.

Huge number of changes (sorry!), but thankfully most are fairly simple.

afeld:
Updated!

cmc333333:
Very nice

PR #450, opened 2015-07-06
92032848 / Progress in adding client side validation

anthonygarvan:
This pull request covers integrating a client side validation framework, Parsley.js, into the NCR form. So far, it works for all text fields, the selectize fields, and the radio button. I also added reasonable limits on string length that I suggested a few days ago and is now associated with story 98243294.

Couldn't quite close this one out due to the custom validation required for the B80 / RWA Number logic. but I thought it would be good to issue a PR now since I might not have time to get back to it.


afeld:
Replaced by #451.

PR #451, opened 2015-07-06
92032848 / Progress in adding client side validation

anthonygarvan:
This pull request covers integrating a client side validation framework, [Parsley.js](http://parsleyjs.org/), into the NCR form. I also added reasonable limits on string length that I suggested a few days ago (98243294).

I think I covered all the validations for this form:
- required text fields
- required radio buttons
- required selectize fields
- conditional required for BA80 / RWA Number
- min ($0) and max ($3000) values for Amount

Overall, I am pretty happy with Parsley as a client side validation framework - the easy things were easy, and the hard things (eg., integrating with selectize.js, conditional validations) were possible by overriding or extending the default behavior.

Also, I would recommend creating a tech debt story / chore to remove server side validation once this story is accepted. With such a small team you definitely don't want to maintain the same validations in two different places, and it would be very confusing for new developers to see that in the code.

### DIFF
--- a/app/decorators/proposal_decorator.rb
+++ b/app/decorators/proposal_decorator.rb
@@ -56,10 +56,6 @@ class ProposalDecorator < Draper::Decorator
     "#{number_approved} of #{total_approvers} approved."
   end
 
-  def email_subject
-    "Request #{proposal.public_identifier}"
-  end
-
   def email_msg_id
     "<proposal-#{self.id}@#{DEFAULT_URL_HOST}>"
   end
--- a/app/mailers/communicart_mailer.rb
+++ b/app/mailers/communicart_mailer.rb
@@ -113,9 +113,29 @@ class CommunicartMailer < ActionMailer::Base
 
     mail(
       to: to_email,
-      subject: @proposal.email_subject,
+      subject: proposal_subject(@proposal),
       from: from_email || default_sender_email,
       template_name: template_name
     )
+  end
+
+  def proposal_subject(proposal)
+    params = proposal.as_json
+    #todo: replace with public_id once #98376564 is fixed
+    params[:public_identifier] = proposal.public_identifier
+    # Add in requester params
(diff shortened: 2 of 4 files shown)


cmc333333:
I played with Parsley on another project and like it a lot. Ideally, we'd be defining validation logic once (and only once) per field; being more declarative gets a bit closer to that goal.

app/assets/javascripts/selectizer.js.coffee
```diff
@@ -47,3 +47,24 @@ $ ->
     selectizer = new Selectizer(el)
     selectizer.enable()
     selectizer.add_label()
+  
+  $('#new_ncr_work_order').parsley
```
cmc333333: This'd be a bit better if it weren't specific to the work order page. Can you think of a way to implement the same thing more declaratively (e.g. the selectize and filtering data attributes)?

app/views/ncr/work_orders/_ba_80_fields.html.haml
```diff
@@ -1,6 +1,7 @@
 .form-group{data: {filter_key: 'expense-type', filter_value: 'BA80'}}
   = f.label :rwa_number, class: "required"
-  = f.text_field :rwa_number, class: 'form-control'
+  = f.text_field :rwa_number, class: 'form-control', 'data-parsley-conditionalrequired':'#ncr_work_order_expense_type_ba80', 'data-parsley-validate-if-empty':'true', 'data-parsley-conditionalrequired-message':'RWA Number is required for BA80 expense types'
```
cmc333333: Maybe we have the field required, but only if visible? That seems like a general-purpose solution

cmc333333:
Would you mind using the HTML5 attributes instead of the `data-` throughout? E.g. use `maxlength` and `max`/`min`/`range` instead of `data-parsley-maxlength`, `data-parsley-min`, etc.

Just a bit cleaner

PR #454, opened 2015-07-07
WIP [Delivers #98024644] More flexible approvals

cmc333333:
Currently, we have two types of approval flows: `linear` and `parallel`. This implements a wider variety of approval chains, e.g.
```
# Approval hierarchy version of needing two of the following:
# 1) Amy AND Bob
# 2) Carrie
# 3) Dan THEN Erin
```

It accomplishes this by specifying a tree of approvals, with each proposal containing one `root`. Nodes in this tree can be `Individual`, `Parallel`, or `Serial` (with the potential to add more types in the future.) The later two may contain child nodes; the parent controls the communications with their children. Kicking off the approval process is now just a matter of `proposal.root_approval.make_actionable!`.

Much logic has been consolidated -- this removes the need for the linear dispatcher and proposal's "flow" field is made mostly redundant.

There's a lot of code change here, but the vast majority is around accessibly slightly different fields or using slightly different constructors.

~~This isn't *quite* ready to merge. There's still a data migration missing.~~  Additional cleanup can be made around the "flow" field and dispatchers, but I'll hold off until a later PR.

TODOs:
* [ ] Address comments
* [x] Add approved_at field


db/migrate/20150707214004_add_root_approvals.rb
```diff
@@ -0,0 +1,28 @@
...
+    SQL
+    execute <<-SQL
+      UPDATE approvals
+      SET parent_id = (SELECT root.id
+                       FROM approvals root
+                       WHERE root.proposal_id = approvals.proposal_id
+                       AND root.type <> 'Approvals::Individual')
+      WHERE type = 'Approvals::Individual';
+    SQL
+  end
```
afeld: This migration seems complex enough that we may want to run it against a snapshot of production data...thoughts?

app/models/approval.rb
```diff
@@ -6,10 +6,8 @@ class Approval < ActiveRecord::Base
     end
     state :actionable do
       event :approve, transitions_to: :approved
-      event :reject, transitions_to: :rejected
     end
     state :approved
-    state :rejected
```
afeld: Is this change related? If not, mind if we move it to a separate PR?

app/models/approval.rb
```diff
@@ -53,14 +54,14 @@ def approved_at
...
-    Dispatcher.on_approval_approved(self)
+  # notify parents if we've been approved. Notified as a callback so that it
+  # will be present even if subclasses override workflow
+  def on_approved_entry(old_state, event)
+    if self.parent
+      self.parent.child_approved!
+    else
+      self.proposal.approve!
+    end
+    self.reload   # Account for proposal changes
```
afeld: This seems like it shouldn't be necessary in here...more the responsibility of the caller.

app/models/approvals/individual.rb
```diff
@@ -0,0 +1,14 @@
+module Approvals
+  class Individual < Approval
```
afeld: Please

* [ ] Add a comment explaining what this class is for
* [ ] Add a validation for `validates :user, presence: true` (assuming I understand it's purpose correctly)

app/models/approvals/parallel.rb
```diff
@@ -0,0 +1,24 @@
...
+        event :make_actionable, transitions_to: :actionable
+      end
+      state :actionable do
+        on_entry { self.child_approvals.each(&:make_actionable!) }
+        event :child_approved, transitions_to: :approved do
+          halt unless self.min_required_met?
+        end
+      end
+      state :approved do
+        event :child_approved, transitions_to: :approved do halt end   # additional approvals do nothing
```
afeld: Mind splitting to three lines so it shows up in code coverage?

app/models/approvals/parallel.rb
```diff
@@ -0,0 +1,24 @@
+module Approvals
+  class Parallel < Approval
+    workflow do
+      state :pending do
+        event :make_actionable, transitions_to: :actionable
+      end
+      state :actionable do
+        on_entry { self.child_approvals.each(&:make_actionable!) }
+        event :child_approved, transitions_to: :approved do
+          halt unless self.min_required_met?
```
afeld: Ditto the statement below.

app/models/approvals/serial.rb
```diff
@@ -0,0 +1,23 @@
...
+  class Serial < Approval
+    workflow do
+      state :pending do
+        event :make_actionable, transitions_to: :actionable
+      end
+      state :actionable do
+        on_entry { self.trigger_next_child }
+        event :child_approved, transitions_to: :approved do
+          self.trigger_next_child
+          halt unless self.child_approvals.where.not(status: 'approved').empty?
```
afeld: Minor: Mind moving into an `all_children_approved?` method?

app/models/proposal.rb
```diff
@@ -14,17 +11,17 @@ class Proposal < ActiveRecord::Base
...
       event :restart, :transitions_to => :pending
     end
   end
 
   has_one :cart
+  has_one :root_approval, ->{ where(parent_id: nil) }, class_name: 'Approval'
   has_many :approvals
-  has_many :approvers, through: :approvals, source: :user
-  has_many :api_tokens, through: :approvals
+  has_many :user_approvals, ->{ where.not(user_id: nil) }, class_name: 'Approval'
```
afeld: Just double-checking – we don't actually need to walk the tree, right?

spec/models/approval_spec.rb
```diff
@@ -20,4 +20,86 @@
...
+      proposal.root_approval = Approvals::Parallel.create(min_required: 2)
+      and_clause = Approvals::Parallel.create(parent: proposal.root_approval, proposal: proposal)
+      then_clause = Approvals::Serial.create(parent: proposal.root_approval, proposal: proposal)
+
+      FactoryGirl.create(:approval, user: amy, proposal: proposal, parent: and_clause)
+      FactoryGirl.create(:approval, user: bob, proposal: proposal, parent: and_clause)
+      FactoryGirl.create(:approval, user: carrie, proposal: proposal, parent: proposal.root_approval)
+      FactoryGirl.create(:approval, user: dan, proposal: proposal, parent: then_clause)
+      FactoryGirl.create(:approval, user: erin, proposal: proposal, parent: then_clause)
+      proposal.root_approval.make_actionable!
```
afeld: Minor: Let's do all of this in a `before` block.

spec/models/approval_spec.rb
```diff
@@ -20,4 +20,86 @@
...
+    }
+
+    it "approves via Amy, Bob, Carrie" do
+      expect(proposal.reload.approved?).to be false
+      proposal.approval_for(amy).approve!
+      expect(proposal.reload.approved?).to be false
+      proposal.approval_for(bob).approve!
+      expect(proposal.reload.approved?).to be false
+      proposal.approval_for(carrie).approve!
+      expect(proposal.reload.approved?).to be true
```
afeld: Shouldn't Dan and Erin have to approve, too?

spec/models/approval_spec.rb
```diff
@@ -20,4 +20,86 @@
...
+      FactoryGirl.create(:approval, user: amy, proposal: proposal, parent: and_clause)
+      FactoryGirl.create(:approval, user: bob, proposal: proposal, parent: and_clause)
+      FactoryGirl.create(:approval, user: carrie, proposal: proposal, parent: proposal.root_approval)
+      FactoryGirl.create(:approval, user: dan, proposal: proposal, parent: then_clause)
+      FactoryGirl.create(:approval, user: erin, proposal: proposal, parent: then_clause)
+      proposal.root_approval.make_actionable!
+      proposal
+    }
+
+    it "approves via Amy, Bob, Carrie" do
```
afeld: How about wording this "single approval in an AND block" or something? Just to be more clear about what it's _actually_ testing.

spec/models/approval_spec.rb
cmc333333: No, it only needs *two* of these three paths

1.  Amy AND Bob
1. Carrie
1. Dan THEN Erin

spec/models/approval_spec.rb
```diff
@@ -20,4 +20,86 @@
...
+      expect(approval.proposal.approved?).to eq false
+      expect(approval.proposal.pending?).to eq true
+    end
+  end
+
+  describe "complicated approval chains" do
+    # Approval hierarchy version of needing two of the following:
+    # 1) Amy AND Bob
+    # 2) Carrie
+    # 3) Dan THEN Erin
```
afeld: This made me think: now that we have this more...flexible...structure, we should have a more expressive way to set up the approval trees, that looks more like this comment than the huge block below. Not a blocker.

spec/models/approvals/parallel_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    expect(third.reload.status).to eq('approved')
+
+    second.approve!
+    expect(root.reload.status).to eq('approved')
+    expect(first.reload.status).to eq('approved')
+    expect(second.reload.status).to eq('approved')
+    expect(third.reload.status).to eq('approved')
+  end
+
+  it 'can be used for disjunctions' do
```
afeld: Mind defining this in a comment? Not sure that I understand what it means.

spec/models/approval_spec.rb
cmc333333: Indeed, a DSL would be nice. First task is it make it work, though.

afeld:
Phew, ok. I didn't review the _entire_ thing – mostly focused on the models and the specs. A few things that I wrote down:

* The `min_required` thing is unintuitive to me.
* The `Approval` subclasses could use some validations.
* Not necessarily a blocker, but mind adding a bit of an architectural overview about how the approvals work somewhere under `doc/`?

Let me know when you're done with the merge from `master` and I can take another pass.

spec/models/approval_spec.rb
afeld: :+1:

app/models/proposal.rb
```diff
@@ -92,43 +92,31 @@ def users
...
+        existing.parent = new_approval.parent   # this assumes the parent hasn't been replaced
+        existing
+      else
+        new_approval
+      end
+    end
+    self.approvals = new_approvals
+    # position might be out of whack, so we reset it
+    new_approvals.each_with_index do |approval, idx|
+      approval.set_list_position(idx + 1)   # start with 1
```
cmc333333: As the approvals are potentially being modified, we'll want to add an explicit field for approval approved time.

app/models/approval.rb
cmc333333: Removing `rejected` made this code easier, but you're right, a separate PR would be preferable and will cut down the size of this PR.

cmc333333:
This PR touches so much that it takes more than a day to resolve merge conflicts with master. I'm going to try to split this into smaller, forward compatible PRs, which we can merge in sooner.

afeld:
Closing since it sounds like this is being split up, but feel free to re-open if you want to use this PR again!

PR #456, opened 2015-07-08
Feature/change approvals with expense change

nickbristow:
This fixes the bug where AO wasn't updated when the approving official was changed on the form. 

It also correctly adds the new approving officials when changing expense type


app/mailers/communicart_mailer.rb
```diff
@@ -27,6 +27,18 @@ def notification_for_approver(to_email, approval, alert_partial=nil)
...
+    @approval = approval
+    @alert_partial = alert_partial
+    proposal = approval.proposal
+    send_proposal_email(
+      from_email: user_email_with_name(proposal.requester),
+      to_email: to_email,
+      proposal: proposal,
+      template_name: 'proposal_notification_email'
+    )
+  end
```
cmc333333: Should there be a template or something associated with this email? Right now it's identical to `notification_for_approver`

app/models/ncr/work_order.rb
```diff
@@ -194,6 +201,24 @@ def self.ba80_budget_mailbox
...
+        end
+        system_approvers.each do |email|
+          self.add_approver(email)
+        end
+        approvals = self.approvals
+        if(approvals.first.approved?)
+          approvals.second.make_actionable!
+        end
+      end
+    end
```
cmc333333: Can this be combined with `add_approvals` and `update_approver`? It seems there are a lot of ways to set and reset the approval list. I think we can make one entry point that accounts for all combinations?

app/models/proposal.rb
```diff
@@ -102,6 +102,13 @@ def add_approver(email)
     approval
   end
 
+  def remove_approver(email)
+    user = User.for_email(email)
+    approval = self.approvals.find_by(user_id: user.id)
```
cmc333333: I think there's an `approval_for`

cmc333333:
#454 may make this logic a bit simpler. It sends an email when an approval becomes actionable, so you won't need to include all of that logic inside the controller and model.

cmc333333:
The variable names are a little misleading, but I think the logic is sound. We'll tweak it further after the approval hierarchies are introduced.

PR #458, opened 2015-07-09
Updated test to check for , updated selectizer options

nickbristow:
Selectize now splits on `XxxxxXXxxxxX` which has a far less likely error rate than splitting on `,`


cmc333333:
:+1:

PR #462, opened 2015-07-10
[Completes #98751750] Added NYC to 18f config

phirefly:


afeld:
:shipit: (when tests pass)

PR #465, opened 2015-07-10
[Delivers #97288506] includes previous value in change log

annalee:
comments on work orders post-approval.

"As a USER, I can see the before and after values on a request modified post-approval."

paired with @phirefly


app/models/ncr/work_order.rb
```diff
@@ -203,9 +203,10 @@ def record_changes
       comment_texts = []
       bullet = changed_attributes.length > 1 ? '- ' : ''
       changed_attributes.each do |key, value|
+        former = property_to_s(self.send(key + "_was")) if self.proposal.approved?
```
afeld: I'd suggest

* Adding a test for the case where it _shouldn't_ be present
* Don't do the `if` inline, so that the body shows up separately in line coverage reports

phirefly: @afeld I believe this is the test where it isn't present. It doesn't explicitly look for the absence of the text, but it is the case where it is not displayed:
https://github.com/18F/C2/blob/master/spec/models/ncr/work_order_spec.rb#L256-L265

cmc333333:
> "As a USER, I can see the before and after values on a request modified post-approval."

Still seems super weird. Why not have the before and after values for _all_ modifications?

app/models/ncr/work_order.rb
afeld: :+1:

afeld:
Agreed.

PR #468, opened 2015-07-12
Admins can search and view

phirefly:
Adds ability for Admin and App Admin roles to search and view detail pages.
• Admins can can see any requests that match their client slug
• App admins can see all requests


app/policies/proposal_policy.rb
```diff
@@ -83,25 +83,31 @@ def initialize(user, scope)
...
-        -- observer
-        OR EXISTS (SELECT id FROM observations
-                   WHERE proposal_id = proposals.id AND user_id = :user_id)
-        SQL
-      @scope.where(where_clause, user_id: @user.id)
+      if @user.app_admin?
+        @scope.all
+      elsif @user.admin?
+        @scope.where("client_data_type ilike ?", "#{@user.client_slug}%")
+      else
```
cmc333333: Looks like you're changing the meaning of `admin?`. If so, mind changing all of the places it is called and replacing with `app_admin?`

Also, why is this a prefix match? Wouldn't an exact match be less error prone?

spec/controllers/proposals_controller_spec.rb
```diff
@@ -39,19 +39,49 @@
...
+
+      it "allows admins to view requests of same client" do
+        ENV['ADMIN_EMAILS'] = "#{user.email_address}"
+        proposal = FactoryGirl.create(:proposal, requester_id: 5555, client_data_type:"SomeCompany::SomethingApprovable")
+        user.update_attributes(client_slug: 'somecompany')
+
+        get :show, id: proposal.id
+        expect(response).not_to redirect_to(proposals_path)
+        expect(response.request.fullpath).to eq(proposal_path proposal.id)
+      end
```
cmc333333: Should probably also test that an `ADMIN_EMAILS` user cannot see _other_ client slugs.

app/policies/proposal_policy.rb
```diff
@@ -83,25 +83,31 @@ def initialize(user, scope)
...
+            WHERE proposal_id = proposals.id
+              -- TODO make visible to everyone involved
+              AND status <> 'pending'
+              AND (user_id = :user_id OR assignee_id = :user_id)
+          )
+          -- observer
+          OR EXISTS (SELECT id FROM observations
+                     WHERE proposal_id = proposals.id AND user_id = :user_id)
+          SQL
+        @scope.where(where_clause, user_id: @user.id)
```
cmc333333: Instead of doing the `OR`s in Rails, it might make more sense to do them in SQL. Right now, `admin` users can't see proposals from other clients _even if they are associated with that proposal_

spec/controllers/proposals_controller_spec.rb
phirefly: ADMIN_EMAILS (in our case, developers), are allowed to see other client slugs. 
The test is a little further down -- https://github.com/18F/C2/pull/468/files#diff-194dc4c24800956935b202e086179cc7R74

@cmc333333  Does this address your concern?

phirefly:
@cmc333333 Updated! Let me know if you have any other feedback. TY.

cmc333333:
@phirefly :+1: 'd the last commit. Good to go

PR #470, opened 2015-07-13
[#96904782] NCR email subject should only include codes

cmc333333:
It does this by reusing `self.organization.code` and applying a regex for buildings.

Also updated some specs which were testing the executing environment via `with_env_vars`, an extension of `with_env_var`.

### DIFF
--- a/spec/models/ncr/work_order_spec.rb
+++ b/spec/models/ncr/work_order_spec.rb
@@ -304,4 +304,33 @@ describe Ncr::WorkOrder do
       expect(comment.user).to eq(modifier)
     end
   end
+
+  describe "#org_id" do
+    it "pulls out the organization id when present" do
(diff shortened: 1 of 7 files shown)


spec/models/ncr/work_order_spec.rb
```diff
@@ -304,4 +304,33 @@
...
+  end
+
+  describe "#building_id" do
+    it "pulls out the building id when an identifier is present" do
+      wo = FactoryGirl.build(:ncr_work_order, building_number: "AB1234CD then some more")
+      expect(wo.building_id).to eq("AB1234CD")
+    end
+
+    it "defaults to the whole building number" do
+      wo = FactoryGirl.build(:ncr_work_order, building_number: "Another String")
```
phirefly: Great!

phirefly:
Good stuff. :shipit:

PR #471, opened 2015-07-13
[Delivers #98941214] Allow observers to modify ncr requests

cmc333333:
Just adds another OR clause to permissions checking.


phirefly:
:ship: it

PR #472, opened 2015-07-13
[#Delivers 98973752] Use default puma/newrelic config as it works out of the box

cmc333333:
We had a [configuration](https://github.com/puma/puma/issues/128#issuecomment-21050609) that worked for an older version of the newrelic gem; according to the [current docs](https://docs.newrelic.com/docs/agents/ruby-agent/troubleshooting/new-relic-puma) this is no longer required.

Tested by pushing to staging; started logging almost immediately.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,7 +17,7 @@ gem 'haml'
 gem 'html_pipeline_rails'
 gem 'jquery-rails'
 gem 'jquery-turbolinks'
-gem 'newrelic_rpm', require: false # https://github.com/Tomohiro/plate/commit/0a5a072c8d882f32a48d8bdbfa325e83719f58f8
+gem 'newrelic_rpm'
 gem 'omniauth-myusa', git: 'https://github.com/18F/omniauth-myusa.git'
 gem 'paperclip'
 gem 'peek'
--- a/config/newrelic.yml
+++ b/config/newrelic.yml
@@ -193,7 +193,6 @@ development:
   <<: *default_settings
   # Turn on communication to New Relic service in development mode
   monitor_mode: true
-  app_name: C2 (Development)
 
   # Rails Only - when running in Developer Mode, the New Relic Agent will
   # present performance information on the last 100 transactions you have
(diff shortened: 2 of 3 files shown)


afeld:
Did you try running this? Didn't work for me, despite that it supposedly should:

https://github.com/puma/puma/issues/614#issuecomment-117712457

cmc333333:
Yep, both on staging and locally:

```
vagrant@debian8:~/C2$ git status
On branch 98973752-newrelic-monitoring
nothing to commit, working directory clean
vagrant@debian8:~/C2$ rails s -b 0.0.0.0 -p8080
=> Booting Puma
=> Rails 4.2.3 application starting in development on http://0.0.0.0:8080
=> Run `rails server -h` for more startup options
=> Ctrl-C to shutdown server
Puma 2.11.3 starting...
* Min threads: 0, max threads: 16
* Environment: development
* Listening on tcp://0.0.0.0:8080
```

afeld:
Ah, I remember now – try `bundle exec puma`. This is the command [we need to have Cloud Foundry use](https://github.com/18F/C2/blob/master/Procfile), because [`rails server` ignores the Puma config file](https://github.com/puma/puma/issues/512).

afeld:
Ok, just tested both ways – the warning doesn't seem to cause any issues. Thanks!

cmc333333:
For the curious, I also don't get a warning when running puma directly:
```
vagrant@debian8:~/C2$ bundle exec puma -p8080
[30795] Puma starting in cluster mode...
[30795] * Version 2.11.3 (ruby 2.2.2-p95), codename: Intrepid Squirrel
[30795] * Min threads: 5, max threads: 5
[30795] * Environment: development
[30795] * Process workers: 1
[30795] * Preloading application
[30795] * Listening on tcp://0.0.0.0:8080
[30795] * Listening on tcp://0.0.0.0:3000
[30795] Use Ctrl-C to stop
[30795] - Worker 0 (pid: 30798) booted, phase: 0
```

PR #477, opened 2015-07-14
Fix/extra approval delegtion bug

nickbristow:


app/models/ncr/work_order.rb
```diff
@@ -110,6 +109,25 @@ def update_approvers(approver_email=nil)
...
+            system_approver = User.for_email(approver)
+            current_approver = User.for_email(current_approvers[i])
+            if !system_approver.delegates_to?(current_approver)
+              match = false
+              break
+            end
+          end
+      end
+      match
+    end
```
cmc333333: Does this work?
```ruby
def approvers_match?
  paired = self.approvers.zip(system_approvers.map { |e| User.for_email(e) })
  each_match = paired.all { |cur, sys| crr == sys || sys.delegates_to?(cur) }
  self.approvers.count == sys_approvers.length && each_match
end
```

cmc333333:
:+1:

PR #478, opened 2015-07-15
refactor approver updating

afeld:
Hopefully made the methods a bit easier to follow.


app/controllers/ncr/work_orders_controller.rb
```diff
@@ -54,12 +52,7 @@ def suggested_approver_email
...
 
     def approver_email_frozen?
-      if @model_instance
-        approval = @model_instance.approvals.first
-        approval && !approval.actionable?
-      else
-        false
-      end
+      @model_instance.try(:approver_email_frozen?)
     end
```
cmc333333: Maybe we can get rid of this method, since it's just a passthrough?

PR #479, opened 2015-07-16
[#99054450] Make approver updates less error prone

cmc333333:
This replaces a slew of methods on `Proposal` and `WorkOrder` with effectively three:
* `Proposal#kickstart_approvals` - similar to `Proposal#initialize_approvals` except that it can be called after editing or creating
* `WorkOrder#setup_approvals_and_observers` - a single entry point for creating and editing work order approvals/observers
* `Proposal#approvers=` - a single entry point for creating and editing a proposal's approvers. It takes a user list as its argument and reuses existing approvals, even if out of order.

This combines approver creation and editing into a single idempotent operation, and keeps that inside the Proposal. Analogously, it centralizes the logic around frozen approvers and emergencies when determining a Work Order's approver list.

~~I'd like to write a few more specs and call `reset_status()` but this is basically ready for review.~~ It _looks_ like a lot of new code, but we're net neutral if we exclude the specs.


cmc333333:
Updated with some more tests. Ready for review.

app/models/ncr/work_order.rb
```diff
@@ -87,73 +87,26 @@ def approver_email_frozen?
...
+      else
+        emails.unshift(approving_official_email)
+      end
+
       if self.emergency
-        emails.each {|email| self.add_observer(email) }
+        emails.each{|e| self.add_observer(e)}
         # skip state machine
-        self.proposal.update_attribute(:status, 'approved')
+        self.proposal.update(status: 'approved')
```
afeld: Is this still skipping the state machine?

app/models/ncr/work_order.rb
```diff
@@ -87,73 +87,26 @@ def approver_email_frozen?
...
       else
-        emails.each {|email| self.add_approver(email) }
-        self.proposal.initialize_approvals()
+        approvers = emails.map{|e| User.for_email(e)}
```
afeld: TODO figure out how to do this as a single database call.

```ruby
approvers = User.where(email_address: emails) # SELECT "users".* FROM "users" WHERE "users"."email_address" IN (...)
```

would retrieve the users, but wouldn't maintain the order, unfortunately.

app/models/proposal.rb
```diff
@@ -107,18 +108,46 @@ def remove_approver(email)
...
+    approvals = approver_list.each_with_index.map do |approver, idx|
+      approval = self.approval_for(approver)
+      approval ||= Approval.new(user: approver, proposal: self)
+      approval.position = idx + 1   # start with 1
+      approval
+    end
+    self.approvals = approvals
+    self.kickstart_approvals()
+    self.reset_status()
+  end
```
afeld: Feels a little dirty to override a built-in accessor, even though it throws an error since this is a `through` relation. Maybe just make a note of it?

app/models/proposal.rb
```diff
@@ -69,11 +69,12 @@ def delegate?(user)
...
-    self.approvals.find do |approval|
-      approver = approval.user
-      approver == user || approver.outgoing_delegates.exists?(assignee_id: user.id)
-    end
+    where_clause = <<-SQL
+      user_id = :user_id
+      OR user_id IN (SELECT assigner_id FROM approval_delegates WHERE assignee_id = :user_id)
+      OR user_id IN (SELECT assignee_id FROM approval_delegates WHERE assigner_id = :user_id)
+    SQL
+    self.approvals.where(where_clause, user_id: user.id).first
```
afeld: :+1:

app/models/ncr/work_order.rb
cmc333333: Indeed; it also wouldn't add new users when missing. We could write a nice wrapper and put it in `User` though

app/models/proposal.rb
cmc333333: Sure

app/models/proposal.rb
```diff
@@ -107,18 +108,46 @@ def remove_approver(email)
...
 
-  def initialize_approvals()
-    if self.linear? && self.approvals.any?
-      self.approvals.update_all(status: 'pending')
-      self.approvals.first.make_actionable!
-    elsif self.parallel?
-      self.approvals.update_all(status: 'actionable')
+  # Set the approver list, from any start state
+  def approvers=(approver_list)
+    approvals = approver_list.each_with_index.map do |approver, idx|
```
afeld: WHO SAYS YOU'RE NOT A RUBYIST

slow clap

app/models/proposal.rb
```diff
@@ -69,11 +69,12 @@ def delegate?(user)
   end
 
   def approval_for(user)
```
afeld: Maybe we should rename `existing_approval_for`, just to be explicit?

app/models/ncr/work_order.rb
cmc333333: Yep; the state machine will only be triggered with events (`xxx!`)

app/models/proposal.rb
```diff
@@ -107,18 +108,46 @@ def remove_approver(email)
...
+
+  def reset_status()
+    unless self.cancelled?   # no escape from cancelled
+      if self.all_approved?
+        self.update(status: 'approved')
+      else
+        self.update(status: 'pending')
+      end
     end
   end
```
afeld: Is there an edge case here around emergency requests? Sorry if I'm losing track of things.

app/models/proposal.rb
cmc333333: Sure thing

app/models/ncr/work_order.rb
afeld: Mind removing the comment above then?

app/models/ncr/work_order.rb
afeld: _Meh._

app/models/proposal.rb
```diff
@@ -107,18 +108,46 @@ def remove_approver(email)
...
 
   def add_observer(email)
     user = User.for_email(email)
-    self.observations.create!(user_id: user.id)
+    if self.observations.where(user: user).empty?
+      self.observations.create!(user_id: user.id)
+    end
```
afeld: Minor: I'd use `#find_or_create_by`.

app/models/proposal.rb
cmc333333: I don't _think_ so, but glad to have the second review. Emergency requests wouldn't have any approvers, so `all_approved?` would be triggered (unless the request was cancelled), keeping this in the `approved` status.

app/models/proposal.rb
```diff
@@ -190,7 +219,8 @@ def version
   def restart
     # Note that none of the state machine's history is stored
     self.api_tokens.update_all(expires_at: Time.now)
-    self.initialize_approvals()
+    self.approvals.update_all(status: 'pending')
+    self.kickstart_approvals()
```
afeld: Don't want to use `Proposal#reset_status` here?

afeld: Sorry, I mean instead of `self.approvals.update_all(status: 'pending')`.

app/models/ncr/work_order.rb
cmc333333: Sure. Should I remove them elsewhere, too, do you think?

spec/models/proposal_spec.rb
```diff
@@ -81,4 +81,144 @@
...
   end
+
+  describe '#approvers=' do
+    it 'sets initial approvers' do
+      proposal = FactoryGirl.create(:proposal)
+      approvers = [FactoryGirl.create(:user), FactoryGirl.create(:user), FactoryGirl.create(:user)]
+      proposal.approvers = approvers
+      expect(proposal.approvals.count).to be 3
+      expect(proposal.approvers).to eq approvers
+    end
```
afeld: Minor style thing: for specs that are longer than ~3 lines, I like to put whitespace between the "setup", "action", and "assertion" sections. In other words,

```ruby
proposal = FactoryGirl.create(:proposal)
approvers = [FactoryGirl.create(:user), FactoryGirl.create(:user), FactoryGirl.create(:user)]

proposal.approvers = approvers

expect(proposal.approvals.count).to be 3
expect(proposal.approvers).to eq approvers
```

spec/models/proposal_spec.rb
```diff
@@ -81,4 +81,144 @@
       expect(proposal.users).to eq([proposal.requester])
     end
   end
+
+  describe '#approvers=' do
+    it 'sets initial approvers' do
+      proposal = FactoryGirl.create(:proposal)
+      approvers = [FactoryGirl.create(:user), FactoryGirl.create(:user), FactoryGirl.create(:user)]
```
afeld: Minor: slightly more DRY using

```ruby
approvers = 3.times.map{ FactoryGirl.create(:user) }
```

app/models/proposal.rb
cmc333333: This just maintains the existing logic, which is convoluted. Previously, `initialize_approvals` meant resetting all approvals to the pending state and then triggering the first (or all, if parallel). We don't want to `reset_status` as that's already accounted for in the state transition.

app/models/ncr/work_order.rb
afeld: If no longer applicable, please!

afeld: Derp, totally misread your first reply – disregard the rest!

spec/models/proposal_spec.rb
cmc333333: much clearer

spec/models/proposal_spec.rb
cmc333333: Sounds good

app/models/proposal.rb
cmc333333: Roger; that's a new one for me

app/models/proposal.rb
cmc333333: The inline diff didn't update, but added the comment
```ruby
# This overrides the `through` relation but provides parity to the accessor
```

app/models/proposal.rb
afeld: Just checking – the change to the logic here was intentional?

cmc333333: Yep. Previously we'd only needed to look up the relationship when a delegate was trying to approve. Here we also need to look the other way: is there a delegate who has already acted?

PR #480, opened 2015-07-16
remove redundant approver_email_frozen? helper method

afeld:
Per https://github.com/18F/C2/pull/478/files#r34739175.


afeld:
Merging, since @cmc333333 did the identical thing in #479.

afeld:
Actually, jk, will go with 1701f98.

PR #481, opened 2015-07-16
[Delivers 96826060] show requester email in request list

annalee:
As as USER, I can view the email address of the requester in the request list on the My Request page so that I can easily scan who is making the request.
screen shot 2015-07-16 at 3 52 27 pm


annalee:
Switching out the class="first"s for class="next"s, per @cmc333333's suggestion. Updated screenshot:
screen shot 2015-07-16 at 4 19 48 pm

PR #487, opened 2015-07-18
Updating setup doc to include MyUSA redirect uri and DATABASE_URL

anthonygarvan:
I have setup C2 twice now, and both times it was necessary to put in the proper Redirect_URI in the MyUSA application configuration. Also, docs should reflect simplified database configuration.


phirefly:
Looks good to me. :ship:

PR #490, opened 2015-07-20
[WIP] create an ApprovalManager class

afeld:
Just a refactoring, to try and slim down the Proposal class, and move business logic around the approvers to a standalone place. While mergeable (the simplest thing that would work), this was a very quick first pass. Let me know if you think it's worth going further with.

### DIFF
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -104,30 +104,12 @@ class Proposal < ActiveRecord::Base
     approval.destroy
   end
 
-  # Set the approver list, from any start state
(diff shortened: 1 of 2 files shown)


cmc333333:
This is a good idea; it'll be particularly useful with #454. That said, how about we merge this in but pause on additional tweaks to it until #454 can be updated?

afeld:
Heh, totally forgot about that. Will review this afternoon!

cmc333333:
@afeld also oldish. Can we close?

PR #491, opened 2015-07-20
[Delivers #99336494] fix flaky specs

afeld:
For the cancel specs, my guess is that the multiple `login_as` calls for certain specs was the problem, but I'm not entirely sure. Re-ran the entire suite multiple times with

```bash
# tweaked from http://unix.stackexchange.com/a/82602
n=0
until [ $n -ge 5 ]
do
  bundle exec rspec || break
  n=$[$n+1]
done
```

* [x] Investigate failure in `./spec/decorators/proposal_decorator_spec.rb:5`

### DIFF
--- a/spec/decorators/proposal_decorator_spec.rb
+++ b/spec/decorators/proposal_decorator_spec.rb
@@ -6,7 +6,7 @@ describe ProposalDecorator do
     if array.size > 1
       loop do
         new_array = array.shuffle
-        return new_array if new_array == array
+        return new_array if new_array != array
       end
     else
       array
--- a/spec/decorators/proposal_decorator_spec.rb
+++ b/spec/decorators/proposal_decorator_spec.rb
@@ -1,12 +1,26 @@
 describe ProposalDecorator do
   let(:proposal) { FactoryGirl.build(:proposal).decorate }
 
+  # if there is more than one element, return an array with a different order than the original
+  def randomize(array)
+    if array.size > 1
+      loop do
+        new_array = array.shuffle
+        return new_array if new_array == array
+      end
+    else
+      array
+    end
+  end
+
   describe '#approvals_by_status' do
     it "orders by approved, actionable, pending" do
       # make two approvals for each status, in random order
       statuses = Approval.statuses.map(&:to_s)
       statuses = statuses.dup + statuses.clone
-      users = statuses.shuffle.map do |status|
+      statuses = randomize(statuses)
+
+      users = statuses.map do |status|
         user = FactoryGirl.create(:user)
         FactoryGirl.create(:approval, proposal: proposal, status: status, user: user)
         user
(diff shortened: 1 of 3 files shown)


afeld:
Ready for :eyes:! That attachment spec flakiness may still be outstanding, but I can look into that separately.

spec/decorators/proposal_decorator_spec.rb
```diff
@@ -1,12 +1,26 @@
 describe ProposalDecorator do
   let(:proposal) { FactoryGirl.build(:proposal).decorate }
 
+  # if there is more than one element, return an array with a different order than the original
+  def randomize(array)
+    if array.size > 1
+      loop do
+        new_array = array.shuffle
+        return new_array if new_array == array
```
cmc333333: Should this be a `!=`?

spec/decorators/proposal_decorator_spec.rb
```diff
@@ -1,12 +1,26 @@
...
+    end
+  end
+
   describe '#approvals_by_status' do
     it "orders by approved, actionable, pending" do
       # make two approvals for each status, in random order
       statuses = Approval.statuses.map(&:to_s)
       statuses = statuses.dup + statuses.clone
-      users = statuses.shuffle.map do |status|
+      statuses = randomize(statuses)
```
cmc333333: This seems kind of messy; what do you think about removing the reliance on randomness and just providing a set of cases?

spec/factories/ncr/work_orders.rb
```diff
@@ -11,6 +11,10 @@
     project_title "NCR Name"
     association :proposal, flow: 'linear'
 
+    trait :with_approver do
+      association :proposal, :with_approver, flow: 'linear'
+    end
```
cmc333333: Work Orders would never have only one approver, right?

spec/features/cancel_spec.rb
```diff
@@ -1,13 +1,9 @@
...
-  let (:user) { FactoryGirl.create(:user, id: 123456) }
-
-  before do
-    login_as(proposal.requester)
-  end
+  let!(:client_data) { FactoryGirl.create(:ncr_work_order, :with_approver) }
+  let!(:proposal) { client_data.proposal }
 
   it "shows a cancel link for the requester" do
+    login_as(proposal.requester)
```
cmc333333: Are you sure this is needed (and that the factory fix above didn't resolve)? It seems wrong to repeat this in every test

afeld: The reason is that we need different users logged in for different tests (an approver vs. the requester, etc.) I had set up `contexts` around who was logged in and was doing this in `before` blocks, but it made the test grouping somewhat confusing. I think this is a case where being DRY isn't actually helpful.

spec/decorators/proposal_decorator_spec.rb
afeld: I hear you, though I am trying to avoid potential false positives. I am actually a fan of randomness testing for things like sorting, provided I actually write the `randomize` method properly :wink:

afeld:
Updated!

PR #493, opened 2015-07-21
Remove 'rejected' state completely

cmc333333:
Split out from #454, this removes references to the rejected state from our code base.


phirefly:
Nice. :shipit:

PR #495, opened 2015-07-21
Frontpage nav css fix

annalee:
Fixes a css bug that was causing the frontpage nav to show up an unreadable dark blue instead of white. Also moves the header nav css into its own stylesheet.

Paired with @afeld.

Before the fix:
screen shot 2015-07-21 at 4 35 12 pm

After:
screen shot 2015-07-21 at 4 35 30 pm

### DIFF
--- a/app/assets/stylesheets/common/header.scss
+++ b/app/assets/stylesheets/common/header.scss
@@ -22,12 +22,6 @@ header {
   }
 }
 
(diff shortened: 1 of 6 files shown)


cmc333333:
:+1: Should resolve https://www.pivotaltracker.com/story/show/99519420, too

PR #496, opened 2015-07-21
Commenters added to observer if not tracking proposal

nickbristow:
1. Delegate leaves comment on proposal
2. System checks if the delgate is receiving emails
3. If not, adds the delegate as an observer


app/models/comment.rb
```diff
@@ -31,6 +32,15 @@ def self.attributes
...
   end
 
+  def add_user_as_observer
+    proposal = self.proposal
+    user = self.user
+    if !proposal.is_user_tracking? user
+      observer = Observation.new(user: user, proposal: proposal)
+      observer.save
+    end
+  end
```
cmc333333: Seems like this logic should belong in proposal? Something like `self.proposal.subscribe(self.user)`

Also, there's a `proposal.add_observer`

app/models/proposal.rb
```diff
@@ -206,6 +206,14 @@ def partial_approve
     end
   end
 
+  # Is the user getting emails for this proposal
+  def is_user_tracking? user
+    observers = self.observations.map {|o| o.user}
+    users = self.users - self.delegates + observers
+    users.include? user
```
cmc333333: Note that `self.observers` obviates the need for line 211. Given the math in 212, seems like spelling out exactly who you want might make more sense than trying to remove one group and add a potentially-intersecting group back in.

There's a nuance here; comments are only sent to the requester, observers, and approvers _who have been notified already_ (see `Comment#listeners`). For example, if a tier 2 approver added a comment, they wouldn't be "subscribed" until after tier 1 approved.

cmc333333:
A very related idea is that these observers might need to be removed when switching from delegate to approver (further that the mailbox might need to become an observer at that point). Do you think either are required for this?

app/models/proposal.rb
```diff
@@ -206,6 +206,26 @@ def partial_approve
...
 
+  # Is the user getting emails for this proposal
+  def is_user_tracking? user
+    users = [self.requester]
+    users += self.approvers + self.observers
+    users.include? user
+  end
+
+  def subscribe user
+    if !self.is_user_tracking? user
```
cmc333333: Do we need this if the emails are getting deduplicated later? If not, less logic to carry around.

Note that you'll want to merge in master for https://github.com/18F/C2/blob/6ce10e31faae61e5a428763ccaa3d9f48b1c4827/app/models/proposal.rb#L146

PR #497, opened 2015-07-21
Add materialized "approved_at" field to approvals

cmc333333:
Split from #454, this creates an `approved_at` field for approvals. The field is set when the approval enters the approved state. Also backfills existing data.

Also simplifies options on the approval factory (rather than setting non-useful defaults).

### DIFF
--- a/spec/models/approval_spec.rb
+++ b/spec/models/approval_spec.rb
@@ -35,6 +35,8 @@ describe Approval do
       approval.make_actionable!
       approval.approve!
       expect(approval.approved_at).not_to be_nil
+      approval.reload
+      expect(approval.approved_at).not_to be_nil
     end
   end
 end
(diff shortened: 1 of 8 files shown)


spec/models/approval_spec.rb
```diff
@@ -20,4 +20,21 @@
...
+
+    it 'is nil when actionable' do
+      approval.make_actionable!
+      expect(approval.approved_at).to be_nil
+    end
+
+    it 'is set when approved' do
+      approval.make_actionable!
+      approval.approve!
+      expect(approval.approved_at).not_to be_nil
```
afeld: Might want to check that it's actually persisted.

afeld:
:shipit: 

(needs merge from `master`)

spec/models/approval_spec.rb
cmc333333: Can do

cmc333333:
Updated w/ master and persistence test

PR #498, opened 2015-07-22
[Delivers #99321148] Show requester in request list for client-specific users

annalee:
Fixes a bug where the requester email address is hidden in the "my requests" table for users with a client slug set.


cmc333333:
:+1:

PR #500, opened 2015-07-22
[WIP] [Delivers #96622322] add zero-downtime deployment script

afeld:
* [x] Finish testing the app switchover
* [x] Call from the `release` script

### DIFF
--- a/script/deploy
+++ b/script/deploy
@@ -19,20 +19,28 @@ GREEN="${BLUE}-B"
 MANIFEST=$(mktemp -t "${BLUE}_manifest")
 cf create-app-manifest $BLUE -p $MANIFEST
 
-DOMAIN=$(cat $MANIFEST | grep domain: | awk '{print $2}')
+# try/catch
+# http://stackoverflow.com/a/22010339/358804
+{
+  DOMAIN=$(cat $MANIFEST | grep domain: | awk '{print $2}')
 
-# create the GREEN application
-cf push $GREEN -f $MANIFEST -n $GREEN
-# ensure it starts
-curl --fail -I "https://${GREEN}.${DOMAIN}"
+  # create the GREEN application
+  cf push $GREEN -f $MANIFEST -n $GREEN
+  # ensure it starts
+  curl --fail -I "https://${GREEN}.${DOMAIN}"
 
-# add the "green" application to the "blue" load balancer
-cf map-route $GREEN $DOMAIN -n $BLUE
+  # add the "green" application to the "blue" load balancer
+  cf map-route $GREEN $DOMAIN -n $BLUE
 
-# cleanup
-# TODO consider 'stop'-ing the BLUE instead of deleting it, so that depedencies are cached for next time
-cf delete $BLUE -f
-cf rename $GREEN $BLUE
-cf delete-route $DOMAIN -n $GREEN -f
+  # cleanup
+  # TODO consider 'stop'-ing the BLUE instead of deleting it, so that depedencies are cached for next time
+  cf delete $BLUE -f
+  cf rename $GREEN $BLUE
+  cf delete-route $DOMAIN -n $GREEN -f
 
-echo "DONE"
+  echo "DONE"
+} || {
+  # we don't want to keep the sensitive information around
+  rm $MANIFEST
+  exit 1
+}
--- a/script/deploy
+++ b/script/deploy
@@ -10,37 +10,48 @@
 #   https://github.com/dlapiduz/step-cloud-foundry-deploy/blob/master/run.sh
 
 set -e
+set -o pipefail
 set -x
 
 BLUE=$1
 GREEN="${BLUE}-B"
 
+
+finally ()
+{
+  # we don't want to keep the sensitive information around
+  rm $MANIFEST
+}
+
+on_fail () {
+  finally
+  echo "DEPLOY FAILED - you may need to check 'cf apps' and 'cf routes' and do manual cleanup"
+}
+
+
 # pull the up-to-date manifest from the BLUE (existing) application
 MANIFEST=$(mktemp -t "${BLUE}_manifest")
 cf create-app-manifest $BLUE -p $MANIFEST
 
-# try/catch
-# http://stackoverflow.com/a/22010339/358804
-{
-  DOMAIN=$(cat $MANIFEST | grep domain: | awk '{print $2}')
+# set up try/catch
+# http://stackoverflow.com/a/185900/358804
+trap on_fail ERR
 
-  # create the GREEN application
-  cf push $GREEN -f $MANIFEST -n $GREEN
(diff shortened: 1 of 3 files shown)


cmc333333:
Very nice! Logic makes sense to me. 

* It might make sense to not proceed if the curl fails
* Migrations might throw a wrench in this system, as they'd affect the shared db. I'm not sure there's a clean solution there, but worth thinking about at least

afeld:
Updated! @dlapiduz @ozzyjohnson Mind taking a look?

> It might make sense to not proceed if the curl fails

`set -e` stops the script if _any_ commands fail.

> Migrations might throw a wrench in this system

True – would love to think about a general solution for this. Made a note in the docs.

cmc333333:
> set -e stops the script if any commands fail.

Very cool. It doesn't look like 400s (possibly 500s) result in error codes for curl, though?
```bash
(parser)vagrant@debian8:~/C2$ curl notaurl
curl: (6) Could not resolve host: notaurl
(parser)vagrant@debian8:~/C2$ echo $?
6
(parser)vagrant@debian8:~/C2$ curl http://dfdsfdsd.18f.gov/
404 Not Found: Requested route ('dfdsfdsd.18f.gov') does not exist.
(parser)vagrant@debian8:~/C2$ echo $?
0
```

script/deploy
```diff
@@ -0,0 +1,38 @@
...
+#
+# Based on
+#   http://www.cloudfoundry.rocks/blue-green-deployment-with-cloudfoundry/
+#   http://docs.pivotal.io/pivotalcf/devguide/deploy-apps/blue-green.html
+#   https://github.com/dlapiduz/step-cloud-foundry-deploy/blob/master/run.sh
+
+set -e
+set -x
+
+PRIMARY=$1
```
khandelwal: I wouldn't call it primary, secondary. That can imply that one is better than the other. We're replacing the primary with the secondary sounds like we're replacing it with something inferior.

dlapiduz: I was going to agree at this point but since the point of the script is to delete the primary app and make the secondary primary then it kind of makes sense.

afeld: Changed to blue/green to make it consistent with the documentation.

afeld:
Updated!

> It doesn't look like 400s (possibly 500s) result in error codes for curl

Derp, good catch – I had only tried the former.

dlapiduz:
Only one other thing, can you delete the manifest file when finished? That way we don't have passwords lurking around...

cmc333333:
@afeld I'm good merging unless you want to explictly delete the manifest file

afeld:
Updated!

Pretty proud of this script – will definitely link to it from https://docs.18f.gov/apps/continuous-deployment/, and ideally think of a better way to distribute it than "copy-and-paste this into your project".

script/deploy
```diff
@@ -0,0 +1,46 @@
...
+
+# try/catch
+# http://stackoverflow.com/a/22010339/358804
+{
+  DOMAIN=$(cat $MANIFEST | grep domain: | awk '{print $2}')
+
+  # create the GREEN application
+  cf push $GREEN -f $MANIFEST -n $GREEN
+  # ensure it starts
+  curl --fail -I "https://${GREEN}.${DOMAIN}"
```
dhcole: Just tried this on a deploy that failed here, but the script kept executing:

```
0 of 1 instances running, 1 failing
FAILED
Start unsuccessful

TIP: use 'cf logs federalist-B --recent' for more information
+ curl --fail -I https://federalist-B.18f.gov
curl: (22) The requested URL returned error: 404 Not Found
+ cf map-route federalist-B 18f.gov -n federalist
Creating route federalist.18f.gov for org federalist / space prod as david.cole@gsa.gov...
OK
Route federalist.18f.gov already exists
Adding route federalist.18f.gov to app federalist-B in org federalist / space prod as david.cole@gsa.gov...
OK
+ cf delete federalist -f
Deleting app federalist in org federalist / space prod as david.cole@gsa.gov...
OK
+ cf rename federalist-B federalist
Renaming app federalist-B to federalist in org federalist / space prod as david.cole@gsa.gov...
OK
+ cf delete-route 18f.gov -n federalist-B -f
Deleting route federalist-B.18f.gov...
OK
+ echo DONE
DONE

```

afeld: Yeah, apparently `set -e` doesn't apply to functions because _why would it_. 

rage

Gave it another pass, using `trap`. Mind trying again?

afeld: P.S. I tested fake failure scenarios and the real success scenario, but would love to verify that a real failure behaves as expected.

script/deploy
```diff
@@ -0,0 +1,57 @@
...
+# create the GREEN application
+cf push $GREEN -f $MANIFEST -n $GREEN
+# ensure it starts
+curl --fail -I "https://${GREEN}.${DOMAIN}"
+
+# add the "green" application to the "blue" load balancer
+cf map-route $GREEN $DOMAIN -n $BLUE
+
+# cleanup
+# TODO consider 'stop'-ing the BLUE instead of deleting it, so that depedencies are cached for next time
```
dlapiduz: That is exactly why you do blue green... this is not really blue/green but I like that apps are rebuilt

afeld: So you're saying you prefer it this way?

afeld:
Just used this script to deploy `c2-prod`, and it didn't map the `cap.18f.gov` route. Yeesh.

```
$ ./script/deploy c2-prod
...
"DONE"
$ curl --fail -I https://cap.18f.gov
curl: (22) The requested URL returned error: 404 Not Found
$ cf routes
Getting routes as aidan.feldman@gsa.gov ...

space     host         domain    apps
general   c2-staging   18f.gov   c2-staging
general   c2-dev       18f.gov   c2-dev
general   cap          18f.gov
general   c2-prod      18f.gov   c2-prod
$ cf map-route c2-prod 18f.gov -n cap
Creating route cap.18f.gov for org cap / space general as aidan.feldman@gsa.gov...
OK
Route cap.18f.gov already exists
Adding route cap.18f.gov to app c2-prod in org cap / space general as aidan.feldman@gsa.gov...
OK
$ curl --fail -I https://cap.18f.gov
HTTP/1.1 200 OK
```

* [ ] Fix route re-mapping

afeld:
I _think_ I've got it this time (tested by adding a dummy custom domain to development), but am going to deploy to production with the script at night just to be sure before we merge.

afeld:
:boom: Worked swimmingly!

dhcole:
Found that if you are not auto-detecting the buildpack, you need to have `cf cli` version 6.12.1 or higher, because of this bug: https://www.pivotaltracker.com/n/projects/892938/stories/96041780

afeld:
Tricksy. Will make a note about it when I make a page on https://docs.18f.gov.

PR #502, opened 2015-07-23
Hide modification row when proposal is cancelled

nickbristow:


cmc333333:
Mind adding a test?

nickbristow:
@cmc333333 not at all

phirefly:
:shipit:

PR #503, opened 2015-07-24
[#98024644] Approvals::Individual and a single approval chain entry point

cmc333333:
Split out from #454 

This doesn't change too much of `app` - it migrates all approvals to the `Approvals::Individual` subclass and removes all but one entry point for setting the approval chain.

Most of the code updates are to the specs, which were heavy users of `add_approver`

### DIFF
--- a/app/models/approvals/individual.rb
+++ b/app/models/approvals/individual.rb
@@ -14,9 +14,5 @@ module Approvals
       end
       state :approved
     end
-
-    def self.for_email(email)
-      Approvals::Individual.new(user: User.for_email(email))
-    end
   end
 end
--- a/app/models/approvals/individual.rb
+++ b/app/models/approvals/individual.rb
@@ -1,6 +1,8 @@
 # Represents a single user's ability to approve, the "leaves" in an approval chain
 module Approvals
   class Individual < Approval
+    validates :user, presence: true
+
     workflow do
       on_transition { self.touch } # https://github.com/geekq/workflow/issues/96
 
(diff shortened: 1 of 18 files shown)


app/models/approvals/individual.rb
```diff
@@ -0,0 +1,22 @@
...
+      end
+      state :actionable do
+        event :approve, transitions_to: :approved
+      end
+      state :approved
+    end
+
+    def self.for_email(email)
+      Approvals::Individual.new(user: User.for_email(email))
+    end
```
afeld: `User#for_email` will create the record if it doesn't exist yet...thinking we want to be consistent about the behavior for both, or at least make a note that they are different. Doesn't look like this method is being used, though?

spec/features/display_status_spec.rb
```diff
@@ -1,6 +1,8 @@
 describe "Display status text" do
   let(:proposal) { FactoryGirl.create(:proposal, :with_parallel_approvers) }
   before do
+    proposal.approvers.first.update(first_name: "Uniquely", last_name: "Named")
+    proposal.approvers.second.update(first_name: "Onlyof", last_name: "Itskind")
```
afeld: Is this necessary?

afeld:
Just a couple small things – otherwise, looks good!

app/models/approvals/individual.rb
cmc333333: Easy enough to get rid of now and bring it back under a different name if it's needed later

spec/features/display_status_spec.rb
cmc333333: Yes (the "lino" name wasn't too unique); I'll update the relevant tests to make it more apparent how this is used

cmc333333:
Updated

PR #504, opened 2015-07-24
Cancelled request no longer show in completed

nickbristow:
Request that are cancelled no longer show in the completed list.


cmc333333:
Might be worthwhile to remove "closed" altogether, since it seems its definition has changed.

PR #506, opened 2015-07-27
[#96621098] Declarative model for list views

cmc333333:
Represent individual html tables (/listing/list views/etc.) of query sets as a `Listing` object. This will allow us to centralize filtering, sorting, paging, etc. of these types of listings into a single location. Configuration of these listing objects is defined declaratively in a configuration file (which can be overwritten with client-specific instances).

Removes the various client-specific proposal list templates as they are no longer needed. There was a bit of variation in their variables, but it turns out that those variables (`limit`, `role`, etc.) were not used or could be easily replaced. 

Needs some tests around the `Listing` object, but it's not doing anything super interesting yet, so we can wait until it's sorting/etc. if this looks good now.


afeld:
Is there any change to the UI? If so, mind including a screenshot?

lib/listing.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    # Walk the object tree to get the field in question
+    def value_in(row)
+      field_components = @field.split(".")
+      field_components.inject(row) do |object, field|
+        object.try(field)
+      end
+    end
+  end
```
afeld: Mind moving this to a separate file?

lib/listing.rb
```diff
@@ -0,0 +1,42 @@
+class Listing
```
afeld: `Listing` sounds equivalent to `row` to me...how about `Table` or `ListingSet` or something?

afeld:
This is great. Just a couple minor suggestions.

lib/listing.rb
cmc333333: Can do. They're pretty tightly coupled, though, so I'll make a module.

cmc333333:
Updated

> Is there any change to the UI? If so, mind including a screenshot?

UI should be almost identical. The known exception is that all columns are marked with the same width here (previously, the "Request" column was slightly larger than the others). If it's problematic, we can code it up, but it doesn't seem super pressing at the moment.

Before:
screen shot 2015-07-27 at 5 50 50 pm

After:
screen shot 2015-07-27 at 5 47 14 pm

PR #509, opened 2015-07-30
[WIP] Display relative proposal times + other changes (probably)

istanfield09:


app/helpers/proposals_helper.rb
```diff
@@ -11,4 +11,13 @@ def datespan_header(start_date, end_date)
...
   end
+
+  def truncate_link(link)
+    if (link.size > 70)
+      truncated_link = link[0...69] << "..."
+      return truncated_link
+    else
+      return link
+    end
+  end
```
afeld: Rails actually has a built-in [`truncate`](http://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-truncate) method...does the same thing, but mind using that instead?

app/views/default/_proposal_properties.html.erb
```diff
@@ -10,7 +10,11 @@
         <strong><%= name %>
       </td>
       <td>
-        <%= property_to_s(value) %>
+        <% if name == "Link to product" %>
+          <%= link_to truncate_link(value), value %>
+        <% else %>
+          <%= property_to_s(value) %>
+        <% end %>
```
afeld: Wondering if there is another way to go about this...I don't think we want to hard-code that one property name in this partial, but I could see the use for truncating certain fields.

app/helpers/proposals_helper.rb
istanfield09: Thanks for the advice! I will amend my code to reflect that. Slightly embarrassed : X

afeld:
Looks like the relative time piece (https://www.pivotaltracker.com/story/show/93639992) is pretty distinct from the truncation piece...would you mind if we split each into its own pull request, just to keep them focused?

Thanks for submitting this – excited to have you contributing!

istanfield09:
I'll go ahead and roll back my PR several commits, and split truncation and relative times into two separate PRs. Thanks for the critiques! : )

PR #510, opened 2015-07-30
[#98024644] Parallel/Serial approvals, partial hierarchy

cmc333333:
This patch adds the `Parallel` & `Serial` subclasses. It moves user-centric logic into the `Individual` subclass and creates an association for grabbing only the `Individual` approvals (which is a very common need in our tests).

This includes a migration which adds the `parent` field (allowing approvals to be nested) and the `min_children_required` field (which is used by `Parallel` to simulate ORs). Tests have been added to verify that complex approval chains can work.

Importantly, this does _not_ affect `WorkOrder`s or `Procurement`s, and doesn't implement the final protocol for communicating with the proposal (it continues to send `partial_approve!` and forces the proposal to check its`flow` field).

@afeld you mentioned pulling out the persistence layer as a way to test everything. I think it'd be more trouble than it's worth to implement, given the current inheritance hierarchies and extra book-keeping. I hope the changes in the next PR (which resolve the issues in the previous paragraph) should show it all working without passing through that intermediary step.


app/models/concerns/proposal_delegate.rb
```diff
@@ -4,7 +4,8 @@ module ProposalDelegate
   included do
     has_one :proposal, as: :client_data
     has_many :approvals, through: :proposal
-    has_many :approvers, through: :approvals, source: :user
+    has_many :approvals_individual, ->{ individual }, class_name: 'Approvals::Individual', through: :proposal
```
afeld: I think this would read better as `individual_approvals`.

app/models/api_token.rb
```diff
@@ -3,7 +3,7 @@ class ApiToken < ActiveRecord::Base
 
   before_create :generate_token
 
-  belongs_to :approval
+  belongs_to :approval, class_name: 'Approvals::Individual'
```
afeld: Out of curiosity, why is this needed? Does it enforce what class can be assigned?

cmc333333: It's needed to resolve the `has_one through:` relationship of `:user`. I bet there's a better way?

cmc333333:
Updated with master and s/approvals_indidivual/individual_approvals

app/models/api_token.rb
afeld: Meh, not a big deal.

app/models/approvals/parallel.rb
```diff
@@ -0,0 +1,56 @@
...
+        end
+      end
+    end
+
+    def on_actionable_entry(old_state, event)
+      if self.child_approvals.exists?
+        self.child_approvals.each(&:initialize!)
+      else
+        self.force_approve!
+      end
```
afeld: Would it be safe to assume/enforce that non-Individual approvals should always have more than one child?

app/models/approvals/serial.rb
```diff
@@ -0,0 +1,54 @@
+# A node in an approval chain that requires its children be approved in order
+module Approvals
+  class Serial < Approval
+    workflow do
+      on_transition { self.touch } # https://github.com/geekq/workflow/issues/96
```
afeld: Mind adding "sets the `updated_at`" in the comment?

afeld:
Split out from #454.

> the min_children_required field (which is used by Parallel to simulate ORs)

Sorry if I didn't bring it up before, but this is the only part of the original PR that didn't quite feel right to me. Doesn't seem like it will be useful to us yet anyway, correct? Mind taking it out, and we can talk through how to tackle that problem separately?

---

Not a blocker, but would be great to add some info about the structure to [`doc/overview.md`](https://github.com/18F/C2/blob/master/doc/overview.md).

app/models/approvals/parallel.rb
cmc333333: I don't think that restriction gains us anything. Why not tackle the general case?

app/models/approvals/serial.rb
cmc333333: Sure

cmc333333:
> Sorry if I didn't bring it up before, but this is the only part of the original PR that didn't quite feel right to me. Doesn't seem like it will be useful to us yet anyway, correct? Mind taking it out, and we can talk through how to tackle that problem separately?

I'd feel more comfortable removing this functionality if I knew more what you didn't like about it. As it stands, it implements a use case we know we will need.

> Not a blocker, but would be great to add some info about the structure to doc/overview.md.

Let's wait until the next PR is in. As it stands, any documentation would be either inaccurate or need to be rewritten then.

app/models/approvals/parallel.rb
afeld: `self.child_approvals.exists?` (maybe `any?` would be more appropriate?) should theoretically always be true, right? Was just thinking there shouldn't be a need for that check. Not a big deal.

afeld:
Re: the `min_children_required`, it sounds like the implementation would be useful for a [two-man rule](https://en.wikipedia.org/wiki/Two-man_rule). Do we have a use case where `> 1` but `< all` child approvals would be required? If not, seems a bit indirect of a way to achieve the one vs. all specification. I suppose it doesn't involve that much code, so we could change it later if needed.

app/models/approvals/parallel.rb
cmc333333: I hear what you're saying, and agree that it'd be suspicious if one of these didn't have children. I think there's the potential for it to arise without error (e.g. as an intermediate step when building the hierarchy) however, so accounting for the general case is a safer bet.

cmc333333:
> Re: the min_children_required, it sounds like the implementation would be useful for a two-man rule. 

Or three or four or ...

> Do we have a use case where > 1 but < all child approvals would be required? 

Not specifically, but again, why not just solve the general case? If we can require one out of two, why not three out of four?

> If not, seems a bit indirect of a way to achieve the one vs. all specification.

It's indirect if you think about the problem as one vs. all, sure. If you think of it as a ratio instead, it's a very direct solution.

afeld:
Fair enough.

cmc333333:
Updated to use `any?` instead of `exists?` and to mention `updated_at`

PR #513, opened 2015-07-31
Updated dispatcher to notify observers for work order changes

nickbristow:


cmc333333:
I think this makes sense and I'd be willing to merge, but wanted to make to comments/suggestions:

1. Would renaming the approval variable to `subscription` in the template be any cleaner? You'd pass in either an approval or an observation. This might make the templates more complicated, though, if there's conditional logic.
2. At a higher level, it seems like the "determine which of these observers is not really an observer" logic could be reduced to a simpler question. Given a proposal and a user, determine that user's "role" for the proposal. You could imagine something like `user.role_on(proposal).subscriber?` or `user.role_on(proposal).active_approver?`. This overlaps some with @annalee's recent thinking.

As mentioned, neither's a blocker, but while we're working on this, they might be valuable to ponder? LMK how you'd like to proceed.

cmc333333:
@nickbristow 'll be occupied this week, so I'll merge and leave those questions in the air :)

nickbristow:
@cmc333333 oh ok that works. thanks CM

PR #514, opened 2015-07-31
[#96621098] Sorting for tabular data

cmc333333:
This extends tabular data views to allow the user to sort by column. To implement that, the tabular data needed to be able to map the visible columns into database columns, so there's now a connection to Arel columns.

Presently, this allows for sorting by columns on the immediate table and columns in any tables that can be attached via a simple `joins()`. That excludes `client_data` at the moment (as it is polymorphic). That said, hopefully the configuration file and calling syntax in the proposal controller show the simplicity of use.

Looks like:
screen shot 2015-07-31 at 6 24 23 pm
screen shot 2015-07-31 at 6 24 36 pm
screen shot 2015-07-31 at 6 24 51 pm

Next steps are to make it work for the `total_price` field and determine the best way to "sort" the status field, which can contain complex content.


cmc333333:
I've been working on a cleaner implementation that relies less on arel. I'll close this and open that PR when it gets to a good stage

PR #515, opened 2015-08-01
validate email addresses

afeld:


cmc333333:
Clean! Had to fix a failing test, sorry about that.

PR #517, opened 2015-08-03
Feature/add role on method on user

nickbristow:


lib/role.rb
```diff
@@ -0,0 +1,38 @@
+class Role 
+  attr_accessor :role_type, :proposal, :user
+  def initialize(user,proposal)
+    self.user = user
+    self.proposal = proposal
```
cmc333333: I'm not sure it's helpful to carry around the user or proposal after the initialization?

lib/role.rb
```diff
@@ -0,0 +1,38 @@
...
+    if self.user == self.proposal.requester
+      role = 'requester'
+    elsif self.proposal.is_active_approver?(self.user)
+      role = 'active approver'
+    elsif self.proposal.observers.include?(self.user)
+      role = 'observer'
+    elsif self.proposal.approvers.include?(self.user)
+      role = 'approver'
+    end
+    self.role_type = role
```
cmc333333: Probably want to use symbols rather than strings.

More importantly, can't a single user play multiple roles on a proposal? If so, maybe you want to store multiple roles and encourage a capabilities model. Right now, we're asking "Are you an approver?" Maybe we should ask, instead "Can you approve?" "Are you subscribed?". This gets close to Pundit, so I'm not quite sure how to separate the two.

Be sure to reach out to hash this through with @annalee who's been thinking about this more.

cmc333333:
@nickbristow I'm okay merging this, but it sounds like it'll be replaced? LMK how you want to proceed

cmc333333:
I'm going to merge early, as this will be useful for https://www.pivotaltracker.com/story/show/99523048 . Thanks @nickbristow !

PR #518, opened 2015-08-04
[Delivers #96621098] System for sorting tabular data

cmc333333:
This extends tabular data views to allow the user to sort by column. To allow that, we need to specify database columns (or expressions) on which to sort. The beauty here is that the same columns can be used for filtering, when the time comes.

The configuration syntax allows arbitrary tables to be joined in and columns to be queried. For example, we can order on the `total_price` field across tables due to the use of `COALESCE` and we can order by `proposals.status` using a switch-case expression. The syntax also allows for "virtual" columns, columns which only exist in ruby land and hence cannot be sorted.

Looks like (from the `/proposals/query` page):
screen shot 2015-08-04 at 5 47 47 pm
screen shot 2015-08-04 at 5 48 16 pm


lib/tabular_data/container.rb
```diff
@@ -2,24 +2,89 @@ module TabularData
...
+      @name = name
+      self.init_query(config[:engine].constantize, config.fetch(:joins, []))
+      self.init_columns(config.fetch(:column_configs, {}), config.fetch(:columns, {}))
+      self.set_sort(config[:sort])
+    end
+
+    def alter_query
+      @query = yield(@query)
+      self
     end
```
afeld: :zap:

afeld:
This is awesome. Nice work!

PR #519, opened 2015-08-05
Relative proposal times

istanfield09:
In reference to https://www.pivotaltracker.com/n/projects/1149728/stories/93639992.

### DIFF
--- a/app/helpers/value_helper.rb
+++ b/app/helpers/value_helper.rb
@@ -1,8 +1,14 @@
 module ValueHelper
   include ActionView::Helpers::NumberHelper
+
   def date_with_tooltip(time)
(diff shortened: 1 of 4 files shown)


cmc333333:
@istanfield09 thanks for the PR! What do you think about adding a parameter (e.g. `ago`) to the `date_with_tooltip` function instead? By explicitly testing for "ProposalsController", you're coupling previously independent modules.

afeld:
This looks good! Any interest in writing tests for the `ago` scenario and without? Could be very similar to this one:

https://github.com/18F/C2/blob/master/spec/helpers/value_helper_spec.rb

afeld:
Failing test was due to a separate issue that was fixed on master...merged manually.

PR #520, opened 2015-08-07
use cf-blue-green for zero-downtime deployment

afeld:
Moved our script to a separate package.


afeld:
Verified with successful production deploy. Happy to do a follow-up PR if anything needs tweaking!

PR #521, opened 2015-08-07
add manifest file

afeld:
Moves the non-sensitive Cloud Foundry configuration into a config file, meaning that it's now version controlled.


afeld:
Validated this works, and need it for some other deploy stuff, so going to merge now. Happy to address any feedback in a follow-up PR!

PR #522, opened 2015-08-07
Feature/email removed approvers

nickbristow:
So im not a 100% sure the test is good enough. 

Not exactly sure the best way to test this.


app/models/ncr/work_order.rb
```diff
@@ -104,7 +104,11 @@ def setup_approvals_and_observers(approving_official_email)
         self.proposal.update(status: 'approved')
       else
         approvers = emails.map{|e| User.for_email(e)}
+        original_approver_groups = self.proposal.approvers.each.map{|approver| 
+          {user: approver, active: self.proposal.is_active_approver?(approver)}
+        }
         self.proposal.approvers = approvers
+        email_removed_approvers(original_approver_groups)
```
afeld: I think this could be simplified to

```ruby
removed_approvers = self.proposal.approvers - approvers
Dispatcher.on_approver_removal(self.proposal, removed_approvers)
```

lib/ncr_dispatcher.rb
```diff
@@ -34,4 +34,10 @@ def on_proposal_update(proposal)
       end
     }
   end
+
+  def on_approver_removal(proposal, approvers)
+    approvers.each{|approver|
+      CommunicartMailer.notification_for_subscriber(approver.email_address,proposal,"removed").deliver_now
+    }
+  end
```
afeld: Why not have this in the `Dispatcher`? Doesn't seem like NCR-specific functionality.

afeld:
> im not a 100% sure the test is good enough

I would check that given a Proposal with approvers, when a new overlapping set of approvers is assigned, the removal email is sent to the right people. Happy to pair on it if you like.

nickbristow:
@afeld so your comment above about making it simpler. That would work, but then users who haven't receive email will receive a notification that they have been removed.

afeld:
I don't follow... The idea is that you find the difference between the sets of the old and new lists of approvers, then send to those people.

nickbristow:
@afeld yes, but those users don't know they have been added to the work order. They have yet to be notified. We can still notify them, but they haven't received the first email yet. I thought that was the approach we were going with. @phirefly ?

afeld:
In that case, it could be something like

```ruby
existing_approvers = self.proposal.approvers
removed_approvers = existing_approvers - approvers
added_approvers = approvers - existing_approvers

self.proposal.approvers = approvers

Dispatcher.on_approver_removal(self.proposal, removed_approvers)
Dispatcher.on_approvers_added(self.proposal, added_approvers) # or whatever
```

Sorry if I'm misunderstanding something...

phirefly:
@nickbristow We do not want to send a 'you have been removed' email if they are not aware of the request (i.e. they have not received any emails about the request). That is correct.

app/models/ncr/work_order.rb
```diff
@@ -103,7 +103,9 @@ def setup_approvals_and_observers(approving_official_email)
         self.proposal.update(status: 'approved')
       else
         approvers = emails.map{|e| User.for_email(e)}
+        removed_approvers_to_notify = self.proposal.approvals.non_pending.map(&:user) - approvers
```
afeld: Not a blocker, but would probably be useful to have a `scope :non_pending_approvers` on the `Approval` model.

afeld: At the very least, maybe use a variable with that name, just for clarity?

afeld:
Looks good! Just the one minor note – mind fixing the Code Climate error, as well?

PR #523, opened 2015-08-07
[Delivers #98024644] Use approval chains

cmc333333:
This completes the first iteration of approval chains. Notably,
* Approvals are created by setting `proposal.root_approval = xxx`
* A wrapper was written for proposals with a single approver: `proposal.approver = xxx`
* Both `WorkOrder`s and `Procurement`s use this new interface. `WorkOrder`s now include a root, `Serial` approval
* "partial_approve" has been removed in favor of the root approval updating the proposal's state
* Resolved an issue in which old `Approval`s were not being deleted when their proposal was set to nil
* Data migration for `WorkOrder`s (`Procurement`s do not need it) to add a `Serial`

Tested relatively thoroughly locally, but it'd be worth while to test the branch in a VM before merging.


app/models/approval.rb
```diff
@@ -27,11 +27,15 @@ def notify_parent_approved
...
-      self.proposal.partial_approve!
+      self.proposal.approve!
     end
   end
 
   def children_approved?
     self.child_approvals.where.not(status: "approved").empty?
   end
+
+  def preorder_list
```
afeld: Not sure if I understand the method name...does this mean that it shows the approvals in order, whether or not that order is relevant?

app/models/proposal.rb
```diff
@@ -93,36 +99,38 @@ def users
...
-    if self.parallel?
-      pending.update_all(status: 'actionable')
-    elsif self.linear? && actionable.empty? && pending.any?
-      pending.first.initialize!
+  # convenience wrapper for setting a single approver
+  def approver=(approver)
+    # Don't recreate the approval
+    existing = self.existing_approval_for(approver)
+    if existing.nil?
+      self.root_approval = Approvals::Individual.new(user: approver)
```
afeld: Doesn't this need to clean up the existing `root_approval`, if there is one?

app/models/approval.rb
cmc333333: Pre-order tree traversal. I'll add a comment.

app/models/proposal.rb
cmc333333: Good call; it should, yep.

app/models/ncr/work_order.rb
```diff
@@ -102,8 +102,17 @@ def setup_approvals_and_observers(approving_official_email)
...
+          user = User.for_email(email)
+          # Reuse existing approvals, if present
+          if existing = self.proposal.existing_approval_for(user)
+            existing
+          else
+            Approvals::Individual.new(user: user)
+          end
+        end
+
+        self.proposal.root_approval = Approvals::Serial.new(child_approvals: individuals)
```
afeld: Do we need to worry about a `root_approval` already existing here, or can we safely assume this will only be called on the initial setup?

app/models/proposal.rb
```diff
@@ -93,36 +99,38 @@ def users
...
+    # position may be out of whack, so we reset it
+    approval_list.each_with_index do |approval, idx|
+      approval.set_list_position(idx + 1)   # start with 1
+    end
+
+    old_approvals.each do |old|
+      unless approval_list.include?(old)
+        old.destroy()
+      end
     end
```
afeld: @nickbristow I think this method can be used for #522.

app/models/proposal.rb
cmc333333: Existing approvals get destroyed in `#root_approval=`, so this is accounted for already.

afeld: Oh nevermind, looks like that might be taken care of in `#root_approval=`...? Anyway, we should test for it.

afeld: JINX

app/models/ncr/work_order.rb
cmc333333: This is most certainly called more often than just the initial setup, but the existing `root_approval` will get removed in `#root_approval=`

app/models/proposal.rb
```diff
@@ -205,25 +214,10 @@ def restart
     # Note that none of the state machine's history is stored
     self.api_tokens.update_all(expires_at: Time.now)
     self.approvals.update_all(status: 'pending')
-    self.kickstart_approvals()
+    self.root_approval.initialize! if self.root_approval
```
afeld: Minor: mind splitting to multiple lines, for code coverage purposes?

db/migrate/20150807204802_delete_approvals_without_proposals.rb
```diff
@@ -0,0 +1,7 @@
+class DeleteApprovalsWithoutProposals < ActiveRecord::Migration
+  def up
+    execute <<-SQL
+      DELETE FROM approvals WHERE proposal_id is null;
+    SQL
+  end
+end
```
afeld: Wonder if this makes more sense as a general cleanup Rake task?

doc/overview.md
```diff
@@ -4,14 +4,17 @@ C2 is, at its core, a state machine wrapped in email notifications. The system c
...
-    * Once the request is submitted, all approvers receive a notification.
-* Linear (a.k.a serial)
-    * Once the request is submitted, it goes to the first approver. Iff they approve, it goes to the next, and so forth.
+* `Individual` - representing a single user's approving decision
+* `Parallel` - whose children can approve in any order
+* `Serial` - whose children must approve in sequence
+
+The `Parallel` node has an additional configuration (`min_children_needed`), which allows it to require X out of Y approvals, for example requiring only a majority or only one of its children to approve.
+
+The approval chains are designed so that the `Proposal` needs to only communicate with the _root_ of the chain. See the approval specs for examples.
```
afeld: Thanks for documenting!

afeld:
Looks good overall! Happy to help get a copy of the DB set up if you need.

db/migrate/20150807204802_delete_approvals_without_proposals.rb
cmc333333: Added a validation to approvals instead

cmc333333:
Updated so I can fix up the code climate issues. Still needs manual testing, too.

cmc333333:
Updated and manually tested with a relatively comprehensive local database. This should be ready for merging.

app/models/ncr/work_order.rb
```diff
@@ -102,10 +102,14 @@ def setup_approvals_and_observers(approving_official_email)
...
-        self.proposal.approvers = approvers
-        Dispatcher.on_approver_removal(self.proposal, removed_approvers_to_notify)
+        original_approvers = self.proposal.individual_approvals.non_pending.map(&:user)
+        individuals = emails.map do |email|
+          user = User.for_email(email)
+          # Reuse existing approvals, if present
+          self.proposal.existing_approval_for(user) || Approvals::Individual.new(user: user)
+        end
+        self.proposal.root_approval = Approvals::Serial.new(child_approvals: individuals)
+        self.notify_removed_approvers(original_approvers)
```
afeld: Probably worth breaking this section out to one or more named methods.

cmc333333:
Updated w/ latest master and splitting a method

afeld:
Well, I've looked through this a few times and it seems good to me...still thinking it might be worth running on a copy of the production database though. Would you have any interest in setting aside an hour or something and pairing on that?

cmc333333:
Sorry @afeld , I really don't have the capacity. If you feel apprehensive though, go for the test with prod data. The only time constraint is that I know this affects @annalee and @pkarman's work.

PR #524, opened 2015-08-09
Added CL output and tag if the pending request is older than 5 days

phirefly:
• Adds CL number to the report
• Prepends asterisks to requests that are greater than 5 days old

===
Conflicts:
	app/views/report_mailer/budget_status.text.erb


cmc333333:
The `TabularData` construct might help remove some of the duplication here

PR #527, opened 2015-08-10
[Delivers #99002900] don't show all Proposals to admins / client admins

afeld:
...though they can still visit them directly.

* [ ] Deploy branch to test on dev/staging


cmc333333:
This removes those proposals from search, too. Is that the intent?

afeld:
Hmm, good point. Maybe it would make more sense to think of this as a filter rather than a Policy?

afeld:
...and by "this" I mean "the intended behavior".

afeld:
Thinking we'll go about this a different way.

PR #529, opened 2015-08-10
[Delivers #99523048] Allow observers to be destroyed

cmc333333:
This adds a section to the proposal profile that lists all of the users associated with this proposal, along with a link to add another and (if an admin, or the observer) delete the observer. Also removes some repetition in the `Role` model and makes sure `proposals.user` returns distinct users.

Looks like:
screen shot 2015-08-10 at 3 46 51 pm
screen shot 2015-08-10 at 3 47 53 pm


app/views/proposals/show.html.haml
```diff
@@ -32,6 +32,29 @@
...
+              = user.full_name
+              = "(#{role_str})" if role_str
+            %td.righted
+              - if observation && policy(observation).can_destroy?
+                = button_to "Delete", proposal_observation_path(@proposal, observation), method: :delete, data: {confirm: "Are you sure?"}
+        - if policy(@proposal).can_edit?
+          %tr
+            %td{colspan: 2}
+              - #todo: probably want to in-line this
+              = link_to "Add subscriber", proposal_observations_path(@proposal)
```
afeld: Was gonna say...the only reason for putting them on a separate page was to reduce clutter. Definitely don't need both.

app/decorators/proposal_decorator.rb
```diff
@@ -31,6 +31,20 @@ def approvals_in_list_order
     end
   end
 
+  def subscribers_list
+    userXrole = object.users.map {|u| [u, Role.new(u, object)] }
+    schwartzian_uXrXo = userXrole.map do |user, role|
```
afeld: schwartz

app/decorators/proposal_decorator.rb
```diff
@@ -31,6 +31,20 @@ def approvals_in_list_order
     end
   end
 
+  def subscribers_list
+    userXrole = object.users.map {|u| [u, Role.new(u, object)] }
```
afeld: Minor: Can we just retrieve the User from the Role object?

app/decorators/proposal_decorator.rb
```diff
@@ -31,6 +31,20 @@ def approvals_in_list_order
...
+      if role.requester?
+        ["0#{user.full_name}", user, "Requester", nil]
+      elsif role.approver?
+        ["1#{user.full_name}", user, "Approver", nil]
+      else
+        ["2#{user.full_name}", user, nil, object.observations.find_by(user: user)]
+      end
+    end
+    schwartzian_uXrXo.sort_by!(&:first).map{|tuple| tuple[1..-1]}
+  end
```
afeld: Not totally understanding what's happening in this method...is it just for sorting the list of users? Probably worth adding some method/inline comments.

app/models/proposal.rb
```diff
@@ -224,10 +224,10 @@ def partial_approve
     end
   end
 
-  # Returns True if the user is an approver and has acted on the proposal
-  def is_active_approver? user
-    current_approver = self.approvals.find_by user_id: user.id
-    current_approver && current_approver.status != "pending"
+  # Returns True if the user is an approver or has acted on the proposal
```
afeld: "is an _actionable_ approver", right?

app/policies/observation_policy.rb
```diff
@@ -0,0 +1,9 @@
+class ObservationPolicy
+  include ExceptionPolicy
+
+  def can_destroy!
+    role = Role.new(@user, @record.proposal)
+    check(@user.id == @record.user_id || role.client_admin? || @user.admin?,
+          "Only the observer can delete themself")
```
afeld: Hmm, I would think that anyone who can edit the request should be able to delete the observation, in case they accidentally add the wrong person or something.

app/views/proposals/show.html.haml
cmc333333: Sure; will replace with a dropdown for selection

app/decorators/proposal_decorator.rb
cmc333333: I intentionally removed `user` from the `Role` interface, but I can see an argument for adding it back. Will do

app/decorators/proposal_decorator.rb
cmc333333: You're right; in this case it's probably simpler just to spell out the order.

app/models/proposal.rb
cmc333333: :+1:

app/policies/observation_policy.rb
cmc333333: Sure. I started with the most restrictive rule, but I don't see any harm in loosening it up the match creation

app/decorators/proposal_decorator.rb
afeld: This graphic was meant to convey "I have no idea what that variable means" :blush:

app/views/proposals/show.html.haml
afeld: Cool, feel free to remove the old `index` page, too.

app/decorators/proposal_decorator.rb
cmc333333: https://en.wikipedia.org/wiki/Schwartzian_transform , though it'll be removed with the update.

cmc333333:
Updated to address comments and to work in-line using selectize. Looks like:

screen shot 2015-08-14 at 6 32 57 pm
screen shot 2015-08-14 at 6 33 29 pm
screen shot 2015-08-14 at 6 33 40 pm

app/controllers/observations_controller.rb
```diff
@@ -5,23 +5,23 @@ class ObservationsController < ApplicationController
...
 
-  def index
-    @observer = User.new
-    @observation = Observation.new(user: @observer, proposal: @proposal)
-  end
-
   def create
-    observation = @proposal.add_observer(params[:observation][:user][:email_address])
+    cleaned = params.permit(observation: { user: [:email_address] })
+    email = cleaned.require(:observation).require(:user).require(:email_address)
```
afeld: Not saying it's wrong, just seems strange that both of these lines are needed.

lib/role.rb
```diff
@@ -1,50 +1,35 @@
...
-  end
-
-  def observer?
-    self.role_types.include? :observer
   end
 
   # Active Observer means, their main role is an observer
   def active_observer?
-    self.role_types.include?(:observer) && !self.role_types.include?(:active_approver) && !self.role_types.include?(:requester)
+    self.observer? && !self.active_approver? && !self.requester?
```
afeld: Seems like we should have validation that a single user can't have multiple roles on a single Proposal...

cmc333333: A user can certainly have multiple roles on the same proposal. For example, a user who comments becomes an observer, regardless of other roles. Further, when this gets extended to account for the client-specific role types @annalee and @nickbristow were working on, we should expect many more situations.

app/controllers/observations_controller.rb
cmc333333: I'd be happy just sending `:email_address` if you'd prefer.

lib/role.rb
afeld: Ah, didn't realize. No worries then.

app/controllers/observations_controller.rb
afeld: Yeah, the whole strong parameters thing is only useful if you're sending the whole hash to `.create()` or `#update()` or whatever. I think this reads simpler as

```ruby
email = params[:observation][:user][:email_address]
```

afeld: Actually, don't worry about it – just now seeing the tests for the required params.

PR #532, opened 2015-08-10
[Delivers #100901060, Possibly #100944736] Interaction b/w TabularData and ProposalSearch

cmc333333:
* Remove the ability to sort on the search page; we can rely on the search ranking
* Simplify `ProposalSearch` so that it doesn't have a separate projection (and hence doesn't need to `select *`)

The latter resolves an explosion when searching.

~~This needs tests.~~ @annalee I _think_ this might resolve #100944736 too but I'm not positive.

### DIFF
--- a/lib/tabular_data/container.rb
+++ b/lib/tabular_data/container.rb
@@ -1,9 +1,10 @@
 module TabularData
   class Container
-    attr_reader :columns
+    attr_reader :columns, :frozen_sort
 
     def initialize(name, config)
       @name = name
+      @frozen_sort = false
       self.init_query(config[:engine].constantize, config.fetch(:joins, []))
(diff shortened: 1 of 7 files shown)


lib/tabular_data/container.rb
```diff
@@ -1,9 +1,10 @@
 module TabularData
   class Container
-    attr_reader :columns
+    attr_reader :columns, :frozen_sort
 
     def initialize(name, config)
       @name = name
+      @frozen_sort = false
```
afeld: How about passing this through the `config`?

cmc333333: Sure

cmc333333:
Updated with tests to account for both sorting functionality and the explosion (which resulted from proposal ids being nil in the search results)

PR #538, opened 2015-08-12
refactor Proposal listing functionality

afeld:
Probably easiest to understand going through commit-by-commit. Basically just moving things around to make the classes more specialized.


app/policies/proposal_policy/scope.rb
```diff
@@ -0,0 +1,20 @@
...
+      @user = user
+      @scope = scope
+    end
+
+    # returns the entire list of Proposals that are visible to the user
+    def resolve
+      if @user.admin?
+        @scope.all
+      elsif @user.client_admin?
+        # TODO include all Proposals that the user is involved in
```
afeld: This is the only (known) change in behavior: client admins won't be able to see their Proposals that they are involved with from other clients. This is a very narrow edge case, and it greatly simplifies the query to not worry about it.

cmc333333: Both `Query::` classes are effectively wrappers around existing active record relations (an OO-approach). What do you think about a more declarative style, in which individual clauses can be combined arbitrarily using Arel? This could be implemented, for example, via modules which contained individual Arel "where" clauses (or functions returning such clauses). You could say something like:
```ruby
Proposals.where(Matches::admin(@user).or
                Matches::client_admin(@user).or
                Matches::involved(@user))
```

cmc333333:
I'm cool merging this whenever you're ready, @afeld . Just needs a conflict resolved

PR #539, opened 2015-08-12
add a Rubocop configuration file

afeld:
Yesterday, I enabled Code Climate to run on our pull requests, which runs [Rubocop](https://github.com/bbatsov/rubocop) by default. Ruboop is aggressive, but luckily Code Climate only shows the "offsenses" on changed lines with pull requests. Example:

https://codeclimate.com/github/18F/C2/pull/538#style

Explicitly excluded some Cops (ones that I saw in the page above and don't agree with – will probably want to add more as we go), but used `--auto-gen-config` to generate the `.rubocop_todo.yml` to "zero the scale" (so to speak).

Would like to get this in before #538.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,15 @@
+AllCops:
+  Exclude:
+    - 'bin/**/*'
+    - 'db/**/*'
+  RunRailsCops: true
+Metrics/LineLength:
+  Enabled: false
+Style/Documentation:
+  Enabled: false
+Style/DotPosition:
+  Enabled: false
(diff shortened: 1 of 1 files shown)


afeld:
Also, I propose that going forward, we try considering Code Climate failures as blockers to merging, which requires fixing the code or the Rubocop configuration.

afeld:
Removed the `.rubocop_todo.yml` file and squashed – turns out that hid all of our offenses on Code Climate.

cmc333333:
> Metrics/LineLength:
;)

Solid

PR #540, opened 2015-08-12
validate the client_data_type on the Proposal

afeld:
Split out from #538.

### DIFF
--- a/spec/support/models.rb
+++ b/spec/support/models.rb
@@ -1,0 +1,9 @@
+module AbcCompany
+  class SomethingApprovable
+  end
+end
+
+module SomeCompany
+  class SomethingApprovable
+  end
+end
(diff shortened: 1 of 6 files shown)


spec/support/models.rb
```diff
@@ -0,0 +1,9 @@
+module AbcCompany
+  class SomethingApprovable
+  end
+end
+
+module SomeCompany
+  class SomethingApprovable
+  end
+end
```
cmc333333: Are these required, or is the string field enough?

afeld: Tried removing them – they're required because of https://github.com/18F/C2/blob/4a1b9c180aafd96ab1966c85808514d0175e9e18/app/policies/proposal_policy.rb#L113.

cmc333333: Roger

PR #541, opened 2015-08-13
fix client admins not being able to see all Proposals they are involved with

afeld:
Issue introduced in #538.

More importantly, converted the queries to Arel, which makes the various conditions for the subsets of Proposals composable. http://www.scuttle.io/ was super helpful in doing the conversion. Want to play around with them more to try and make them more readable, but I think they're ok for now.

Both pull requests are in service of https://www.pivotaltracker.com/story/show/99002900.

### DIFF
--- a/app/models/approval.rb
+++ b/app/models/approval.rb
@@ -23,6 +23,7 @@ class Approval < ActiveRecord::Base
   self.statuses.each do |status|
     scope status, -> { where(status: status) }
   end
+  scope :non_pending, -> { where.not(status: 'pending') }
 
   default_scope { order('position ASC') }
 
--- a/app/models/approval.rb
+++ b/app/models/approval.rb
@@ -9,9 +9,6 @@ class Approval < ActiveRecord::Base
   end
 
   belongs_to :proposal
-  belongs_to :user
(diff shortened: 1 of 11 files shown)


app/models/approval.rb
```diff
@@ -9,6 +9,9 @@ class Approval < ActiveRecord::Base
   end
 
   belongs_to :proposal
+  belongs_to :user
+  has_many :delegations, through: :user, source: :outgoing_delegates
+  has_many :delegates, through: :delegations, source: :assignee
```
afeld: Didn't end up using this, but I think it's useful anyway.

cmc333333: Should these be on `Approvals::Individual` instead?

afeld:
Updated!

cmc333333:
I had issues reading this (I followed the method names like I'd follow GOTOs); there is probably room for improvement in the method names, though I don't have a specific suggestion. That said, I think it's a step in the right direction. We'll be able compose these Arel clauses with client-specific clauses to create reports, alerts, etc. relatively easily.

afeld:
> there is probably room for improvement in the method names

:+1:

PR #542, opened 2015-08-14
rearrange the code for Proposal listing

afeld:
Probably easiest to review commit-by-commit.


app/models/user.rb
```diff
@@ -1,6 +1,10 @@
 class User < ActiveRecord::Base
   has_paper_trail
 
+  validates :client_slug, inclusion: {
+    in: ->(_) { Proposal.client_slugs },
+    allow_blank: true
+  }
```
afeld: This is the only (intended) change in behavior in this PR.

cmc333333:
It'd be worth taking a look at https://github.com/camertron/arel-helpers to reduce some of the repeated lines

PR #544, opened 2015-08-14
[DO NOT MERGE] make JOIN of client models for proposal pages implicit

afeld:
This is more of an idea than code that should actually be merged, but in the past few refactoring I've been keeping an eye out for how we can make the relationships between Proposals and the client models more implicit; in other words, how we reduce the amount of code specific to each. This is an example, where the `LEFT JOIN`s between them will (presumably) be the same for all client models, so we shouldn't have to write it out by hand. Would love ideas about a better way to achieve this.

### DIFF
--- a/lib/tabular_data/container.rb
+++ b/lib/tabular_data/container.rb
@@ -71,14 +71,21 @@ module TabularData
 
     def init_query(engine, joins)
       @query = engine.all()     # convert into a query
-      joins.each do |name, config|
-        if config == true
-          join_tables = engine.joins(name).join_sources
-          join_tables[-1].left.table_alias = name   # alias the table
-          @query = @query.joins(join_tables).includes(name)
-        else  # String config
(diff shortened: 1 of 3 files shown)


lib/tabular_data/container.rb
```diff
@@ -71,14 +71,21 @@ def set_sort(field)
...
+
+      Proposal::CLIENT_MODELS.each do |client_model|
+        table_name = client_model.table_name
+        join = <<-SQL
+          LEFT JOIN #{table_name}
+          ON (#{table_name}.id = proposals.client_data_id
+          AND proposals.client_data_type = '#{client_model}')
+        SQL
+        @query = @query.joins(join)
       end
```
afeld: Not sure of the best place to put this.

afeld:
Moving to a story to pick up later: https://www.pivotaltracker.com/story/show/105638870

PR #546, opened 2015-08-17
[Delivers #100955526] allow anyone who can see the Proposal to add other observers

afeld:
We might want to tighten this up a bit down the road, but I think it's fine for now.

### DIFF
--- a/app/controllers/observations_controller.rb
+++ b/app/controllers/observations_controller.rb
@@ -1,7 +1,7 @@
 class ObservationsController < ApplicationController
   before_action :authenticate_user!
   before_action :find_proposal
-  before_action ->{authorize @proposal, :can_edit!}
+  before_action ->{authorize @proposal, :can_show!}
   rescue_from Pundit::NotAuthorizedError, with: :auth_errors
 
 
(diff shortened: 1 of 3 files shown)


app/controllers/observations_controller.rb
```diff
@@ -1,7 +1,7 @@
 class ObservationsController < ApplicationController
   before_action :authenticate_user!
   before_action :find_proposal
-  before_action ->{authorize @proposal, :can_edit!}
+  before_action ->{authorize @proposal, :can_show!}
```
afeld: Wondering if it's worthwhile to break out a separate ObservationPolicy...

cmc333333:
Double checking, do https://github.com/18F/C2/blob/master/app/controllers/observations_controller.rb#L4 and https://github.com/18F/C2/blob/master/app/policies/ncr/work_order_policy.rb#L11 make this change not required?

afeld:
Good point, though it sounds like we want to make observers add-able more broadly.

https://www.pivotaltracker.com/story/show/99421886

cmc333333:
Perfect; this'll do for now, then

PR #549, opened 2015-08-17
create an ObservationPolicy class

afeld:
Centralizes the authorization logic around adding/removing observers.

### DIFF
--- a/app/controllers/observations_controller.rb
+++ b/app/controllers/observations_controller.rb
@@ -1,7 +1,8 @@
 class ObservationsController < ApplicationController
   before_action :authenticate_user!
   before_action :find_proposal
-  before_action ->{authorize @proposal, :can_show!}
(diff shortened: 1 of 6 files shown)


app/controllers/observations_controller.rb
```diff
@@ -1,7 +1,8 @@
 class ObservationsController < ApplicationController
   before_action :authenticate_user!
   before_action :find_proposal
-  before_action ->{authorize @proposal, :can_show!}
+  before_action ->{authorize Observation.new(proposal: @proposal)}, only: :create
```
afeld: Kinda wish there was a cleaner way to do this...

https://github.com/elabs/pundit#additional-context

cmc333333: What about modifying `#observation` to provide this empty object if there's no id?

afeld:
Updated!

PR #551, opened 2015-08-18
convert the HTML mailer templates to HAML

afeld:
Worth noting that I didn't manually test each of these.


app/views/communicart_mailer/comment_added_email.html.haml
```diff
@@ -0,0 +1,12 @@
+%table.w-container.main-container{width: '800'}
+  - proposal = @comment.proposal
+  %tr
+    %td.w-container.html-email-message
+      %p
+        Action need on Request \##{proposal.public_identifier} (#{proposal.name}):
```
cmc333333: I think we can remove the first pound, as the `public_identifier` includes it, hey?

app/views/communicart_mailer/_comments.html.haml
```diff
@@ -0,0 +1,18 @@
...
+  %table.submodel-section
+    %tr
+      %td.w-container{colspan: 2}
+        %strong Comments
+    - comments.each do |c|
+      %tr{class: ('comment-update' if c.update_comment)}
+      %td.comment-detail.first{valign: 'top'}
+        %p
+          = l(c.created_at.to_date)
+      %td.comment-detail.first{valign: 'top'}
```
cmc333333: Looks like "first" was added as a class here?

app/views/communicart_mailer/_comments.html.haml
```diff
@@ -0,0 +1,18 @@
+- if comments.any?
+  %table.submodel-section
+    %tr
+      %td.w-container{colspan: 2}
+        %strong Comments
+    - comments.each do |c|
+      %tr{class: ('comment-update' if c.update_comment)}
```
cmc333333: Good; this was one of the issues with the automated conversion

app/views/communicart_mailer/_comments.html.haml
afeld: Damn, good catch! I think I copied-and-pasted that line.

afeld:
Updated!

PR #550, opened 2015-08-17
[#100958394] let observers know who they have been added by

afeld:

PR #552, opened 2015-08-18
refactor mailers

afeld:
Builds on #550. Diff:

https://github.com/18f/c2/compare/100958394_observer_context...communicart-mailer-refactor


afeld:
Updated! Might be easier to review commit-by-commit than as a whole, but the whole idea is trying to split up the CommunicartMailer.

afeld:
Updated!

cmc333333:
Great stuff

PR #553, opened 2015-08-19
[Delivers #99002900] only display Proposals a user is involved with on /proposals

afeld:
In other words, app/client admins won't see _all_ Proposals on that page.

### DIFF
--- a/app/controllers/proposals_controller.rb
+++ b/app/controllers/proposals_controller.rb
@@ -18,8 +18,9 @@ class ProposalsController < ApplicationController
 
   def index
     @CLOSED_PROPOSAL_LIMIT = 10
+
     @pending_data = self.listing.pending
-    @approved_data = self.listing.approved(@CLOSED_PROPOSAL_LIMIT)
+    @approved_data = self.listing.approved.alter_query{ |rel| rel.limit(@CLOSED_PROPOSAL_LIMIT) }
     @cancelled_data = self.listing.cancelled
   end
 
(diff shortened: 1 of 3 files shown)


afeld:
Deployed to dev – seems to work!

lib/query/proposal/listing.rb
```diff
@@ -59,6 +59,15 @@ def proposals_container(name, extra_config = {})
         container
       end
 
+      # returns a Container that is limited to what the user should see on /proposals, even if the ProposalPolicy::Scope allows them to see more
+      def index_visible_container(name)
+        container = self.proposals_container(name)
+        container.alter_query do |rel|
+          condition = Query::Proposal::Clauses.which_involve(self.user)
+          rel.where(condition)
+        end
```
cmc333333: There used to be a nice syntax for doing this, but I think it was removed:
```ruby
self.proposals_container(name) do |rel|
    rel.where(Query::Proposal::Clauses.which_involve(self.user))
end
```

app/controllers/proposals_controller.rb
```diff
@@ -18,8 +18,9 @@ def show
 
   def index
     @CLOSED_PROPOSAL_LIMIT = 10
+
     @pending_data = self.listing.pending
-    @approved_data = self.listing.approved(@CLOSED_PROPOSAL_LIMIT)
+    @approved_data = self.listing.approved.alter_query{ |rel| rel.limit(@CLOSED_PROPOSAL_LIMIT) }
```
cmc333333: :+1: Ideally this'll become something you could pass in to the config

PR #554, opened 2015-08-19
[Delivers #101210264] background email sending

afeld:
This takes the sending of emails out of the request cycle, which has a few benefits:

* Response times will be faster
* Emails failing to send will no longer fail the overall request
* Failed email sends will be retried, and if those fail, we have a record of it

Updated gems, as well.


phirefly:
:shipit:

PR #557, opened 2015-08-20
move API token creation into the mailer

afeld:
Don't _love_ the idea of the mailer being responsible for this, but it centralizes the logic that's needed for the mail to be sent successfully.

Came up because the delayed email sends are failing on staging because they don't have a token (which should have been created before, but is now missing for whatever reason). This sidesteps the issue of creating the token if it doesn't exist right before the email is sent.

### DIFF
--- a/app/mailers/communicart_mailer.rb
+++ b/app/mailers/communicart_mailer.rb
@@ -8,8 +8,9 @@ class CommunicartMailer < ActionMailer::Base
   add_template_helper MarkdownHelper
 
   # Approver can approve/take other action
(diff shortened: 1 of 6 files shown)


afeld:
Ready! 9651a25230ca51d00c592342b18d8c16eab818b5 is the important change here...the others are just refactoring.

afeld:
Tested the branch on staging – emails sent successfully.

cmc333333:
> (which should have been created before, but is now missing for whatever reason)

Might be a good idea to create this token once the approval becomes active

PR #561, opened 2015-08-22
Specifies DAP agency field, adds forceSSL

konklone:
This specifies an `agency` of `GSA` for DAP reporting, and adds a missing `forceSSL` flag to GA settings.


afeld:
Thanks!

PR #563, opened 2015-08-24
[Delivers #100956196] send to a different mailbox for Office of Leasing BA80 requests

afeld:
This pull request is a bit gross to me (adding group-specific conditional logic), but not sure how to make it better :confused:


cmc333333:
> This pull request is a bit gross to me (adding group-specific conditional logic), but not sure how to make it better :confused:

I'd argue that this configuration should live in the database

afeld:
Can you elaborate?

cmc333333:
Sure. The basic idea is to convert the logic (currently encoded as Ruby) into data, which can then be manipulated on the fly. Think if-this-then-that or a serializable rules engine.

Start with a "rule" (or "trigger", "condition", "scenario", etc.). It'd be good to be more generic, but a quick first pass of a rule would look something like:
```
(123, "organization", "0123456")    # (rule id, field, value)
```

To make a flexible system, we'd want to create a set of potential "actions", but for this scenario, you could just link to approval chains directly. For example, storing all of the rule ids required for a particular approval chain to be used. Then, to determine which approval chain to use, you run through the associated rules and use (e.g.) the first applicable chain.

Those rules and actions could be configured though files, but eventually should be surfaced through an admin interface. The upshot to this approach is that there's no client-specific code, deployments, environmental variables etc. for these types of changes. This would allow C2 to scale a bit easier as the number of configurations grows. The downside is that the configuration is harder to reproduce (though that can be ameliorated via the ability to import/export settings).

I briefly explored the applicability of this approach to C2 when implementing dynamic fields. I had a prototype up in about a day, so I don't think there's a high cost. I'm a proponent of generalizing basically everything, though.

pkarman:
+1 to serializing rules in the db. I'm just dipping my toes into my familiarity with this app, but at first blush, the scalability of defining complex workflow rules seems to me suited to some kind of serialized DSL.

PR #568, opened 2015-08-27
Search UI modifications

istanfield09:
In response to [remove search box from `show` templates](https://www.pivotaltracker.com/story/show/95181828) and [Ability to clear out existing search results](https://www.pivotaltracker.com/story/show/95162638). 

I am not sure this an exact solution to the latter story. The "Clear search terms" link just leads back to `/proposals`, effectively clearing the search box.

I am also pretty sure my Haml can be improved. Thanks for taking a look! :)

### DIFF
--- a/app/views/proposals/query.html.haml
+++ b/app/views/proposals/query.html.haml
@@ -2,10 +2,9 @@
   = render partial: 'search_ui'
   %h1
     Proposals participated in
-    %h2
-    %strong Currently displaying results for: 
-    = property_to_s(@text)
-    %br
-    %a{ :href => "/proposals" } Clear search terms
-    = datespan_header(@start_date, @end_date)
+  %strong Currently displaying results for: 
+  = property_to_s(@text)
+  %br
+  %a{ :href => proposals_path } Clear search terms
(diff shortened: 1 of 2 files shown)


app/views/proposals/query.html.haml
```diff
@@ -2,5 +2,10 @@
   = render partial: 'search_ui'
   %h1
     Proposals participated in
+    %h2
+    %strong Currently displaying results for: 
+    = property_to_s(@text)
+    %br
+    %a{ :href => "/proposals" } Clear search terms
```
cmc333333: Your `h2` doesn't have a body, so it can be removed. Also, maybe replace the anchor with
```
%a{href: proposals_path} Clear search terms
```
so as to not hard-code the urls.

cmc333333:
Good stuff! We'll go ahead and merge in a bit unless you want to respond to that suggestion. A screenshot would also be lovely :)

istanfield09:
screenshot 2015-08-31 14 23 11

Here is a screenshot, thanks for taking a look at my code!

afeld:
Thanks for the contribution! :us: :rocket: :guitar:

PR #570, opened 2015-08-28
send feedback emails 'from' the submitter

afeld:


cmc333333:
Might make sense to _also_ CC the submitter

PR #573, opened 2015-09-02
[Delivers 101461556] Optional reason when adding observer

yozlet:
One thing to note is the `observation.save!` in `observations_controller.rb`:11, which I don't like but needed to make it work (otherwise the Observation received by the Dispatcher and mailer didn't include the `reason`). I suspect I should be doing that a better way, right?


pkarman:
The schema change to the observations table is going to conflict with https://github.com/18F/C2/tree/100427542-user-roles where the observations table is dropped altogether. We can refactor the user-roles branch if this PR gets merged first; just wanted to note it here.

Also I note that the observation.reason stores the same content as the proposal.comment. Could we instead add a `ref_type` and `ref_id` (kinda polymorphic but not quite) to the comments table so we can cross-reference the "thing" that originated the comment? Seems more normalized and extensible.

yozlet:
Superseded by #577.

PR #574, opened 2015-09-02
adding favicon to home page and application template

anthonygarvan:
Tested on Firefox and Chrome.


afeld:
This looks fine, though there's already a favicon under [`public/`](https://github.com/18F/C2/blob/master/public/favicon.ico)...might be as simple as replacing that file.

anthonygarvan:
Oh yeah, I saw that a while back but forgot about it. It was just an empty image, but there were not any links to it from anywhere so it was not coming through. In some dark corners of the internet, I can find reports that just having a favicon.ico in the project root will magically make favicons show up, but most places recommend putting in a link to it. Anyway, the recommended thing for rails apps seems to be to have it in assets and link to it with this helper as I have done. I removed the dead favicon to avoid confusion in the future.

PR #575, opened 2015-09-02
[Delivers #100427542] Moving user roles to the DB

annalee:
Per https://www.pivotaltracker.com/story/show/100427542 , we're moving user roles like admin and client admin out of environment variables and into the DB. This will make it easier to manage user roles as we scale.

We're also working on moving Proposal-scoped roles, such as Observer and Approver, into the new ProposalRole table. This PR moves Observer over. Approver will be tackled separately (it's more complicated, because it needs to hook into @cmc333333's new approval flow).


spec/lib/query/proposal/listing_spec.rb
```diff
@@ -8,28 +10,68 @@
...
       context "with an arbitrary client" do
         before do
           user.update_attribute(:client_slug, 'ncr')
+          another_user.update_attribute(:client_slug, 'ncr')
+          yet_another_user.update_attribute(:client_slug, 'gsa18f')
         end
 
-        it "only displays that user's Proposals when they are a client admin" do
-          proposal = FactoryGirl.create(:proposal, requester: user, status: status)
+        describe "when user is not an admin of any kind" do
```
annalee: Oop, @pkarman and I wrote the wrong set of tests here. This is meant to test the proposal listing on the user dash--Aidan's original test was correct. We'll need to move our permissions tests elsewhere and update them.

yozlet:
Whoops, sorry - deleted & closed wrong PR. Nothing to see here...

app/models/proposal.rb
```diff
@@ -135,10 +135,26 @@ def reset_status()
...
+  def add_observer(email_or_user)
+    # polymorphic
+    if email_or_user.is_a?(User)
+      user = email_or_user
+    else
+      user = User.for_email(email_or_user)
+    end
+
+    # no duplicates
+    observer = self.observers.select{|o| o.id == user.id}.first
```
afeld: Can we do this via SQL, and as a method on the Proposal?

```ruby
def existing_observation_for(user)
  self.observations.find_by(user: user)
end
```

db/migrate/20150828150524_remove_approval_groups.rb
```diff
@@ -0,0 +1,7 @@
+class RemoveApprovalGroups < ActiveRecord::Migration
+  def change
+    drop_table :approval_roles
+    drop_table :approval_groups
+    drop_table :approval_groups_users
+  end
+end
```
afeld: Would you mind pulling this out to a separate PR?

pkarman: the evolution of this branch required it be here. the original table was `user_roles` which we renamed to `approval_roles` and then realized it wasn't used at all. The user_roles namespace was what we wanted to use.

We can pull out this particular migration into another PR, but I considered it a 'house-cleaning' prior to the new roles' tables structures.

app/models/proposal.rb
```diff
@@ -135,10 +135,26 @@ def reset_status()
...
+
+    if !observer
+      observer_role = Role.find_or_create_by(name: 'observer')
+      observer = Observation.new(user_id: user.id, role_id: observer_role.id, proposal_id: self.id)
```
afeld: Mind calling this `observation`, just to be consistent?

db/migrate/20150828150524_remove_approval_groups.rb
afeld: Ok, no worries.

afeld:
The code here looks fine, but I have some questions about the overall goal(s). I promise I'm not trying to be a jerk :wink:

Does anyone have strong feelings about a `roles` table, instead of using a string column? The latter is simpler in that it saves a `JOIN` each time you need to look one up. We do this for `Proposal#flow` and `Ncr::WorkOrder#expense_type`, for example.

In my mind, the UserRole makes the most sense in relation to a particular client, e.g. "I'm a client admin for NCR". Thoughts?

> We're also working on moving Proposal-scoped roles, such as Observer and Approver, into the new ProposalRole table

Heh, this is how it used to be – see https://github.com/18F/C2/pull/257. What's the benefit of re-combining them? While I understand the implementation, it's still not super clear to me how these roles would be used. I think I had asked this in a past sprint planning or something, but for the `ProposalRole`s specifically: "what problem are we solving?" For starters, what would other examples of roles be?

pkarman:
Strong feeling? Not strong, but yes, feelings. Maybe habits and my feeling of attachment to them is a better description for me.

How to manage lookup values (db vs code) has always felt to me more art than science. My bias is the database, and further, to exercise as many of the data integrity features of the database as possible. I realize that is not the Rails Way, which is more to treat the database as a sophisticated key/value store (NoSQL) and use naming conventions rather than actual FK constraints, e.g.

That said, I have often used code constants to manage enum-like columns (e.g. `type` must be one of (A, B, C)) where code clarity is (I think) helped by naming the constant with something like (TYPE_A, TYPE_B, TYPE_C) and storing it as a single char rather than an id pointing at another db table. There's a storage trade-off (a char is one byte, and id is by default 4) but I won't defend that scientifically. It just *feels* right to me that small known value ranges (enums) would be managed in code.

So we started this branch with a bias: roles should be normalized as much as possible. Which led us to thinking about it from a db schema perspective, since both @annalee and I are not Rails folks by training, with little knowledge of the history of the project.

I could see doing this instead:

```
create table user_roles (
  user_id integer,
  role_name varchar(255)
)
```

and then managing `role_name` via code, with no master Roles table. It *feels* a little dirty to me, somehow, to not go the fully monty and normalize away the role name to a master Roles table. But I can see the appeal of avoiding the extra JOIN.

As far as #257 and Observers and Approvers, I asked the same question when we started: what problem are we attempting to solve. We did talk about this at some length, because it seemed that in lieu of a "role" there was an entire table named for each particular role. I don't have a specific example at this point of another role besides "observer" or "approver" and so we could be accused of premature optimization by normalizing away the role-as-table to the structure proposed in this PR. I guess the lynchpin there is "premature" because while I agree that it is an optimization, maybe, whether it reflects a certain point in the maturity of the project is a more subjective thing and I am new enough to not have sufficient perspective. I guess history will be the judge, whether in the course of this national ramp-up we discover another role (or two) that need to be reflected in the code, and then do we add a new table or insert a new Role row. If we find in 6 mos that we have a need for more roles it feels somehow more sustainable to have done that normalizing work now, assuming that inserting rows rather than creating tables is "more sustainable".

So I don't feel strongly about this approach. It fits my mental habits and the patterns I have worked with before, but that doesn't mean I'm so attached to those habits that this is a ditch I want to die in.

afeld:
As I remember, the motivation for the story was to more explicitly represent different user responsibilities/permissions, such as (for NCR/PBS):

* Card holders
* Approving officials
* Reconcilers
* Budget approvers
* Budget mailboxes
* Regional supervisors (a.k.a. "client admins")
* etc.

The `roles` table vs enum thing is minor...totally fine with merging the `UserRole` stuff as-is, but less sure about how the `ProposalRole`s get us closer to the above.

pkarman:
Understood. There are 3 different cases where a User has a Role: in relation to the app as a whole, in relation to a client, and in relation to a specific proposal. The `UserRole` table handles the first 2 cases. The `ProposalRole` addresses the 3rd. The only win I could and can see for `ProposalRole` is architectural consistency: all the roles are managed via a master table and a join table. Whether that "consistency" is optimal is debate-able, and I can only say that as a person coming to the app for the first time, I would find it easier to grok if it were consistent. It took me a bit to understand that the existing `user_roles` table was vestigal, that `approvals` was managing both the approver and the approval itself and therefor is provisioned in advance of the approval actual taking place, and that observations was really defining the relationship between the observer (User) and the proposal.

The old joke is that there are only 2 hard things in computer science, and naming things is one of them. This feels like a case of that.

cmc333333:
Quick note on another benefit of a distinct `roles` table -- it can be referenced separately in the future. For example, if we ever wanted to make permissions more dynamic, we'd need to refer to roles in the database. Likewise, we could attach approvals to a specific role rather than an individual users.

If, instead, role were just a denormalized string on each of these tables, maintenance would be a PITA.

db/migrate/20150901141746_move_client_admin_role.rb
```diff
@@ -0,0 +1,11 @@
+class MoveClientAdminRole < ActiveRecord::Migration
+  def change
+    role = Role.find_or_create_by(name: 'client_admin')
+
+    User.client_admin_emails.each do |email_address|
```
afeld: We should avoid using the `.client_admin_emails` method here, since it will presumably be removed and thus break this migration in the future.

db/migrate/20150901190853_move_admin_env_var_to_role.rb
```diff
@@ -0,0 +1,11 @@
+class MoveAdminEnvVarToRole < ActiveRecord::Migration
+  def change
+    role = Role.find_or_create_by(name: 'admin')
+
+    User.admin_emails.each do |email_address|
```
afeld: Ditto above.

afeld:
Just two minor comments and one Code Climate issue (which it might be worthwhile have Code Climate / rubocop [ignore](https://github.com/bbatsov/rubocop#disabling-cops-within-source-code)). Otherwise, :shipit:

db/migrate/20150901141746_move_client_admin_role.rb
afeld: Come to think of it, mind removing that method?

afeld:
Migration failed on a copy of the production database :cry: 

```
OUT Migrating to MoveObservationsToRoles (20150831193037)
OUT == 20150831193037 MoveObservationsToRoles: migrating ==========================
OUT PG::NotNullViolation: ERROR:  null value in column "user_id" violates not-null constraint
OUT DETAIL:  Failing row contains (103, 1, null, 142).
OUT : INSERT INTO "proposal_roles" ("proposal_id", "role_id") VALUES ($1, $2) RETURNING "id"
ERR rake aborted!
ERR StandardError: An error has occurred, this and all later migrations canceled:
ERR PG::NotNullViolation: ERROR:  null value in column "user_id" violates not-null constraint
ERR DETAIL:  Failing row contains (103, 1, null, 142).
ERR : INSERT INTO "proposal_roles" ("proposal_id", "role_id") VALUES ($1, $2) RETURNING "id"
```

cmc333333:
Seems likely that's due to the manual user deletion. Yay adding constraints to the db!

Fix would be to delete those bad observations.

afeld:
Grrrrr

```
> ap Observation.where(user_id: nil)
[
    [0] #<Observation:0x007f424ba7d2f0> {
                 :id => 108,
        :proposal_id => 142,
            :user_id => nil,
         :created_at => Wed, 22 Jul 2015 19:36:39 UTC +00:00,
         :updated_at => Wed, 22 Jul 2015 19:36:39 UTC +00:00
    },
    [1] #<Observation:0x007f424ba7d188> {
                 :id => 369,
        :proposal_id => 286,
            :user_id => nil,
         :created_at => Thu, 13 Aug 2015 13:08:32 UTC +00:00,
         :updated_at => Thu, 13 Aug 2015 13:08:32 UTC +00:00
    },
    [2] #<Observation:0x007f424ba7d048> {
                 :id => 372,
        :proposal_id => 484,
            :user_id => nil,
         :created_at => Thu, 13 Aug 2015 14:28:47 UTC +00:00,
         :updated_at => Thu, 13 Aug 2015 14:28:47 UTC +00:00
    },
    [3] #<Observation:0x007f424ba7cf08> {
                 :id => 446,
        :proposal_id => 522,
            :user_id => nil,
         :created_at => Mon, 17 Aug 2015 12:13:21 UTC +00:00,
         :updated_at => Mon, 17 Aug 2015 12:13:21 UTC +00:00
    }
]
```

https://www.pivotaltracker.com/story/show/102723774

cmc333333:
It'd be a good idea to check invalid references too -- deleting the user might not unset it's user_id in the `observations` table.

afeld:
Ran an `Observer.where(user_id: nil).destroy_all`, but got this on the next deploy:

```
2015-09-04T00:16:12.21-0400 [App/0]      OUT Migrating to MoveObservationsToRoles (20150831193037)
2015-09-04T00:16:12.21-0400 [App/0]      OUT == 20150831193037 MoveObservationsToRoles: migrating ==========================
2015-09-04T00:16:13.15-0400 [App/0]      OUT PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_proposal_roles_on_role_id_and_user_id_and_proposal_id"
2015-09-04T00:16:13.15-0400 [App/0]      OUT DETAIL:  Key (role_id, user_id, proposal_id)=(13, 457, 552) already exists.
2015-09-04T00:16:13.15-0400 [App/0]      OUT : INSERT INTO "proposal_roles" ("proposal_id", "user_id", "role_id") VALUES ($1, $2, $3) RETURNING "id"
```

PR #576, opened 2015-09-02
102087948 timestamps

pkarman:
Use the browser's timezone for displaying timestamps in the browser. Use UTC for dates and times in emails, and make the timezone explicit.

https://www.pivotaltracker.com/story/show/102087948

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,7 @@ gem 'autoprefixer-rails'
 gem 'awesome_print'
 gem 'aws-sdk-v1'    # remaining on v1 due to https://github.com/thoughtbot/paperclip/issues/1764
 gem 'bootstrap-sass'
+gem 'browser-timezone-rails'
 gem 'clockwork', require: false
 gem 'daemons' # for delayed_job
 gem 'delayed_job_active_record'
--- a/app/mailers/report_mailer.rb
+++ b/app/mailers/report_mailer.rb
@@ -3,7 +3,7 @@ class ReportMailer < ApplicationMailer
 
   def budget_status
     to_email = ENV.fetch('BUDGET_REPORT_RECIPIENT')
-    date = Time.now.in_time_zone('Eastern Time (US & Canada)').strftime("%a %m/%d/%y")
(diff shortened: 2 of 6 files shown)


Gemfile
```diff
@@ -8,6 +8,7 @@ gem 'autoprefixer-rails'
 gem 'awesome_print'
 gem 'aws-sdk-v1'    # remaining on v1 due to https://github.com/thoughtbot/paperclip/issues/1764
 gem 'bootstrap-sass'
+gem 'browser-timezone-rails'
```
afeld: Ha, @kbaum (the creator) and I run [a meetup](http://hackerhours.org/) together! Small world.

afeld:
Looks good! Mind fixing [the Code Climate issues](https://codeclimate.com/github/18F/C2/pull/576)?

pkarman:
thanks for noticing that @afeld. I am new to code climate so will watch for that in future.

app/mailers/report_mailer.rb
```diff
@@ -3,7 +3,7 @@ class ReportMailer < ApplicationMailer
 
   def budget_status
     to_email = ENV.fetch('BUDGET_REPORT_RECIPIENT')
-    date = Time.now.in_time_zone('Eastern Time (US & Canada)').strftime("%a %m/%d/%y")
+    date = Time.now.utc.strftime("%a %m/%d/%y (%Z)")
```
afeld: Does this mean the times will show up in emails as UTC?

Gemfile
kbaum: I hope it still works! :smile: 

We still use it but we are on rails 3.  Don't tell anyone!

app/mailers/report_mailer.rb
pkarman: It does. It's not optimal and would be better if it showed up in the timezone of the respondent. However, for this particular case, there is no time: just m/d/y and day of the week. I added the timezone so it was clear what timezone was being used, but in practice, as long as the email is sent before 8pm ET, it will have the same ymd no matter whether it was UTC or ET or CT or PT.

PR #578, opened 2015-09-02
101217598 vendor form fix

pkarman:
Make a new vendor string value sticky on form error.

https://www.pivotaltracker.com/story/show/101217598

NOTE there is no accompanying test for this because (a) I could find no prior art in spec/ for testing HTML responses and (b) there was no regression apparent in existing tests. I would be happy to blaze that trail (testing HTML responses) if folks think it is needed.

### DIFF
--- a/app/helpers/ncr/work_orders_helper.rb
+++ b/app/helpers/ncr/work_orders_helper.rb
@@ -13,13 +13,13 @@ module Ncr
       all.uniq.sort
     end
 
-    def vendor_options(vendor=nil)
+    def vendor_options(vendor = nil)
       all_vendors = Ncr::WorkOrder.where.not(vendor: nil).pluck('DISTINCT vendor')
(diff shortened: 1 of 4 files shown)


afeld:
> I could find no prior art in spec/ for testing HTML responses

Hmm, I thought I remember some test(s) around the dropdown in [`spec/features/ncr_work_orders_spec.rb`](https://github.com/18F/C2/blob/master/spec/features/ncr_work_orders_spec.rb), but I'm not seeing anything at a quick glance. Maybe spend a bit of time (timebox to an hour or something?) to see if you can write a reasonable test for it (using [capybara](https://github.com/jnicklas/capybara) and `js: true`), but otherwise, go ahead and merge.

pkarman:
thanks @afeld for pointing out where those tests were. I added one for this feature, and it passing. Ok then to merge?

PR #579, opened 2015-09-03
adding browser support table

anthonygarvan:
Let's continue the discussion on browser support here. Again, not sure how relevant it is for you guys since I don't know the details of your contracts etc., but typically it's nice to have a written policy on these things before it's too late and you get roped into supporting older stuff, or newer stuff that just acts funky like iPads.

### DIFF
--- a/doc/overview.md
+++ b/doc/overview.md
@@ -78,3 +78,11 @@ The NCR use case was built around GSA service centers (paint shops, landscapers,
 ## Performance
 
 C2 has the [Peek](https://github.com/peek/peek) performance bar loaded in development, and in production for admins. It's hidden by default – press the backtick (`` ` ``) key to display it.
+
+## Supported Browsers
+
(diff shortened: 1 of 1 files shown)


afeld:
@phirefly Do you know if we've made any promises around what browsers we support?

Here are our traffic stats for the last month:

screen shot 2015-09-03 at 11 58 42 am

IE versions:

screen shot 2015-09-03 at 11 59 03 am

phirefly:
No promises at this point. Chrome is approved to use in NCR. We verified that all folks in trainings for NCR had access to it when we told them it was the recommended browser. 11% of the users on IE is still significant. We will be pushing this discussion forward as we do discovery on the PBS national rollout.

phirefly:
https://www.pivotaltracker.com/story/show/102679526

phirefly:
This is likely also worth some content from the FAQ. I will add a story.

:ship:ping this.

PR #581, opened 2015-09-04
try to avoid sql constraint errors

pkarman:
100427542 - migration bug due to orphaned FK records and dupes.

### DIFF
--- a/db/migrate/20150831193037_move_observations_to_roles.rb
+++ b/db/migrate/20150831193037_move_observations_to_roles.rb
@@ -4,19 +4,9 @@ class MoveObservationsToRoles < ActiveRecord::Migration
     role = Role.find_or_create_by(name: 'observer')
 
     # We have to use raw sql because we've subclassed the observation class
-    pgres = ProposalRole.connection.execute("SELECT proposal_id, user_id from observations")
-    seen_proles = {}
+    pgres = ProposalRole.connection.execute("SELECT DISTINCT(proposal_id, user_id) FROM observations WHERE user_id IN (SELECT id FROM users)")
     pgres.each_row do |row|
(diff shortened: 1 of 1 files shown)


db/migrate/20150831193037_move_observations_to_roles.rb
```diff
@@ -5,11 +5,21 @@ def change
 
     # We have to use raw sql because we've subclassed the observation class
     pgres = ProposalRole.connection.execute("SELECT proposal_id, user_id from observations")
+    seen_proles = {}
```
cmc333333: Minor: should `seen_proles` be a set?

A bigger concern is the n+1 queries. Even though this is only run once, each migration has a timeout (due to being included in the cloud foundry start command).
What do you think about
```SQL
SELECT proposal_id, user_id, users.id FROM observations LEFT JOIN users ON (user_id = users.id)
```
Then if the third column were null, we'd know the user doesn't exist?

You could also group by proposal_id and user_id (or `DISTINCT`) if you wanted to get rid of duplicates via SQL rather than in Ruby land.

pkarman: thanks @cmc333333. I was not familiar with Ruby Set class. (My Perl background showing here, which doesn't have a built-in Set type.) 

Just read up on it. I like that the syntax is clearer. http://blog.x-aeon.com/2012/02/07/using-data-sets-in-ruby-performance-of-hash-vs-set/ has an interesting comparison.

Re: SQL suggestions, yes I see the advantages. The code was easier for me to write than the SQL, but I can rewrite if you think the timeouts are going to be an issue.

pkarman:
@cmc333333 that SQL look better?

PR #583, opened 2015-09-04
102087948 timezone coercion

pkarman:
The helper method that humanizes the timestamp was operating on a string which lacked the timezone, so it assumed UTC. Passing it a Time object instead, with proper localized timezone, should help it Do The Right thing in its date math.


phirefly:
:ship: it

PR #585, opened 2015-09-04
use #each instead of #each_row in migration

afeld:
Follow-up to #581.

`#each_row` returns each row as an array, whereas `#each` returns them as a Hash :facepalm: This migration is getting on my nerves.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -123,9 +123,9 @@ ActiveRecord::Schema.define(version: 20150901190853) do
     t.datetime "created_at"
     t.datetime "updated_at"
     t.boolean  "direct_pay"
-    t.string   "cl_number"
-    t.string   "function_code"
-    t.string   "soc_code"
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -123,9 +123,9 @@
     t.datetime "created_at"
     t.datetime "updated_at"
     t.boolean  "direct_pay"
-    t.string   "cl_number"
-    t.string   "function_code"
-    t.string   "soc_code"
+    t.string   "cl_number",       limit: 255
+    t.string   "function_code",   limit: 255
+    t.string   "soc_code",        limit: 255
```
afeld: https://www.pivotaltracker.com/story/show/102785168

afeld:
Fixed the issue when run locally, and ran successfully on a new copy of the production database.

afeld:
Well, just deployed this to production accidentally (oy), so I'm just gonna merge and walk away from my computer.

PR #588, opened 2015-09-07
Model validations

istanfield09:
[Relevant Pivotal Tracker story.](https://www.pivotaltracker.com/story/show/102723774)

Is it correct that some of these validations should `allow_blank`, or should I amend the tests in order to get them to pass?

Thanks!

### DIFF
--- a/app/models/api_token.rb
+++ b/app/models/api_token.rb
@@ -9,8 +9,6 @@ class ApiToken < ActiveRecord::Base
 
   # TODO validates :access_token, presence: true
   validates :approval_id, presence: true
-  validates :access_token, presence: true, allow_blank: true, length: { minimum: 0, maximum: 255 }
-  validates :user, presence: true, allow_blank: true
 
   scope :unexpired, -> { where('expires_at >= ?', Time.now) }
   scope :expired, -> { where('expires_at < ?', Time.now) }
--- a/app/models/api_token.rb
+++ b/app/models/api_token.rb
@@ -9,6 +9,8 @@ class ApiToken < ActiveRecord::Base
 
   # TODO validates :access_token, presence: true
   validates :approval_id, presence: true
+  validates :access_token, presence: true, allow_blank: true, length: { minimum: 0, maximum: 255 }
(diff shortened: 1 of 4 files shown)


app/models/api_token.rb
```diff
@@ -9,6 +9,8 @@ class ApiToken < ActiveRecord::Base
 
   # TODO validates :access_token, presence: true
   validates :approval_id, presence: true
+  validates :access_token, presence: true, allow_blank: true, length: { minimum: 0, maximum: 255 }
+  validates :user, presence: true, allow_blank: true
```
afeld: I think we want to require the association with the user...?

app/models/approval.rb
```diff
@@ -16,6 +16,7 @@ class Approval < ActiveRecord::Base
 
   scope :individual, -> { where(type: 'Approvals::Individual') }
 
+  validates :proposal, presence: true, allow_blank: true
```
afeld: @cmc333333 Will all Approvals be associated with a Proposal?

Also, I think `presence: true, allow_blank: true` is basically canceling itself out :wink:

app/models/observation.rb
```diff
@@ -4,6 +4,9 @@ class Observation < ActiveRecord::Base
   belongs_to :proposal
   belongs_to :user
 
+  validates :proposal, presence: true 
+  validates :user, presence: true, allow_blank: true
```
afeld: Shouldn't `allow_blank`.

app/models/approval.rb
cmc333333: At this point in time, yep. Hypothetically, we might want to create some sort of "model" approval chain in the future, but that's a bit away.

istanfield09: Haha thanks. I was a little confused writing some of these, and figured allowing blank would be defeating the purpose. If I write a validation, and it causes the test to fail with the reason being that `:proposal` cant be blank, should I change the test?

afeld: Probably...? Might need to create associated models in the fixtures by default or something.

PR #589, opened 2015-09-08
do not notify user with email when they update a request

pkarman:
Relevant story: https://www.pivotaltracker.com/story/show/100957216

I'm not convinced that this doesn't affect *too many* notification cases, despite that tests are passing (including new test).

### DIFF
--- a/lib/ncr_dispatcher.rb
+++ b/lib/ncr_dispatcher.rb
@@ -28,7 +28,10 @@ class NcrDispatcher < LinearDispatcher
     }
 
     proposal.observers.each{|observer|
-      next if modifier and observer.id == modifier.id # https://www.pivotaltracker.com/story/show/100957216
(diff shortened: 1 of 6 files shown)


lib/ncr_dispatcher.rb
```diff
@@ -28,6 +28,7 @@ def on_proposal_update(proposal)
     }
 
     proposal.observers.each{|observer|
+      next if modifier and observer.id == modifier.id  # https://www.pivotaltracker.com/story/show/100957216
```
afeld: Mind adding a comment to the effect of "don't notify the person who triggered the notification"?

afeld:
Tests failed :cry:

pkarman:
The lone failing test was comparing arrays that assumed a specific order, when the order really didn't matter. The rspec fix was just to use match_array() rather than eq() for comparison.

PR #590, opened 2015-09-08
cancellation emails

pkarman:
Story: https://www.pivotaltracker.com/story/show/100733040

Only send cancellation emails to non-pending approvers.

### DIFF
--- a/app/models/approval.rb
+++ b/app/models/approval.rb
@@ -15,8 +15,6 @@ class Approval < ActiveRecord::Base
   belongs_to :parent, class_name: 'Approval'
   has_many :child_approvals, class_name: 'Approval', foreign_key: 'parent_id'
 
-  belongs_to :user
-
   scope :individual, -> { where(type: 'Approvals::Individual') }
 
 
--- a/app/models/approval.rb
+++ b/app/models/approval.rb
@@ -15,6 +15,8 @@ class Approval < ActiveRecord::Base
   belongs_to :parent, class_name: 'Approval'
   has_many :child_approvals, class_name: 'Approval', foreign_key: 'parent_id'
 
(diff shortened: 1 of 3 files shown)


lib/dispatcher.rb
```diff
@@ -36,8 +36,10 @@ def deliver_new_proposal_emails(proposal)
   end
 
   def deliver_cancellation_emails(proposal)
-    proposal.approvers.each do |approver|
-      CommunicartMailer.cancellation_email(approver.email_address, proposal).deliver_later
+    proposal.approvals.each do |approval|
+      next if approval.pending?  # https://www.pivotaltracker.com/story/show/100733040
```
afeld: Mind adding a comment of something like "don't notify people who haven't heard about the proposal yet"?

lib/dispatcher.rb
```diff
@@ -36,8 +36,10 @@ def deliver_new_proposal_emails(proposal)
...
+      next unless approval.user
```
afeld: Maybe loop through `proposal.individual_approvals` instead?

app/models/approval.rb
```diff
@@ -15,6 +15,8 @@ class Approval < ActiveRecord::Base
   belongs_to :parent, class_name: 'Approval'
   has_many :child_approvals, class_name: 'Approval', foreign_key: 'parent_id'
 
+  belongs_to :user
```
afeld: Don't think we want this in this model...see comment below.

afeld:
As an overall thought: this kind of problem has come up a bunch of times, so maybe we should centralize the logic of "these are the people who have been notified about the proposal already." Thoughts? Not a blocker for this PR.

PR #594, opened 2015-09-09
allow for MAX_AMOUNT and MIN_AMOUNT to be set with env vars

pkarman:
story: https://www.pivotaltracker.com/story/show/101608800

Since the story asks for amount to change on a particular date, refactor code to allow for overriding defaults with env vars, so deployment can be async with code merge.

### DIFF
--- a/app/models/gsa18f/procurement.rb
+++ b/app/models/gsa18f/procurement.rb
@@ -5,7 +5,7 @@ module Gsa18f
   end
 
   DATA = YAML.load_file("#{Rails.root}/config/data/18f.yaml")
-
+  MAX_AMOUNT = (ENV['GSA18F_MAX_AMOUNT'] || 3000).to_f
 
   class Procurement < ActiveRecord::Base
     URGENCY = DATA['URGENCY']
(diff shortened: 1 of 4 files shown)


afeld:
Strong feelings about this being an environment variable, rather than having a conditional in the code that changes it automatically (and that we can remove later)? I'm just afraid of forgetting to change it at the right time – that requires us to re-deploy that night.

@phirefly Are the values always going to be the same for 18F and NCR? If so, maybe we should just make a shared constant.

pkarman:
No strong feelings about env var per se. I was trying to come up with a way to merge this code to master in advance of the Oct 1 deadline. It felt awkward to me to hardcode a date-specific conditional in the code, since that would need to be cleaned up (as you indicate) sometime after Oct 1.

I was assuming that it was possible to set an env var with CF w/o deploying, just restarting the container(s), similar to how heroku works. But yes, would need to remember to do it. Seemed like there were other FY issues that needed to be remembered (e.g. https://www.pivotaltracker.com/story/show/101217754) so maybe we'd just plan a Sept 30 deploy in any case?

pkarman:
My real preference would be to store this kind of thing in the db, with a new **clients** table, where we would start to accumulate client-specific configuration. Since env vars seem to be the precedent for this kind of control, though, I went that route. Not at all wedded to it.

pkarman:
Story has been de-prioritized.

PR #595, opened 2015-09-09
[Delivers #103020810] randomize timezone in tests, fix failing test that assumed ET

pkarman:
story: https://www.pivotaltracker.com/story/show/103020810

Seems to be a regression from earlier timezone work. Adds Zonebie gem to randomize timezone on every rspec run, and fixes test that assumed ET.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -71,6 +71,7 @@ group :test do
   gem 'simplecov'
   gem 'timecop'
   gem 'turnip'
+  gem 'zonebie'
 
   # For better test reporting in CircleCI
   # http://blog.circleci.com/announcing-detailed-test-failure-reporting/
--- a/spec/helpers/value_helper_spec.rb
+++ b/spec/helpers/value_helper_spec.rb
@@ -21,9 +21,11 @@ describe ValueHelper do
     end
 
     it "converts time to relative time when specified" do
-      date = DateTime.new(2015,8,15,4,5,6).in_time_zone("Eastern Time (US & Canada)").strftime("%b %-d, %Y at %l:%M%P")
+      date = DateTime.new(2015,8,15,4,5,6).in_time_zone
+      date_str = date.strftime("%b %-d, %Y at %l:%M%P")
+      puts "Time.zone==#{Time.zone}"
       relative_date = time_ago_in_words(date)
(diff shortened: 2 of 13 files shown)


spec/helpers/value_helper_spec.rb
```diff
@@ -21,9 +21,11 @@
     end
 
     it "converts time to relative time when specified" do
-      date = DateTime.new(2015,8,15,4,5,6).in_time_zone("Eastern Time (US & Canada)").strftime("%b %-d, %Y at %l:%M%P")
+      date = DateTime.new(2015,8,15,4,5,6).in_time_zone
+      date_str = date.strftime("%b %-d, %Y at %l:%M%P")
+      puts "Time.zone==#{Time.zone}"
```
cmc333333: Mind removing the `puts`?

spec/helpers/value_helper_spec.rb
```diff
@@ -21,9 +21,11 @@
...
       relative_date = time_ago_in_words(date)
-      expect(helper.date_with_tooltip(date,true)).to eq('<span title="Aug 15, 2015 at 12:05am">' + relative_date + ' ago</span>')
+      expect(helper.date_with_tooltip(date,true)).to eq('<span title="' + date_str + '">' + relative_date + ' ago</span>')
```
cmc333333: I'm not sure this is testing the right thing -- as `date_with_tooltip` calls `strftime`, calling it here doesn't help us much, hey?

pkarman:
thanks @cmc333333, puts zapped.

cmc333333:
Nothing serious holding this back -- if the `to_s` is critical, lmk and I'll merge

pkarman:
I think the `to_s` only affects the tests, since that is where DateTime is being used instead of Time. Everything I've read suggests those 2 classes are interoperable but the code was explicitly checking for `Time`.

pkarman:
the failing tests from circleci are troubling. taking this back for more investigation.

pkarman:
This should be ready to merge now that I got all tests passing again. They were failing anytime the random timezone was +UTC since that resulted in records being created in the "previous year".

Gemfile
```diff
@@ -71,6 +71,7 @@ group :test do
   gem 'simplecov'
   gem 'timecop'
   gem 'turnip'
+  gem 'zonebie'
```
afeld: Awesome gem, thanks @alindeman! :us:

PR #599, opened 2015-09-10
[Delivers #100957216] do not email approver if they modify a request

pkarman:
story: https://www.pivotaltracker.com/story/show/100957216

filter out any approver from email notification if they are the user modifying a request.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,8 +13,6 @@ Style/DotPosition:
   Enabled: false
 Style/EmptyElse:
   Enabled: false
-Style/EmptyLinesAroundBody:
-  Enabled: false
 Style/GuardClause:
   Enabled: false
 Style/IfUnlessModifier:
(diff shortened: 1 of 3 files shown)


pkarman:
These codeclimate complaints seem wrong to me, but not sure how to quiet them.

cmc333333:
Should the requester get notified?

Regarding Code Climate -- there's a complaint about existing code and one about the new lines. Would you mind reducing the complexity of that function (fixing our mistakes :) )? As for the spaces, could they be removed if using `unless` or a filter? I agree that the spaces make the `next` more prominent -- perhaps that means we shouldn't use `next`, though?

pkarman:
The spaces are there on purpose, along with the (identical) comments, to increase legibility. Code climate seems to treat extra blank lines as naughty. Maybe it's just my old eyes, but I like more, rather than less, whitespace, especially when there are blocks of comments.

`Should the requester get notified?` I don't know. That seems outside the scope of the story though; the story asks that the modifier *not* get notified.

pkarman:
@cmc333333 that make it easier to read?

cmc333333:
Thanks @pkarman !

PR #600, opened 2015-09-10
[Delivers #101291648] Notify users when attachments are added to a proposal

annalee:
Currently, we do not send an email if an attachment is uploaded to a request, meaning users have to post a comment or manual email to let people know when they add an attachment.

This introduces an automatic email to all users associated with a proposal* when someone adds an attachment.

*As with cancellation emails, pending approvers are not emailed.

Paired with @afeld


app/views/communicart_mailer/new_attachment_email.html.haml
```diff
@@ -0,0 +1,9 @@
+%div
+  %p
+    Hello,
+    %br
+    An attachment has been added to #{@proposal.name} (#{@proposal.public_id}).
+
+%p
+  %strong
+    = link_to "View This Request", proposal_url(@proposal), {target: 'C2'}
```
afeld: @phirefly Any strong feelings about what the email should say? We're currently sending this same thing to everyone. Thinking it should at least have the name of (and a link to) the attachment, but @annalee suggested that we might want to attach it to the email directly.

afeld:
* [ ] Add plain text email template

app/views/communicart_mailer/new_attachment_email.html.haml
phirefly: The only thing that I would include is the title of the document. Can we add a line break after the "An attachment has been added" sentence and link to the doc using the attachment filename?

lib/dispatcher.rb
```diff
@@ -35,6 +35,15 @@ def deliver_new_proposal_emails(proposal)
     self.email_sent_confirmation(proposal)
   end
 
+  def deliver_attachment_emails(proposal)
+    proposal.users.each do |user|
+      # do not send email to approvers who have not yet heard about the proposal
+      approval = proposal.approvals.find_by(user_id: user.id)
+      next if approval && approval.pending?
```
cmc333333: Can we merge this with the two or three other ways to determine which users should be notified? I'm thinking of the methods on comments and on updates to a proposal.

afeld: Maybe a `Proposal#previously_notified_users` method or something.

cmc333333:
I'm going to take the non-response as a negative to "unless you want to address" ;)

PR #602, opened 2015-09-10
[Delivers #102267428] daily budget report

pkarman:
Story: https://www.pivotaltracker.com/story/show/102267428


app/models/ncr/work_order.rb
```diff
@@ -193,19 +193,55 @@ def system_approver_emails
     end
 
     def self.ba61_tier1_budget_mailbox
-      ENV['NCR_BA61_TIER1_BUDGET_MAILBOX'] || 'communicart.budget.approver@gmail.com'
+      users = User.with_role('BA61_tier1_budget_approver').where(client_slug: 'ncr')
+      if !users or users.length == 0
```
afeld: Would `users` ever be `nil`?

app/models/ncr/work_order.rb
```diff
@@ -193,19 +193,55 @@ def system_approver_emails
...
+        user = User.with_email_role_slug!(
+          ENV['NCR_BA61_TIER1_BUDGET_MAILBOX'] || 'communicart.budget.approver@gmail.com',
```
afeld: Not necessarily a blocker for this PR, but if that role already exists this environment variable gets ignored, which could be confusing.

afeld: In other words, creating the record in this method doesn't *quite* feel right. Having required records in databases is strange to me...hence my previous leanings to environment variables :wink:

afeld: Maybe worth moving into `db/seeds.rb`?

http://guides.rubyonrails.org/active_record_migrations.html#migrations-and-seed-data

app/models/ncr/work_order.rb
pkarman: good point. legacy paranoia from my perl days, checking for various flavors of false-ness

app/models/ncr/work_order.rb
pkarman: good point. I need to write a migration that creates that UserRole and then zap that code.

pkarman: Thanks for the reminder about seeds -- need for bootstrapping new checkout.

app/models/user.rb
```diff
@@ -21,27 +22,43 @@ class User < ActiveRecord::Base
...
     end
-    user_role = UserRole.new(role: role)
-    roles << user_role
+    role.users
+  end
+
+  def self.sql_for_role_slug(role, slug)
+    sql = "SELECT u.id FROM users AS u INNER JOIN user_roles AS ur ON u.id=ur.user_id INNER JOIN roles AS r ON r.id=ur.role_id "
+    sql += "WHERE r.name=#{self.sanitize(role)} and u.client_slug=#{self.sanitize(slug)}"
+    sql
```
afeld: This can be simplified to

```ruby
User.joins(:roles).where(client_slug: slug, roles: {name: role}).to_sql
```

db/migrate/20150911195321_add_budget_roles.rb
```diff
@@ -0,0 +1,24 @@
+class AddBudgetRoles < ActiveRecord::Migration
+  def change
+    User.with_email_role_slug!(
```
afeld: Can we not use the model method in the migration? If the method name changes or goes away, the migration breaks.

lib/ncr/reporter.rb
```diff
@@ -21,18 +21,18 @@ def self.ba80_proposals
       budget_proposals("BA80", 1.week.ago)
     end
 
-    def self.proposals_pending_approving_official
+    def self.proposals_pending_approving_official(approval_status = 'actionable')
```
afeld: Is the new argument used anywhere? If not, let's change it back.

lib/ncr/reporter.rb
```diff
@@ -21,18 +21,18 @@ def self.ba80_proposals
...
       # TODO convert to SQL
       Proposal.pending
               .where(client_data_type: 'Ncr::WorkOrder')
-              .select{ |p| p.individual_approvals.pluck(:status)[0] == 'actionable' }
+              .select{ |p| p.individual_approvals.pluck(:status)[0] == approval_status }
     end
 
-    def self.proposals_pending_budget
+    def self.proposals_pending_budget(approval_status = 'actionable')
```
afeld: Ditto.

app/models/ncr/work_order.rb
```diff
@@ -193,19 +193,35 @@ def system_approver_emails
     end
 
     def self.ba61_tier1_budget_mailbox
-      ENV['NCR_BA61_TIER1_BUDGET_MAILBOX'] || 'communicart.budget.approver@gmail.com'
+      users = User.with_role('BA61_tier1_budget_approver').where(client_slug: 'ncr')
+      if users.length == 0
```
afeld: Minor: you can use `users#empty?`.

app/models/ncr/work_order.rb
```diff
@@ -193,19 +193,35 @@ def system_approver_emails
...
+        raise "Missing User with role BA61_tier1_budget_approver -- did you run rake db:migrate and rake db:seed?"
+      end
+      users.first.email_address
```
afeld: Since all of these methods do basically the same thing, can we share the logic somewhere?

db/migrate/20150911195321_add_budget_roles.rb
```diff
@@ -0,0 +1,24 @@
...
+      'BA80_budget_approver',
+      'ncr'
+    )
+    User.with_email_role_slug!(
+      ENV['NCR_OOL_BA80_BUDGET_MAILBOX'] || 'communicart.budget.approver@gmail.com',
+      'OOL_BA80_budget_approver',
+      'ncr'
+    )
+  end
+end
```
afeld: Do we need this here *and* in the seeds file?

lib/ncr/reporter.rb
```diff
@@ -43,5 +43,23 @@ def self.budget_proposals(type, timespan)
...
+
+      approver_sql = "SELECT a.proposal_id FROM approvals AS a WHERE a.status='actionable' AND a.user_id IN (#{tier_one_sql})"
+
+      work_order_sql = "SELECT id FROM ncr_work_orders AS nwo "
+      work_order_sql += "WHERE nwo.org_code!='#{Ncr::Organization::WHSC_CODE}' AND nwo.expense_type IN ('BA60','BA61')"
+
+      sql = "SELECT * FROM proposals AS p "
+      sql += "WHERE p.status='pending' AND p.client_data_type='Ncr::WorkOrder' "
+      sql += "AND p.client_data_id IN (#{work_order_sql}) AND p.id IN (#{approver_sql})"
+      sql
```
afeld: Not a blocker, but I think we can dramatically condense this query.

app/models/user.rb
afeld: *Updated to only select the IDs.*

lib/ncr/reporter.rb
afeld: Also, FYI that you can do multi-line SQL with heredocs.

```ruby
<<-SQL
  SELECT * FROM proposals AS p
  WHERE p.status='pending'
  AND p.client_data_type='Ncr::WorkOrder'
  AND p.client_data_id IN (#{work_order_sql})
  AND p.id IN (#{approver_sql})
SQL
```

db/migrate/20150911195321_add_budget_roles.rb
pkarman: yes. we need to migrate existing data (dev, stage, prod) and also provide base fixtures for a clean checkout.

If `db:seed` were part of a regular deploy, we wouldn't need both. But afaik, we never seed an existing, working db.

db/migrate/20150911195321_add_budget_roles.rb
pkarman: does this comment apply to the db/seeds.rb file as well? The method is also used there, and run on every rspec.

How best to DRY with code/logic used in migrations and elsewhere?

pkarman:
@afeld updated -- please review

db/migrate/20150911195321_add_budget_roles.rb
afeld: The best approach may be to have the logic under `lib/` called from both places...the benefit being that you can easily write tests for it.

In past projects, any time required seed data has been added, I've just seen an announcement sent to all of the developers to "make sure to run `rake db:seed`". Having it in a migration is probably better because it's enforced by Rails, but has the downside of being inadvertently broken. See also:

https://twitter.com/aidanfeldman/status/643550323538313216

db/migrate/20150911195321_add_budget_roles.rb
afeld: > provide base fixtures for a clean checkout

That's done as part of the [bootstrap](https://github.com/18F/C2/blob/36842b72024c5b387a7f96c8dbbabbcfa6f6ed5d/script/bootstrap#L13).

> If `db:seed` were part of a regular deploy

We could certainly add this, but would be nice if we didn't need those environment variables set indefinitely.

> we never seed an existing, working db

We don't, but that's because we don't currently have any seed data. If we do, ideally running `rake db:seed` should be idempotent (by using `find_or_create_by`).

db/migrate/20150911195321_add_budget_roles.rb
pkarman: I expect that a migration will only ever run once in any given environment, and that `db:seed` would never be run on an existing, populated database. If the garden is already growing, you don't seed it. It's for starting from scratch (as when running the rspec suite, which runs DatabaseCleaner and then loads the seeds to initialize a clean db instance on each run).

IME there are two kinds of migrations: (1) schema and (2) data. Most often it is (1) as we alter the db definitions, but occasionally it's (2) as we need to move existing data from one structure to another as the architecture changes.

I'd be happy to pull the relevant method out of `app/model` and put it in `lib/` if that's the objection here.

pkarman: Because "I expect that a migration will only ever run once in any given environment," I also expect that the environment variables are really only relevant in production, where they can be unset after the migration has run.

pkarman: An alternative to putting this data migration in `db/migrations` would be to put it in `lib/tasks` where it would then need to be run manually as part of a deployment. I'm also fine with that, it just means one extra one-time step on deployment.

bipartisanMas0n:
Tier Definition: Yes Correct, TIER 1 PREVIOUS PENDING RECORDS DAMAGE THE HOST'S INTERIOR OF SURVIVAL, BY POKING PIN-TIP SIZE HOLES IN A MORAL-TIER: 1, AND UP IN COMBATIVE RESPONSE...due to the numerous pending requests back listed.

app/models/ncr/work_order.rb
```diff
@@ -193,19 +197,27 @@ def system_approver_emails
...
 
     def self.ba61_tier2_budget_mailbox
-      ENV['NCR_BA61_TIER2_BUDGET_MAILBOX'] || 'communicart.ofm.approver@gmail.com'
+      self.approver_with_role('BA61_tier2_budget_approver')
+    end
+
+    def self.approver_with_role(role_name)
+      users = User.with_role(role_name).where(client_slug: 'ncr')
+      if users.empty?
+        fail "Missing User with role #{role_name} -- did you run rake db:migrate and rake db:seed?"
```
phirefly: I'm wondering if there's a way to easily see which roles are required at a higher level. In this case, the `BA61_tier2_budget_approver` is required but that can easily be missed.

pkarman: by a "higher level" do you mean when grokking the code? Or something programmatic when calling that method?

Check out `db/seeds` for example.

phirefly: Moreso the former. Perhaps I'm looking for an equivalent of what we do in env.rb, where we indicate optional and required env vars. We no longer have that visibility. Not a huge deal for this PR of course, but just something that I'm thinking about now that we've moved it out of the codebase. Maybe it's something we surface into the user management admin.

pkarman: ah ok. I think `db/seeds.rb` is the easiest place to grok what data is now being created with a "fresh" database (as on every rspec).

app/mailers/report_mailer.rb
```diff
@@ -5,10 +5,27 @@ def budget_status
...
+    build_attachments
+
     mail(
       to: to_email,
       subject: "C2: Daily Budget report for #{date}",
       from: self.sender_email
     )
   end
+
+  # rubocop:disable all
```
afeld: That's cheating! In all seriousness, we have Rubocop enabled expressly for the purpose of keeping us honest. Please don't disable!

afeld: Minor: Also, can we make it a `protected` method?

db/seeds.rb
```diff
@@ -5,3 +5,28 @@
 #
 #   cities = City.create([{ name: 'Chicago' }, { name: 'Copenhagen' }])
 #   Mayor.create(name: 'Emanuel', city: cities.first)
+
+Role.find_or_create_by(name: 'observer')
+Role.find_or_create_by(name: 'client_admin')
+Role.find_or_create_by(name: 'admin')
+
+User.with_email_role_slug!(
+  ENV['NCR_BA61_TIER1_BUDGET_MAILBOX'] || 'communicart.budget.approver@gmail.com',
```
afeld: If the environment variable isn't set on production, is the desired behavior that it uses this test email address? Don't want to hold up this PR any more, but worth thinking about longer-term.

afeld:
With #620 in, cool with removing the migration and undoing the change to the `schema.rb`? Otherwise, feel free to :ship: when the build passes.

app/mailers/report_mailer.rb
```diff
@@ -5,10 +5,29 @@ def budget_status
...
+    date = Time.now.utc.strftime('%Y-%m-%d')
+    { 'approved-ba60-week' => Ncr::Reporter.as_csv(Ncr::Reporter.ba60_proposals),
+      'approved-ba61-week' => Ncr::Reporter.as_csv(Ncr::Reporter.ba61_proposals),
+      'approved-ba80-week' => Ncr::Reporter.as_csv(Ncr::Reporter.ba80_proposals),
+      'pending-at-approving-official' => Ncr::Reporter.as_csv(Ncr::Reporter.proposals_pending_approving_official),
+      'pending-at-budget' => Ncr::Reporter.as_csv(Ncr::Reporter.proposals_pending_budget),
+      'pending-at-tier-one-approval' => Ncr::Reporter.as_csv(Ncr::Reporter.proposals_tier_one_pending),
+    }.each do |name, csv|
+      attachments[name + '-' + date + '.csv'] = csv
+    end
```
afeld: This can be simplified by doing the CSV conversion in the loop:

```ruby
{ 'approved-ba60-week' => Ncr::Reporter.ba60_proposals,
  'approved-ba61-week' => Ncr::Reporter.ba61_proposals,
  'approved-ba80-week' => Ncr::Reporter.ba80_proposals,
  'pending-at-approving-official' => Ncr::Reporter.proposals_pending_approving_official,
  'pending-at-budget' => Ncr::Reporter.proposals_pending_budget,
  'pending-at-tier-one-approval' => Ncr::Reporter.proposals_tier_one_pending,
}.each do |name, proposals|
  csv = Ncr::Reporter.as_csv(proposals)
  attachments[name + '-' + date + '.csv'] = csv
end
```

app/mailers/report_mailer.rb
```diff
@@ -5,10 +5,29 @@ def budget_status
...
     mail(
       to: to_email,
       subject: "C2: Daily Budget report for #{date}",
       from: self.sender_email
     )
   end
+
+  private
+
+  # rubocop:disable AbcSize, TrailingComma
```
afeld: Feel free to disable the `TrailingComma` check in the `.rubocop.yml`, but please fix the `AbcSize` issue!

lib/ncr/reporter.rb
```diff
@@ -21,6 +21,25 @@ def self.ba80_proposals
       budget_proposals("BA80", 1.week.ago)
     end
 
+    # rubocop:disable all
```
afeld: Please don't do this!

lib/ncr/reporter.rb
```diff
@@ -43,5 +62,34 @@ def self.budget_proposals(type, timespan)
               .where('created_at > ?', timespan)
               .select { |pr| pr.client_data.expense_type == type }
     end
+
+    # rubocop:disable all
```
afeld: Or this!

app/mailers/report_mailer.rb
afeld: You might consider breaking the Hash out to an `attachments_data` method or something.

lib/ncr/reporter.rb
```diff
@@ -43,5 +62,34 @@ def self.budget_proposals(type, timespan)
...
+    def self.proposals_tier_one_pending_sql
+      tier_one_sql = User.sql_for_role_slug('BA61_tier1_budget_approver', 'ncr')
+
+      approver_sql = <<-SQL.gsub(/^ {8}/, '')
```
afeld: FYI that you can use [`strip_heredoc`](http://api.rubyonrails.org/classes/String.html#method-i-strip_heredoc). Not sure that it's necessary for SQL, though...

db/seeds.rb
pkarman: When this was a migration, it didn't matter, because it only ran once, when the env vars were correctly defined.

Now that it runs on every deploy via seeds, yes you are correct, the env vars must continue to be correctly set in production or we will get test users for production budget approvers. That sort of defeats the purpose of moving away from env vars, if they must continue to be set indefinitely just for deploys to work.

pkarman: My 2c: revert the change that runs db:seed on every deploy, and restore the migration. That fixes production once and only once, and allows the tests to continue working using the test addresses.

afeld: > When this was a migration, it didn't matter, because it only ran once

Based on all of the feedback in https://twitter.com/aidanfeldman/status/643550323538313216 and https://github.com/testdouble/feedback/issues/7, I'm leaning towards not thinking of migrations in that way. Also, almost everyone in those conversations advised not referencing models from migrations.

/cc https://github.com/18F/C2/pull/607 

---

What if we only create the records in `seeds` when the environment variables are set (or maybe use the defaults in non-production?), and print out a warning otherwise?

pkarman: let's move the old migration to a one-time rake task then, to avoid polluting the db/migration space with model-speak.

Making `db/seeds.rb` Rails.env aware seems to defeat the purpose, imo. I would rather we didn't run it on deploy (for existing dbs) and only used it for seeding a clean, new database. That would be the least surprising thing, imo. As described here, that's what db:seed is for:
http://www.xyzpub.com/en/ruby-on-rails/3.2/seed_rb.html
```
With the file db/seeds.rb, the Rails gods have given us a way of feeding default values easily and quickly to a fresh installation.
```

afeld: > let's move the old migration to a one-time rake task then

:+1: Moving it out of this file too, yes?

afeld: > Making `db/seeds.rb` Rails.env aware seems to defeat the purpose

_Meh._

pkarman: see #630 and c628b7d

afeld: In case I haven't mentioned, my top pet peeve is duplication; I feel more strongly that it should only be in one place than where that place is.

lib/tasks/convert.rake
```diff
@@ -0,0 +1,27 @@
...
+        'ncr'
+      )
+      User.with_email_role_slug!(
+        ENV['NCR_OOL_BA80_BUDGET_MAILBOX'] || 'communicart.budget.approver@gmail.com',
+        'OOL_BA80_budget_approver',
+        'ncr'
+      )
+    end
+  end
+end
```
afeld: If this is required locally and we _don't_ keep it in `seeds` or a migration, we should run the rake command when [bootstrapping](https://github.com/18F/C2/blob/master/script/bootstrap).

pkarman: how about d74a81c4b806c076ee19f385c550a241dbef0606

pkarman:
5fd983273e3b85b425a3e3531a3d494027575a0e should allow `db:seed` to run multiple times in the same environment with and w/o environment variables.

lib/roles_conversion.rb
```diff
@@ -0,0 +1,40 @@
+class RolesConversion
```
afeld: Why "conversion"?

app/models/user.rb
```diff
@@ -80,6 +89,38 @@ def self.for_email(email)
...
+    end
+
+    # unique triple -- check if any other user with role+slug already exists
+    return if self.exists_with_role_slug?(role, slug)
+
+    user.client_slug = slug
+    user.add_role(role)
+    user.save!
+    user
+  end
```
afeld: Can we move these to the `RoleConversation` class, since that's the only place they are used?

app/models/user.rb
```diff
@@ -80,6 +89,38 @@ def self.for_email(email)
...
+    return false unless the_role
+    triple_exists = false
+    the_role.users.each do |u|
+      if u.client_slug == slug
+        triple_exists = true
+        break
+      end
+    end
+    triple_exists
+  end
```
afeld: Would it make sense to do this as a validation on the `Role`? (Legitimately asking.)

lib/roles_conversion.rb
pkarman: I was trying to avoid the word "migration".

The class represents one-time transformation of data storage from environment variable to User-with-Role. I guess it could be `RolesTransformer` or `RolesWithButter` or something else innocuous. I imagine it might be used again when transforming GSA18F_PURCHASER_EMAIL and GSA18F_APPROVER_EMAIL in a similar way. I am not attached to what it is called.

app/mailers/report_mailer.rb
```diff
@@ -5,10 +5,31 @@ def budget_status
...
+      'pending-at-approving-official' => Ncr::Reporter.as_csv(Ncr::Reporter.proposals_pending_approving_official),
+      'pending-at-budget' => Ncr::Reporter.as_csv(Ncr::Reporter.proposals_pending_budget),
+      'pending-at-tier-one-approval' => Ncr::Reporter.as_csv(Ncr::Reporter.proposals_tier_one_pending),
+    }
+  end
+
+  def build_attachments
+    date = Time.now.utc.strftime('%Y-%m-%d')
+    csv_reports.each do |name, csv|
+      attachments[name + '-' + date + '.csv'] = csv
```
afeld: Minor: how about doing the `Ncr::Reporter.as_csv` call in here so that you don't need it on every line above?

app/models/user.rb
pkarman: Sure, will do.

fwiw, my rationale for putting them in the model was that the model is where I tend to put convenience or utility methods that operate on a particular data model. If that's not the convention here, cool.

app/models/user.rb
```diff
@@ -80,6 +89,38 @@ def self.for_email(email)
...
+    the_role = role.is_a?(Role) ? role : Role.find_by_name(role)
+    return false unless the_role
+    triple_exists = false
+    the_role.users.each do |u|
+      if u.client_slug == slug
+        triple_exists = true
+        break
+      end
+    end
+    triple_exists
```
afeld: This last section can be simplified to

```ruby
the_role.users.exists?(client_slug: slug)
```

app/models/user.rb
pkarman: I'm failing to imagine what that might look like?

app/mailers/report_mailer.rb
pkarman: it was intentionally a (probably premature) optimization to avoid needing to keep each array of db object results around any longer than necessary, and instead convert them into a (presumably) smaller text string as quickly as possible.

But if it makes it easier to read, sure, I'll change.

app/models/user.rb
afeld: Also, can we add some tests for this method?

lib/tasks/convert.rake
```diff
@@ -0,0 +1,8 @@
+namespace :convert do
+  namespace :roles do
+    desc "migrate NCR env var budget approvers to Roles"
+    task ncr_budgets: :environment do
+      RolesConversion.ncr_budget_approvers
+    end
+  end
+end
```
afeld: Minor: Do we need this, since it's being run as part of the seeds?

afeld:
Phew! I think we're close. Thanks for being patient with all the back-and-forth on this.

lib/roles_conversion.rb
afeld: Well, it's really `NcrRoleCreation`, right? "Create if it doesn't exist." Anyway, not too worried about it.

app/models/user.rb
afeld: Not a blocker. Nevermind :wink:

app/mailers/report_mailer.rb
afeld: Well, ideally they would just be the Relations, but no worries.

pkarman: ah, that makes more sense. I keep forgetting how ActiveRecord defers fetching things till you actually *do* something with them.

pkarman:
thanks for all the thorough review. I think I've committed all the changes you noted.

afeld:
:hugs:

PR #569, opened 2015-08-28
[Delivers #101069160] surface paper_trail data through the UI

afeld:
Not amazing yet, but you get the idea.

screen shot 2015-08-28 at 3 26 32 am

PR #603, opened 2015-09-10
refactor version diff from template to decorator

afeld:
Refactoring attempt of #569.

This is still pretty massive. My next step would be to make individual decorator classes for the different types of messages (basically, one corresponding to each `when`), but that seems like it may be overkill. Would love to hear feedback/ideas.

/cc @cmc333333 @jessieay @monfresh


app/decorators/c2_version_decorator.rb
```diff
@@ -0,0 +1,75 @@
+class C2VersionDecorator < Draper::Decorator
```
jessieay: yeahhhh decoratorsss :+1: 

if there are already decorators in here using Draper, that makes total sense to continue on with. If not, I would consider using SimpleDelegator since I've probably ripped Draper out of projects more often than I've added it in 

http://hashrocket.com/blog/posts/using-simpledelegator-for-your-decorators

app/decorators/c2_version_decorator.rb
```diff
@@ -0,0 +1,75 @@
...
+  def to_html
```
jessieay: test for this method?

afeld:
Phew, ok, I think this is ready for another look.

spec/decorators/base_decorator_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+  def escape(str)
+    # equivalent to what's used in templates
+    ERB::Util.html_escape(str)
+  end
+
+  describe '.combine_html' do
+    it "escapes content" do
+      output = BaseDecorator.combine_html(["<script>alert()</script>"])
+      expect(escape(output)).to eq("&lt;script&gt;alert()&lt;/script&gt;")
+    end
```
cmc333333: Thanks for this test! It's unclear from the term `SafeBuffer` whether this was a buffer of safe entities or a buffer of entities which could be made safe. This test clears that up nicely.

cmc333333:
Nice to see this broken up into classes. Good job!

PR #605, opened 2015-09-14
[Delivers #100957216] notify requester on proposal change

pkarman:
It appears that requesters were only getting notified on comments, not changes, to a proposal. This PR adds code + test to verify requesters are getting notified when a proposal changes.


bipartisanMas0n:
NEED VETERAN PRACTITIONERS' CARD

PR #607, opened 2015-09-15
have CircleCI run migrations

afeld:
This pull request intentionally raises a question: do we want to ensure all past migrations continue to work indefinitely? If they are broken (such as referencing a class name that is removed), tests will fail, as seen here. See discussion in https://github.com/testdouble/feedback/issues/7 for more background.

/cc @cmc333333

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -2,8 +2,6 @@
 # https://github.com/testdouble/feedback/issues/7
 database:
   override:
-    - export RAILS_ENV="test"
-    - export RACK_ENV="test"
     - bundle exec rake db:create db:migrate --trace
 dependencies:
   pre:
--- a/circle.yml
+++ b/circle.yml
@@ -1,4 +1,10 @@
+# run all migrations
+# https://github.com/testdouble/feedback/issues/7
+database:
+  override:
+    - export RAILS_ENV="test"
+    - export RACK_ENV="test"
+    - bundle exec rake db:create db:migrate --trace
 dependencies:
   pre:
     - cp .env.example .env
-


bipartisanMas0n:
DO NOT BREAK FIRST MODEL CHAIN! systemically branch outward with another, seperate from lotto-affiliations / binding a vet non-profit charity players team based on 5+6=[+11] we'll run this data set more so if God grants me the rest of the week, later on ill teach a course to delegate player strengths and combative taunts based on skills and naval ranking.

bipartisanMas0n:
keep stats at exit migration and then we will layer on down the road per dependent in need of updated stats; FILES ATTAINABLE PER MIGRATE REMOVED SHELLS.

jessieay:
I think ideally we would write migrations that can run indefinitely...or at least for a while. There are some good strategies for this here: http://railsguides.net/change-data-in-migrations-like-a-boss/

That being said, not sure it makes sense to have Circle fail if we have migrations that can no longer run since our `bootstrap` script is running `rake db:setup` which does a schema load rather than a migrate

afeld:
We could certainly change the bootstrap script...strong feelings about how to proceed?

pkarman:
I'm -1 on this for a few reasons:

* Data migrations are intended to be run against an existing database. It's an integration exercise, not a unit test. There has to be something to migrate. Since CI always starts with a a fresh db instance, it's just practically the wrong place to be testing migrations.

* Migrations are ephemeral, one-off tasks, run sequentially. It's their sequential-ness that is important. They run once-per-environment. If at some future point (months, years, later) an old migration stops working, it could/should simply be deleted as irrelevant. That's why `db:migrate` always updates the `db/schema.rb` file when it runs, because the `db/schema.rb` file is the thing that really matters in the ongoing life of an app.

* I think the problem you're trying to solve is a valid problem: how do we mitigate risk when we introduce migrations. I think that should be part of the QA process, since it requires an existing database. The solution feels to me like a process solution, not a technical solution.

jessieay:
I think Peter's arguments are persuasive. Especially:

> I think the problem you're trying to solve is a valid problem: how do we mitigate risk when we introduce migrations. I think that should be part of the QA process, since it requires an existing database. The solution feels to me like a process solution, not a technical solution.

Let's make sure we write reversible migrations and define classes within migrations (or use raw SQL), but let's not fail CI when past migrations are no longer runnable.

afeld:
Ok, thanks folks.

PR #610, opened 2015-09-15
[Delivers #101608800] allow for MAX_AMOUNT and MIN_AMOUNT to be set with env vars

pkarman:
Story: https://www.pivotaltracker.com/n/projects/1149728/stories/101608800

See also previous comments on #594


pkarman:
going to re-approach this in a simpler way

afeld:
Moved to #614.

PR #611, opened 2015-09-15
add documentation for how to give a user admin authz

pkarman:


doc/production.md
```diff
@@ -53,3 +53,22 @@ We will be moving to Mandrill for sending transactional emails.
...
 You should now be signed in to Mandrill with the shared account.
+
+## Admin accounts
+
+18F developers can give admin access to users in the system. Here is an example:
+
+```
+# on localhost
+% cd /tmp
+% git checkout git@github.com:18F/C2.git
```
afeld: Need to `cd` into the folder.

afeld:
If we start to use roles for more things, this is enough of a pain that we should prioritize adding some admin UI. Might be as simple as adding one of the [off-the-shelf admin interfaces](https://www.ruby-toolbox.com/categories/rails_admin_interfaces).

pkarman:
+1 to admin UI. I have used activeadmin with good feelings.

PR #613, opened 2015-09-15
[Delivers #103357204] Feedback emails cc the feedback writer

annalee:
Tested it out on Dev:
c2-cc-screenshot

### DIFF
--- a/app/mailers/feedback_mailer.rb
+++ b/app/mailers/feedback_mailer.rb
@@ -7,6 +7,7 @@ class FeedbackMailer < ApplicationMailer
       to: self.class.support_email,
       subject: 'Feedback submission',
       from: from,
+      cc: from,
       body: message
     )
   end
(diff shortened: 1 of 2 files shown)


app/mailers/feedback_mailer.rb
```diff
@@ -7,6 +7,7 @@ def feedback(sending_user, form_values)
       to: self.class.support_email,
       subject: 'Feedback submission',
       from: from,
+      cc: from,
```
phirefly: I could be missing something here, but the `from` and `cc` are the same. On the test email sent to gatewaycommunicator, these values are different.

annalee: Yeah, I'm not sure where that from header is being overwritten so that it doesn't send from the user. I didn't want to change the from value in the bugfix because it wasn't necessary to actually resolve the issue, but we can flag it as a # FIXME to change that value so that it reflects the actual behavior of the application?

phirefly:
:shipit:

PR #614, opened 2015-09-15
[Delivers #101608800] amount limit tied to FY start

pkarman:
Story: https://www.pivotaltracker.com/story/show/101608800


pkarman:
need to fix tests

pkarman:
tests ok

app/models/ncr/work_order.rb
```diff
@@ -8,6 +8,13 @@ def self.table_name_prefix
 
   EXPENSE_TYPES = %w(BA60 BA61 BA80)
   BUILDING_NUMBERS = YAML.load_file("#{Rails.root}/config/data/ncr/building_numbers.yml")
+  FY16 = Time.zone.parse('2015-10-01')
```
phirefly: Should we share, since this is also being used with gsa18f?

pkarman: share that one line? `FY16 = ...` ?

the logic for setting MAX_AMOUNT?

or share the whole `validates` block?

if the latter, I think the only way to do that would be with a Concern mix-in. Right now there are 2 different validation messages (the default, and our custom one in `Ncr::WorkOrder`). Do we want to have the same message across both classes?

afeld: "Yes" to all of these things. Can call the Concern `PurchaseCardUseCase` or `PurchaseCardMixin` or `PurchaseCardRequest` or something.

app/models/concerns/purchase_card_mixin.rb
```diff
@@ -0,0 +1,33 @@
+module PurchaseCardMixin
```
afeld: I'd add a comment at the top saying "requires `.purchase_amount_column_name` method to be defined before `include`ing".

afeld:
Just one minor comment...otherwise :shipit:

pkarman:
@afeld thanks, commented added.

PR #615, opened 2015-09-15
[Delivers #102278486] archive link

pkarman:
Story: https://www.pivotaltracker.com/story/show/102278486

resurrect the archive link.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -48,7 +48,15 @@ class ApplicationController < ActionController::Base
   private
 
   def current_user
-    @current_user ||= User.find_or_create_by(email_address: session[:user]['email']) if session[:user] && session[:user]['email']
+    @current_user ||= find_current_user
+  end
+
+  def find_current_user
+    if ENV['FORCE_USER_ID']
+      User.find ENV['FORCE_USER_ID']
+    else
+      User.find_or_create_by(email_address: session[:user]['email']) if session[:user] && session[:user]['email']
+    end
   end
 
   def sign_in(user)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -48,15 +48,7 @@ class ApplicationController < ActionController::Base
   private
 
   def current_user
-    @current_user ||= find_current_user
-  end
-
-  def find_current_user
-    if ENV['FORCE_USER_ID']
-      User.find ENV['FORCE_USER_ID']
-    else
-      User.find_or_create_by(email_address: session[:user]['email']) if session[:user] && session[:user]['email']
-    end
+    @current_user ||= User.find_or_create_by(email_address: session[:user]['email']) if session[:user] && session[:user]['email']
   end
 
   def sign_in(user)
--- a/app/views/proposals/_archive_link.html.erb
+++ b/app/views/proposals/_archive_link.html.erb
@@ -1,5 +1,5 @@
 <div class="col-sm-2">
-  <%- if @approved_data.rows.size > @CLOSED_PROPOSAL_LIMIT %>
+  <%- unless @approved_data.rows.size < @CLOSED_PROPOSAL_LIMIT %>
     <%= link_to "View the archive", url_for(:controller => 'proposals', :action => 'archive') %>
   <%- end %>
 </div>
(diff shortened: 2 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -48,7 +48,15 @@ def peek_enabled?
   private
 
   def current_user
-    @current_user ||= User.find_or_create_by(email_address: session[:user]['email']) if session[:user] && session[:user]['email']
+    @current_user ||= find_current_user
+  end
+
+  def find_current_user
+    if ENV['FORCE_USER_ID']
```
phirefly: Can you talk about this? I'm not sure I follow why this is necessary.

pkarman: This is a development hack.

In order simulate what a user sees, w/o being able to authenticate as that user, you can set the `FORCE_USER_ID` environment variable to a User id value when starting the rails server. E.g.:
```
FORCE_USER_ID=1234 rails server
```

This is most useful in dev when using a db snapshot from production, when trying to re-create what a user is seeing. Since we can't authenticate as that user, we can mimic the login by forcing the server to treat each request as if the current_user is someone other than yourself.

afeld: > In order simulate what a user sees

Have you needed to do this regularly? Seems like the real issue here is a lack of admin tools on production.

>  in dev when using a db snapshot from production

Not sure that we should be making this typical practice...not sure how sensitive our clients are about their data. They might not be thrilled about copies of the database roaming around on our laptops.

app/views/proposals/_archive_link.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="col-sm-2">
-  <%- if @approved_data.rows.size > @CLOSED_PROPOSAL_LIMIT %>
+  <%- unless @approved_data.rows.size < @CLOSED_PROPOSAL_LIMIT %>
```
afeld: :facepalm: Nice catch!

afeld: Mind adding a test for this? Also, would be great to have this as a separate PR from the `FORCE_USER_ID` thing.

app/controllers/application_controller.rb
pkarman: I've used it on 3 different branches in the last 24 hours. It's very helpful ime.

Point well taken about data privacy however.

I think we really need some secure way to act-as-another-user, in order to debug authz rules, display issues, etc. I'm very open to suggestions. It's just a basic debugging/testing aide, imo, to be able to reproduce.

I can probably use/extend `lib/populator.rb` to do more of the dev work; for debugging, though, I'm not sure how best to solve this need.

afeld: Can we at least add a check for `!Rails.env.production?`. Ok to merge this in otherwise until we think of a better solution.

pkarman: there is, in `config/initializers/env_check.rb`

afeld: Mind also listing in [`.env.example`](https://github.com/18F/C2/blob/master/.env.example)?

config/initializers/env_check.rb
```diff
@@ -22,4 +22,8 @@ def get_keys(env_file)
   raise "Please set DEFAULT_URL_HOST"
 end
 
+if ENV['FORCE_USER_ID'] && Rails.env.production?
+  raise "FORCE_USER_ID not allowed in production environment"
+end
```
afeld: Hmm, don't love the idea of keeping the server from starting in production...

pkarman:
simplified to just the one line change -- please review

afeld:
* [ ] Add test for archive link appearing

Other changes moved to #619.

pkarman:
@afeld test added. thanks for the nudge.

spec/features/archive_link.rb
```diff
@@ -0,0 +1,21 @@
...
+      wo = FactoryGirl.create(:ncr_work_order, project_title: "Work Order #{i}")
+      wo.proposal.update(requester: user)
+      wo.proposal.individual_approvals.create!(user: approver, status: 'actionable')
+      approval = wo.proposal.existing_approval_for(approver)
+      approval.approve!
+      wo.proposal
+    end
+    visit '/proposals'
+    expect(page).to have_content('View the archive')
+  end
```
afeld: Can we add a test for "doesn't display archive link for fewer than 10 results"?

phirefly:
:shipit:

PR #616, opened 2015-09-16
[Delivers #99618664] pending proposal default sort

pkarman:
Story: https://www.pivotaltracker.com/story/show/99618664

There are a couple of unrelated test fixes in here as well.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -26,6 +26,7 @@ MYUSA_SECRET: consumer_secret_key
 # NOTIFICATION_FROM_EMAIL
 # PORT
 # RESTRICT_ACCESS=true
+# TEST_DATABASE_URL=postgresql://localhost/c2_test
 # WEB_CONCURRENCY
 
 # production only
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,6 +5,8 @@ AllCops:
   RunRailsCops: true
 Metrics/LineLength:
   Enabled: false
+Metrics/ClassLength:
+  Enabled: false
 Style/CommentAnnotation:
   Enabled: false
 Style/Documentation:
--- a/config/database.yml
+++ b/config/database.yml
@@ -13,5 +13,7 @@ development:
   database: c2_development
 test:
   <<: *default
-  url: postgresql://localhost/c2_test
+# use 'url' rather than 'database' so that DATABASE_URL env var is not merged in.
+# see URL at top of this file for docs.
+  url: <% ENV['TEST_DATABASE_URL'] || 'postgresql://localhost/c2_test' %>
 production: *default
--- a/config/database.yml
+++ b/config/database.yml
@@ -13,5 +13,5 @@ development:
   database: c2_development
 test:
   <<: *default
-  database: c2_test
+  url: postgresql://localhost/c2_test
 production: *default
--- a/lib/query/proposal/listing.rb
+++ b/lib/query/proposal/listing.rb
@@ -9,7 +9,7 @@ module Query
         @user = user
       end
 
-      def has_table_param?(name)
+      def table_param?(name)
         params[:tables] && params[:tables][name]
       end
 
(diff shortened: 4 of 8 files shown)


config/database.yml
```diff
@@ -13,5 +13,5 @@ development:
   database: c2_development
 test:
   <<: *default
-  database: c2_test
+  url: postgresql://localhost/c2_test
```
phirefly: Did you intend on committing this?

pkarman: yes very intentional.

right now, if you have a DATABASE_URL defined in your local .env file, it will override the test db name and nuke your dev db while tests are running. Not good.

afeld: Iiiinteresting. One might argue that is the expected behavior, but having blown away a production database that same way, very much a :+1: on being extra careful in preventing people from doing it.

.rubocop.yml
```diff
@@ -5,6 +5,8 @@ AllCops:
   RunRailsCops: true
 Metrics/LineLength:
   Enabled: false
+Metrics/ClassLength:
+  Enabled: false
```
afeld: Hmm, I'm actually a fan of this one. Open to ignoring Rubocop failures for this PR if you don't want to be sidetracked with refactoring a particular class when adding a single line to a class.

config/database.yml
afeld: Also, maybe worth an inline comment explaining the reason, since the Rails convention is to just have the `database` specified.

lib/tabular_data/container.rb
```diff
@@ -25,6 +26,12 @@ def rows
       results
     end
 
+    # rubocop:disable all
+    def rows=(new_rows)
+      @rows = new_rows
+    end
+    # rubocop:enable all
```
afeld: Just use [`attr_accessor :columns`](http://ruby-doc.org/core-2.0.0/Module.html#method-i-attr_accessor) at the top, and you get this method for free.

afeld:
Raised an overall question about this approach from a UI perspective in [the story](https://www.pivotaltracker.com/n/projects/1149728/stories/99618664).

config/database.yml
pkarman: I was referencing http://edgeguides.rubyonrails.org/configuring.html#connection-preference

What I've seen before is a TEST_DATABASE_URL env var supported in the `database.yml` explicitly, so that if you want to override the default in `.env` you can, but with a different var explicitly for automated tests.

I may do that instead, and add comments.

.rubocop.yml
pkarman: I discovered more than one class already exceeding the limit, so it appeared to me that we were not obeying that rule.

lib/tabular_data/container.rb
pkarman: I tried that first, and things blew up, I think because the assignment syntax was not expected. I.e. this:
```
container.rows( new_rows )  # ok
container.rows = new_rows  # not ok
```

.rubocop.yml
afeld: Oh, totally, we are breaking a _lot_ of our enforced rules:

https://codeclimate.com/github/18F/C2/issues

Luckily Code Climate only fails the build for style rules in lines that you change. I've only been disabling rules that I disagree with or don't think are important.

config/database.yml
```diff
@@ -13,5 +13,7 @@ development:
   database: c2_development
 test:
   <<: *default
-  database: c2_test
+# use 'url' rather than 'database' so that DATABASE_URL env var is not merged in.
+# see URL at top of this file for docs.
+  url: <% ENV['TEST_DATABASE_URL'] || 'postgresql://localhost/c2_test' %>
```
afeld: :+1:

lib/query/proposal/listing.rb
```diff
@@ -9,9 +9,42 @@ def initialize(user, params, proposals = ::Proposal.all)
...
+      end
+
       def pending
-        self.index_visible_container(:pending).alter_query(&:pending)
+        self.sort_pending_container(self.index_visible_container(:pending).alter_query(&:pending))
+      end
+
+      # this complex default sort requires extra SQL per proposal so currently performed post-query.
+      # TODO incorporate into the TabularData::Container itself as the definition of "status".
+      # rubocop:disable all
```
afeld: :-1: on disabling :wink:

.env.example
```diff
@@ -26,6 +26,7 @@ MYUSA_SECRET: consumer_secret_key
 # NOTIFICATION_FROM_EMAIL
 # PORT
 # RESTRICT_ACCESS=true
+# TEST_DATABASE_URL=postgresql://localhost/c2_test
```
afeld: Maybe worth adding a "test only" section?

pkarman:
re-approaching this.

PR #617, opened 2015-09-16
[Delivers #98729742] prefix sandbox emails

pkarman:
Story: https://www.pivotaltracker.com/story/show/98729742

### DIFF
--- a/app/views/gsa18f/procurements/form.html.erb
+++ b/app/views/gsa18f/procurements/form.html.erb
@@ -88,7 +88,11 @@
       <%= f.label :additional_info, 'Please include any additional information you feel we need to correctly execute the purchase' %>
       <%= f.text_area :additional_info, class: 'form-control' %>
     </div>
-    <%= f.submit class: 'form-button' %>
+    <% unless ENV['DISABLE_SANDBOX_WARNING'] == 'true' %>
+      <%= f.submit value: 'Submit test request for approval', class: 'form-button' %>
+    <% else %>
+      <%= f.submit class: 'form-button' %>
+    <% end %>
     <% if @model_instance.persisted? %>
       <%= link_to 'Discard Changes', proposal_url(@model_instance.proposal) %>
     <% end %>
--- a/app/views/gsa18f/procurements/form.html.erb
+++ b/app/views/gsa18f/procurements/form.html.erb
@@ -88,11 +88,7 @@
       <%= f.label :additional_info, 'Please include any additional information you feel we need to correctly execute the purchase' %>
       <%= f.text_area :additional_info, class: 'form-control' %>
     </div>
-    <% unless ENV['DISABLE_SANDBOX_WARNING'] == 'true' %>
-      <%= f.submit value: 'Submit test request for approval', class: 'form-button' %>
-    <% else %>
-      <%= f.submit class: 'form-button' %>
-    <% end %>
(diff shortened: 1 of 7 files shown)


app/views/gsa18f/procurements/form.html.erb
```diff
@@ -88,7 +88,11 @@
       <%= f.label :additional_info, 'Please include any additional information you feel we need to correctly execute the purchase' %>
       <%= f.text_area :additional_info, class: 'form-control' %>
     </div>
-    <%= f.submit class: 'form-button' %>
+    <% unless ENV['DISABLE_SANDBOX_WARNING'] == 'true' %>
+      <%= f.submit value: 'Submit test request for approval', class: 'form-button' %>
+    <% else %>
+      <%= f.submit class: 'form-button' %>
+    <% end %>
```
afeld: Since we have the same logic in both forms, can we make a shared partial/helper for it?

config/initializers/mailer.rb
```diff
@@ -13,3 +13,12 @@
...
 }
+
+unless ENV['DISABLE_SANDBOX_WARNING'] == 'true'
+  class PrefixEmailSubject
+    def self.delivering_email(mail)
+      mail.subject = "[TEST] " + mail.subject
+    end
+  end
+  ActionMailer::Base.register_interceptor(PrefixEmailSubject)
+end
```
afeld: Faaaancy. Maybe worth adding a little comment above saying something like "indicate that this is not a real request in the email subjects".

spec/rails_helper.rb
```diff
@@ -8,6 +8,9 @@
 require 'steps/user_steps'
 require 'steps/approval_steps'
 
+# mimic production env for view+mail
+ENV["DISABLE_SANDBOX_WARNING"] = 'true'
```
afeld: Minor: Hmmm...is it worth setting this explicitly for the tests that it affects?

pkarman: I think you're right. The test failures we currently have are expecting it to *not* be set.

pkarman:
please review

spec/support/environment_spec_helper.rb
```diff
@@ -1,6 +1,8 @@
 module EnvironmentSpecHelper
   # @todo rewrite using with_env_vars
   def with_18f_procurement_env_variables
+    ENV['DISABLE_SANDBOX_WARNING'] = 'true'
```
afeld: I know you were just following what the rest of the rest of the method is doing, but mind converting all of them to use `with_env_vars`?

afeld: Nevermind, took care of it in #625. Whichever gets merged first should tweak this though.

app/views/shared/_form_submit_button.html.erb
```diff
@@ -0,0 +1,5 @@
+<% unless ENV['DISABLE_SANDBOX_WARNING'] == 'true' %>
+  <%= form.submit value: 'Submit test request for approval', class: 'form-button' %>
+<% else %>
+  <%= form.submit class: 'form-button' %>
+<% end %>
```
afeld: Surprised this didn't get caught by Rubocop (I guess it doesn't read ERB?), but `if/else` is preferred to `unless/else`.

afeld:
The tweaks were minor, so I went ahead and made them.

PR #618, opened 2015-09-17
[Delivers #102256768] client slug authz

pkarman:


afeld:
Want feedback on this yet?

pkarman:
@afeld please. definitely wip though.

pkarman:
QA for this branch should definitely include trying with multiple user accounts, assigned to different client_slug values. This affects not only NCR but any endpoints that inherit from the `use_case_controller` so gsa18f et al.

config/environments/development.rb
```diff
@@ -25,7 +25,7 @@
   # Debug mode disables concatenation and preprocessing of assets.
   # This option may cause significant delays in view rendering with a large
   # number of complex assets.
-  config.assets.debug = true
+  config.assets.debug = ENV['ASSETS_DEBUG'] ? true : false
```
yozlet: Nice. Add this to the setup docs?

app/policies/proposal_policy.rb
```diff
@@ -20,8 +20,7 @@ def can_show!
   end
 
   def can_create!
-    # TODO restrict by client_slug
-    true
+    slug_matches? || @user.admin?
```
yozlet: I'm still new to this area, so please forgive what may be a stupid question:
When would this be false? And do we have a test for that situation?

pkarman: It would be false in the following cases, most (but not all) of which have tests.

The user does not have a 'admin' role and one of the following is true:

* the user is missing a client_slug
* the user requests /ncr/* endpoint but has gsa18f client_slug
* the user requests /gsa18f/* endpoint but has ncr client_slug

yozlet: Would it be easy to add the missing tests for those cases?

pkarman: But of course! added in 6f05e73

yozlet: Yay, thanks!

yozlet:
:shipit:

config/environments/development.rb
afeld: @pkarman Curious about the motivation for this...in my experience, it's preferable to have the files served separately in development mode.

pkarman: Mostly I was trying to cut down on the noise in my terminal when I was trying to debug something. All the static assets being served were logging a success response and so it seemed less optimal to see them on every request.

PR #619, opened 2015-09-17
[Delivers #103623422] add FORCE_USER_ID env var for overriding session user

pkarman:
development-only hack to mimic current_user as someone other than authenticated user


pkarman:
needs tests, story

app/views/home/me.html.erb
```diff
@@ -0,0 +1,3 @@
+<div class="inset">
+ <h2>Welcome, <%= current_user.email_address %></h2>
+</div>
```
afeld: Not _super_ against this, but couldn't you just check the header?

app/controllers/application_controller.rb
```diff
@@ -48,7 +48,15 @@ def peek_enabled?
...
   def current_user
-    @current_user ||= User.find_or_create_by(email_address: session[:user]['email']) if session[:user] && session[:user]['email']
+    @current_user ||= find_current_user
+  end
+
+  def find_current_user
+    if ENV['FORCE_USER_ID'] && !Rails.env.production?
+      User.find ENV['FORCE_USER_ID']
+    else
+      User.find_or_create_by(email_address: session[:user]['email']) if session[:user] && session[:user]['email']
```
afeld: Minor: how about moving the trailing `if` to an `elsif` above?

app/views/home/me.html.erb
pkarman: interestingly, the header uses `session['user']['email']` instead of `current_user` so when `FORCE_USER_ID` is on, the header shows the authenticated user, and the /me page shows the effective user.

afeld: Ah. In that case, game to change the logic in the header to make everything go through `current_user`?

pkarman: Done.

I'd still like to keep the `/me` endpoint though, not least because I think it might be helpful longer term for UX to populate it with User details and link to from the header. No story for that yet, though.

app/controllers/application_controller.rb
pkarman: Done, thank for catching that.

spec/features/force_user_id_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  let(:user) { FactoryGirl.create(:user, client_slug: 'gsa18f') }
+
+  it "respects session for current_user" do
+    wo = FactoryGirl.create(:ncr_work_order, :with_approvers)
+    login_as(wo.proposal.requester)
+    visit '/me'
+    expect(page.find('h2')).to have_content(wo.proposal.requester.email_address)
+  end
+
+  # can't use with_env here because we need to use factory object as env var value
```
afeld: Not true! You can use it around _or_ within specs.

app/views/home/me.html.erb
afeld: Okey doke.

.env.example
```diff
@@ -14,6 +14,7 @@ MYUSA_SECRET: consumer_secret_key
 # DEFAULT_URL_HOST
 # DEFAULT_URL_PORT
 # DEFAULT_URL_SCHEME
+# FORCE_USER_ID=123456
```
afeld: Can we document how to use this?

afeld: `/me recognizes not all of the others have corresponding documentation`

spec/features/force_user_id_spec.rb
pkarman:
```
diff --git a/spec/features/force_user_id_spec.rb b/spec/features/force_user_id_spec.rb
index 993e2d0..fc49537 100644
--- a/spec/features/force_user_id_spec.rb
+++ b/spec/features/force_user_id_spec.rb
@@ -8,13 +8,13 @@ describe "Acts as a different User in request" do
     expect(page.find('h2')).to have_content(wo.proposal.requester.email_address)
   end
 
-  # can't use with_env here because we need to use factory object as env var value
-  it "respects FORCE_USER_ID to override current_user" do
-    ENV['FORCE_USER_ID'] = user.id.to_s
-    wo = FactoryGirl.create(:ncr_work_order, :with_approvers)
-    login_as(wo.proposal.requester)
-    visit '/me'
-    expect(page.find('h2')).to have_content(user.email_address)
+  with_env('FORCE_USER_ID', user.id.to_s) do
+    it "respects FORCE_USER_ID to override current_user" do
+      login_as(wo.proposal.requester)
+      visit '/me'
+      expect(page.find('h2')).to have_content(user.email_address)
+    end
   end
 
 end
```

yields this error:

```
in `method_missing': `user` is not available on an example group (e.g. a `describe` or `context` block). It is only available from within individual examples (e.g. `it` blocks) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc). (RSpec::Core::ExampleGroup::WrongScopeError)
```

afeld: Try

```ruby
it "respects FORCE_USER_ID to override current_user" do
  with_env_var('FORCE_USER_ID', user.id) do
    wo = FactoryGirl.create(:ncr_work_order, :with_approvers)
    login_as(wo.proposal.requester)
    visit '/me'
    expect(page.find('h2')).to have_content(user.email_address)
  end
end
```

afeld: RSpec doesn't allow you to access `let`s outside of `it`/`before`/`after`s.

pkarman: ah! good suggestion, thanks.

phirefly:
Fwiw, I do this in development by simply setting up a new Chrome user in `chrome://settings/` and logging into that account. I'm not opposed to having this in our codebase but it seems like it's doing the same thing, isn't it?

pkarman:
@phirefly that works great when you want to test as anyone-other-than-your-real-self. This feature allows for testing as a specific, real user, that isn't you. E.g. someone submitting a help request who can't see a proposal, and you want to understand why.

phirefly:
But that real user would be on production, which we don't allow. Is that correct?

pkarman:
the use case here is masquerading as a user in a production db snapshot used in a sandbox. 

You are correct in noting that the feature is disabled when `Rails.env.production?` is true.

phirefly:
Got it. That was the missing piece. Thanks!

phirefly:
:shipit:

PR #620, opened 2015-09-17
run `rake db:seed` on deploy

afeld:
Addressing an issue raised in #602.

For any changes to `seeds.rb`, we want to ensure that those records are created on our various environments, without needing to duplicate the seeding in a migration. This change makes a `start` script, which should be used on development machines or servers, which

* Runs migrations
* Runs seeds
* Starts the server with the background workers

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,7 +1,1 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the rake db:seed (or created alongside the db with db:setup).
-#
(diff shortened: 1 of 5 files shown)


pkarman:
Running `db:seed` on every deploy does not make semantic sense to me, but google tells me that it is not uncommon for Rails projects to do that. That proves that I am an old codger.

:shipit:

afeld:
I hear ya – was considering making a separate rake task, but really it would just do the same thing.

PR #622, opened 2015-09-17
require appname on deploy, to avoid deploying everywhere

pkarman:
Per anecdote from recent deploy mishap, require explicit appname in `script/deploy`.

### DIFF
--- a/script/deploy
+++ b/script/deploy
@@ -1,5 +1,8 @@
 #!/bin/sh
 
+USAGE="$0: appname"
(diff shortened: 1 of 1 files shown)


afeld:
Was going to comment to say "please split up to multiple lines to make it easier to read", but felt bad about the amount of grief I've been giving on your PRs recently, so went ahead and did it myself :heart:

afeld:
(Let me know if you prefer I don't do that!)

afeld:
Tested locally.

PR #623, opened 2015-09-18
suggested changes to review and testing workflow

pkarman:


doc/team_practices.md
```diff
@@ -42,20 +42,40 @@ See the [story tracker documentation](https://www.pivotaltracker.com/help/api?ve
...
+1. The Reviewer should also test the code. This can be done on the `c2-dev` or `c2-staging` environment.
+Check with your teammates to see which environment might already be in use.
+Example flow:
+
+  1. `% cd /tmp && mkdir deploy-qa && cd deploy-qa`
+  1. `% git clone git@github.com:18F/C2.git`
+  1. `% cd C2`
+  1. `% git checkout -b qa-123456-fix-timezones`
+  1. `% git merge -m 'temp qa branch' origin/123456-fix-timezones`
+  1. `% script/deploy c2-dev`
```
afeld: Minor: I think this would look better as a [fenced code block](https://gist.github.com/clintel/1155906).

doc/team_practices.md
```diff
@@ -42,20 +42,40 @@ See the [story tracker documentation](https://www.pivotaltracker.com/help/api?ve
...
+
+1. The Reviewer should test out the bugfix or new functionality. The Owner might provide testing
```
afeld: I'd move this second instruction up to the step where the PR is created.

doc/team_practices.md
```diff
@@ -42,20 +42,40 @@ See the [story tracker documentation](https://www.pivotaltracker.com/help/api?ve
 
 1. Click **Finish** on the story in the story tracker.
 
-1. Someone from the team should review the PR. If you do not get feedback within 24 hours, assign the PR to a team member.
-The Reviewer is encouraged to indicate a "Ship it" (or your favorite Ship It emoticon) before merging the pull request. 
-Consider pointing out the awesomeness of your teammate's code, too.
-
-1. The Reviewer merges the PR. Pro tip: If the PR title follows the conventions above, 
+1. Someone from the team should review the PR. If you do not get feedback within 24 hours, 
+assign the PR to a team member for review. The Reviewer should consider the automated
```
afeld: Nice compromise :+1:

afeld:
This is great, thanks for writing it up. The one thing I want to bring up for discussion is whose responsibility it is to QA. I see a few options:

* The feature requester
   * The person who came up with (or at least wrote down) the idea _probably_ has the best idea of whether the implementation has met the requirements.
* The project manager
    * This person is most aware of the business needs.
    * Since this makes a single-person bottleneck, I'm giving a big fat :-1: off the bat.
* The code reviewer
    * This person is already in the mindset of looking for problems with the implementation, so in that sense it would be natural for them to test it as a user.
* The implementer
    * The person who wrote the code hopefully has the best idea of all the edge cases, especially after code review.

At my last company, the convention was to use the implementer, unless it was a UI-focused feature, in which case the requester would also take a look at it live. Totally open to other approaches though.

@18F/cap Anyone have strong feelings? Since this is a process change: 

* [ ] Before this is merged, get everyone on the team to "sign off" on it, to ensure they are a) aware of and b) cool with the changes.

doc/team_practices.md
pkarman: ah thanks for that link. I had struggled to get the ordered numbering to continue correctly. Now I know I need to indent the block properly.

phirefly:
I like that we're calling out and defining the different roles. Currently, we do share the QA responsibility, so to expect this from any one role would not be desirable. I can't think of a scenario where I'd want the  implementer do the final QA. They can certainly do a deploy to dev/staging to do their own QA, but I don't see that person doing the QA that results in Accept/Reject of the story.

afeld:
@phirefly In that case, thoughts about using the requester vs. the code reviewer?

phirefly:
@afeld: In my mind, both requester and code reviewer are equally valid QA contributors.

I would like to add that I've previously been on teams with dedicated QA engineers. The QA engineer role would be involved in the writing stories, developing acceptance criteria, manual testing, and often time writing automated tests. In my experience, this was very successful. It'd be great to introduce this model into our team. Until then, I'm excited about this common understanding of what it means to share the QA responsibilities.

Anyone else have similar experiences with dedicated QA roles?

pkarman:
My experience with QA is similar to @phirefly's. The QA engineer was ultimately responsible for marking the story (ticket) as accept/reject, having done the QA work on stage.

Having a dedicated QA person has been a life-saver for me, frankly. An independent advocate for the user who was the gatekeeper for code going into master.

afeld:
> I've previously been on teams with dedicated QA engineers...It'd be great to introduce this model into our team.

Funny you say this, since I've heard a lot of people say that the industry seems to be moving away from the model of dedicated QA folks.

http://www.drdobbs.com/testing/the-shrinking-role-of-qa/240003928

---

Coincidentally/tangentially, just came across this project:

https://github.com/guardian/prout

which might be interesting to try out. From their README:

> Prout comes from the tenet that:
>
> > Developers are responsible for checking their changes on Production

Note: I don't have super strong feelings on this.

phirefly:
Just a quick follow-up on this. Can we agree that any of the roles mentioned, except for the implementer of the story, is a valid candidate for doing QA on a story? If so, I would recommend adding that bit and then ship it.

afeld:
Might suffice to say "it should be tested on dev/staging prior to merge", and we'll see how we evolve.

pkarman:
So the "reviewer" and the "tester" are different roles then? Ok, I will edit to reflect that. The part that concerned me was merging to master before QA. Currently we merge to master *in order to* QA. That's the part that I want this PR to reflect a change to.

doc/team_practices.md
```diff
@@ -38,24 +42,45 @@ and [view the doc folder](https://github.com/18F/C2/tree/master/doc).
...
+testing status and Code Climate reports, in addition to checking the code for architectural
+consistency, style and legibility. Code reviews are encouraged early in the development process;
+you can always create a PR with a `[WIP]` prefix before you are ready to deliver, and ask for review
+of work-in-progress.
+
+1. The Reviewer should also test the code. This can be done on the `c2-dev` or `c2-staging` environment.
```
afeld: My main concern here is that it increases the burden on the review. We _already_ aren't doing a great job keeping up with reviewing pull requests, and if there is additional effort required on that side, I could see myself subconsciously procrastinating on doing reviews.

pkarman: That's a very valid point, and it bolsters the idea (imo) of having dedicated QA folks who can keep the pipeline moving wrt PRs.

I guess this comes down to risk and mitigation and how we feel about master. I like master HEAD to always be deployable, since it is easiest on the deployer. Our current process makes that much riskier, because there is zero integration or user testing done before merging to master. Process is always about the "appropriate amount of friction", and it's still unclear to me whether the risk associated with an unproven HEAD is acceptable or not, and if not, where we need to add appropriate friction.

afeld: Right- I meant to imply that I wouldn't rule out the idea of the author doing the QA, since they are probably the person who is most invested in getting it merged. Happy to iterate on this, though. 

Credit to @jessieay for that idea.

yozlet:
+1 to having dedicated QA; I think 18F as a whole could benefit hugely. I've worked with agile teams before who replace dedicated QA with automated tests, but then have no one doing **exploratory** testing. Especially with web apps, decent QA people can find all kinds of problems that you'd otherwise have no idea are there. The point is to find these problems before your customers/users do.

afeld:
K...what is the preference in the meantime?

pkarman:
My preference is anyone-but-the-owner (committer) should QA.

pkarman:
I have made some more edits to clarify that the Tester role is distinct from the Reviewer role, though they might be played by the same person.

afeld:
Since this affects everyone, please check the box below to confirm that you've read (and are ok with) the changes!

* [x] @afeld 
* [ ] @annalee 
* [ ] @anthonygarvan 
* [ ] @phirefly 
* [x] @pkarman 
* [ ] @yozlet 

_Guess I should have asked before merging, but :sparkles: YOLO :sparkles:_

PR #626, opened 2015-09-18
turn off TrailingComma check

pkarman:
c.f. https://github.com/bbatsov/ruby-style-guide/issues/273


afeld:
FWIW, I don't do this, but am not completely opposed to it.

PR #630, opened 2015-09-18
no db:seed on app start

pkarman:


afeld:
I dunno, I still like the idea of running the seeds on server start.

pkarman:
I have updated the affected branch so that `db:seed` can be run multiple times in the same environment:
https://github.com/18F/C2/pull/602#issuecomment-142066439

I will close this PR and add some comments to the db/seeds.rb file.

PR #632, opened 2015-09-21
fix wording of Comment comment

afeld:
@cmc333333 Mind double-checking this is what you meant?

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -40,7 +40,7 @@ class Comment < ActiveRecord::Base
   end
 
   # All of the users who should be notified when a comment is created
-  # This is basically Proposal.users _minus_ future approvers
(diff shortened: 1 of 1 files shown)


cmc333333:
Hrm. No, I meant "future" -- where a "future" approver is one who hasn't seen the proposal yet.

afeld:
Hmm, `currently_awaiting_approvers` is minus both...updated.

cmc333333:
Right `currently_awaiting_approvers` should be all actionable approvers -- those who can take action now. That gets added to the `approved` approvers, which should be all approvers who took action in the past. All that's missing, then, is the future approvers.

This is one of the multiple places where this kind of "subscriber" logic is being calculated -- each time the logic is ever so slightly different.

afeld:
Oh. Derp.

Thanks!

PR #634, opened 2015-09-21
[Delivers #99460178] require budget re-approval for certain changes on WorkOrders

afeld:
A bunch of the early commits are refactoring...the change of behavior is in https://github.com/18f/c2/compare/82f8b1a7e68c9ccf04203896b64a512f7250852b...99460178-post-approval-notifications.


spec/features/ncr/work_orders/edit_spec.rb
```diff
@@ -0,0 +1,331 @@
...
+    it "requires re-approval for a non-budget-approver adding a Function code" do
+      login_as(requester)
+
+      visit "/ncr/work_orders/#{work_order.id}/edit"
+      fill_in 'Function code', with: 'foo'
+      click_on 'Update'
+
+      expect_budget_approvals_restarted
+    end
+  end
```
afeld: @phirefly Can you check (the descriptions of) the specs in this block? Just want to make sure I am understanding the story tasks correctly.

spec/features/ncr/work_orders/edit_spec.rb
```diff
@@ -0,0 +1,331 @@
...
+    def expect_budget_approvals_restarted
+      work_order.reload
+      expect(work_order.status).to eq('pending')
+      approval_statuses = work_order.individual_approvals.pluck(:status)
+      expect(approval_statuses).to eq(%w(
+        approved
+        actionable
+        pending
+      ))
+      # TODO check who gets notified
```
afeld: @phirefly Who should be notified when the budget approval process gets restarted? What do the notifications need to say?

spec/features/ncr/work_orders/edit_spec.rb
```diff
@@ -0,0 +1,331 @@
...
+
+      visit "/ncr/work_orders/#{work_order.id}/edit"
+      fill_in 'Function code', with: 'foo'
+      click_on 'Update'
+
+      work_order.reload
+      expect(work_order.status).to eq('approved')
+    end
+
+    it "requires re-approval for a non-budget-approver adding a Function code" do
```
phirefly: Great catch on the role consideration. Asking our budget folks for validation...

phirefly:
Follow-up from convos with Budget folks:

*1. Can we assume that if a change is made to any of these values -- Org Code / Soc / Function / Building number / RWA -- that re-approval by budget is only needed if the change is made by a non-budget person? In other words, do we need anyone to re-approve if a budget person makes the change to any of those attributes?*

Correct - only if non-budget would want to have reapproved
if budget person FYI only is good.

 
*2. When triggering a reapproval, we will need to send a couple of new messages to Budget: a) A message that the request you had approved has been modified and that it is no longer in "Approved" status, b) Another email, informing you of the new version to be approved.*

Assuming these messages would come back to back don't see an issue we would have everything needed to go in and revisit.

afeld:
Ok, I _think_ this is good to go. Rebased off of `master`, because of some major conflicting changes (see https://github.com/18F/C2/pull/700#issuecomment-148389795). This is the email that goes to the budget approver when the request is modified by someone _other_ than budget post-approval.

screen shot 2015-10-18 at 6 20 09 pm

Note that it doesn't say anything explicitly about "this was approved but then they changed it and now it needs you to approve it again", but not sure that it needs to. Thoughts?

## Testing notes

To get a request in a fully-approved state, run the following from the console:

```ruby
proposal = ...
proposal.individual_approvals.outstanding.each do |approval|
  approval.reload
  approval.approve!
end
```

* [ ] Modify as a budget approver – the Proposal should remain `approved`.
* [ ] Modify as a requester (or someone else) – the Proposal should switch to `pending`, and the budget approvers should be required to re-approve.

pkarman:
@afeld alas, branch conflicts again.

afeld:
Fixed!

app/controllers/ncr/work_orders_controller.rb
```diff
@@ -15,7 +15,7 @@ def create
 
     def edit
       if self.proposal.approved?
-        flash[:warning] = "You are about to modify a fully approved request. Changes will be logged and sent to approvers but this request will not require re-approval."
+        flash[:warning] = "You are about to modify a fully approved request. Changes will be logged and sent to approvers but this request will not necessarily require re-approval."
```
jessieay: ah! this is an important copy change.

app/controllers/ncr/work_orders_controller.rb
```diff
@@ -74,5 +71,50 @@ def add_approvals
...
+
+    def reapprove_if_necessary
+      if requires_budget_reapproval?
+        @model_instance.restart_budget_approvals
+        flash[:success] = "Successfully modified! This request now needs to be re-approved by budget."
+      end
+    end
+
+    def after_update
+      if @model_changing && !@model_instance.emergency # skip approvals if emergency
```
jessieay: I thought `emergency` didnt matter? or was that a different type of model instance?

phirefly:
Testing results:
• Modifying the request and Increasing the amount resulted in a 500 error.
Steps to reproduce: Clicked `Modify request` and modified amount to a higher amount 
(example: Modifying https://c2-dev.18f.gov/proposals/208 redirected with 500 error to  https://c2-dev.18f.gov/ncr/work_orders/187)

• Also reloaded https://c2-dev.18f.gov/proposals/208 to verify the new amount in in the request, but approval from communicart.budget.approver@gmail.com was still approved. Expected this to be set to Pending.

app/controllers/ncr/work_orders_controller.rb
afeld: Good catch! Screwed that up in the merge.

afeld:
@phirefly Mind giving it another shot?

phirefly:
A couple of issues that I'm seeing:
* Getting a 500 when the approver re-approves (email and web). The request does get reapproved though.
* Updating from blank value (Soc/Function) triggers the re-approval. It should only do this if there is an existing value for Soc/Function code

I will continue to test and add if anything else comes up.

spec/features/ncr/work_orders/post_approval_modification_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    work_order.reload
+    expect(work_order.status).to eq('approved')
+    expect(work_order.proposal.root_approval.status).to eq('approved')
+    expect(approval_statuses).to eq(%w(
+      approved
+      approved
+      approved
+    ))
+  end
```
afeld: I normally don't create end-to-end feature specs like this (logging in as various people and performing different actions), but in this case it did help me identify a bug. Open to feedback about whether to keep it or not, or whether it should be slimmed down.

lib/ncr/work_order_updater.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    def after_update
+      if model_changing
+        work_order.setup_approvals_and_observers
+        reapprove_if_necessary
+        Dispatcher.on_proposal_update(proposal, work_order.modifier)
+      end
+    end
+  end
+end
```
afeld: I don't love how this class turned out, but it keeps the logic out of the controller, and makes the functionality more directly testable (even though I don't have any unit tests for it). Open to ideas about how to improve.

afeld:
> Update the following values should retrigger a re-approval but currently do not: `Org Code` `Building number` `RWA Number`

Would changing any of these require re-approval from the approving official?

afeld:
And what if it's budget approver who changes them?

phirefly:
re-approval from the approving official  is not required

lib/ncr/work_order_reapproval_checker.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    def requires_budget_reapproval?
+      work_order.approved? && (
+        self.amount_increased? || (
+          self.protected_fields_changed? &&
+          !self.budget_approver?
+        )
+      )
+    end
```
afeld: @phirefly This is the important part, in terms of determining if a request needs to be re-approved or not.

lib/ncr/work_order_reapproval_checker.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    protected
+
+    def protected_fields
+      results = Ncr::WorkOrder.budget_code_fields + [:building_number, :org_code, :rwa_number]
```
phirefly: budget_code_fields includes `:cl_number` which isn't in the list of fields that need to trigger re-approval

phirefly:
Here are my latest findings:
• CL Number should not trigger re-approval but it currently does
• When changing a `Soc code` or `Function code` from empty value, it triggers re-approval. Re-approval should only occur if there is an existing value, not when adding a new value

_(The next two may be caused by the same thing: state of the request?)_
• Email template used when sending the re-approval email does not have an ‘Approve’ button
• The request detail page does not have an approve button for the approve to re-approve. Additionally, once this is fixed, we need to make sure that re-approving puts the request back into ‘approved’ state, which was a previous issue.

afeld:
* [ ] Verify that all non-approval Steps get reset

afeld:
> When changing a `Soc code` or `Function code` from empty value, it triggers re-approval

I couldn't reproduce this with a manual or automated test. Mind double-checking?

afeld:
Just thought of a future problem here. Now that #722 is in, suppose we had a "reconcile" Step at the end. With `linear` flows (like WorkOrders have), we are assuming that each step must be completed after the previous one. In other words, if the budget approvals were reset, the WorkOrder would need to be re-reconciled, from C2's point of view.

Not a problem yet, but something that may bite us down the road.

/cc @yozlet

phirefly:
Looks good! Maybe I didn't pull before deploying.

afeld:
* [x] QA
* [ ] Code review

app/controllers/ncr/work_orders_controller.rb
pkarman: could we instead say "...but this request may require re-approval."
?

app/models/ncr/work_order.rb
```diff
@@ -171,8 +177,12 @@ def final_approver
       end
     end
 
-    def email_approvers
-      Dispatcher.on_proposal_update(self.proposal, self.modifier)
+    def budget_approvals
+      self.individual_approvals.offset(1)
```
pkarman: do we have any other way of identifying the flavor of approval it is, other than by its relative position? that feels brittle to me.

app/models/step.rb
```diff
@@ -25,9 +25,11 @@ class Step < ActiveRecord::Base
     scope status, -> { where(status: status) }
   end
   scope :non_pending, -> { where.not(status: "pending") }
+  scope :outstanding, -> { where.not(status: "approved") }
 
   default_scope { order("position ASC") }
 
+  # TODO make a protected method
```
pkarman: why can't we do it now?

spec/features/ncr/work_orders/post_approval_modification_spec.rb
pkarman: I like it.

pkarman:
a few nit comments, non blocking. Since this is such a big change, maybe another set of :eyes:  on it for review too?

app/models/ncr/work_order.rb
afeld: I totally hear you, but mind if we create a story for it and take care of it separately? Just don't want to hold up this change any longer.

afeld:
Updated!

spec/features/ncr/work_orders/post_approval_modification_spec.rb
jessieay: I like it too! The only comment I have on this spec is that we are trying to move to a place where feature specs perform and check for user-facing items (see https://www.pivotaltracker.com/story/show/105780488) and this is checking the DB in the expectations rather than UI elements. But I think the end-to-end-ness of it is :+1: :+1: :+1:

spec/support/proposal_spec_helper.rb
```diff
@@ -1,10 +1,22 @@
...
+
   def fully_approve(proposal)
     proposal.individual_approvals.each do |approval|
       approval.reload
       approval.approve!
     end
     expect(proposal.reload).to be_approved # sanity check
+
+    # sanity checks
+    expect(proposal.status).to eq('approved')
```
jessieay: this seems like the same as L11?

app/models/ncr/work_order.rb
pkarman: yes, fine to merge as-is.

Only outstanding comment is https://github.com/18F/C2/pull/634/files#r44310834 afaik @afeld

yozlet:
@afeld: Good question, and not sure. I don't understand the story well enough yet: Would we _want_ the request re-reconciled if it needs re-approval? If not, maybe there's a way we should be storing completed Steps so that if the Step gets re-triggered, it can check (in this case, against the original Cost/Price value vs whatever it currently is, maybe?) and automatically mark itself completed.

spec/features/ncr/work_orders/post_approval_modification_spec.rb
afeld: @jessieay I hear you, though it would be very difficult to do exclusively through th's tricky to do exclusively through the UI because

* The code to verify this in the markup would be tricky, trying to distinguish which item words like "pending" and "approve" refer to. I could imagine this being very brittle.
    * This probably speaks to the 508 compliance of the Proposal show page, but that's a whole separate issue.
* It's checking for internal state (the `root_step`), which isn't reflected in the UI.
    * Checking for this directly revealed a bug, so I would be very reluctant to take it out.

Is there a compromise, like moving the checks into methods to describe what it's doing?

app/models/ncr/work_order.rb
afeld: Storied: https://www.pivotaltracker.com/story/show/107771800

afeld:
@yozlet Yeeeeeahhh...that's a philosophical conversation I'd love to avoid for the moment :wink:

spec/features/ncr/work_orders/post_approval_modification_spec.rb
jessieay: I say we can keep as-is for now since the Big Refactor of Tests hasn't occured yet. Just wanted to point out the direction we are hoping to head in future for consideration as we continue to write feature specs etc etc :pear:

jessieay:
Looks ready to :ship:

PR #638, opened 2015-09-22
[Delivers #102764816] db constraints

pkarman:
Create unique indexes, foreign key constraints, clean up dirty/orphaned/deprecated data.


app/models/approval.rb
```diff
@@ -13,7 +13,7 @@ class Approval < ActiveRecord::Base
   validates :proposal, presence: true
 
   belongs_to :parent, class_name: 'Approval'
-  has_many :child_approvals, class_name: 'Approval', foreign_key: 'parent_id'
+  has_many :child_approvals, class_name: 'Approval', foreign_key: 'parent_id', dependent: :delete_all
```
afeld: I would go with `destroy_all` as a general rule, so that it cascades.

http://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-dependent

app/models/proposal.rb
```diff
@@ -115,14 +115,18 @@ def root_approval=(root)
...
+    self.reset_status()
+  end
+
+  def clean_up_old_approvals(old_approvals, approval_list)
     old_approvals.each do |old|
       unless approval_list.include?(old)
-        old.destroy()
+        old.destroy() if Approval.exists?(old.id)
       end
     end
```
afeld: Minor: this could be simplified to

```ruby
approvals_to_remove = old_approvals - approval_list
approvals_to_remove.each(&:destroy!)
```

_Not that much simpler, now that I look at it. Up to you._

db/migrate/20150921183423_add_db_constraints.rb
```diff
@@ -0,0 +1,44 @@
...
+    execute "ALTER TABLE comments DROP CONSTRAINT user_id_fkey"
+    execute "ALTER TABLE comments DROP CONSTRAINT proposal_id_fkey"
+    execute "ALTER TABLE proposal_roles DROP CONSTRAINT user_id_fkey"
+    execute "ALTER TABLE proposal_roles DROP CONSTRAINT proposal_id_fkey"
+    execute "ALTER TABLE proposal_roles DROP CONSTRAINT role_id_fkey"
+    execute "DROP INDEX roles_name_idx"
+    execute "ALTER TABLE user_roles DROP CONSTRAINT user_id_fkey"
+    execute "ALTER TABLE user_roles DROP CONSTRAINT role_id_fkey"
+  end
+end
```
afeld: Out of curiosity, have you seen or used something that can check for or create these constraints automatically? Easy for them to get out of sync with the models. Here's what I found on Google:

* https://github.com/SchemaPlus/schema_auto_foreign_keys
* https://github.com/SchemaPlus/schema_validations
* https://github.com/nullobject/rein
* https://github.com/trptcolin/consistency_fail (via https://robots.thoughtbot.com/the-perils-of-uniqueness-validations)
* https://www.ruby-toolbox.com/categories/Active_Record_Index_Assistants

Not necessarily a blocker.

pkarman: I had not seen or used any of those before.

It wasn't clear to me in reading through the docs for each of them whether adding them at this point would create a discrete migration (before/after) or whether they would silently add the missing constraints the first time the app is started. Do you have any experience or insight there?

pkarman:
@afeld thanks for the review. I've made the changes you suggested wrt code.

As for using additional gems in the model space to automate this, I'm totally open to experimenting with those. Maybe we could do that via an additional story, and use this raw sql to establish a clean-data baseline?

db/migrate/20150921183423_add_db_constraints.rb
```diff
@@ -0,0 +1,44 @@
...
+    execute "CREATE UNIQUE INDEX approvals_user_proposal_idx ON approvals USING btree (user_id, proposal_id)"
+    execute "ALTER TABLE attachments ADD CONSTRAINT user_id_fkey FOREIGN KEY (user_id) REFERENCES users (id)"
+    execute "ALTER TABLE attachments ADD CONSTRAINT proposal_id_fkey FOREIGN KEY (proposal_id) REFERENCES proposals (id)"
+    execute "ALTER TABLE comments ADD CONSTRAINT user_id_fkey FOREIGN KEY (user_id) REFERENCES users (id)"
+    execute "ALTER TABLE comments ADD CONSTRAINT proposal_id_fkey FOREIGN KEY (proposal_id) REFERENCES proposals (id)"
+    execute "ALTER TABLE proposal_roles ADD CONSTRAINT user_id_fkey FOREIGN KEY (user_id) REFERENCES users (id)"
+    execute "ALTER TABLE proposal_roles ADD CONSTRAINT proposal_id_fkey FOREIGN KEY (proposal_id) REFERENCES proposals (id)"
+    execute "ALTER TABLE proposal_roles ADD CONSTRAINT role_id_fkey FOREIGN KEY (role_id) REFERENCES roles (id)"
+    execute "ALTER TABLE proposals ADD CONSTRAINT requester_id_fkey FOREIGN KEY (requester_id) REFERENCES users (id)"
+    execute "CREATE UNIQUE INDEX roles_name_idx ON roles USING btree (name)"
```
yozlet: Isn't this table just a few kilobytes? An index wouldn't hurt, but I doubt it'll help much. What am I missing?

yozlet:
Added a comment about a possibly-unnecessary index, but whether we need that or not, this is good to go.

pkarman:
The index is for the unique constraint, not performance.

yozlet:
Doh! OK, sorry. :shipit:

db/migrate/20150921183423_add_db_constraints.rb
afeld: I haven't tried any of them, but I think it depends on the gem.

afeld:
> As for using additional gems in the model space to automate this...Maybe we could do that via an additional story

:+1:

PR #110, opened 2014-11-26
[#79556070] send emails using Mandrill

afeld:
Pretty easy!

screen shot 2014-11-26 at 1 53 00 pm

PR #639, opened 2015-09-22
[Delivers #79556070] Mandrill for SMTP

pkarman:
c.f. #110


pkarman:
ugh. lot's of spurious conflicts. starting over.

PR #641, opened 2015-09-22
[Delivers #103334544] feedback

pkarman:
Document SUPPORT_EMAIL env var and switch to uservoice.com


app/mailers/feedback_mailer.rb
```diff
@@ -13,6 +13,6 @@ def feedback(sending_user, form_values)
   end
 
   def self.support_email
-    ENV['SUPPORT_EMAIL'] || 'gatewaycommunicator@gsa.gov' # not sensitive, so hard coding
+    ENV['SUPPORT_EMAIL'] || 'tickets@18f.uservoice.com' # not sensitive, so hard coding
```
afeld: * [ ] Look into using a custom domain for support emails

app/controllers/feedback_controller.rb
```diff
@@ -8,7 +8,10 @@ def create
     unless form_values.empty?
       FeedbackMailer.feedback(current_user, form_values).deliver_later
```
afeld: When we settle on a support service, we may want to use an API rather than going through a emails, which are harder to capture errors for.

afeld:
Looks good so far!

afeld:
* [ ] Set the `SUPPORT_EMAIL` to send to UserVoice _and_ `gatewaycommunicator@gsa.gov` when we deploy, until we fully commit to the new system

app/mailers/feedback_mailer.rb
pkarman: something like support@c2.18f.gov ? That particular domain is pointed at mandrill which does not seem able to simply redirect mail to a different destination. We can continue to use gatewaycommunicator@gsa.gov as our published address and forward via gmail afaik.

afeld: Cool, that works for now. Maybe `help.c2.18f.gov` eventually if we need a different subdomain.

PR #643, opened 2015-09-23
keep track of deployments in New Relic

afeld:
Tested on dev:

screen shot 2015-09-23 at 2 05 39 am

It's particularly helpful in the graph views, where the deployments show up as vertical lines:

screen shot 2015-09-23 at 2 08 09 am

Not that interesting here since there's no real traffic to dev in the middle of the night, but you get the idea.

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -4,7 +4,7 @@
 # NOTE: any changes to this file require running `cf push <appname>` directly, rather than `cf-blue-green <appname>`.
 
 # general configuration
-command: script/start
+command: script/server_start
 domain: 18f.gov
 instances: 1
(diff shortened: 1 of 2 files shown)


pkarman:
looks good to me, :shipit:

pkarman:
I just tried this on `c2-dev` with master and do not see the deployment listed under the .../deployments page. Only the 2 that you did earlier this week.

Does deploying with `cf-blue-green` affect whether this shows up or not? I used that.

afeld:
Heh. Not perfect yet.

https://github.com/18F/cf-blue-green#manifests

PR #640, opened 2015-09-22
[HOLD] [Delivers #79556070] Mandrill SMTP config

pkarman:
Story: https://www.pivotaltracker.com/story/show/79556070

PR #647, opened 2015-09-24
[Delivers #89824332] incoming mail handler

pkarman:
Story: https://www.pivotaltracker.com/story/show/89824332

This PR integrates Mandrill's incoming mail API with C2, allowing C2 users to comment on proposals simply by replying to C2-originating emails.

The one security consideration I am aware of is whether we should somehow sign emails to help mitigate the possibility that someone could spam the comments of a proposal simply by spoofing an email from an existing user. I'm not exactly sure how to do that without introducing so much friction in to the user's workflow that the entire feature becomes moot.


spec/features/incoming_mail_handler_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+                        "description" : "Delivered to internal network by a host with no rDNS"
+                     }
+                  ],
+                  "score" : 1.3
+               }
+            },
+            "ts" : 1443043366
+         }
+      ]
+    END
```
afeld: Minor: I'd put this in `spec/fixtures/mandrill_incoming.json`, just to make the spec file easier to read.

afeld:
Looking good so far! Will definitely need a good amount of manual testing.

spec/features/incoming_mail_handler_spec.rb
pkarman: @afeld: good call; done in 3df6f1e

config/routes.rb
```diff
@@ -11,6 +11,9 @@
 
   resources :help, only: [:index, :show]
 
+  # mandrill-rails
+  resource :inbox, :controller => 'inbox', :only => [:show, :create]
```
jessieay: new hash syntax!? LETS DO IT

lib/incoming_mail.rb
```diff
@@ -0,0 +1,110 @@
+module IncomingMail
+  # mail_types
+  UNKNOWN = 0
+  REQUEST = 1
+
+  class Handler
+    # parses a Mandrill callback object and Does The Right Thing
```
jessieay: Given that I am not sure what "The Right Thing" is,  I think we can do without this comment :)

lib/incoming_mail.rb
```diff
@@ -0,0 +1,110 @@
+module IncomingMail
```
jessieay: I feel like `lib` is becoming kind of a junk drawer for this codebase. Is there anywhere else it might make sense to put this?

lib/incoming_mail.rb
```diff
@@ -0,0 +1,110 @@
...
+  # mail_types
+  UNKNOWN = 0
+  REQUEST = 1
+
+  class Handler
+    # parses a Mandrill callback object and Does The Right Thing
+
+    def initialize(params = {})
```
jessieay: since this module has an `initialize` method, might it make sense to make it a class?

lib/incoming_mail.rb
```diff
@@ -0,0 +1,110 @@
...
+      if payload.is_a?(Array) and payload[0]['event'] and payload[0]['event'] == 'inbound'
+        create_response(payload[0])
+      elsif payload.is_a?(Mandrill::WebHook::EventDecorator)
+        create_response(payload)
+      else
+        fail "Invalid Mandrill event payload. Must be event==inbound"
+      end
+    end
+
+    private
```
jessieay: I believe `private` has no effect in modules -- another reason to re-CLASSify this file ;)

lib/incoming_mail.rb
```diff
@@ -0,0 +1,110 @@
...
+      else
+        resp.action = Response::DROPPED
+      end
+      resp
+    end
+
+    def identify_mail_type(payload)
+      subject = payload['msg']['subject']
+      references = payload['msg']['headers']['References']
+      if subject.match(/Request (#|FY)\d+/)
```
jessieay: the meaning of those line is not very clear to me - love what you're doing here with breaking out logic into private methods. Perhaps we can continue on with that pattern and throw this line into a well-named private method that reveals what this mysterious regex is all about?

lib/incoming_mail.rb
```diff
@@ -0,0 +1,110 @@
...
+        return IncomingMail::REQUEST
+      elsif references and references.match(/<proposal-\d+/)
+        return IncomingMail::REQUEST
+      else
+        return IncomingMail::UNKNOWN
+      end
+    end
+
+    def create_comment(msg)
+      proposal     = find_proposal(find_public_id(msg)) or return
```
jessieay: is aligning the `=`s here a style we are using in this project? (I usually do not do this but am unsure what this alignment is called and therefore am having trouble googling for rubocop related rules)

lib/incoming_mail.rb
```diff
@@ -0,0 +1,110 @@
...
+      else
+        # yes, user adds self as observer, which also generates comment
+        proposal.add_observer(comment_user, comment_user, comment_text)
+        # return most recent comment (TODO race condition here?)
+        proposal.comments.last
+      end
+    end
+
+    def find_proposal(public_id)
+      Proposal.find_by_public_id(public_id) || Proposal.find(public_id)
```
jessieay: I tend to use the `Proposal.find_by(public_id: public_id)` style in places like this because I always forget which dynamic finders were deprecated in Rails 4. Don't feel strongly either way, though https://github.com/rails/activerecord-deprecated_finders

lib/incoming_mail.rb
```diff
@@ -0,0 +1,110 @@
...
+      if references.match(ref_re)
+        return references.match(ref_re)[1]
+      end
+
+      sbj_re = /Request\ #?([\w\-]+)/
+      if msg['subject'].match(sbj_re)
+        return msg['subject'].match(sbj_re)[1]
+      end
+
+      fail "Failed to find public_id in msg #{msg.inspect}"
```
jessieay: why not do an `if` `elsif` `else` here rather than explicitly doing a `return` for the first 2 cases?

lib/incoming_mail.rb
```diff
@@ -0,0 +1,110 @@
...
+      EmailReplyParser.parse_reply(msg['text'])
+    end
+
+    def find_comment_user(msg)
+      from = msg['from_email']
+      User.find_by_email_address(from)
+    end
+  end
+
+  class Response
```
jessieay: I think it would be awesome to have this little baby class in it's own file so it's more findable -- what do you think?

lib/incoming_mail.rb
afeld: The two options for Ruby files in Rails apps are (in my experience) `app/` or `lib/`... what would you suggest? That being said, I think we could be grouping more of the files under `lib/` into subfolders.

app/controllers/inbox_controller.rb
```diff
@@ -0,0 +1,17 @@
+class InboxController < ApplicationController
+  include Mandrill::Rails::WebHookProcessor
+
+  def handle_inbound(event_payload)
+    handler = IncomingMail::Handler.new
+    resp = handler.handle(event_payload)
+    Rails.logger.warn(resp.inspect)
+    if resp.action == IncomingMail::Response::ERROR
+      # TODO do we want mandrill to keep trying on our error?
+      head 500
```
jessieay: is there an error tracking system we can use to log this with?

lib/incoming_mail.rb
afeld: Take a look again :wink: `IncomingMail` is just the namespace.

app/controllers/inbox_controller.rb
```diff
@@ -0,0 +1,17 @@
+class InboxController < ApplicationController
+  include Mandrill::Rails::WebHookProcessor
+
+  def handle_inbound(event_payload)
+    handler = IncomingMail::Handler.new
+    resp = handler.handle(event_payload)
+    Rails.logger.warn(resp.inspect)
```
jessieay: did you mean to leave this in here?

lib/incoming_mail.rb
afeld: I don't do it either, FWIW. [`Style/MultilineOperationIndentation`](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/style/multiline_operation_indentation.rb) is what you're looking for.

lib/incoming_mail.rb
```diff
@@ -0,0 +1,110 @@
...
+    def initialize(params = {})
+      params.each do |key, value|
+        instance_variable_set("@#{key}", value)
+      end
+    end
+
+    def handle(payload)
+      if payload.is_a?(Array) and payload[0]['event'] and payload[0]['event'] == 'inbound'
+        create_response(payload[0])
+      elsif payload.is_a?(Mandrill::WebHook::EventDecorator)
```
jessieay: interested in why there are 2 different possibilities here - perhaps a test for each case would clarify?

lib/incoming_mail.rb
afeld: :+1: to only one class per file.

lib/incoming_mail.rb
jessieay: you are right!

app/controllers/inbox_controller.rb
afeld: Yep, we use New Relic to (automatically) track exceptions. I'd allow it to bubble up, and use [`rescue_from`](http://guides.rubyonrails.org/action_controller_overview.html#rescue-from) if you need some special handling of it.

app/controllers/inbox_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+    if resp.action == IncomingMail::Response::ERROR
+      # TODO do we want mandrill to keep trying on our error?
```
afeld: Definitely.

lib/incoming_mail.rb
jessieay: I think we could have more subfolders within `app` .... even `app/modesls`...I like the model suggested in this blog post: http://blog.codeclimate.com/blog/2012/02/07/what-code-goes-in-the-lib-directory/

Basically, anything that would potentially be re-usable by another app could go in `lib`. But this parser, for example, feels very particular to this app. Perhaps we could add it to `app/models/` if we make this a class. This is what they do in the griddler demo (griddler is a lib for parsing incoming mail, too) https://github.com/thoughtbot/griddler-demo/blob/master/app/models/comment_via_email.rb

lib/incoming_mail.rb
jessieay: in that case, let's unalign these `=`! :green_apple:

jessieay: @afeld thanks for the link! Will try to remember "multiline operation indentation" for the future

pkarman:
Thanks for the thorough reviews, @afeld and @jessieay. I think I've caught all the suggestions in my latest refactor.

lib/incoming_mail/handler.rb
```diff
@@ -0,0 +1,98 @@
...
+
+    def initialize(params = {})
+      params.each do |key, value|
+        instance_variable_set("@#{key}", value)
+      end
+    end
+
+    def handle(payload)
+      # raw JSON
+      if payload.is_a?(Array) and payload[0]['event'] and payload[0]['event'] == 'inbound'
```
jessieay: this is super small / not a blocker but one idea in a situation like this would be to have a private method called `is_raw_json?` and include `payload.is_a?(Array) and payload[0]['event'] and payload[0]['event'] == 'inbound'` in there so the code itself explains the branching here (same for the `elsif`)

jessieay: and then we would no longer need a comment :clap: :clap: :clap:

lib/incoming_mail/handler.rb
```diff
@@ -0,0 +1,98 @@
...
+      proposal = find_proposal(find_public_id(msg)) or return
+      comment_text = find_comment_text(msg)
+      comment_user = find_comment_user(msg)
+      if proposal.existing_observation_for(comment_user) || proposal.requester_id == comment_user.id
+        # already in the loop, just add comment.
+        proposal.comments.create(comment_text: comment_text, user: comment_user)
+      else
+        # yes, user adds self as observer, which also generates comment
+        proposal.add_observer(comment_user, comment_user, comment_text)
+        # return most recent comment (TODO race condition here?)
```
jessieay: add this to tracker rather than to do?

jessieay:
Looks so awesome. I say :ship: but also code climate is still running so kinda a TBD :ship:

pkarman:
thanks @jessieay for the review.

I am going to merge this branch and #640 into a QA branch for stage today. Assuming it tests out ok there, I think this feature can go straight into prod since it won't be actually used till we change the relevant ENV vars to use a reply-to address @c2.18f.gov.

pkarman:
Incoming messages that the app cannot handle (for whatever reason) are not re-sent to `NOTIFICATION_FALLBACK_EMAIL` which defaults to `communicart.sender@gsa.gov`. This should allow us to not lose anything, and the headers are re-written so that replying to a message will go to the original sender.

lib/incoming_mail/handler.rb
```diff
@@ -0,0 +1,103 @@
+module IncomingMail
+  class Handler
+    def initialize(params = {})
+      params.each do |key, value|
+        instance_variable_set("@#{key}", value)
+      end
+    end
+
+    def handle(payload)
+      if payload_is_raw?(payload)
```
jessieay: YES this is so readable :heart:

spec/features/incoming_mail_handler_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      'text'       => mail_msg.text_part.body.encoded,
+      'html'       => mail_msg.html_part.body.encoded,
+      'raw_msg'    => mail_msg.to_s,
+      'headers'    => headers,
+    }
+    [ { 'event' => 'inbound', 'msg' => msg } ]
+  end
+
+  with_env_vars(NOTIFICATION_FALLBACK_EMAIL: 'nowhere@some.gov', NOTIFICATION_FROM_EMAIL: 'noreply@some.gov') do
+    it "should forward non-app email to NOTIFICATION_FALLBACK_EMAIL" do
```
jessieay: this is more of a general thought and not necessarily something to take action on now: what about removing "should" from our spec descriptions:

> Finally, don’t begin examples names with the word ‘should’.  It’s redundant and results in hard to read spec output. 

> Write examples by starting with a present tense verb that describes the behavior.

http://blog.carbonfive.com/2010/10/21/rspec-best-practices/

spec/features/incoming_mail_handler_spec.rb
```diff
@@ -0,0 +1,79 @@
+describe "Handles incoming email" do
+  let(:proposal) { FactoryGirl.create(:proposal, :with_parallel_approvers) }
+  let(:approval) { proposal.individual_approvals.first }
+  let(:mail) { CommunicartMailer.actions_for_approver(approval) }
+  let(:mandrill_inbound_noapp) { File.read(RSpec.configuration.fixture_path + '/mandrill_inbound_noapp.json') }
+
+  def mandrill_payload_from_message(mail_msg)
```
jessieay: I usually put private methods for specs at the bottom of the file so specs are at the top!

spec/features/incoming_mail_handler_spec.rb
```diff
@@ -0,0 +1,79 @@
+describe "Handles incoming email" do
+  let(:proposal) { FactoryGirl.create(:proposal, :with_parallel_approvers) }
```
jessieay: thoughts on private methods instead of `let` per convo here https://github.com/18F/C2/pull/676#issuecomment-146062260

pkarman: I have no preference for let-vs-private. I figure if the feature is in rspec (let) then it won't be a surprise for newcomers to see it. Private methods, while possibly more performant, feel, I dunno, less idiomatic? +0 from me at a philosophical level. Happy to follow whatever consensus emerges.

spec/features/incoming_mail_handler_spec.rb
pkarman: This is a habit from my C days. The "main" stuff at the bottom, like a pyramid.

Happy to change if private-at-the-bottom is the convention.

pkarman:
deployment note: this requires creating an inbound email entry at mandrillapp.com. The target URL will be https://cap.18f.gov/inbox and the app must be deployed *before* you create that entry, because the URL is checked on creation.

spec/features/incoming_mail_handler_spec.rb
jessieay: Perhaps all of us should sit down and discuss at some point. Yoz and I had a 10 minute convo about it. I find the arguments in this blog post to be persuasive: https://robots.thoughtbot.com/lets-not

Agree that many many people who write rspec use `let` statements so they are not confusing to new people. But I do prefer to have all of the dependencies for each test called explicitly in each test so I know what I'm dealing with.

pkarman:
How should we handle "out of office" auto-replies?

jessieay:
@pkarman that is a very interesting question! Te issue is that we want to create a comment for regular "replies via email" but not for out of office auto replies, right?

pkarman:
that is the question. it might be relevant if the OOO reply is from the only approver (e.g.), in which case a comment might be appropriate.

pkarman:
fwiw, I believe the current behavior as written would be that OOO autoreplies will create comments, because they will contain one or more of the regex patterns that match the original Request.

@phirefly if we think that will create too much noise, I can add a check for the OOO pattern(s) and treat them as pass-through. Or just drop them?

lib/incoming_mail.rb
afeld: > anything that would potentially be re-usable by another app could go in `lib`

I would say anything that would be potentially reusable should be a gem :trollface:

Files under `lib/tasks/` are almost always app-specific, so I'm not sure I agree with that distinction. Honestly, if I were building Rails from scratch, I'm not sure that there's much need for both `app/` and `lib/` directories, other than that certain directories beneath them (`app/views/`) have special meaning.

> this parser, for example, feels very particular to this app. Perhaps we could add it to `app/models/`

Hmm, I think of `app/models/` as specific to classes that deal with retrieving/storing data (and their `concerns/`), which seems like a stretch here.

spec/features/incoming_mail_handler_spec.rb
afeld: Commented in the article about the "slow tests" argument no longer applying (as they are lazily loaded), but I hear you about the setting up dependencies explicitly. I try to use `let` judiciously – for simple data that is in fact shared by all of the tests – rather than using them to create every variable. That being said, I totally hear the Mystery Guest argument.

afeld:
I think we are all on the same page that we want to handle Out of Office emails in a special way (if it's not super hard to do so reliably), but let's tackle that [separately](https://www.pivotaltracker.com/story/show/83540466).

pkarman:
The OOO issue may actually be a non-issue, because OOO replies are only sent once to each sending email address. Since all our notifications come from the same From address, that means we will get infrequent OOO autoreplies, which imo makes them unreliable.

https://support.google.com/mail/answer/25922?hl=en

phirefly:
@pkarman Good point. Interesting.

lib/incoming_mail.rb
jessieay: @afeld Agreed on `lib/tasks` -- I see that the distinction we having going her is that `app/models` is for AR classes and `lib` is for everything else. I think that is a fine distinction to run with, although not one I've seen in the past.

If we do continue to put all non-AR classes into `lib`, perhaps a good middle ground would be to have more sub-folders inside `lib` so we don't have a bunch of unrelated classes thrown into the same directory?

afeld: :+1:

pkarman:
can I get an amen ... err ship-it?

jessieay:
I said :ship: awhile back but CI hadn't passed! looks like we are on the same BOAT now hashtagLOL

pkarman:
all green again. please merge if you approve.

jessieay:
@pkarman MERGED

pkarman:
@jessieay GRACIAS

PR #648, opened 2015-09-25
[Delivers #98787550] sort gsa18f procurements by urgency

thebenedict:
Allows gsa18F `Procurement`s to be sorted by `urgency`in the table at /proposals, and changes the type of `urgency` from `:string` to `:integer` to support sorting. 

:warning: Includes a migration to update existing urgency values to integers, but only for the 3 urgency strings currently in 18f.yaml. Please confirm that only these 3 values exist before running the migration in production. Happy to update the migration if there are other values.

*Context/notes to self:*
- If a user can have multiple clients (i.e. work w/ multiple orgs/agencies), /proposals should probably become a tabbed view. Sorting by urgency is unlikely to make sense across clients. 
- Spacing out the integer values (10/20/30 vs 1/2/3) is so folks can add new urgency levels later w/o a data migration.
- With shorter, more symbol-like labels this would be a candidate for ActiveRecord enums.
- `Procurement#urgency_string` should be moved to a decorator if there are more than 1 or 2 similar methods

### DIFF
--- a/app/models/gsa18f/procurement.rb
+++ b/app/models/gsa18f/procurement.rb
@@ -69,6 +69,10 @@ module Gsa18f
       self.product_name_and_description
     end
 
+    def urgency_string
+      URGENCY[urgency]
+    end
+
     def self.approver_email
       ENV.fetch('GSA18F_APPROVER_EMAIL')
     end
--- a/app/views/gsa18f/procurements/form.html.erb
+++ b/app/views/gsa18f/procurements/form.html.erb
@@ -80,7 +80,7 @@
      <div class="form-group">
       <%= f.label :urgency %>
       <%= f.collection_select(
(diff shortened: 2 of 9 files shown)


cmc333333:
Stellar! Thanks @thebenedict

pkarman:
This is a great PR, @thebenedict and thank you.

I have confirmed that production data contains only the 3 affected strings, so the migration should be fine as is.

Your UI comment is well-taken. We have some designs already around that very thing.

I also like your comments about using enum and we should consider how to do that both here and elsewhere in future. This PR is fine, though, as-is.

I'm going to QA separately from this review and then will merge to master.

thebenedict:
Thanks much for the feedback @pkarman @cmc333333. Please ping me on slack if other tasks come up that an external person can do, happy to jump in when I can.

PR #577, opened 2015-09-02
[Delivers #101461556] Add an optional reason when adding an observer

yozlet:
The same functionality as previously implemented by #573, but this time on top of the user roles work in #575 (on which this PR is dependent, so that needs to be merged first)

PR #649, opened 2015-09-25
[Delivers #101461556] Add an optional reason when adding an observer (supplemental)

yozlet:
Two additional tweaks to #577:

1. Use JS to only show the "reason" field when an observer has been selected in the dropdown.
2. Instead of adding a proper Comment to the proposal, we want the reason to show up in the Proposal's paper trail log. There's now a `comment` field in `proposal_roles` which stores the reason. I'd especially like feedback on this - @afeld?


pkarman:
1. I really like the JS to only show the comment box when appropriate. That feels like a UX win, and the code looks fine.

1. Adding the comment field to `proposal_roles` table feels awkward to me, from a data modeling perspective. We have a `comments` table already, and I would expect it there. Is the reason it's not there because the paper_trail code doesn't know how to relate a specific comment to a specific action (adding a `proposal_roles` row)? What if instead the `comments` table was made somewhat polymorphic, so that it could refer to something in addition to a proposal? I.e., add a `ref_type` and `ref_id` to the `comments` table and point at the relevant `proposal_roles` record? That would do a couple things: (1) keep all the comments in the comments table, and (2) allow for more one than one comment on a single `proposal_roles` record (e.g. "added this person" then later "removed this person").

In the interest of moving the original PR/story along, maybe you could split the comment storage piece off into its own story while the modeling gets hashed out? Or does the modeling prevent proper display of the comment in the timeline later?

yozlet:
@pkarman: Yup, makes a lot of sense. I'll close this PR and do separate ones for the JS and the comment refactor. Currently we're doing comments as plain old Comments with no papertrail or other associations, and we can leave it like that for the moment while working out how we want comments to work.

afeld:
Sorry for just responding to this.

> We have a `comments` table already...Is the reason it's not there because the paper_trail code doesn't know how to relate a specific comment to a specific action (adding a proposal_roles row)?

Correct, though Paper Trail is somewhat tangential here. Paper Trail comes into the picture because we should be able to use it for a user-facing activity log (which `update_comment`s partially do now).

> Currently we're doing comments as plain old Comments with no papertrail or other associations, and we can leave it like that for the moment while working out how we want comments to work.

:+1:

PR #650, opened 2015-09-25
[Delivers #102267428] pick highest pending approver on the stack

pkarman:
fixes bug reported in QA for https://www.pivotaltracker.com/story/show/102267428

### DIFF
--- a/app/models/ncr/work_order.rb
+++ b/app/models/ncr/work_order.rb
@@ -116,8 +116,8 @@ module Ncr
 
     # the highest approver on the stack, pending preferred if status indicates
     def current_approver_email_address
-      if self.individual_approvals.pending.first
-        self.individual_approvals.pending.first.user.email_address
+      if self.pending?
+        self.individual_approvals.where(status: 'actionable').first.user.email_address
       else
         self.approving_official.email_address
       end
--- a/app/models/ncr/work_order.rb
+++ b/app/models/ncr/work_order.rb
@@ -114,9 +114,13 @@ module Ncr
       self.approvers.first
     end
 
-    # the highest pending approver on the stack
+    # the highest approver on the stack, pending preferred if status indicates
     def current_approver_email_address
-      self.individual_approvals.pending.first.user.email_address
+      if self.individual_approvals.pending.first
+        self.individual_approvals.pending.first.user.email_address
+      else
+        self.approving_official.email_address
+      end
     end
 
     def email_approvers
--- a/app/models/ncr/work_order.rb
+++ b/app/models/ncr/work_order.rb
@@ -114,8 +114,9 @@ module Ncr
       self.approvers.first
     end
 
-    def approving_official_email_address
-      approving_official ? approving_official.email_address : self.system_approver_emails.first
+    # the highest pending approver on the stack
+    def current_approver_email_address
+      self.individual_approvals.pending.first.user.email_address
     end
 
     def email_approvers
--- a/spec/models/ncr/work_order_spec.rb
+++ b/spec/models/ncr/work_order_spec.rb
@@ -453,4 +453,19 @@ describe Ncr::WorkOrder do
       expect(wo.building_id).to be_nil
     end
   end
+
+  describe "#current_approver_email_address" do
+    it "returns the first pending approval's email address" do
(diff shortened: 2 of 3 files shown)


phirefly:
Looks like this is blowing up https://gist.github.com/phirefly/4f725ae6de6a4b43933f#file-error-2015-09-25-L7

likely because these records don't have pending approvals:
https://github.com/18F/C2/blob/master/app/mailers/report_mailer.rb#L20-L22

pkarman:
I think bf1170c should fix it, whenever we return to QAing this.

yozlet:
:shipit:

annalee:
:shipit:

pkarman:
@phirefly if/when this passes for you in stage, feel free to merge.

app/models/ncr/work_order.rb
```diff
@@ -114,8 +114,13 @@ def approving_official
       self.approvers.first
     end
 
-    def approving_official_email_address
-      approving_official ? approving_official.email_address : self.system_approver_emails.first
+    # the highest approver on the stack, pending preferred if status indicates
```
jessieay: what do you think about not having a comment here? I think this method's tests make it clear why it works this way, and having the comment could lead to more complications down the line.

For more details on this line of thinking, this is a killer blog post on why code comments can be tricky https://robots.thoughtbot.com/letting-your-code-speak-for-itself by the one and only @adarsh

spec/models/ncr/work_order_spec.rb
```diff
@@ -453,4 +453,19 @@
...
     end
   end
+
+  describe "#current_approver_email_address" do
+    it "returns the first pending approval's email address" do
+      wo = FactoryGirl.create(:ncr_work_order, :with_approvers)
+      expect(wo.current_approver_email_address).to eq(wo.individual_approvals.first.user.email_address)
+      wo.individual_approvals.first.approve!
+      expect(wo.current_approver_email_address).to eq(wo.individual_approvals.last.user.email_address)
+    end
```
jessieay: add a newline between tests?

spec/models/ncr/work_order_spec.rb
```diff
@@ -453,4 +453,19 @@
       expect(wo.building_id).to be_nil
     end
   end
+
+  describe "#current_approver_email_address" do
+    it "returns the first pending approval's email address" do
```
jessieay: when you say `first` -- what are users being ordered by? just `id`?

spec/models/ncr/work_order_spec.rb
```diff
@@ -453,4 +453,19 @@
...
+      wo = FactoryGirl.create(:ncr_work_order, :with_approvers)
```
jessieay: Might be outside the scope of this PR, but what do you think about include FactoryGirl syntax methods so this can just be `wo = create(:ncr_work_order, :with_approvers)` ? Would be a lot less typing!!

https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#configure-your-test-suite

app/models/ncr/work_order.rb
adarsh: :thumbs_up:

On Monday, October 5, 2015, Jessie A. Young <notifications@github.com>
wrote:

> In app/models/ncr/work_order.rb
> <https://github.com/18F/C2/pull/650#discussion_r41155349>:
>
> > @@ -114,8 +114,13 @@ def approving_official
> >        self.approvers.first
> >      end
> >
> > -    def approving_official_email_address
> > -      approving_official ? approving_official.email_address : self.system_approver_emails.first
> > +    # the highest approver on the stack, pending preferred if status indicates
>
> what do you think about not having a comment here? I think this method's
> tests make it clear why it works this way, and having the comment could
> lead to more complications down the line.
>
> For more details on this line of thinking, this is a killer blog post on
> why code comments can be tricky
> https://robots.thoughtbot.com/letting-your-code-speak-for-itself by the
> one and only @adarsh <https://github.com/adarsh>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/18F/C2/pull/650/files#r41155349>.
>


-- 


-- 
@adarshp <http://www.twitter.com/adarshp>
313.454.1515 (mobile)
See when I am free: http://bit.ly/adarshcal

PR #652, opened 2015-10-02
[Delivers #103696912] add .about.yml file per https://github.com/18F/about_yml/

pkarman:


.about.yml
```diff
@@ -0,0 +1,108 @@
...
+  id: pkarman
+  role: developer
+- github: yozlet
+  id: yoz
+  role: developer
+
+# Partners for whom the project is developed
+partners:
+- NCR
+- 18F
```
jessieay: GSA, too?

pkarman: probably `CAP` is most accurate. @phirefly ?

Gemfile
```diff
@@ -1,6 +1,7 @@
 source 'https://rubygems.org'
 ruby '2.2.2' # this should match `.ruby-version` and docs/setup.md
 
+gem 'about_yml'
```
afeld: Do we actually need this to be installed with the project?

.about.yml
afeld: Went ahead and expanded the partner names.

Gemfile
pkarman: good question. I was following https://github.com/18F/about_yml/ README which mentions adding to the project Gemfile, but in retrospect, it really doesn't make much sense to have it as part of the app since it was most useful on my local system only, for bootstrapping the file.

afeld: Take another look:

https://github.com/18F/about_yml/#installation

Maybe they expanded that section since you looked at it? Totally fine with including it (in `development` and `test` only, with `require: false`) if you want to run it as part of CI.

.about.yml
```diff
@@ -0,0 +1,109 @@
+---
+# .about.yml project metadata
+#
+# Short name that acts as the project identifier (required)
+name: C2
+
+# Full proper name of the project (required)
+full_name: C2
```
mbland: This wouldn't be "Communicart"?

.about.yml
```diff
@@ -0,0 +1,109 @@
...
+# The type of content in the repo
+# values: app, docs, policy
+type: app
+
+# Describes whether a project team, working group/guild, etc. owns the repo (required)
+# values: guild, working-group, project
+owner_type: project
+
+# Name of the main project repo if this is a sub-repo; name of the working group/guild repo if this is a working group/guild subproject
+parent: C2
```
mbland: You can delete/comment this out if _this_ is the main repo.

.about.yml
```diff
@@ -0,0 +1,109 @@
...
+# values: true, false
+testable: true
+
+# Team members contributing to the project (required)
+# Items:
+# - github: GitHub user name
+#   id: Internal team identifier/user name
+#   role: Team member's role; leads should be designated as 'lead'
+team:
+- github: afeld
```
mbland: Technically you don't need _both_ `github:` and `id:` here, but so long as it's certain that they're both correct, there's no harm done.

I'd also maybe push the `lead:` to the top, but it doesn't really matter, if you want to keep things alphabetized as you have them here.

.about.yml
```diff
@@ -0,0 +1,109 @@
...
+- 18F
+- GSA Common Acquisition Platform
+
+# Brief descriptions of significant project developments
+milestones:
+- 
+
+# Technologies used to build the project
+stack:
+- Rails
```
mbland: May also want to throw `- Ruby` in here, and any other relevant technologies you can think of.

.about.yml
```diff
@@ -0,0 +1,109 @@
...
+impact:
+
+# Services used to supply project status information
+# Items:
+# - name: Name of the service
+#   category: Type of the service
+#   url: URL for detailed information
+#   badge: URL for the status badge
+services:
+-
```
mbland: You got lots of badges in your README; should list them here. (Which gets me thinking, perhaps we need an `about_yml_update_readme` command...)

.about.yml
```diff
@@ -0,0 +1,109 @@
...
+services:
+- 
+
+# Licenses that apply to the project and/or its components (required)
+# Items by property name pattern:
+#   .*:
+#     name: Name of the license from the Software Package Data Exchange (SPDX): https://spdx.org/licenses/
+#     url: URL for the text of the license
+licenses: 
+- name: "CC0-1.0"
```
mbland: This field is a little funky; it should look like:

```yaml
licenses:
  C2:
    name: "CC0-1.0"
    url: https://github.com/18F/C2/blob/master/LICENSE.md
```

.about.yml
```diff
@@ -0,0 +1,109 @@
...
+# Blogs or websites associated with project development
+blog:
+- 
+
+# Links to project artifacts
+# Items:
+# - url: URL for the link
+#   text: Anchor text for the link
+links:
+- "https://c2.18f.gov/"
```
mbland: These should be of the form:

```yaml
links:
- url: https://c2.18f.gov/
- text: Communicart website
```

.about.yml
```diff
@@ -0,0 +1,109 @@
...
+
+# Email addresses of points-of-contact
+contact:
+- gatewaycommunicator@gsa.gov
```
mbland: The new format for `contact:` is:

```yaml
- url: mailto:gatewaycomunicator@gsa.gov
  text: C2 Gateway Communicator
- url: https://github.com/18F/C2/issues
  text: C2 GitHub issues
```

pkarman:
Thanks for the reviews, @afeld and @mbland -- I think I've addressed them all.

.about.yml
pkarman: The product name is changing.

PR #654, opened 2015-10-02
[Delivers #104925358] Hide reason field until an observer has been selected

yozlet:
Broken out from #649. @pkarman has already reviewed & approved the code. @phirefly, OK to merge?


app/assets/javascripts/hidden_until_select.js.coffee
```diff
@@ -0,0 +1,16 @@
+class HiddenUntilSelect
+  constructor: ($root, $hidden_el) ->
+    @$hidden_el = $hidden_el
+    @$controller = $root.find("##{ $hidden_el.attr('data-hide-until-select') }")
+    @$controller.keyup => @checkHide()
+    @$controller.change => @checkHide()
```
afeld: Are both of these needed? If so, maybe worth an inline comment.

app/assets/javascripts/hidden_until_select.js.coffee
```diff
@@ -0,0 +1,16 @@
+class HiddenUntilSelect
+  constructor: ($root, $hidden_el) ->
+    @$hidden_el = $hidden_el
```
afeld: Two very minor things:

* CoffeeScript methods can take a property name prefixed with `@`, which automatically does assignment. These two lines could be collapsed to

    ```javascript
    constructor: ($root, @$hidden_el) ->
    ```

* In JS, the convention is to use camelCase names for variables (`$hiddenEl` vs. `$hidden_el`)

app/assets/javascripts/hidden_until_select.js.coffee
```diff
@@ -0,0 +1,16 @@
...
+    @$controller.keyup => @checkHide()
+    @$controller.change => @checkHide()
+    @checkHide()
+
+  checkHide: ->
+    @$hidden_el.toggle(@$controller.val())
+
+$ ->
+  # @todo - better scope
+  $scope = $(document)
```
afeld: Minor: how about `document.body`?

cmc333333:
Quick note: there's already a `Filter` class. It might make sense to just extend that to fit your use case?

app/assets/javascripts/hidden_until_select.js.coffee
afeld: Also, `@$hidden_el` is a somewhat misleading name for an element that will eventually be shown. Maybe `@$dependent`?

spec/features/observer_spec.rb
```diff
@@ -53,4 +53,28 @@
...
+    proposal = FactoryGirl.create(:proposal)
+    observer = FactoryGirl.create(:user)
+    login_as(proposal.requester)
+
+    visit "/proposals/#{proposal.id}"
+    submit_button = find("#add_subscriber")
+    expect(submit_button).to be_disabled
+    fill_in_selectized('observation_user_email_address', observer.email_address)
+    expect(submit_button).to_not be_disabled
+  end
```
afeld: Later TODO: do front-end testing after #658 is in.

app/assets/javascripts/hidden_until_select.js.coffee
```diff
@@ -0,0 +1,16 @@
...
+  $scope.find("[data-hide-until-select]").each (idx, el) ->
+    new HiddenUntilSelect($scope, $(el))
```
afeld: Not necessarily a blocker, but this whole class seems like very similar functionality to the existing [Filter](https://github.com/18F/C2/blob/master/app/assets/javascripts/filter.js.coffee)...we should look to combine them.

afeld: Jinx @cmc333333.

yozlet: So, after looking at Filter and Aidan's upcoming work, and trying to work out the best way to refactor it all to cover this case, I couldn't see how to do it without adding twice as much code. What Filter does is similar but far enough from identical that it requires a load of extra thought to cover this case.

So I'm going to stick with this solution for the moment, and then we can refactor once we get greater overlap in functionality.

yozlet:
I added a comment which Github seems to have eaten, the gist of which was: while Filter has similar functionality, it's far enough from what's needed here that adding the functionality to Filter would require twice as much code as simply adding this class. But if there's more overlap in the future, we should definitely merge the two.

app/assets/javascripts/hidden_until_select.js.coffee
```diff
@@ -0,0 +1,14 @@
...
+  constructor: ($root, @$dependent) ->
+    @$controller = $root.find("##{ @$dependent.attr('data-hide-until-select') }")
+    @$controller.change => @checkHide()
+    @checkHide()
+
+  checkHide: ->
+    @$dependent.toggle(@$controller.val())
+
+$ ->
+  # @todo - better scope
```
jessieay: remove `todo` ?

jessieay: my view on todos: if it's important but not in scope, it should be in Tracker. If it's important but in scope: should be done now. If it's not important: don't do it!

spec/support/feature_spec_helper.rb
```diff
@@ -10,4 +10,11 @@ def login_as(user)
     setup_mock_auth(:myusa, user)
     visit '/auth/myusa'
   end
+
+  # adapted from http://stackoverflow.com/a/25047358
+  def fill_in_selectized(key, *values)
```
jessieay: let's put this in the `observer_spec` file as a method inside the `describe` block so it's scoped. Easier to see what's going on that way!

spec/features/observer_spec.rb
yozlet: @afeld: Not entirely sure what you mean here. Fancy adding it as a story?

afeld: [Done](https://www.pivotaltracker.com/story/show/105635126).

coffeelint.json
```diff
@@ -0,0 +1,135 @@
...
+  "space_operators": {
+    "level": "ignore"
+  },
+  "spacing_after_comma": {
+    "level": "ignore"
+  },
+  "transform_messes_up_line_numbers": {
+    "level": "warn"
+  }
+}
```
afeld: This should probably be in a separate pull request, but please hook it up as part of the CircleCI build! Feel free to move that setup (with the addition of this file) to its own story if you like.

afeld: Yeah, actually, I'm not sure that I agree with all of these settings so if we could have that discussion separately, that would be great :wink:

afeld:
Just the one comment, to move the adding of `coffeelint.json` to its own PR...otherwise :ship:!

PR #656, opened 2015-10-04
[Delivers #96851664] simplify form specification

afeld:
This pull request accomplishes a few things:

* Uses [Simple Form](https://github.com/plataformatec/simple_form) to generate the appropriate form fields and labels with things like `required` attributes automatically, based on the schema and validations
* Converts the use case forms to HAML
* (Most) written text was moved from the form templates to internationalization config files, separating them from the form logic and making life easier if we ever wanted to provide alternate translations at some point

While the overall number of lines went up (mostly from the generated Simple Form config files), 172 lines were removed from `app/views/`, and IMHO those files are now much easier to follow.

While there were some minor changes to the form itself (fixing capitalization inconsistencies, etc.), both should otherwise be the same. I've looked at the old and new forms side-by-side to verify, but would love to be double-checked by whomever is doing QA. Thanks!


lib/templates/erb/scaffold/_form.html.erb
```diff
@@ -0,0 +1,13 @@
...
+  <div class="form-inputs">
+  <%- attributes.each do |attribute| -%>
+    <%%= f.<%= attribute.reference? ? :association : :input %> :<%= attribute.name %> %>
+  <%- end -%>
+  </div>
+
+  <div class="form-actions">
+    <%%= f.button :submit %>
+  </div>
+<%% end %>
```
afeld: This file was generated by Simple Form...I've never customized [scaffold templates](http://guides.rubyonrails.org/generators.html#customizing-your-workflow-by-changing-generators-templates) before, and am somewhat indifferent about whether we want to keep it or not. Thoughts?

config/locales/simple_form.en.yml
```diff
@@ -0,0 +1,38 @@
...
+        additional_info: "Please include any additional information you feel we need to correctly execute the purchase"
+      ncr_work_order:
+        direct_pay: "I am going to be using direct pay for this transaction"
+        emergency: "This request was an emergency and I received a verbal Notice to Proceed (NTP)"
+    placeholders:
+      ncr_work_order:
+        building_number: "Type here to search or add"
+        project_title: "e.g. \"Blue paint for the Blue Room\""
+        vendor: "Type here to search or add"
+    # TODO have simple_form set up the aria-describedby
```
afeld: This is the only known regression by switching to Simple Form.

afeld:
Some testing notes (will add as I think of things):

* [ ] Check the Work Order form, relative to the current
* [ ] Check the Procurements form, relative to the current
* [ ] Ensure that the Work Order expense type radio buttons still work as expected

app/models/gsa18f/procurement.rb
```diff
@@ -19,10 +19,12 @@ def self.purchase_amount_column_name
     include ProposalDelegate
     include PurchaseCardMixin
 
+    validates :cost_per_unit, presence: true
     validates :quantity, numericality: {
       greater_than_or_equal_to: 1
-    }
+    }, presence: true
     validates :product_name_and_description, presence: true
+    # TODO add conditional validations for recurring charges
```
jessieay: why not do this now!?

app/views/gsa18f/procurements/form.html.haml
```diff
@@ -0,0 +1,25 @@
+- content_for :title, "GSA 18f Purchase Request"
```
jessieay: this form is BEAUTIFUL (esp compared to old form) ...are we generally moving everything to HAML in this project?

app/views/ncr/work_orders/_attachments.html.haml
```diff
@@ -0,0 +1,6 @@
+.form-group
+  = field_set_tag 'Attachments' do
+    %ul.attachments{ data: { add_minus: true } }
+      - 10.times do
```
jessieay: Right now, `10` feels like a mystery number. I'm sure there is a reason behind allowing 10 attachments, but since I just see it here, it feels like maybe it is a typo or something random...what about using a constant to store the value?

config/initializers/simple_form.rb
```diff
@@ -0,0 +1,166 @@
+# Use this setup block to configure all options available in SimpleForm.
```
jessieay: I think it would be so very lovely to remove the comments from this initializer since it is long and realistically we will revisit the documentation to update the `simple_form` configuration anyway (not to mention that these comments will become outdated as we upgrade our version of `simple_form` over time)

config/locales/en.yml
```diff
@@ -36,6 +36,8 @@ en:
         create: "Add"
   activerecord:
     attributes:
+      gsa18f/procurement:
+        recurring: "Recurring charge"
```
jessieay: so we are using I18n! Interesting. How strict are we being about using it? Seems like there is quite a bit of text in the view templates...

config/locales/simple_form.en.yml
```diff
@@ -0,0 +1,38 @@
...
+    hints:
+      gsa18f_procurement:
+        link_to_product: "Search <a href='https://www.gsaadvantage.gov/advantage/main/start_page.do' target='_blank'>GSA Advantage</a> first; then <a href='http://www.amazon.com' target='_blank'>Amazon;</a> then other vendor. Paste link here."
+        recurring_length: "e.g. Number of Days, Months, Years"
+      ncr_work_order:
+        description: "Include request details (e.g. Room Number, Estimated number of hours)"
+        org_code: "Select your Service Center."
+    # include_blanks:
```
jessieay: remove commented code?

lib/templates/erb/scaffold/_form.html.erb
jessieay: huh! I don't usually use these. Seems like a cool idea...but if we're moving to HAML, I might suggested HAMLizing it. And if that's too much work, then maybe it isn't worth keeping around. 

If someone else has used this and loved it, I would love to hear more!

spec/features/18f_procurement_spec.rb
```diff
@@ -27,7 +27,7 @@
       expect(Dispatcher).to receive(:deliver_new_proposal_emails)
 
       visit '/gsa18f/procurements/new'
-      fill_in "Product Name and Description", with: "buying stuff"
+      fill_in "Product name and description", with: "buying stuff"
```
jessieay: this is where having full usage of I18n would be awesome - could change copy and not have to update associated specs

app/models/gsa18f/procurement.rb
afeld: Trying to stay focused! Geez. :wink: Will take care of it in a follow-up PR.

lib/templates/erb/scaffold/_form.html.erb
afeld: I can't imagine us doing much scaffolding moving forward...feel like there's a greater of it getting out-of-date. :bomb:

app/views/gsa18f/procurements/form.html.haml
afeld: [Yep](https://www.pivotaltracker.com/story/show/98506386)! [Slowly](https://github.com/18F/C2/pull/488).

app/views/ncr/work_orders/_attachments.html.haml
afeld: I believe it _is_ an arbitrary number...correct @cmc333333? https://github.com/18F/C2/pull/380 Moving to a constant.

config/locales/en.yml
afeld: We haven't been trying at all...I was just looking for ways to take as much as possible out of the templates. The fact that it would make internationalization easier is a (distant) secondary benefit.

spec/features/18f_procurement_spec.rb
afeld: Would you prefer this used the element `name`/`id` instead?

jessieay: Not sure I understand what you're asking? I was just noticing that we had to touch a lot of code (in specs) to make a small copy change, didn't intend to comment on the change itself

afeld: I mean changing these lines to

```ruby
fill_in 'gsa_18f_procurement[product_name_and_description]', with: "buying stuff"
```

The downside there is that you're relying on generated `name` attributes, which seem like they should be more of an implementation detail. How would you do it?

afeld:
Updated!

spec/features/18f_procurement_spec.rb
jessieay: Yeah I think the current state is better than using `name` or another field that is not visible to the user.

I would do it by using I18n! so something like

 `fill_in t('forms.product.name_and_description'), with: buying stuff"`

...but if we aren't using a lot of I18n, that might not make sense for this PR

afeld: Mind if we punt on that? :football: :runner: `t()` doesn't work unless we enter translations for each attribute explicitly, which doesn't seem worth the effort.

jessieay: No prob at all to punt - this was more of a general comment about what we might want to more toward in the future. I kind of hate I18n because the translation keys can get super long, but once you have it all set up it is nice to manage all app copy in one place....so something to think about

config/locales/en.yml
jessieay: I vote that we either try to move as much to the locales or try not to use them at all. Being split on I18n could get kinda confusing :hammer:

app/models/ncr/work_order.rb
```diff
@@ -26,6 +26,7 @@ def self.purchase_amount_column_name
     before_validation :normalize_values
     before_update :record_changes
 
+    validates :amount, presence: true
```
jessieay: spec for this? Looks like we don't have any validation specs in the spec file for this model -- would love to hear about why! Is there a philosophical opposition to specs for AR validations (I know many hold the view that they are not important because it's basically just double checking that you wrote a validation rather than testing behavior)

app/models/gsa18f/procurement.rb
jessieay: What about putting it in tracker!? 

Totally makes sense to do things separately, but because there is so little room for context here it becomes a to do for *you* rather than anyone on the team who might want to pitch in - LET US HELP (it may be the case that others could guess which conditional validations you are referring to, but I am so new here that I really have no idea)

app/controllers/ncr/work_orders_controller.rb
```diff
@@ -1,5 +1,8 @@
 module Ncr
   class WorkOrdersController < UseCaseController
+   # arbitrary number...number of upload fields that "ought to be enough for anybody"
+    MAX_UPLOADS_ON_NEW = 10
```
jessieay: :clap:

jessieay:
@afeld this is failing :(

config/locales/en.yml
afeld: Well, my thinking is that we put all _overrides_ for fields and such in i18n files...I don't think we need to do have entries for every single piece of text that gets displayed in the app. Again, the goal with this PR was making the use case form templates easy to read/change.

app/models/ncr/work_order.rb
afeld: Yeah, I normally don't write specs for "simple" validations for that reason (I would for things like `format`, like below), but am open to adding if you think it's important.

jessieay: I think not having it for this is fine since we don't have any other simple validation specs. 

I know a lot of people don't like simple validation specs because it's kind of like testing Rails, but I do like it them because using them makes specs a full record of documentation for the code they cover. It's nice being able to look into a spec file and know everything you need to know about an object. 

BUT that is something for next time - for now, seems like a validation isn't needed.

afeld:
Updated!

app/controllers/ncr/work_orders_controller.rb
```diff
@@ -26,7 +29,7 @@ def update
 
       super
 
-      if @model_changing && !@model_instance.emergency  # skip approvals if emergency
+      if @model_changing && !@model_instance.emergency # skip approvals if emergency
```
jessieay: I think `@model_instance.not_emergency?` would prevent the need for a comment here and would be clearer to read since it's easier to miss the `!`

config/initializers/simple_form_bootstrap.rb
```diff
@@ -0,0 +1,148 @@
+SimpleForm.setup do |config|
```
jessieay: I still know very little about 508 compliance, so just wondering if there might be problems with using bootstrap in that context? I read somewhere that bootstrap isn't the greatest for accessibility, but I am not sure what using `simple_form_bootstrap` is doing exactly here so not sure if that will be a problem

afeld: > I read somewhere that bootstrap isn't the greatest for accessibility

Yeah, I think that is more about the [Components](http://getbootstrap.com/components/) (which I don't think we're using...?) than the [CSS](http://getbootstrap.com/css/), but I don't know too many details. Googling brought up

* http://getbootstrap.com/getting-started/#accessibility
* https://paypal.github.io/bootstrap-accessibility-plugin/

> I am not sure what using simple_form_bootstrap is doing exactly here

Yeah, I find the SimpleForm element config syntax very confusing :confused: but this is what it generated out of the box.

> not sure if that will be a problem

We were already using Bootstrap, and the generated HTML is very similar to what we had before (though my investigation wasn't exhaustive). At the very least there shouldn't be any regressions in terms of accessibility, besides https://www.pivotaltracker.com/story/show/105961466.

jessieay: thanks for the info!

jessieay:
Code looks good! :ship:

jessieay:
copy-pasting from Slack: I have a process question (and am happy to update doc after I get an answer) - how do we indicate who is doing QA for something if it is not the reviewer? I am about to QA https://github.com/18F/C2/pull/656 — should I assign myself the PR? make a comment on the PR saying I am QAing? Make a comment on pivotal

jessieay:
Only difference I found in forms is procurements form: new is on left, old is on right. Don't think this should be problematic 

screen shot 2015-10-19 at 3 14 49 pm

jessieay:
QA done! MERGING

Issue #565, opened 2015-08-25
onboarding for @pkarman


afeld:
## Tasks for @pkarman

* [x] [Set up your development environment](https://github.com/18F/C2/blob/master/doc/setup.md)
* [x] Go through the rest of the [general documentation](https://github.com/18F/C2#general)

### Later

* [x] [Set up the Cloud Foundry CLI](https://docs.18f.gov/getting-started/setup/) (which you will need to deploy)
* [x] [Learn more about Cloud Foundry](https://docs.18f.gov)
* [x] Deploy C2 to `c2-dev` (or `c2-staging`)

## Tasks for @afeld

* [x] Add to Slack channels
* [x] [Add to Pivotal Tracker](https://www.pivotaltracker.com/projects/1149728/memberships)
* [x] Add to [@18F/cap](https://github.com/orgs/18F/teams/cap) team on GitHub
* [x] Add to calendar items: standups, IPM, and story grooming sessions
* [x] Schedule a code walkthrough
* [x] Set up pairing session
* [x] Give intro to [current stories](https://pivotaltracker.com/n/projects/1149728)
* [x] Give intro to weekly ceremonies and team workflow

### Later

* [x] [Add them to `cap` organization on Cloud Foundry](https://docs.cloudfoundry.org/adminguide/cli-user-management.html#org-roles) (`cf set-org-role USERNAME cap OrgManager`)
* [ ] Admin access to C2 production (add to `ADMIN_EMAILS`)

pkarman:
I think I've done everything, and just created #584 to reflect changes to this for the next person.

PR #658, opened 2015-10-04
[Delivers #98157690] add client-side form validation

afeld:
Depends/builds on #656. The diff:

https://github.com/18f/c2/compare/96851664-simplify-forms...98157690-client-side-validation

## Changes

* Enables automatic HTML5 form validation via Simple Form
* Includes a refactoring of the client-side form filtering code
* `disable`s the form fields that are hidden by the filter, so that they:
    * Don't get submitted with the form, e.g. sending the `recurring_interval` value when `recurring` isn't checked
    * Don't trigger client-side validation warnings
* Set up JS testing using [Konacha](https://github.com/jfirebaugh/konacha)
* Add tests around the filtering

Yay for limited-internet time to work on this! :steam_locomotive:

## Testing notes

* [ ] Ensure that client-side validation works (or at least doesn't break the form) for the normal fields (ideally, in multiple browsers)
* [x] Ensure that empty `required` hidden fields (e.g. the BA80 fields when `BA61` is selected) don't block submissions of the form
* [x] Ensure that values from hidden fields aren't getting sent to the server
* [x] Ensure that Selectize-d fields (the ones with the fancy dropdowns) aren't broken


jessieay:
Will review after #565 is finished/merged since it seems like there is a good amount of overlap

afeld:
Ready for some :eyes: !

Gemfile
```diff
@@ -53,6 +53,7 @@ gem 'workflow'
 group :test, :development do
   gem 'bullet', require: false # use BULLET_ENABLED=true
   gem 'database_cleaner'
+  gem 'konacha'
```
jessieay: niiiice

Gemfile
```diff
@@ -61,6 +62,7 @@ end
 
 group :development do
   gem 'guard-rspec', require: false
+  gem 'guard-shell', require: false
```
jessieay: seems like this is specific to a particular dev setup and not needed in the Gemfile. 

> dependencies of a developer utility should not be able to interfere with the dependencies of your app.

https://github.com/ddollar/foreman/pull/437#issuecomment-41110407

Guardfile
```diff
@@ -15,3 +15,8 @@ guard :rspec, cmd: 'bin/rspec' do
   # Capybara features specs
   watch(%r{^app/views/(.+)/.*\.(erb|haml|slim)$})     { |m| "spec/features/#{m[1]}_spec.rb" }
 end
+
+guard :shell do
```
jessieay: I am not using guard locally - I am fine with keeping this (and guard-shell) gem locally, but just an FYI that it isn't necessary for all (read: my) dev environments

app/assets/javascripts/field_filter.js.coffee
```diff
@@ -0,0 +1,17 @@
+class @FieldFilter
+  constructor: (@$fieldOrWrappers) ->
+
+  toggle: (showOrHide) ->
+    # https://www.paciellogroup.com/blog/2012/05/html5-accessibility-chops-hidden-and-aria-hidden/
+    @$fieldOrWrappers.attr('aria-hidden', !showOrHide)
+
+    # disable inputs so they aren't submitted with the form
+    if @$fieldOrWrappers.is(':input')
```
jessieay: what about putting this in a well-named function rather than adding a comment about what it does!?

app/assets/javascripts/filter.js.coffee
```diff
@@ -1,45 +1,29 @@
...
-      key = el.getAttribute('data-filter-control')
-      if !filters.hasOwnProperty(key)
-        filters[key] = new Filter($scope, key)
-    filters
+  @generateIn = ($scope) ->
+    $scope.find('[data-filter-control]').map (idx, control) ->
+      new Filter($scope, $(control))
 
 $ ->
   #  @todo - better scope
```
jessieay: remove comment since scope was changed?

app/assets/javascripts/filter_set.js.coffee
```diff
@@ -0,0 +1,27 @@
+class @FilterSet
+  constructor: (@$root, @key, @val) ->
+
+  $: (selector) ->
+    @$root.find(selector)
+
+  children: ->
+    @$("[data-filter-key=#{ @key }][data-filter-value=#{ @val }]")
+
+  adjacentChildren: ->
```
jessieay: siblings?

app/assets/javascripts/filter_set.js.coffee
```diff
@@ -0,0 +1,27 @@
...
+    @$("[data-filter-key=#{ @key }][data-filter-value!=#{ @val }]")
+
+  showChildren: ->
+    filter = new FieldFilter(@children())
+    filter.show()
+
+  hideAdjacentChildren: ->
```
jessieay: `hideSiblings`?

app/assets/javascripts/selectizer.js.coffee
```diff
@@ -3,6 +3,10 @@ class Selectizer
     @$el = $(el)
     @dataAttr = @$el.attr('data-attr') || 'default_field'
 
+    # `required` inputs don't work with Selectize
```
jessieay: I think linking to the gh issue is great for documenation; summary of gh issue not necesssary

doc/setup.md
```diff
@@ -79,7 +79,8 @@ have it in your PATH. This is used for javascript and interface testing.
 ### Running the entire suite once
 
 ```bash
-rake
+./bin/rake
```
jessieay: rake is so much easier to type! what about including `konacha:run` in default rake task so we can use it!?

Gemfile
afeld: Mind if we have that discussion separately?

app/assets/javascripts/filter_set.js.coffee
afeld: > Siblings (brothers or sisters) are nodes with the same parent

http://www.w3schools.com/js/js_htmldom_navigation.asp

This method could be named `childrenOfSiblingsOfParent`, but that doesn't seem much clearer :confused: Thoughts?

app/assets/javascripts/filter_set.js.coffee
afeld: Ditto above.

app/assets/javascripts/selectizer.js.coffee
afeld: Not necessary, but I think it's helpful in this case.

afeld:
Updated!

doc/setup.md
afeld: > rake is so much easier to type!

...but it doesn't run via Spring, which makes the startup significantly faster.

afeld: > what about including `konacha:run` in default rake task so we can use it!?

:+1:

pkarman:
I will QA

pkarman:
In general this looks good and doesn't seem to introduce any regressions.

There are some bugs though. Notably, an empty "Approver" field does not seem to trigger the "required" check. At first glance this seems to be because the HAML markup for that particular field is different than the others.

Likewise, empty "Building Code" and "Vendor" dropdown fields are not caught as required either.

Server-side validation still works as expected, though, so this might be acceptable in this first pass.
I'd be willing to merge this as a baseline implementation, and then open some bug stories to describe the missing pieces. Or, I can reject the existing story and you can give it another pass. Up to you @afeld.

Gemfile
jessieay: sure!

app/assets/javascripts/filter_set.js.coffee
jessieay: ah interesting - wonder if this might be a good use for `aria-owns` http://tink.uk/using-the-aria-owns-attribute/ ? (truly not sure, just saw that post and thought it was interesting)

afeld: hahaha just Googled "children of siblings of parent" to see if there was a word for it...yeeeeah :grin: 

`aria-owns` is interesting (didn't know about that before), but I think that's tangential.

afeld:
Yeah, unfortunately client-side validation doesn't work with Selectize fields (https://github.com/brianreavis/selectize.js/issues/733). Client-side validation is a convenience, so I personally don't think it's a blocker if it doesn't work across every field.

PR #659, opened 2015-10-05
Updated #current_approver_email_address 

phirefly:
After testing on staging, I'm still getting the same output as previous. I think we need to check for 'actionable'

### DIFF
--- a/app/models/ncr/work_order.rb
+++ b/app/models/ncr/work_order.rb
@@ -116,8 +116,8 @@ module Ncr
 
     # the highest approver on the stack, pending preferred if status indicates
     def current_approver_email_address
-      if self.individual_approvals.pending.first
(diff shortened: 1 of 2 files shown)


app/models/ncr/work_order.rb
```diff
@@ -116,8 +116,8 @@ def approving_official
 
     # the highest approver on the stack, pending preferred if status indicates
     def current_approver_email_address
-      if self.individual_approvals.pending.first
-        self.individual_approvals.pending.first.user.email_address
+      if self.pending?
+        self.individual_approvals.where(status: 'actionable').first.user.email_address
```
phirefly: Checking for actionable here instead of grabbing the first pending approval. I tested this with proposal ID 265 on staging.

pkarman:
you're right: actionable it is.

PR #660, opened 2015-10-05
refactor WorkOrder model/controller to use an attribute for the approving official email

afeld:
Builds/depends on #656. The diff:

https://github.com/18f/c2/compare/96851664-simplify-forms...96851664-simplify-forms-B

### DIFF
--- a/app/controllers/ncr/work_orders_controller.rb
+++ b/app/controllers/ncr/work_orders_controller.rb
@@ -4,12 +4,12 @@ module Ncr
     MAX_UPLOADS_ON_NEW = 10
 
     def new
-      @approver_email = self.suggested_approver_email
+      @model_instance.approving_official_email = self.suggested_approver_email
       super
     end
 
     def create
-      @approver_email = params[:approver_email]
+      @model_instance.approving_official_email = params[:approver_email]
       super
     end
 
@@ -17,29 +17,30 @@ module Ncr
       if self.proposal.approved?
         flash[:warning] = "You are about to modify a fully approved request. Changes will be logged and sent to approvers but this request will not require re-approval."
       end
-      first_approver = self.proposal.approvers.first
-      @approver_email = first_approver.try(:email_address)
 
       super
     end
 
     def update
-      @approver_email = params[:approver_email]
+      # TODO remove need for this
+      @model_instance.approving_official_email = params[:approver_email]
       @model_instance.modifier = current_user
 
       super
 
       if @model_changing
-        @model_instance.setup_approvals_and_observers(@approver_email)
+        # TODO remove need for this
+        @model_instance.setup_approvals_and_observers(@model_instance.approving_official_email)
         @model_instance.email_approvers
       end
     end
 
-    def attribute_changes?
-      super || @model_instance.approver_changed?(@approver_email)
-    end
-
     protected
+
+    def attribute_changes?
+      # TODO remove need for passing in the approver_email
+      super || @model_instance.approver_changed?(@model_instance.approving_official_email)
+    end
 
     def model_class
       Ncr::WorkOrder
@@ -61,17 +62,18 @@ module Ncr
 
     def errors
       results = super
-      if @approver_email.blank? && !@model_instance.approver_email_frozen?
+      if @model_instance.approving_official_email.blank? && !@model_instance.approver_email_frozen?
         results += ["Approver email is required"]
       end
       results
     end
 
-    # @pre: @approver_email is set
+    # @pre: @model_instance.approving_official_email is set
     def add_approvals
       super
       if self.errors.empty?
-        @model_instance.setup_approvals_and_observers(@approver_email)
+        # TODO remove need for passing in the approver_email
+        @model_instance.setup_approvals_and_observers(@model_instance.approving_official_email)
       end
     end
   end
--- a/app/controllers/ncr/work_orders_controller.rb
+++ b/app/controllers/ncr/work_orders_controller.rb
@@ -34,8 +34,7 @@ module Ncr
     protected
 
     def attribute_changes?
-      # TODO remove need for passing in the approving_official_email
-      super || @model_instance.approver_changed?(@model_instance.approving_official_email)
+      super || @model_instance.approver_changed?
     end
 
     def model_class
--- a/app/controllers/ncr/work_orders_controller.rb
+++ b/app/controllers/ncr/work_orders_controller.rb
@@ -56,14 +56,6 @@ module Ncr
       params.require(:ncr_work_order).permit(:project_title, :approving_official_email, *fields)
     end
 
-    def errors
-      results = super
-      if @model_instance.approving_official_email.blank? && !@model_instance.approver_email_frozen?
-        results += ["Approver email is required"]
-      end
-      results
-    end
-
     # @pre: @model_instance.approving_official_email is set
     def add_approvals
       super
--- a/app/controllers/ncr/work_orders_controller.rb
+++ b/app/controllers/ncr/work_orders_controller.rb
@@ -9,7 +9,6 @@ module Ncr
     end
 
     def create
-      @model_instance.approving_official_email = params[:approver_email]
       super
     end
 
@@ -22,8 +21,6 @@ module Ncr
     end
 
     def update
-      # TODO remove need for this
-      @model_instance.approving_official_email = params[:approver_email]
       @model_instance.modifier = current_user
 
       super
@@ -38,7 +35,7 @@ module Ncr
     protected
 
     def attribute_changes?
-      # TODO remove need for passing in the approver_email
+      # TODO remove need for passing in the approving_official_email
       super || @model_instance.approver_changed?(@model_instance.approving_official_email)
     end
 
@@ -57,7 +54,7 @@ module Ncr
       if @model_instance
         fields.delete(:emergency) # emergency field cannot be edited
       end
-      params.require(:ncr_work_order).permit(:project_title, *fields)
+      params.require(:ncr_work_order).permit(:project_title, :approving_official_email, *fields)
     end
 
     def errors
@@ -72,7 +69,7 @@ module Ncr
     def add_approvals
       super
       if self.errors.empty?
-        # TODO remove need for passing in the approver_email
+        # TODO remove need for passing in the approving_official_email
         @model_instance.setup_approvals_and_observers(@model_instance.approving_official_email)
       end
     end
--- a/app/controllers/ncr/work_orders_controller.rb
+++ b/app/controllers/ncr/work_orders_controller.rb
@@ -26,8 +26,7 @@ module Ncr
       super
 
       if @model_changing
-        # TODO remove need for this
-        @model_instance.setup_approvals_and_observers(@model_instance.approving_official_email)
+        @model_instance.setup_approvals_and_observers
         @model_instance.email_approvers
       end
     end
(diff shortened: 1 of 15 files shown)


app/controllers/ncr/work_orders_controller.rb
```diff
@@ -54,22 +50,14 @@ def permitted_params
...
       end
-      params.require(:ncr_work_order).permit(:project_title, *fields)
-    end
-
-    def errors
-      results = super
-      if @approver_email.blank? && !@model_instance.approver_email_frozen?
-        results += ["Approver email is required"]
-      end
-      results
```
afeld: No longer need to do manual validation in the controller!

app/views/ncr/work_orders/form.html.haml
```diff
@@ -0,0 +1,34 @@
...
+      = expense_type_radio_button(f, 'BA80')
+      = render partial: 'ba_80_fields', locals: {f: f}
+    = f.input :building_number, input_html: { class: 'js-selectize', data: { initial: JSON.generate(building_options) } }
+    = f.input :org_code, collection: Ncr::Organization.all, label_method: :to_s, prompt: :translate, input_html: { class: 'js-selectize' }
+    = f.input :vendor, input_html: { class: 'js-selectize', data: { initial: JSON.generate(vendor_options(@model_instance.vendor)) } }
+    = field_set_tag "Amount", class: 'required' do
+      = f.input :amount, as: :currency, label_html: { class: 'sr-only' }
+      = f.input :not_to_exceed, as: :radio_buttons, collection: [['Exact', false], ['Not to exceed', true]], label: false
+    = f.input :direct_pay
+    = f.input :approving_official_email, collection: approver_options, include_blank: true, disabled: @model_instance.approver_email_frozen?, prompt: :translate, input_html: { class: 'js-selectize' }
```
afeld: Switched to use Simple Form for the field.

afeld:
Ok, ready! This is just a refactoring – no intended change in behavior.

## Changes

* Treat `approving_official_email` as an attribute on the WorkOrder model, which has a few benefits – commented inline.
* Simplified the Ncr::WorkOrdersController

## Testing notes

* [ ] Ensure the approving official dropdown appears the same as before
* [ ] Ensure you can create Work Orders (thoroughly covered by tests)
* [ ] Ensure you can modify Work Orders (thoroughly covered by tests)

pkarman:
@afeld is this PR still valid? It appears to have been against a PR that was already merged, and currently has conflicts.

afeld:
Updated!

app/controllers/use_case_controller.rb
```diff
@@ -66,6 +60,18 @@ def attribute_changes?
...
 
+  def build_model_instance
+    if params[:action] == 'new'
+      filtered_params = {}
+    else
+      filtered_params = self.permitted_params
+    end
+    @model_instance = self.model_class.new(filtered_params)
+
+    # TODO unify with how the factories create model instances
```
pkarman: does this TODO have a PT story also?

jessieay: please let's make all todos into PT stories instead! :pray:

spec/factories/ncr/work_orders.rb
```diff
@@ -9,10 +9,15 @@
     rwa_number "R1234567" # TODO remove, since it's not applicable for BA61
     org_code Ncr::Organization.all[0].to_s
     project_title "NCR Name"
+    sequence(:approving_official_email) {|n| "approver#{User.count}@some-cartoon-show.com" }
     association :proposal, flow: 'linear'
 
     trait :with_approvers do
+      # TODO use wo.setup_approvals_and_observers instead
```
pkarman: PT story for this?

app/controllers/ncr/work_orders_controller.rb
```diff
@@ -4,12 +4,11 @@ class WorkOrdersController < UseCaseController
     MAX_UPLOADS_ON_NEW = 10
 
     def new
-      @approver_email = self.suggested_approver_email
+      @model_instance.approving_official_email = self.suggested_approver_email
```
jessieay: I tend to use `self` for class methods only - thoughts on just using `suggested_approver_email` here?

jessieay: is this controller being used for more than one type of object? of not, I think something like `@ncr_work_order` would be a clearer var name than `@model_instance`

spec/models/ncr/work_order_spec.rb
```diff
@@ -466,32 +468,37 @@
...
-      wo.individual_approvals.first.approve!
-      wo.reload.individual_approvals.last.approve!
-      expect(wo.final_approver).to eq(wo.individual_approvals.last.user)
-    end 
+      fully_approve(wo)
+      expect(wo.final_approver).to eq(wo.approvers.last)
+    end
+  end
+
+  def fully_approve(work_order)
```
pkarman: I had to invent this myself the other day for a different test. Perhaps this should be a method on WorkOrder itself?

app/controllers/ncr/work_orders_controller.rb
```diff
@@ -17,30 +16,28 @@ def edit
...
+        @model_instance.setup_approvals_and_observers
         @model_instance.email_approvers
       end
     end
 
+    protected
+
     def attribute_changes?
-      super || @model_instance.approver_changed?(@approver_email)
+      # TODO remove need for passing in the approving_official_email
```
jessieay: let's put this in Tracker instead! (and add more details on why etc while we're at it!)

app/controllers/ncr/work_orders_controller.rb
jessieay: woot

app/controllers/use_case_controller.rb
```diff
@@ -66,6 +60,18 @@ def attribute_changes?
     !@model_instance.changed_attributes.blank?
   end
 
+  def build_model_instance
+    if params[:action] == 'new'
+      filtered_params = {}
+    else
+      filtered_params = self.permitted_params
```
jessieay: remove `self`?

app/controllers/use_case_controller.rb
```diff
@@ -66,6 +60,18 @@ def attribute_changes?
     !@model_instance.changed_attributes.blank?
   end
 
+  def build_model_instance
+    if params[:action] == 'new'
```
jessieay: wha about putting this conditional into a method named `filtered_params` and then leaving L69 as-is?

app/controllers/use_case_controller.rb
```diff
@@ -66,6 +60,18 @@ def attribute_changes?
...
+      filtered_params = self.permitted_params
+    end
+    @model_instance = self.model_class.new(filtered_params)
+
+    # TODO unify with how the factories create model instances
+    @model_instance.build_proposal(flow: 'linear', requester: current_user)
+  end
+
   def find_model_instance
     @model_instance ||= self.model_class.find(params[:id])
```
jessieay: without knowing much about this controller, my feedback is that all of this `model_class` and `model_instance` stuff is kinda confusing. Don't currently have a suggestion on how to change it because I am not exactly sure on why it is here to begin with, but thought I'd provide that feedback as a semi-new person to this codebase who is seeing things the way that other new people might see it.

app/models/ncr/work_order.rb
```diff
@@ -57,6 +60,10 @@ def self.purchase_amount_column_name
     }
 
     def set_defaults
+      # not sure why the latter condition is necessary...was getting some weird errors from the tests without it. -AF 10/5/2015
```
jessieay: can we try to get to the bottom of this instead of leaving a comment? Or at least file a tracker story about it?

spec/controllers/ncr/work_orders_controller_spec.rb
```diff
@@ -135,8 +162,14 @@
     it 'will not modify emergency status on non-emergencies' do
       expect(work_order.approvals.empty?).to be false
       expect(work_order.observers.empty?).to be true
-      post :update, {id: work_order.id, approver_email: work_order.approvers.first.email_address,
-                     ncr_work_order: {expense_type: 'BA61', emergency: '1'}}
+      post :update, {
```
jessieay: nice, these are so much more readable now

spec/factories/ncr/work_orders.rb
jessieay: to tracker instead of a comment please!

jessieay:
General comment: thoughts on removing all `self` references that are not for class method? I find them confusing because `self` is required for class methods and option for instance methods. I find myself double checking on whether something is a class or instance method whenever I see `self`

app/controllers/use_case_controller.rb
afeld: In my defense, these TODOs were added before our decision to put them in stories instead.

jessieay: ah! did not realize that. 

sorry to badger you on this front. It is out of selfish desire to be able to help with small refactors. Currently feeling like I am still new enough to this project that inline todos are not actionable for me (don't have enough context yet)

app/controllers/ncr/work_orders_controller.rb
afeld: > I tend to use `self` for class methods only - thoughts on just using `suggested_approver_email` here?

I use `self` for calling instance methods to distinguish them from variable references, but I acknowledge that isn't a common pattern in Ruby. The reasoning, if you're curious is that I've been burned by bugs like

```ruby
something = 'foo'
# sets a local variable, where the intended behavior was actually
self.something = 'foo'
```

Using `self` just avoids that ambiguity. Starting to lean towards only using `self` for assignment methods, though. Will remove.

> is this controller being used for more than one type of object?

This inherits from the UseCaseController, where `@model_instance` is set.

> I think something like `@ncr_work_order` would be a clearer var name than `@model_instance`

I hear you, but mind if we take care of that separately? If nothing else, because there are several open PRs that change WorkOrdersController, and :crying_cat_face: :weary: :-1: more merge conflicts.

/cc https://www.pivotaltracker.com/story/show/101710542

app/controllers/use_case_controller.rb
afeld: https://www.pivotaltracker.com/story/show/106197926

spec/factories/ncr/work_orders.rb
afeld: https://www.pivotaltracker.com/story/show/106197926

pkarman:
I like `self` everywhere it refers to a method, regardless of instance or class. Helps disambiguate variables from methods. I rarely care what kind of method something is since Ruby will croak if I try and use it incorrectly. However, since variables autovivify, particularly for cases like @afeld mentioned with assignment methods, I prefer to clearly see `self.foo = bar`.

That said, I know that's not the Ruby way, which imo, favors concision over clarity.

pkarman:
let's push the style questions to a different story; this poor PR has suffered long enough.

pkarman:
@afeld I will QA if you can resolve whatever conflicts are being reported.

app/models/ncr/work_order.rb
afeld: See #736...makes no sense to me.

afeld:
Phew, took a bunch of hours, but finally got this updated. Ready for another quick look, and QA.

pkarman:
looks good. QA'd on c2-dev.

PR #661, opened 2015-10-05
[Delivers #102764816] db data cleanup for constraints

pkarman:
* zap proposal_role rows pointing at non-existent users
* clean up duplicate role records


annalee:
:shipit:

PR #665, opened 2015-10-05
[Delivers #104924964] Update/document rake task for populating dashboard

jessieay:
* Fixes https://www.pivotaltracker.com/story/show/104924964

### DIFF
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -41,6 +41,30 @@ PORT=3000 ASSETS_DEBUG=true ./script/start
 open http://localhost:3000
 ```
 
+## Seeding your dashboard with data
+
+Once you've authed locally, there will be a `user` record associated with your
+email address. There won't be much for you to see until your client slug is set,
+so find your user record and set it to `ncr`:
+
+```bash
+rails c
+user = User.find_by(email_address: 'example@gsa.gov')
+user.update(client_slug: 'ncr')
+```
+
+Now you will see the link to create a new work order locally.
+
+If you'd like to seed your dashboard with work orders, you can run this rake
+task:
+
+```bash
+rake populate:ncr:for_user[example@gsa.gov]
+```
+
+Now you should see 25 pending purchase requests at
+http://localhost:3000/proposals.
+
 ### Viewing the mailers
 
 As emails are sent, they will be visible at http://localhost:3000/letter_opener. If you are working on an email mailer/template, you can view all of them at http://localhost:3000/mail_view/.
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -41,14 +41,14 @@ PORT=3000 ASSETS_DEBUG=true ./script/start
 open http://localhost:3000
 ```
 
-## Seeding your dashboard with data
+## Populating with data
 
 Once you've authed locally, there will be a `user` record associated with your
 email address. There won't be much for you to see until your client slug is set,
 so find your user record and set it to `ncr`:
 
 ```bash
-rails c
+bin/rails c
 user = User.find_by(email_address: 'example@gsa.gov')
 user.update(client_slug: 'ncr')
 ```
@@ -59,7 +59,7 @@ If you'd like to seed your dashboard with work orders, you can run this rake
 task:
 
 ```bash
-rake populate:ncr:for_user[example@gsa.gov]
+bin/rake populate:ncr:for_user[example@gsa.gov]
 ```
 
 Now you should see 25 pending purchase requests at
--- a/lib/populator.rb
+++ b/lib/populator.rb
@@ -1,56 +1,68 @@
-module Populator
-  # singleton
-  # https://practicingruby.com/articles/ruby-and-the-singleton-pattern-dont-get-along
-  extend self
+class Populator
+  include FactoryGirl::Syntax::Methods
 
-  def uniform_ncr_data(n=10)
-    n.times do |i|
-      FactoryGirl.create(:ncr_work_order, :with_approvers)
+  def uniform_ncr_data(n: 10)
+    n.times do
+      create(:ncr_work_order, :with_approvers)
     end
   end
 
-  def random_bool(pct_true=0.5)
+  def random_ncr_data(num_proposals: 50)
+    num_proposals.times do
+      requested_at = rand(3.months.ago..1.day.ago)
+      proposal = create_proposal(requested_at: requested_at)
+      create_work_order(proposal: proposal)
+      add_comments(proposal: proposal, requested_at: requested_at)
+    end
+  end
+
+  def ncr_data_for_user(email:, num_proposals: 25)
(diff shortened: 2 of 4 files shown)


lib/populator.rb
```diff
@@ -1,56 +1,68 @@
...
+      :with_serial_approvers,
+      :with_observers,
+      requester: requester,
+      created_at: requested_at,
+      updated_at: requested_at,
+    )
+  end
 
-      unless work_order.emergency
-        # TODO randomly approve approvals and proposals at different times
```
jessieay: is this an important TODO that should be added to tracker?

lib/populator.rb
```diff
@@ -1,56 +1,68 @@
...
-    num_proposals.times do |i|
-      requested_at = rand(3.months.ago..1.day.ago)
+    num_proposals.times do
+      proposal = create_proposal(requester: user)
+      create_work_order(proposal: proposal)
+      add_comments(proposal: proposal)
+    end
+  end
 
-      # TODO all of these things should have the same created_at/updated_at... use Timecop
```
jessieay: Is this an important TODO that should be added to tracker?

afeld: Was just a random idea about how this generator could be more life-like...not a huge deal. Could be made a task in a story to "improve the populator".

lib/populator.rb
afeld: Ditto.

doc/setup.md
```diff
@@ -41,6 +41,30 @@ PORT=3000 ASSETS_DEBUG=true ./script/start
...
 ```
 
+## Seeding your dashboard with data
+
+Once you've authed locally, there will be a `user` record associated with your
+email address. There won't be much for you to see until your client slug is set,
+so find your user record and set it to `ncr`:
+
+```bash
+rails c
```
afeld: Minor: use `bin/rails c` (and `bin/rake ...` below) so that the application starts faster (via [Spring](https://github.com/rails/spring)).

doc/setup.md
```diff
@@ -41,6 +41,30 @@ PORT=3000 ASSETS_DEBUG=true ./script/start
 open http://localhost:3000
 ```
 
+## Seeding your dashboard with data
+
+Once you've authed locally, there will be a `user` record associated with your
+email address. There won't be much for you to see until your client slug is set,
+so find your user record and set it to `ncr`:
```
afeld: THANK YOU for adding this.

doc/setup.md
```diff
@@ -41,6 +41,30 @@ PORT=3000 ASSETS_DEBUG=true ./script/start
 open http://localhost:3000
 ```
 
+## Seeding your dashboard with data
```
afeld: Minor: how about "Populating with data", so that there isn't confusion that this has anything to do with `db/seeds.rb`?

spec/lib/populator_spec.rb
```diff
@@ -1,19 +1,31 @@
...
+
+  describe '#ncr_data_for_user' do
+    it 'creates proposals for the user with email address passed in' do
+      email = 'test@example.com'
+      num_proposals = 2
+      user = FactoryGirl.create(:user, email_address: email)
+
+      expect {
+        Populator.new.ncr_data_for_user(email: email, num_proposals: num_proposals)
+      }.to change { user.requested_proposals.count }.from(0).to(num_proposals)
```
afeld: Might want to also check that the new Proposals have the right `requester`.

afeld:
QA'd locally. Just a couple minor notes, and a CodeClimate failure (finally!). Otherwise :shipit:

lib/populator.rb
jessieay: OK - can you give me more info on the *why* for both of these todos? They are not actionable right now because I am not sure why they should be done and I am guessing they would continue to not be actionable for anyone else who saw them as tracker stories

doc/setup.md
jessieay: Spring is my mortal enemy. But will change :)

doc/setup.md
jessieay: DE NADA

doc/setup.md
jessieay: on it

spec/lib/populator_spec.rb
jessieay: good thinking!

jessieay: on second thought, since we are looking at the requested proposals for `user`, we are in effect checking that the requester is the `user`. To be more explicit, we could have 2 expectations - 1 to confirm that 2 records were created, 1 to confirm that the requester for each of those is the user. Thoughts?

jessieay:
@afeld questions for you re: meaning of todos (so I can add them to Tracker). Otherwise, should be fixed in Rubocop land and Aidancop land! :rainbow:

jessieay:
@afeld re-review?

lib/populator.rb
afeld: I guess the overall user story would be

> As a developer, I want to be have data in development/testing that is realistic (but not sensitive)

This first TODO actually isn't important, now that I think about it, so no worries here.

lib/populator.rb
afeld: I believe I created the Populator when we were starting work on a reporting dashboard, and having a bunch of Proposals that were approved at the same time wouldn't make for very interesting reports/graphs. The idea with this TODO is that we would add some variance in how far along each Proposal was in the approval process, and how much time passed between each approval. Does that make sense?

lib/populator.rb
jessieay: Hey! OK, I will add a tracker story. What do you mean by sensitive? Still don't know enough about this app to understand why the updated at / created_at dates being similar would make the data sensitive.

spec/lib/populator_spec.rb
afeld: Oh sorry, I think I missed that part of the query above – that works!

lib/populator.rb
afeld: Well, two options for executing that user story would be

1. Download the data from production as-is
1. Download a scrubbed version of the data from production
1. Generate realistic data

In that list, as realistic-ness goes :arrow_down:, the sensitivity goes :arrow_down:. I just meant that "having a Populator isn't the only way to tackle this problem, but it would be more useful if it generated data that looked more like data in production".

afeld: And by "sensitive" I mean "NCR was hesitant about having a list of their buildings publicly accessible, so we should be careful about passing around their purchase history".

jessieay: AH, gotcha! Thought meant sensitive as in "would break something in the app" :+1: 

Added story here https://www.pivotaltracker.com/story/show/105241822

jessieay:
can I Get a :ship: ?

PR #666, opened 2015-10-06
Comment cleanup

pkarman:
This is some tidying up based on comments on #650 after it was merged.


afeld:
I actually like them the way they were...

jessieay:
I like this change! Always makes me happy to see code comments go in the :fire:

PR #667, opened 2015-10-06
use existing logic to retrieve the current approver on a WorkOrder

afeld:
Stumbled across while reviewing #666.


jessieay:
Nice find! Merging.

PR #669, opened 2015-10-06
[HOLD] use the best-guess app URL on Cloud Foundry

afeld:
Cloud Foundry provides a list of the mapped routes to the application:

http://docs.run.pivotal.io/devguide/deploy-apps/environment-variable.html#VCAP-APPLICATION

This pull request makes it so we don't need to specify the `DEFAULT_URL_HOST` explicitly.

## Testing notes

* [x] Run `cf unset-env <appname> DEFAULT_URL_HOST`
* [x] Deploy this branch with `cf push`
* [x] Trigger an email, and ensure that the links in the email point back to the site correctly
* [x] When done, deploy `master` with with `cf push` (so `DEFAULT_URL_HOST` gets re-set)


manifest.yml
```diff
@@ -14,14 +14,18 @@ applications:
...
 - name: c2-staging
   host: c2-staging
   env:
-    DEFAULT_URL_HOST: c2-staging.18f.gov
+    NEW_RELIC_APP_NAME: C2 (Staging)
+    S3_BUCKET_NAME: c2-demo
 - name: c2-prod
   host: cap
   env:
     DEFAULT_URL_HOST: cap.18f.gov
```
afeld: Leaving this set, because otherwise it would pick `c2.18f.gov` once #668 gets in.

afeld:
## Post-deploy

* [ ] Run `cf unset-env c2-dev DEFAULT_URL_HOST`
* [ ] Run `cf unset-env c2-staging DEFAULT_URL_HOST`

lib/cloud_foundry.rb
```diff
@@ -11,6 +11,15 @@ def self.vcap_data
     end
   end
 
+  def self.app_url
+    if self.is_environment?
+      # (arbitrarily) use the shortest assigned route, if there are multiple
+      self.vcap_data['application_uris'].min_by(&:length)
+    else
+      nil
```
jessieay: if we are returning `nil` from an `else` I think we can just remove the `else` (and this method will still return `nil`)

spec/lib/cloud_foundry_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    with_env_var('VCAP_APPLICATION', { application_uris: ['foo.bar.com'] }.to_json) do
+      it "returns the URI" do
+        expect(CloudFoundry.app_url).to eq('foo.bar.com')
+      end
+    end
+
+    with_env_var('VCAP_APPLICATION', { application_uris: ['foo.bar.com', 'short.com', 'baz.longerdomain.com'] }.to_json) do
+      it "returns the shortest URI" do
```
jessieay: interesting! glad we have this test or I would not have understood that this was the case

spec/support/features.rb
```diff
@@ -7,11 +7,23 @@ def in_spec?
 end
 
 def with_env_vars_runner(env)
+  # hold on to old values
   env = env.stringify_keys
```
jessieay: these helper methods are neat !! 

Have you considered using `ClimateControl` instead? https://github.com/thoughtbot/climate_control - I think it basically does what we are doing here but I am not sure it provides the ability to delete an `ENV` var so maybe we should stick with what we have

spec/support/features.rb
```diff
@@ -7,11 +7,23 @@ def in_spec?
...
+      v = v.to_s
+    end
+
+    ENV[k] = v
+  end
+
   yield
+
+  # restore old values
   old_values.each { |k, v| ENV[k] = v }
```
jessieay: what about sticking this in a well-named private method rather than using a code comment to indicate its purpose?

spec/support/features.rb
```diff
@@ -44,12 +56,17 @@ def with_env_var(name, val, &block)
   with_env_vars({ name => val }, &block)
 end
 
+def without_env_var(name, &block)
+  # TODO use ENV.delete
```
jessieay: why not do it now!?

lib/cloud_foundry.rb
```diff
@@ -11,6 +11,15 @@ def self.vcap_data
     end
   end
 
+  def self.app_url
+    if self.is_environment?
+      # (arbitrarily) use the shortest assigned route, if there are multiple
```
jessieay: curious on why shortest rather than first, as long as we're being arbitrary!

lib/cloud_foundry.rb
afeld: Yeah, I just like being explicit about it. Feel strongly about it?

lib/cloud_foundry.rb
afeld: Well, when we do [zero-downtime deploys](https://github.com/18F/cf-blue-green), the "green" applications (and routes) are named `<appname>-B`. I've also manually set up routes like `<appname>-test`. Strong feelings?

spec/support/features.rb
afeld: Of _course_ There's A Thoughtbot Project For That :wink: [Storied](https://www.pivotaltracker.com/story/show/105137620).

jessieay: hah you know I didn't even realize it was from tbot until posting that - I'd used it once or twice and just thought it was some rando lib!

lib/cloud_foundry.rb
jessieay: I think explicitness is a great cause in general but it doesn't seem to me that this is a case where the shorter version is confusing. also, I looked around and we do not appear to be returning `nil` from `else` in other parts of this app, so it feels inconsistent to me.

afeld: For methods where the return value is important, I always do, but very possible that we aren't totally consistent about it.

spec/support/features.rb
afeld: Mind if I skip that, since we're probably going to swap it out for the gem anyway?

afeld:
Updated!

lib/cloud_foundry.rb
pkarman: I too like the explicit return from else when the `nil` is important. Ruby (like Perl) relies on implicit returns from methods and defaults to `nil` but that implicit-ness is both hard on newbies and can be non-obvious on later refactorings.

spec/support/features.rb
jessieay: sure, although adding a private method shouldn't be *too* much work

lib/cloud_foundry.rb
jessieay: > For methods where the return value is important, I always do,

@afeld what would be an example of a method where the return value is not important? 

@pkarman interesting to hear that this can be confusing to people. I definitely prefer explicitness to concision but have never seen people explicitly return `nil` as part of that effort in Ruby!

afeld: > I looked around and we do not appear to be returning `nil` from `else` in other parts of this app

Just searched the code with the regex `^\s*nil$`, and I'm seeing seven places, at least :wink: 

> what would be an example of a method where the return value is not important?

Some examples:

* Most ActiveRecord class methods (`has_many`, `scope`, etc.)
* Controller actions
* [`Ncr::WorkOrder#setup_approvals_and_observers`](https://github.com/18F/C2/blob/d68ca0ed741dcde13a8b66b034493f7b50a9309f/app/models/ncr/work_order.rb#L107-L118)

spec/support/features.rb
afeld: Removed in https://github.com/18F/C2/pull/711.

afeld:
Merge conflict fixed.

lib/cloud_foundry.rb
jessieay: not strong feelings, I just feel like this comment says "this is arbitrary" so why not actually be arbitrary and use `take` to  get a random one?

lib/cloud_foundry.rb
jessieay: I will open a separate style PR about this. My feeling on it as of right now is that it's a style I have not come across in the dozens of ruby apps I've worked on, so I am not crazy about it being a style we use in this app. But not going to hold this PR up over it as it can be discussed separately.

pkarman:
This looks well-reviewed. I will QA.

pkarman:
Looks great. QA'd on c2-dev.

I am going to mark as HOLD until I can add the deploy steps to the PT sprint tasks. Because there are env vars to unset, I don't want to merge until right before we deploy to prod.

PR #670, opened 2015-10-06
move additional environment variables to manifest

afeld:
Came across a few that were set manually which aren't sensitive.


jessieay:
Looks good to me! Merging.

PR #671, opened 2015-10-06
[Delivers #104981410] fix broken budget email report

pkarman:
Story: https://www.pivotaltracker.com/n/projects/1149728/stories/104981410

resurrect the not-nil check for approving_official and default to system_approver_emails when nil (not sure in what legit biz case it would be undefined)

### DIFF
--- a/app/models/ncr/work_order.rb
+++ b/app/models/ncr/work_order.rb
@@ -119,7 +119,7 @@ module Ncr
       if self.pending?
         self.currently_awaiting_approvers.first.email_address
       else
-        self.approving_official.email_address
+        self.approving_official ? self.approving_official.email_address : self.system_approver_emails.first
       end
     end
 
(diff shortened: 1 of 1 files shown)


app/models/ncr/work_order.rb
```diff
@@ -119,7 +119,7 @@ def current_approver_email_address
       if self.pending?
         self.currently_awaiting_approvers.first.email_address
       else
-        self.approving_official.email_address
+        self.approving_official ? self.approving_official.email_address : self.system_approver_emails.first
```
jessieay: this ternary operator is hiding an `if/else` within an `if/else` - any chance we could move the logic ELSEwhere? (see what I did there -- ?? :100: )

jessieay:
should we have a test for this to guard against regressions?

pkarman:
consensus in stand up was to go ahead and merge this now so nightly report can run, and then try and identify a test case to avoid regressions later.

PR #673, opened 2015-10-06
103538862 Annual Report Mailer

anthonygarvan:
This creates a mailer and a rake task which calls the mailer for a CSV summarizing approved requests from the specified financial year.


app/mailers/report_mailer.rb
```diff
@@ -14,6 +14,32 @@ def budget_status
...
   end
 
+  def annual_report(year, to_email)
+    proposals = Proposal.approved.where(client_data_type: "Ncr::WorkOrder")
+        proposals = proposals.select {|p| p.client_data.fiscal_year == year}
+        csv_string = CSV.generate do |csv|
+          csv << ["Amount", "Date Approved", "Org Code", "CL#", 
+            "Budget Activity", "SOC", "Function Code", "Building #", 
+            "Vendor", "Description", "Requestor", "Approver"]
+          for p in proposals
```
jessieay: this totally works but in Ruby we usually see `each` being used to iterate over an Array: http://ruby-doc.org/core-2.2.0/Array.html#method-i-each

so this would be:

```ruby
proposals.each do |proposal|
 ... (stuff with proposal) ...
end
```

pkarman: this report is specific to NCR. Should that be reflected in the method name somehow? I realize there's poor precedent for the reports since they are all NCR at this point.

app/mailers/report_mailer.rb
```diff
@@ -14,6 +14,32 @@ def budget_status
     )
   end
 
+  def annual_report(year, to_email)
+    proposals = Proposal.approved.where(client_data_type: "Ncr::WorkOrder")
+        proposals = proposals.select {|p| p.client_data.fiscal_year == year}
```
jessieay: it might be better to use a different variable name her so we don't get confused -- what about `proposals_for_fiscal_year`?

lib/tasks/annual_report.rake
```diff
@@ -0,0 +1,12 @@
+require "mail"
+# usage: 
+# rake MAILTO=your.email@gsa.gov FY=15 annual_report:annual_report
+
+namespace :annual_report do
+	desc "Emails FY approvals"
```
jessieay: Not sure if we have rubocop configured for this, but I believe we are indenting 2 spaces in this project (so this should be bumped out a bit)

app/mailers/report_mailer.rb
pkarman: is the indent weird here or is that just me?

app/mailers/report_mailer.rb
```diff
@@ -14,6 +14,32 @@ def budget_status
     )
   end
 
+  def annual_report(year, to_email)
+    proposals = Proposal.approved.where(client_data_type: "Ncr::WorkOrder")
```
jessieay: since this appears to be an `ncr`-specific mailer, should we include `ncr` in the mailer method name?

lib/tasks/annual_report.rake
```diff
@@ -0,0 +1,12 @@
...
+	task :annual_report => :environment do
```
jessieay: let's use new hash syntax! `task annual_report: :environment do`

app/mailers/report_mailer.rb
```diff
@@ -14,6 +14,32 @@ def budget_status
...
+              p.client_data.function_code, p.client_data.building_number, p.client_data.vendor, 
+              p.client_data.description, p.requester.full_name, approver_name]
+          end
+        end
+
+      attachments['FY' + year.to_s + '_Report.csv'] = csv_string
+
+      mail(
+        to: to_email,
+        subject: 'FY' + year.to_s + ' Report',
```
jessieay: maybe `NCR Work Order` should be included somewhere in the subject so the content of the CSV is clearer?

lib/tasks/annual_ncr_report.rake
```diff
@@ -0,0 +1,12 @@
...
+# rake MAILTO=your.email@gsa.gov FY=15 annual_ncr_report:annual_ncr_report
+
+namespace :annual_ncr_report do
+	desc "Emails FY approvals"
+	task annual_ncr_report: :environment do
+		ReportMailer.annual_ncr_report(ENV['FY'].to_i, ENV['MAILTO']).deliver_later
+
+		puts "Report has been sent!"
+	end
+end
```
afeld: Two-space indents, please!

afeld: https://www.sublimetext.com/docs/3/indentation.html

app/mailers/report_mailer.rb
```diff
@@ -14,6 +14,32 @@ def budget_status
...
+        "Budget Activity", "SOC", "Function Code", "Building #", 
+        "Vendor", "Description", "Requestor", "Approver"]
+      proposals_for_fiscal_year.each do |p|
+        approver_name = p.client_data.approving_official ? p.client_data.approving_official.full_name : "no approver listed"
+        csv << [p.client_data.amount, p.root_approval.approved_at, p.client_data.org_code, 
+          p.client_data.cl_number, p.client_data.expense_type, p.client_data.soc_code, 
+          p.client_data.function_code, p.client_data.building_number, p.client_data.vendor, 
+          p.client_data.description, p.requester.full_name, approver_name]
+      end
+    end
```
afeld: Could we move all of the CSV [data] generation logic to a module/class?

app/mailers/report_mailer.rb
```diff
@@ -14,6 +14,32 @@ def budget_status
...
+
+    attachments['NCR_Work_Order_Report_FY' + year.to_s + '.csv'] = csv_string
```
afeld: Minor: the convention in Ruby is to use string interpolation rather than concatenation. In other words,

```ruby
"NCR_Work_Order_Report_FY#{year.to_s}.csv"
```

app/mailers/report_mailer.rb
```diff
@@ -14,6 +14,32 @@ def budget_status
     )
   end
 
+  def annual_ncr_report(year, to_email)
+    proposals = Proposal.approved.where(client_data_type: "Ncr::WorkOrder")
+    proposals_for_fiscal_year = proposals.select {|p| p.client_data.fiscal_year == year}
```
afeld: Let's do this in SQL instead, for performance reasons:

```ruby
# app/models/ncr/work_order.rb
FISCAL_YEAR_START_MONTH = 10 # 1-based
scope :for_fiscal_year, ->(year) {
  start_time = Time.zone.local(year, FISCAL_YEAR_START_MONTH, 1)
  end_time = start_time + 1.year
  where(created_at: start_time...end_time)
}

# here
work_orders = Ncr::WorkOrder.approved.for_fiscal_year(year)
```

afeld:
Good start! A few more TODOs:

* [ ] Add tests for the CSV [data] generation
* [ ] Add tests for the `annual_ncr_report` mailer
* [ ] Fix [Code Climate issues](https://codeclimate.com/github/18F/C2/pull/673)

Happy to go through any of this in more detail over screenshare if you like.

afeld:
There are tests for the other report that you can almost copy-and-paste from, if that would be helpful:

* https://github.com/18F/C2/blob/master/spec/lib/ncr/reporter_spec.rb
* https://github.com/18F/C2/blob/master/spec/mailers/report_mailer_spec.rb

afeld:
Since this is a one off-thing outside of the request flow that's somewhat time-sensitive, I'm :+1: with merging...as long as we [work on tests](https://www.pivotaltracker.com/story/show/105300436) next :wink: Ran out of time to manually QA today, unfortunately :confused:

afeld:
Already QA'd in production, apparently.

https://18f.slack.com/archives/cap/p1444424465001467

jessieay:
Thanks @afeld !

PR #674, opened 2015-10-06
Remove comments from spec helper files

jessieay:
* Easier to see what's going on when the files are smaller


phirefly:
:shipit:

PR #675, opened 2015-10-06
new team members should be added to New Relic

afeld:


jessieay:
this looks great - will wait for code climate to pass out of respect for code climate :pray:

PR #676, opened 2015-10-06
Speed up 3 slowest specs

jessieay:
* Did an `rspec` profile and edited the 3 slowest specs
* Made single quotes more consistent while I was there
* Replaced a "sleep" with Timecop
* Inlined `let` statements since those can get hairy

### DIFF
--- a/spec/features/active_admin_spec.rb
+++ b/spec/features/active_admin_spec.rb
@@ -1,16 +1,23 @@
-describe "/admin endpoint" do
-  let(:user) { FactoryGirl.create(:user) }
-  it "requires admin role to access" do
+describe '/admin endpoint' do
+  it 'user is not an admin' do
     login_as(user)
+
     visit '/admin'
+
     expect(page.status_code).to eq(403)
   end
 
-  it "allows admin role to access" do
+  it 'user is an admin' do
     user.add_role('admin')
+
     login_as(user)
     visit '/admin'
+
     expect(page.status_code).to eq(200)
     expect(page).to have_content('Dashboard')
+  end
+
+  def user
+    @user ||= FactoryGirl.create(:user)
   end
 end
--- a/spec/features/cancel_spec.rb
+++ b/spec/features/cancel_spec.rb
@@ -1,80 +1,95 @@
-describe "Canceling a request" do
-  let!(:client_data) { FactoryGirl.create(:ncr_work_order, :with_approvers) }
-  let!(:proposal) { client_data.proposal }
-
-  it "shows a cancel link for the requester" do
+describe 'Canceling a request' do
+  it 'shows a cancel link for the requester' do
+    proposal = FactoryGirl.create(:proposal)
     login_as(proposal.requester)
+
     visit proposal_path(proposal)
-    expect(page).to have_content("Cancel my request")
+
+    expect(page).to have_content('Cancel my request')
   end
 
-  it "does not show a cancel link for non-requesters" do
+  it 'does not show a cancel link for non-requesters' do
+    proposal = FactoryGirl.create(:proposal, :with_approver)
     login_as(proposal.approvers.first)
+
     visit proposal_path(proposal)
-    expect(page).to_not have_content("Cancel my request")
+
+    expect(page).to_not have_content('Cancel my request')
   end
 
-  it "prompts the requester for a reason" do
+  it 'prompts the requester for a reason' do
+    proposal = FactoryGirl.create(:proposal)
     login_as(proposal.requester)
+
     visit proposal_path(proposal)
     click_on('Cancel my request')
+
     expect(current_path).to eq("/proposals/#{proposal.id}/cancel_form")
   end
 
   context 'email' do
-    # TODO this shouldn't actually be needed
-    around(:each) { ActionMailer::Base.deliveries.clear }
-
-    it "send emails when cancellation is complete" do
+    it 'send emails when cancellation is complete' do
+      ActionMailer::Base.deliveries.clear
+      proposal = FactoryGirl.create(:proposal)
       expect(deliveries.length).to eq(0)
 
       login_as(proposal.requester)
       visit proposal_path(proposal)
       click_on('Cancel my request')
-      fill_in "reason_input", with: "This is a good reason for the cancellation."
+      fill_in 'reason_input', with: 'This is a good reason for the cancellation.'
       click_on('Yes, cancel this request')
 
-      expect(deliveries.length).to eq(2)
+      expect(deliveries.length).to eq(1)
     end
   end
 
-  context "entering in a reason cancellation" do
-    it "successfully saves comments, changes the request status" do
+  context 'entering in a reason cancellation' do
+    it 'successfully saves comments, changes the request status' do
+      proposal = FactoryGirl.create(:proposal)
       login_as(proposal.requester)
 
       visit proposal_path(proposal)
       click_on('Cancel my request')
-      fill_in "reason_input", with: "This is a good reason for the cancellation."
+      fill_in 'reason_input', with: 'This is a good reason for the cancellation.'
       click_on('Yes, cancel this request')
 
       expect(current_path).to eq("/proposals/#{proposal.id}")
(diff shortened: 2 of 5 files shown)


spec/features/cancel_spec.rb
```diff
@@ -1,80 +1,95 @@
...
 
       login_as(proposal.requester)
       visit proposal_path(proposal)
       click_on('Cancel my request')
-      fill_in "reason_input", with: "This is a good reason for the cancellation."
+      fill_in 'reason_input', with: 'This is a good reason for the cancellation.'
       click_on('Yes, cancel this request')
 
-      expect(deliveries.length).to eq(2)
+      expect(deliveries.length).to eq(1)
```
jessieay: this expectation was "pending" before because of the `around` above -- I don't see why it should have sent 2 emails -- thoughts?

spec/features/active_admin_spec.rb
```diff
@@ -1,16 +1,23 @@
...
     visit '/admin'
+
     expect(page.status_code).to eq(200)
     expect(page).to have_content('Dashboard')
   end
+
+  def user
+    @user ||= FactoryGirl.create(:user)
+  end
 end
```
pkarman: +1

spec/features/cancel_spec.rb
pkarman: I would expect `proposal.requester` to be the only recipient, but there may be some biz rules in `lib/dispatcher.rb` that affect this? I would run the test and dump `deliveries` to see what might be happening. Maybe dupe emails?

spec/features/cancel_spec.rb
```diff
@@ -1,80 +1,95 @@
...
+      proposal = FactoryGirl.create(:proposal, :with_approver)
       login_as(proposal.approvers.first)
+
       visit cancel_form_proposal_path(proposal)
-      expect(page).to have_content("You are not the requester")
+
+      expect(page).to have_content('You are not the requester')
       expect(current_path).to eq("/proposals/#{proposal.id}")
     end
   end
```
pkarman: +1

spec/features/version_check_spec.rb
```diff
@@ -1,16 +1,19 @@
-describe "Version check" do
-  let(:proposal) { FactoryGirl.create(:proposal, :with_parallel_approvers) }
-
-  it "occurs if the proposal is modified in after seeing the profile page" do
+describe 'Version check' do
+  it 'occurs if the proposal is modified in after seeing the profile page' do
+    proposal = FactoryGirl.create(:proposal, :with_parallel_approvers)
     login_as(proposal.approvers.first)
-    visit "/proposals/#{proposal.id}"
+    time = Time.current
```
pkarman: does this need to be `Time.zone.current` ?

pkarman:
In general looks fine. The speed up is nice. The whitespace is nicer. I'm +0 on the quotes.

spec/features/cancel_spec.rb
jessieay: it passes as-is, but the previous expectation was to have 2 emails. So...this might be something to ask @phirefly , who added this test in https://github.com/18F/C2/commit/60cbf244ed6fa2e2cdc0e1fe429c145d9c14d48c

spec/features/version_check_spec.rb
jessieay: this is the same as `TIme.zone,now` - http://api.rubyonrails.org/classes/Time.html#method-c-current

Is your suggestion that I change to `Time.now`? I am SO bad with timezones. Hope to live in a time zone-less world once day

:earth_africa: :watch:

pkarman: ah cool, did not know that about `Time.current` -- fine to leave as-is.

yozlet:
Could you explain the inlining of `let` statements? What's the value there?

spec/features/cancel_spec.rb
phirefly: The two emails would go to 1) The requester that cancelled the request and 2) The current pending approver

jessieay:
Hey @yozlet !  Thanks for asking. There are a few reasons I like to avoid `let` and `before` statements

1) Mystery guest problem: when I am writing a test that depends on a bunch of stuff happening in a `before` or `let` statement, the connection between the setup and the outcomes of a spec are obscured. More stuff on mystery guest problem: http://xunitpatterns.com/Obscure%20Test.html#Mystery%20Guest

2) Slow tests / extra data - there's a tendency to use `let` statements that include data not actually needed for each spec. In https://github.com/18F/C2/pull/676/files#diff-1087c56dafa7bb9f883f5fb020b058c8L1, for example, I found that we didn't need both the `proposal` and `client_data` records for each spec so, by inlining the `let` statements, we are creating fewer records and speeding up our test suite

3) There are other ways to extract common setup! I am all about creating private methods that extract common setup behavior for a set of specs. I think the benefit of these over a `let` or `before` statement is that you must call the private method in the spec so you are not obscuring the connection between the setup as much as you are with a `let` statement that gets executed automatically. 

Does this make sense? I know a lot of people use `let` statements and like them so I am curious to hear your perspective!

jessieay:
Update: @yozlet and I discussed this IRL (actually, on screenhero).

We agreed that there are benefits and drawbacks to `let` and `before` statements, but that doesn't mean we should go through this app and rip them all out. 

I took them out in this case because they were creating more data for each spec than they needed to. 

I think it's a good practice moving forward to avoid `before` and `let` in favor of private methods for the reasons mentioned in https://github.com/18F/C2/pull/676#issuecomment-146235025, but not a hard and fast convention / rule.

spec/features/cancel_spec.rb
jessieay: discovered why there was just 1 email:

When I updated these specs, I removed the common proposal record. Formerly, the proposal always had an approver but I realized we didn't need one for every spec.

In this case, I assumed we didn't need one because the approver was not mentioned so I assumed we didn't need one. But because of @phirefly 's comment about "the pending approver" receiving an email, I can see that the test was written with the expectation of a proposal with an approver.

When I added an approver, there were actually 3 emails sent. Upon further inspection with @pkarman , it was discovered that a cancellation email goes to the requester and to the approver. The third email is a comment email that goes to the approver as well. 

There is a story already for removing the comment email upon cancellation: https://www.pivotaltracker.com/n/projects/1149728/stories/99273060

For now, I am going to keep the expectation in the spec that there are 3 emails, because that is the current behavior, and I will tackle the "remove comment email" feature next (and update this spec accordingly)

jessieay:
@pkarman update specs for email stuff are at 130718fd2d85aecbe55d90923b5ff0aee9154d3f -- this should be ready to merge!?

pkarman:
@jessieay one nit comment here: https://github.com/18F/C2/commit/130718fd2d85aecbe55d90923b5ff0aee9154d3f#commitcomment-13693865

otherwise I will merge later today.

jessieay:
@pkarman thanks! fixed at f96c26b11cf195c9d3599ee605af52ca2dde9b6d

afeld:
> by inlining the let statements, we are creating fewer records and speeding up our test suite

Not true! From [the docs](https://www.relishapp.com/rspec/rspec-core/v/3-3/docs/helper-methods/let-and-let):

> Note that let is lazy-evaluated: it is not evaluated until the first time the method it defines is invoked. You can use let! to force the method's invocation before each example.

jessieay:
@afeld in this case we are creating fewer because our `let` statements created 2 records but we only needed 2 records for some of the specs, not all (but they were all using them)

spec/support/feature_spec_helper.rb
```diff
@@ -1,6 +1,4 @@
 module FeatureSpecHelper
-  # requires IntegrationSpecHelper
```
afeld: Why remove this? It's still true & useful.

jessieay: We have a lot of spec files that require each other and don't make a comment about each one. Why is this file any different?

afeld: Well, this is a mixin that requires another mixin, neither of which are included in all specs.

PR #678, opened 2015-10-06
Create delegate transient attribute

jessieay:
* Can check that one off the to do list!


phirefly:
I've never used transient before. Nice! :ship:

PR #681, opened 2015-10-07
[Delivers #104981410] consistent approving official email in report

pkarman:
use same method everywhere for reporting the approving official for NCR work orders. 
NOTE the biz rule here is in question and this PR is mostly to prompt a conversation.


pkarman:
@phirefly @afeld @cmc333333 looking for biz history comments on this one.

app/models/ncr/work_order.rb
```diff
@@ -120,11 +122,21 @@ def current_approver_email_address
         self.currently_awaiting_approvers.first.email_address
       elsif self.approving_official
         self.approving_official.email_address
+      elsif self.emergency and self.approvers.empty?
+        EMERGENCY_APPROVER_EMAIL
```
afeld: I would expect it to return `nil` in this case...if this is only meant for display purposes, I would move it to a helper or a Decorator.

app/models/ncr/work_order.rb
```diff
@@ -120,11 +122,21 @@ def current_approver_email_address
...
       else
         self.system_approver_emails.first
       end
     end
 
+    def final_approver_email_address
+      if self.emergency and self.approvers.empty?
+        EMERGENCY_APPROVER_EMAIL
```
afeld: Ditto.

afeld:
:+1: on centralizing the logic, thanks!

app/models/ncr/work_order.rb
pkarman: `nil` is somewhat ambiguous as a return value. Does it mean "there wasn't an email address" or "you were wrong to ask in the first place" or "I couldn't find what you wanted" or "this is a specific case where nil is important"?

I made it a constant though `EMERGENCY_APPROVER_EMAIL` so that we can compare and/or change it easily. If everyone wants the constant value to be `nil` I won't object. Just wanted it to be clear when reading that it was a special case.

afeld: This method could alternatively be called `email_address_of_the_person_the_proposal_is_waiting_on`, which for an `emergency` is no one. `nil` makes sense to me in that case. Totally open to other ideas, but returning an arbitrary string that can't (easily) be disambiguated from an email address seems more confusing.

pkarman: Right now this method is only used in the daily budget report email. I thought it would be important for readers to know why particular proposals were approved, since those approvals were done verbally rather than inside the system. If we don't care about communicating that, then having `nil` email address is fine with me.

@phirefly thoughts?

afeld: Right...see the latter part of https://github.com/18F/C2/pull/681/files#r41535674.

pkarman:
@afeld the Decorator comment made a lot more sense once I looked at the other existing decorators. Thanks. I have refactored to split the display stuff into `Ncr::WorkOrderDecorator`.

app/decorators/ncr/work_order_decorator.rb
```diff
@@ -0,0 +1,28 @@
+module Ncr
+  class WorkOrderDecorator < Draper::Decorator
+    delegate_all
+
+    EMERGENCY_APPROVER_EMAIL = 'Emergency - Verbal Approval'
+    NO_APPROVER_FOUND        = 'No Approver Found'
```
jessieay: I know I mentioned this in another PR, but not sure what the outcome was. Do we want to be lining up `=` ? Seems like we are being inconsistent on that here. I would vote for no.

app/decorators/ncr/work_order_decorator.rb
```diff
@@ -0,0 +1,28 @@
...
+    def final_approver_email_address
+      approver_email_address(final_approver)
+    end
+
+    private
+
+    def approver_email_address(approver)
+      if approver
+        approver.email_address
+      elsif !approver and self.emergency
```
jessieay: I think this can just be `emergency` since L19 means that anything in this `elsif` would not have an approver

app/models/ncr/work_order.rb
```diff
@@ -114,13 +114,23 @@ def approving_official
...
+      elsif self.emergency and self.approvers.empty?
+        nil
       else
-        self.system_approver_emails.first
+        User.for_email(self.system_approver_emails.first)
+      end
+    end
+
+    def final_approver
+      if self.emergency and self.approvers.empty?
```
jessieay: how about:

```ruby
if not_emergency? && no_approvers?
  approvers.last
end
```

I think that's easier to read. Methods used in `if` would need to be defined in order for this to work..

pkarman: easier to read maybe, but logically wrong. If `no_approvers` is true, `approvers.last` will always be `nil`.

fwiw, I found it important to check for `approvers.empty?` and `emergency` as a pair because it might be possible (though bad) for there to be zero approvers on a proposal when `emergency` is false.

app/decorators/ncr/work_order_decorator.rb
pkarman: good catch!

app/models/ncr/work_order.rb
jessieay: AH yes, I see. I think the additional color you provided above gives more evidence that this could be more clearly constructed, because my first reading of what was happening wasn't correct. what about:

```ruby 
def final_approver
  if not_emergency? && approvers.any?
    self.approvers.last
  end
end
```

app/models/ncr/work_order.rb
```diff
@@ -114,13 +114,23 @@ def approving_official
       self.approvers.first
     end
 
-    def current_approver_email_address
+    def current_approver
       if self.pending?
-        self.currently_awaiting_approvers.first.email_address
+        self.currently_awaiting_approvers.first
```
jessieay: In Ruby, I believe it isn't "idiomatic" to use `self` when referring to objects. The only time I usually use self is when defining class methods. Why so much `self` in here?

pkarman: I agree, I find the `self` a little too selfish. I was just following what appeared to be the convention when I looked at other code in that file, preferring consistency over correctness.

app/models/ncr/work_order.rb
pkarman: heh. I find that harder to grok, personally.

this is a classic dilemma: a positive and a negative assertion combined in a single `if`.

```
unless emergency and approvers.empty?
  approvers.last
end
```

same thing, with the `unless` helping? or not?

app/decorators/ncr/work_order_decorator.rb
afeld: FWIW This is one of those things where I don't do `=` alignment, but also don't feel strongly enough that I would tell someone to change it.

app/models/ncr/work_order.rb
jessieay: I find `unless` to be the hardest to understand. I think the easiest way to explain my thinking around this is to explain how I would say each thing in English:


v1: `if self.emergency and self.approvers.empty?`

= "if it is an emergency and there are no approvers, don't sent to anyone. otherwise, send to the last approver"

(in this case, not sure what "otherwise" means without thinking really hard about it)

v2 `if not_emergency? && approvers.any?`

= "if it's not an emergency, and there are some approvers, send it to the last approver"

(pretty easy to understand)

v3 `unless emergency and approvers.empty?`

= "if it is not an emergency and there are no approvers, return the last approver`

(WAIT, that isn't even right in terms of what Ruby will return, but that is how it reads to me)

Given the above interpretations, I think the second is the easiest to understand. But interested in others' opinions as well!

app/models/ncr/work_order.rb
jessieay: change it all to not use self!? (haven't seen that style used in the rest of the project)

app/decorators/ncr/work_order_decorator.rb
pkarman: I trained myself to appreciate vertical alignment when writing Perl. It's in the Perl Best Practices book (https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=perl+best+practices+vertical+align+assignment).

That said, I tend to use it only when I think it makes the code more legible, as when setting constants (e.g.). I'm not wedded to it, just have a habit of it. If the consensus is not to do it, I will try and un-train myself.

pkarman:
@jessieay I have re-worked the logical syntax in 9735c278c6f8009939d8af2169d80416d15c09ad

jessieay:
@pkarman  so much easier to parse! :heart: :heart:

spec/models/ncr/work_order_spec.rb
```diff
@@ -454,19 +454,35 @@
...
       wo.individual_approvals.first.approve!
       wo.reload.individual_approvals.last.approve!
-      expect(wo.current_approver_email_address).to eq(wo.individual_approvals.first.user.email_address)
+      expect(wo.current_approver).to eq(wo.individual_approvals.first.user)
     end
   end
+
+  describe "#final_approver" do
+    it "returns the final approver" do
+      wo = FactoryGirl.create(:ncr_work_order, :with_approvers)
```
jessieay: don't need `FactoryGirl` anymore :)

app/models/ncr/work_order.rb
```diff
@@ -121,13 +121,23 @@ def approving_official
...
       else
-        self.system_approver_emails.first
+        User.for_email(self.system_approver_emails.first)
+      end
+    end
+
+    def final_approver
+      if !emergency and approvers.any?
+        approvers.last
+      else
```
jessieay: don't need this `else` -- I know we discussed this in a previous PR with @afeld but it seems to be that if we want to always do an `else nil` on a conditional in Ruby (which is the behavior whether you do it explicitly or not) then we should do it everywhere and not just selectively. I would vote not to do it.

app/decorators/ncr/work_order_decorator.rb
jessieay: I don't think we are doing `=` alignment in the rest of this project, so I would vote to not do it here. I don't feel strongly about this "rule" per se but I do feel strongly about consistency within a project!

pkarman:
@jessieay as you wish.

jessieay:
@pkarman :ship: :it: !!

PR #682, opened 2015-10-07
[Delivers 97730392] Add threshold info popover to NCR Amount field

yozlet:
WIP, [see story for discussion](https://www.pivotaltracker.com/n/projects/1149728/stories/97730392). Working on this with @jessieay.


app/views/ncr/work_orders/form.html.erb
```diff
@@ -72,10 +72,13 @@
     </div>
     <div class="form-group">
       <%= field_set_tag "Amount", class: "required" do %>
-        <%= f.label :amount, "Dollar Amount", :class => 'sr-only'  %>
+        <%= f.label :amount, "Dollar Amount", class: 'sr-only' %>
```
jessieay: this makes me so happy

spec/features/ncr_work_orders_spec.rb
```diff
@@ -108,6 +108,15 @@
         end
       end
 
+      it "shows hint text for amount field", js: true do
+        visit '/ncr/work_orders/new'
+        page.execute_script "document.getElementById('ncr_work_order_amount').scrollIntoView()"
```
jessieay: WOOF! hate it when we have to do this type of thing in feature specs, but I know it is a necessary evil at times.

To make it more obvious what these two lines are doing, what about putting 113 and 114 into a private method called something like `hover_over_amount_field_popover` ?

app/views/ncr/work_orders/form.html.erb
```diff
@@ -72,10 +72,13 @@
...
         <div class="input-group">
           <div class="input-group-addon">$</div>
-          <%= f.number_field :amount, class: 'form-control', step: 0.01 %>
+          <%= f.number_field :amount, class: 'form-control', step: 0.01,
+                data: { toggle: 'popover', trigger: 'focus', html: true,
+                  title: 'Transaction Amount Thresholds', placement: 'top',
+                  content: "$3,000 for supplies<br>$2,500 for services<br>$2,000 for construction" } %>
```
jessieay: maybe we can throw this text into a helper? Or one of your BFF decorators? Not sure if we've been extracting form or view logic in either of those much up to this point, but it's always nice to have simpler views

app/views/ncr/work_orders/form.html.erb
afeld: Some people like fancy cars and diamond rings...just get @jessieay Ruby 1.9 Hash syntax.

app/views/ncr/work_orders/form.html.erb
yozlet: Good idea!

spec/features/ncr_work_orders_spec.rb
yozlet: Oh yeah, that'd make it much better. On it.

app/helpers/ui_helper.rb
```diff
@@ -0,0 +1,43 @@
+module UiHelper
```
jessieay: wouldn't any helper be a UI helper since these are usually methods used in views? Thoughts on a more specific name, such as `FormHelper` ?

config/locales/en.yml
```diff
@@ -34,6 +34,10 @@ en:
       # overrides
       observation:
         create: "Add"
+    popover:
```
jessieay: yay!

spec/features/ncr_work_orders_spec.rb
```diff
@@ -450,8 +440,8 @@
         click_on 'Update'
         proposal = Proposal.last
 
-        expect(proposal.approvers.first.email_address).to eq (approver.email_address)
-        expect(proposal.individual_approvals.first.actionable?).to eq (true)
+        expect(proposal.approvers.first.email_address).to eq approver.email_address
+        expect(proposal.individual_approvals.first.actionable?).to eq true
```
jessieay: thoughts on using a predicate matcher? https://www.relishapp.com/rspec/rspec-expectations/v/2-2/docs/matchers/predicate-matchers

`expect(proposal.individual_approvals.first).to be_actionable`

reads nicely!

spec/support/feature_spec_helper.rb
```diff
@@ -10,4 +10,9 @@ def login_as(user)
     setup_mock_auth(:myusa, user)
     visit '/auth/myusa'
   end
+
+  def focus_field(field_id)
```
jessieay: since we'er only using this method once right now, I'd suggest putting it in the spec file where it is being used. Easier to check it out that way!

app/helpers/ui_helper.rb
yozlet: Good point - I thought helpers were more general, hence the name, but let's do that.

spec/features/ncr_work_orders_spec.rb
yozlet: ... and in fact, we have use of such a matcher down the page. Didn't know about those - cool!

yozlet:
Note that I renamed `ncr.scss`, which was a mostly-inaccurate name, to `form.scss`, which is slightly less inaccurate but still a fair distance from the glorious plains of Accuracy and Meaning. I chose `form` because I could not think of anything more appropriate, given the variety of uses that file covers. Fact is, our SCSS is in weird badly-apportioned piles and needs fixing all over, but the good news is that [we have a story for that](https://www.pivotaltracker.com/n/projects/1149728/stories/105594432). In the meantime, if anyone has a better name for that file, I'm happy to change it.

app/assets/stylesheets/webapp/form.scss
```diff
@@ -1,4 +1,5 @@
 .controller-ncr-work_orders,
+.controller-gsa18f-procurements,
 .controller-feedback {
```
afeld: Maybe we should just make a custom class for everything we want this applied to? Or is it every `<form>`? Feel free to make a story if you feel it's outside of the scope of this story.

afeld: Nevermind, totally is.

config/locales/en.yml
```diff
@@ -34,6 +34,10 @@ en:
...
+      ncr_amount:
+        title: "Transaction Amount Thresholds"
+        content: "$3,000 for supplies<br>$2,500 for services<br>$2,000 for construction"
```
afeld: Don't think these numbers are still accurate as of FY16...check with @phirefly.

afeld:
QA'd on dev – looks good! We confirmed that the popover is 508 compliant?

yozlet:
I believe it's 508-compliant [thanks to this work](https://github.com/twbs/bootstrap/issues/13482), but I don't know whether it's actually been tested with screen readers. Can/should we test that?

config/locales/en.yml
phirefly: Great catch. Checking.

phirefly: Confirmed these values:
====
$3,500 for supplies
$2,500 for services
$2,000 for construction
====

yozlet:
Comment from @phirefly in Slack: "Anything else on that PR? It looks like we're good on 508. We are not targeting screenreaders at this point, so it's worth a look but not a  blocker to delivering the story."

PR #683, opened 2015-10-08
silence logging of requests for assets in development

afeld:
Asset logging, e.g.

```
Started GET "/assets/icon-logout.png" for ::1 at 2015-10-08 13:18:47 -0400
```

is basically just noise, so added a gem that silences it. Also removed the now-not-really-useful `ASSETS_DEBUG` config flag, which was [apparently](https://github.com/18F/C2/pull/618/files#r41514899) added to reduce the number of asset logging messages received. Now it's a non-issue, and it makes JS debugging easier since they appear in separate files.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,7 +10,6 @@ MYUSA_SECRET: consumer_secret_key
 # optional
 #
 # API_ENABLED=true
-# ASSETS_DEBUG=true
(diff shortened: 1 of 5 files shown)


pkarman:
love it. much better solution.

PR #684, opened 2015-10-08
add explicit buildpack so that Ruby 2.2.3 is supported on cloud.gov

pkarman:
this is necessary until 18F CF buildpacks are updated.

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -8,6 +8,7 @@ command: script/server_start
 domain: 18f.gov
 instances: 1
 memory: 1G
+buildpack: https://github.com/cloudfoundry/ruby-buildpack.git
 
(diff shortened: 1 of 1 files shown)


afeld:
Maybe worth adding a comment? Is there an issue or anything we can link to?

pkarman:
https://github.com/18F/cloud-foundry-notes/commit/6b823c69eadf4128b906b9abde8d1f9e82b05817

jessieay:
LGTM. This is a case where a code comment might make sense because this is something we can remove at a later date when it is no longer needed :ship:

pkarman:
thanks for reviews. I added a comment about why the buildback url is explicit.

PR #685, opened 2015-10-08
Gemfile.lock was out of sync after yanking about_yml

pkarman:


jessieay:
waiting for CI to pass. otherwise :ship:

PR #687, opened 2015-10-09
Use Factory Girl syntax methods

jessieay:
* https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#configure-your-test-suite

### DIFF
--- a/spec/controllers/attachments_controller_spec.rb
+++ b/spec/controllers/attachments_controller_spec.rb
@@ -1,6 +1,6 @@
 describe AttachmentsController do
   describe 'permission checking' do
-    let (:proposal) { FactoryGirl.create(:proposal, :with_parallel_approvers, :with_observers) }
+    let (:proposal) { create(:proposal, :with_parallel_approvers, :with_observers) }
     let (:params) {{
       proposal_id: proposal.id,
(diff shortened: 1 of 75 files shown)


pkarman:
So I like terse and this does that. 

My only concern is that unless you know that the FactoryGirl `create` method is magically imported, it might be harder for a newcomer to know where to look for the available options when adding/modifying a test. E.g. my first inclination would be to google `rspec create` thinking it was some method available through rspec.

So I'm a +0 on this change. Sometimes more typing helps clarity, sometimes it doesn't.

jessieay:
I agree that terseness can be confusing. 

Factory Girl has pretty huge market share in the "factories for RSpec" world, so anyone familiar with Rails would probably guess that we are using FG. And if they weren't sure, they could always visit `spec/factoriess` to check out the factory definitions and would see that we are using FG.

I like this change because it means less typing for all of us!

afeld:
I'm +0 on this. If @pkarman / others are ok with it, feel free to merge manually.

jessieay:
Rebased off of master, so this should be OK for a "merge button" merge once we get consensus

afeld:
I think 2.5 :+1:s count.

PR #690, opened 2015-10-13
Add .idea/ (RubyMine) to gitignore

yozlet:


jessieay:
This seems specific to your dev environment - what about adding it to your own global gitignore? https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

yozlet:
Oh, good point. Will close. Thanks!

PR #692, opened 2015-10-13
Replace callback with DB default

jessieay:
* Let's try to avoid callbacks!
* http://samuelmullen.com/2013/05/the-problem-with-rails-callbacks/

### DIFF
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -59,16 +59,11 @@ class Proposal < ActiveRecord::Base
   scope :closed, -> { where(status: ['approved', 'cancelled']) } #TODO: Backfill to change approvals in 'reject' status to 'cancelled' status
   scope :cancelled, -> { where(status: 'cancelled') }
 
-  after_initialize :set_defaults
   after_create :update_public_id
 
   # @todo - this should probably be the only entry into the approval system
   def root_approval
     self.approvals.where(parent: nil).first
-  end
-
-  def set_defaults
-    self.flow ||= 'parallel'
   end
 
   def parallel?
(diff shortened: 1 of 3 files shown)


baccigalupi:
Great article. I hadn't see that one yet!

One thing I hadn't thought about though is whether the default is initialized into the AR object before save. If there are tests that look for this initialized value, then I am just worrying needlessly.

:+1:

jessieay:
@baccigalupi what do you mean by "default is initialized into the AR object? As far as I know, this change does not change the behavior at all. Example:

Before:
```bash
Proposal.new
=> #<Proposal:0x007ff4238cb210 id: nil, status: nil, flow: "parallel", created_at: nil, updated_at: nil, client_data_id: nil, client_data_type: nil, requester_id: nil, public_id: nil>
```

After:
```bash
Proposal.new
=> #<Proposal:0x007ff4218039d8 id: nil, status: nil, flow: "parallel", created_at: nil, updated_at: nil, client_data_id: nil, client_data_type: nil, requester_id: nil, public_id: nil>
```

...is there something else I should check out? 

I do not believe there are tests for this behavior but I don't usually write tests for DB defaults so I'm not sure about writing specs for this now. What do you think?

yozlet:
This is relevant to the work @annalee and I are doing on workflows, since we're planning on adding an `after_initialize` call to set up the workflow. The linked article's rule is “Use a callback only when the logic refers to state internal to the object” and we'll be sticking to that. From where we are right now I think it's the least unpleasant option, but we'll show you once we have a first version of the code.

afeld:
> The linked article's rule is “Use a callback only when the logic refers to state internal to the object”

As does the `Proposal#set_defaults` method being removed here, but having defaults in the database does make querying easier (since you don't have to account for `null` on those columns). :+1: for this, and changing the other models that use the same pattern.

afeld:
Storied: https://www.pivotaltracker.com/story/show/105636870

PR #693, opened 2015-10-13
[Fixes #99273060] Suppress comment email on proposal cancellation

jessieay:
* When a proposal is cancelled, we send a cancellation email to
  requester and approvers
* Extra email about cancellation comment is repetitive
* Fixes #99273060

### DIFF
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -203,6 +203,8 @@ class Proposal < ActiveRecord::Base
     end
   end
 
+  ## delegated methods ##
+
   def public_identifier
     self.delegate_with_default(:public_identifier) { "##{self.id}" }
   end
@@ -226,6 +228,8 @@ class Proposal < ActiveRecord::Base
       self.client_data.try(:version)
     ].compact.max
   end
+
+  #######################
 
   def restart
     # Note that none of the state machine's history is stored
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -268,8 +268,11 @@ class Proposal < ActiveRecord::Base
     # invalidate relation cache so we reload on next access
     self.observers(true)
     # when explicitly adding an observer using the form in the Proposal page...
-    if adder && reason
-      add_observation_comment(user, adder, reason)
+    if adder
+      if reason
+        add_observation_comment(user, adder, reason)
+      end
+
       Dispatcher.on_observer_added(observation, reason)
     end
 
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -280,13 +280,14 @@ class Proposal < ActiveRecord::Base
   end
 
   def add_observation_comment(user, adder, reason)
-    comment = comments.create(
-      comment_text: I18n.t('activerecord.attributes.observation.user_reason_comment',
-                           user: adder.full_name,
-                           observer: user.full_name,
-                           reason: reason),
+    comments.create(
+      comment_text: I18n.t(
+        'activerecord.attributes.observation.user_reason_comment',
+        user: adder.full_name,
+        observer: user.full_name,
+        reason: reason
+      ),
       user: adder
     )
-    Dispatcher.on_comment_created(comment)
   end
 end
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -203,9 +203,6 @@ class Proposal < ActiveRecord::Base
     end
   end
 
-
-  ## delegated methods ##
-
   def public_identifier
     self.delegate_with_default(:public_identifier) { "##{self.id}" }
   end
@@ -229,9 +226,6 @@ class Proposal < ActiveRecord::Base
       self.client_data.try(:version)
     ].compact.max
   end
-
-  #######################
-
 
   def restart
     # Note that none of the state machine's history is stored
@@ -278,11 +272,13 @@ class Proposal < ActiveRecord::Base
   end
 
   def add_observation_comment(user, adder, reason)
-    self.comments.create(
+    comment = comments.create(
       comment_text: I18n.t('activerecord.attributes.observation.user_reason_comment',
                            user: adder.full_name,
                            observer: user.full_name,
                            reason: reason),
-      user: adder)
+      user: adder
+    )
+    Dispatcher.on_comment_created(comment)
   end
 end
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -264,10 +264,11 @@ class Proposal < ActiveRecord::Base
     # invalidate relation cache so we reload on next access
     self.observers(true)
     # when explicitly adding an observer using the form in the Proposal page...
-    if adder
-      add_observation_comment(user, adder, reason) unless reason.blank?
+    if adder && reason
+      add_observation_comment(user, adder, reason)
       Dispatcher.on_observer_added(observation, reason)
     end
+
     observation
   end
 
--- a/lib/dispatcher.rb
+++ b/lib/dispatcher.rb
@@ -14,7 +14,7 @@ class Dispatcher
     proposal.observations.each do |observation|
       user = observation.user
       if user.role_on(proposal).active_observer?
-       email_observer(observation)
+        email_observer(observation)
       end
     end
   end
--- a/lib/dispatcher.rb
+++ b/lib/dispatcher.rb
@@ -39,8 +39,10 @@ class Dispatcher
   def deliver_attachment_emails(proposal)
     proposal.users.each do |user|
       approval = proposal.approvals.find_by(user_id: user.id)
-      next if approval && approval.pending?
-      CommunicartMailer.new_attachment_email(user.email_address, proposal).deliver_later
+
+      if user_is_not_approver?(approval) || approver_knows_about_proposal?(approval)
+        CommunicartMailer.new_attachment_email(user.email_address, proposal).deliver_later
+      end
     end
   end
 
@@ -84,5 +86,13 @@ class Dispatcher
 
   def send_notification_email(approval)
     CommunicartMailer.actions_for_approver(approval).deliver_later
+  end
+
+  def user_is_not_approver?(approval)
+    approval.blank?
+  end
+
+  def approver_knows_about_proposal?(approval)
+    !approval.pending?
   end
 end
--- a/lib/dispatcher.rb
+++ b/lib/dispatcher.rb
@@ -48,8 +48,9 @@ class Dispatcher
 
   def deliver_cancellation_emails(proposal)
     proposal.individual_approvals.each do |approval|
-      next if approval.pending?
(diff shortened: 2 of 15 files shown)


jessieay:
Implementation note: what I've done here is move the comment created mailer out of a callback on `Comment` - now we are calling the mailer where we need it rather than having it always be called by default when a comment is created. This is also a good refactor because less callbacks is MORE fun!

app/models/proposal.rb
```diff
@@ -203,9 +203,6 @@ def delegate_with_default(method)
     end
   end
 
-
-  ## delegated methods ##
-
```
afeld: Can we re-add this? I think that grouping is useful (though would be open to putting them in a mixin or something).

app/models/proposal.rb
```diff
@@ -270,19 +264,22 @@ def create_new_observation(user, adder, reason)
     # invalidate relation cache so we reload on next access
     self.observers(true)
     # when explicitly adding an observer using the form in the Proposal page...
-    if adder
-      add_observation_comment(user, adder, reason) unless reason.blank?
+    if adder && reason
+      add_observation_comment(user, adder, reason)
```
afeld: Just checking: this change in behavior is intentional?

lib/dispatcher.rb
```diff
@@ -29,15 +29,15 @@ def email_sent_confirmation(proposal)
...
-    self.email_observers(proposal)
-    self.email_sent_confirmation(proposal)
+
+    email_observers(proposal)
+    email_sent_confirmation(proposal)
   end
 
   def deliver_attachment_emails(proposal)
     proposal.users.each do |user|
-      # do not send email to approvers who have not yet heard about the proposal
```
afeld: Why remove this comment (and the one below)? It seems useful, short of doing https://www.pivotaltracker.com/story/show/104647900.

spec/controllers/comments_controller_spec.rb
```diff
@@ -1,18 +1,30 @@
...
+        expect(response).to redirect_to(proposal)
+      end
+
+      it 'sends a comment email to approvers and observers' do
+        ActionMailer::Base.deliveries.clear
+        login_as(proposal.requester)
+
+        post :create, params
+
+        expect(deliveries.length).to eq 4
```
afeld: I would write this as

```ruby
expect {
  post :create, params
}.to change {  deliveries.length }.from(0).to(4)
```

The `.clear` shouldn't be necessary.

spec/features/cancel_spec.rb
```diff
@@ -34,27 +34,34 @@
         proposal = create(:proposal)
 
         login_as(proposal.requester)
-        visit proposal_path(proposal)
-        click_on('Cancel my request')
-        fill_in 'reason_input', with: 'This is a good reason for the cancellation.'
-        click_on('Yes, cancel this request')
+        cancel_proposal(proposal)
+
+        expect(deliveries.length).to eq(1)
```
afeld: Ditto.

spec/features/cancel_spec.rb
```diff
@@ -34,27 +34,34 @@
...
 
         login_as(proposal.requester)
-        visit proposal_path(proposal)
-        click_on('Cancel my request')
-        fill_in 'reason_input', with: 'This is a good reason for the cancellation.'
-        click_on('Yes, cancel this request')
+        cancel_proposal(proposal)
 
-        expect(deliveries.length).to eq(3)
+        expect(deliveries.length).to eq(2)
```
afeld: Ditto.

spec/models/ncr/work_order_spec.rb
```diff
@@ -423,6 +423,16 @@
...
     end
+
+    it 'does not send a comment email for the update comment to proposal listeners' do
+      listener = create(:user)
+      work_order.proposal.add_observer(listener)
+      ActionMailer::Base.deliveries.clear
+
+      work_order.update(vendor: 'TestVendor')
+
+      expect(deliveries.length).to eq(0)
```
afeld: Ditto.

spec/models/proposal_spec.rb
```diff
@@ -277,17 +277,29 @@
...
-    context "with an adding user" do
+
+    it 'adds a comment and sends a comment email when there is an adder and observation reason' do
+      ActionMailer::Base.deliveries.clear
+      reason = "this is required"
+
+      proposal.add_observer(observer_email, user, reason)
+
+      expect(proposal.comments.count).to eq 1
+      expect(deliveries.length).to eq 3
```
afeld: Ditto.

app/models/proposal.rb
jessieay: I removed this because it seems out of sync with the rest of the app - we aren't using comments to organize methods (and I would argue that we shouldn't have to if we have nice lean models!). But I can add back in.

app/models/proposal.rb
jessieay: this should not represent a change in the behavior - can you explain more?

lib/dispatcher.rb
jessieay: I added a test with in `spec/features/cancel_spec.rb`:

```ruby
context 'proposal with pending status' do
   it 'does not send cancellation email to approver' do
```

this test should provide the same context that the comment does and has the added benefit of failing if the behavior changes whereas a comment would do nothing if the behavior changes. What do you think?

spec/controllers/comments_controller_spec.rb
jessieay: nice! much better, will change

spec/features/cancel_spec.rb
jessieay: updated in 00be83c24ee02f381a2b5ab432db234fc0eeadcc

spec/features/cancel_spec.rb
jessieay: updated in 00be83c24ee02f381a2b5ab432db234fc0eeadcc

spec/controllers/comments_controller_spec.rb
jessieay: updated 00be83c24ee02f381a2b5ab432db234fc0eeadcc

app/models/proposal.rb
jessieay: added back in at 00be83c24ee02f381a2b5ab432db234fc0eeadcc

app/models/proposal.rb
afeld: Before, `Dispatcher.on_observer_added` would be called even if `reason == nil`. Seems like we would want it the old way, and worth adding a test around.

spec/models/proposal_spec.rb
jessieay: updated at db5088707ca83c69e4c46f91dd63d820eb0297c0

spec/models/ncr/work_order_spec.rb
jessieay: updated at db5088707ca83c69e4c46f91dd63d820eb0297c0

app/models/proposal.rb
afeld: > we aren't using comments to organize methods

We are, actually...search for `###`.

> I would argue that we shouldn't have to if we have nice lean models!

:+1: Totally agree that having them in a standalone classes/models is better.

app/models/proposal.rb
jessieay: before it was `add_observation_comment(user, adder, reason) unless reason.blank?` -- so it would *not* be called if blank. 

When I remove the `&& reason` here, `spec/models/proposal_spec.rb:304` fails, so it does appear to be covered by tests

BTW, I am assuming you thought the behavior was different because you didn't see the trailing `unless` - this is why I am firmly anti-trailing conditionals!

lib/dispatcher.rb
afeld: Yep, totally understand the value of the test, but the problem the comment solves for is "I'm looking at this code, but don't understand what it's doing," which tests don't really help with. Ideal would be making a self-descriptive method to retrieve the recipients, which is what the story linked to above is about.

app/models/proposal.rb
afeld: > before it was `add_observation_comment(user, adder, reason) unless reason.blank?` -- so it would not be called if blank.

Right...I was talking about the next line :wink: 

> I am firmly anti-trailing conditionals!

:+1:

lib/dispatcher.rb
jessieay: If I am looking at code and don't understand what it's going, I will look at the tests for the code to understand. I can make this method name for self-descriptive now so that it is easier to understand without looking at tests! Was trying to keep this PR small while still doing small cleanups along the way but this seems like a worthwhile thing to take care of now.

afeld: Didn't mean to imply that you should do the refactoring in this PR...up to you!

app/models/proposal.rb
jessieay: oh! the next line isn't a change in behavior - this mailer (`Dispatcher.on_observer_added(observation, reason)`) was formerly called in  callback on `comment#create`, but I removed that callback to give us more control over behavior and make it easier to suppress comment emails in certain situations, like when a proposal is cancelled. 

I did try to add specs for each situation where we create a comment and send a comment email.

jessieay: OH wait how I see what you are saying. Let me double check that this has the outcomes I expected. Am getting all turned around with this `Dispatcher` class

jessieay: reverted change back to old behavior and wrote a test for it in 01cff40012e0a76d31623e124ebed9b04c3686eb

lib/dispatcher.rb
jessieay: added some descriptive private methods here 1115555e27d7566f8f893d71b7962d765b455e35

I think this is actually a win because the previous comment mentioned not sending emails to approvers who don't know about a proposal, but didn't mention that we also want to send emails to all non-approvers. methods make that clearer, I think

jessieay:
@afeld updated and green! FINALLY.

app/models/proposal.rb
```diff
@@ -271,18 +269,25 @@ def create_new_observation(user, adder, reason)
     self.observers(true)
     # when explicitly adding an observer using the form in the Proposal page...
     if adder
-      add_observation_comment(user, adder, reason) unless reason.blank?
+      if reason
```
afeld: Minor: maybe should be `reason.present?`.

lib/dispatcher.rb
```diff
@@ -29,38 +29,35 @@ def email_sent_confirmation(proposal)
...
       approval = proposal.approvals.find_by(user_id: user.id)
-      next if approval && approval.pending?
-      CommunicartMailer.new_attachment_email(user.email_address, proposal).deliver_later
+
+      if user_is_not_approver?(approval) || approver_knows_about_proposal?(approval)
```
afeld: Nice.

PR #694, opened 2015-10-14
adding public_id field to annual report

anthonygarvan:
I had some local changes pending when @afeld closed this case last week, this is the actual code that was used to create the report, which includes an Id column.


afeld:
Follow-up to https://github.com/18F/C2/pull/673.

afeld:
:shipit: on green! Also, bump on https://www.pivotaltracker.com/story/show/105300436 :wink:

jessieay:
Code climate was stuck so I re-ran. now it's green. mergin'

PR #695, opened 2015-10-14
Added Coffeelint config

yozlet:
Did this to bring CodeClimate's Coffeescript linting in line with our existing Rubocop policy. @afeld wants to discuss various config aspects.


coffeelint.json
```diff
@@ -0,0 +1,135 @@
...
+  "no_throwing_strings": {
+    "level": "error"
+  },
+  "no_trailing_semicolons": {
+    "level": "error"
+  },
+  "no_trailing_whitespace": {
+    "level": "error",
+    "allowed_in_comments": false,
+    "allowed_in_empty_lines": true
```
jessieay: we want to allow trailing whitespace in empty lines?

coffeelint.json
```diff
@@ -0,0 +1,135 @@
+{
+  "arrow_spacing": {
+    "level": "ignore"
+  },
+  "braces_spacing": {
+    "level": "ignore",
```
jessieay: I would love if it we did not ignore this!

coffeelint.json
```diff
@@ -0,0 +1,135 @@
...
+    "level": "error"
+  },
+  "empty_constructor_needs_parens": {
+    "level": "ignore"
+  },
+  "ensure_comprehensions": {
+    "level": "warn"
+  },
+  "eol_last": {
+    "level": "ignore"
```
jessieay: I would also love to not ignore this!

coffeelint.json
```diff
@@ -0,0 +1,135 @@
...
+  },
+  "newlines_after_classes": {
+    "value": 3,
+    "level": "ignore"
+  },
+  "no_backticks": {
+    "level": "error"
+  },
+  "no_debugger": {
+    "level": "warn",
```
jessieay: what about `error`? isn't it ideal to make sure debugger statements don't make it into prod?

coffeelint.json
```diff
@@ -0,0 +1,135 @@
...
+    "console": false
+  },
+  "no_empty_functions": {
+    "level": "ignore"
```
jessieay: seems worthwhile to `error` on this one

> Disallows declaring empty functions. The goal of this rule is that unintentional empty callbacks can be detected:

- http://www.coffeelint.org/

yozlet:
Thumbs up to all @jessieay's suggestions. To clarify: This first commit is literally the default generated config with one change (ignoring `max_line_length`), as opposed to something that reflects my own preferences. So have at it.

coffeelint.json
```diff
@@ -0,0 +1,135 @@
...
+  "colon_assignment_spacing": {
+    "level": "ignore",
+    "spacing": {
+      "left": 0,
+      "right": 0
+    }
+  },
+  "cyclomatic_complexity": {
+    "value": 10,
+    "level": "ignore"
```
afeld: I don't think we want to disable this.

afeld:
Some ideas:

* [ ] Run CoffeeLint as part of CI.
* [ ] Slim the `coffeelint.json` to the non-default rules. It can apparently generate this using `coffeelint --trimconfig`.
* [ ] Mention the linting in [the setup docs](https://github.com/18F/C2/blob/master/doc/setup.md#running-tests).

coffeelint.json
```diff
@@ -0,0 +1,135 @@
...
+  },
+  "no_unnecessary_fat_arrows": {
+    "level": "warn"
+  },
+  "non_empty_constructor_needs_parens": {
+    "level": "ignore"
+  },
+  "prefer_english_operator": {
+    "level": "ignore",
```
afeld: On the fence about this one...definitely good for ensuring that `==` isn't used accidentally, but I don't necessarily prefer `is` to `===`. Thoughts?

coffeelint.json
```diff
@@ -0,0 +1,135 @@
...
+    "value": "unix"
+  },
+  "max_line_length": {
+    "value": 80,
+    "level": "ignore",
+    "limitComments": true
+  },
+  "missing_fat_arrows": {
+    "level": "ignore",
```
afeld: On the fence about this one...definitely good for ensuring that instance methods won't accidentally be called in the wrong scope, but it's good practice to take care of ensuring the correct context on the caller side. Thoughts?

pkarman:
I will happily defer to group consensus on all JS formatting guidelines.

pkarman:
@yozlet, @afeld has had changes merged into this PR. If it still reflects what you wanted, I can merge. Please comment here either way.

pkarman:
@yozlet also, @afeld made some suggestions in https://github.com/18F/C2/pull/695#issuecomment-148375257 that probably need addressing, either here or in a separate PR (maybe open a Chore story in PT?).

yozlet:
@pkarman @afeld Yep, this all looks good to me. I confess that I only started this because I got irritated with CodeClimate complaining about one line that was a couple of chars too long, and figured that, hey, if we're enforcing standards for Ruby...

So I *think* that CC will automatically use this config now that it exists, but I may be wrong - I don't know enough about how CC works. If I'm right, that takes care of @afeld's first checkbox, and he's done the second, which only leaves the setup docs.

pkarman:
According to this we might need to explicitly specify our coffeelint config file.
https://github.com/codeclimate/codeclimate-coffeelint/issues/3#issuecomment-123863081

However, I don't see a way to confirm that CC is actually running that engine. Maybe we should introduce an intentional violation to confirm?

afeld:
Should be running:

https://github.com/18F/C2/blob/0933ce24ef2b21c5b595acd94daec45e474831b1/.codeclimate.yml#L10-L13

It is interesting that you can't see the output log from CodeClimate, though...

PR #696, opened 2015-10-14
[Delivers #105695746] Add default `rake` task

jessieay:
* Running `rake` runs test suite


pkarman:
+1 love this. started on similar myself a few weeks ago and got distracted. I will merge as soon as lights go all green.

PR #697, opened 2015-10-14
Add Railroady diagram of models

jessieay:
* https://github.com/preston/railroady
* Part of ATO
* Step of https://www.pivotaltracker.com/story/show/90140624
* Removed unused AR model since it was causing railroady to fail


jessieay:
Approver comments table was dropped here https://github.com/18F/C2/blob/master/db/migrate/20140512190302_drop_approvers_table.rb, but model was not removed

jessieay:
Guessing we should link to this somewhere. Suggestion on where?

afeld:
The [overview](https://github.com/18F/C2/blob/master/doc/overview.md) page is probably the most relevant, but that page might be large enough now that we should split it up (not a blocker).

afeld:
* [ ] Embed the image in an existing/new Markdown doc
* [ ] Include a note about how it was created (and thus, how it can be regenerated)

Otherwise, looks good! :ship:

jessieay:
@afeld 

Diagram embedded at cc6d8c13cac338a7b3d3b4d9d9cebb610613fedc
Generation of diagram explained at e23cb2680977498d1bc36c380e30cf0cff916af6

PR #698, opened 2015-10-14
[Delivers #105128776] Return 401 with logged out user visits '/admin'

jessieay:
* Was returning a 500 because we weren't checking for a current user
  before doing admin role check

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -82,7 +82,7 @@ class ApplicationController < ActionController::Base
 
   def authenticate_admin_user!
     if not_signed_in?
-      render nothing: true, status: 401
+      render "communicarts/authentication_error", status: 401
     elsif current_user.not_admin?
       flash[:error] = 'You need to sign in for access to this page.'
       render 'communicarts/authorization_error', status: 403
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -88,7 +88,7 @@ class ApplicationController < ActionController::Base
     authenticate_user!
 
     if not_signed_in?
-      render nothing: true, status: 403
+      render nothing: true, status: 401
     elsif current_user.not_admin?
       flash[:error] = 'You need to sign in for access to this page.'
       render 'communicarts/authorization_error', status: 403
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -82,10 +82,10 @@ class ApplicationController < ActionController::Base
 
   def authenticate_admin_user!
     if not_signed_in?
-      render "communicarts/authentication_error", status: 401
+      render "communicarts/authentication_error", status: 401, layout: 'application'
     elsif current_user.not_admin?
       flash[:error] = 'You need to sign in for access to this page.'
-      render 'communicarts/authorization_error', status: 403
+      render 'communicarts/authorization_error', status: 403, layout: 'application'
     end
   end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -70,7 +70,7 @@ class ApplicationController < ActionController::Base
   end
 
   def signed_in?
-    current_user?
+    current_user.present?
   end
 
   def not_signed_in?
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -82,7 +86,10 @@ def authenticate_user!
 
   def authenticate_admin_user!
     authenticate_user!
-    unless current_user.admin?
+
+    if not_signed_in?
+      render nothing: true, status: 403
```
pkarman: should that be a 401 instead of 403?

app/models/user.rb
```diff
@@ -82,7 +82,11 @@ def client_admin?
   end
 
   def admin?
-    self.has_role?('admin')
+    has_role?('admin')
+  end
+
+  def not_admin?
+    !admin?
   end
```
pkarman: So I'm curious about `not_admin` vs `!admin` -- one is less typing (which seems to be the Ruby way?) and the other more legible. Do you prefer readability over concision?

pkarman:
Generally looks good. Just a clarification on the status code and some style questions. Looks like tests might be backed up at CircleCI ("something is definitely afoot at the Circle K, Ted").

app/controllers/application_controller.rb
jessieay: right you are! have a test for it but circle is down so I didn't realize it was failing. good :baseball:

app/models/user.rb
jessieay: I do prefer readability over concision. I don't usually describe Ruby as terse, although compared to something in Obj-C it certainly is.

I think prepending a method with `!` can be very readable at times, but I tend to make small helper methods for `!`-prepended methods because it's just more pleasant to read. Plus, it makes it much different from something like `!!admin?` which is probable my least favorite thing in Ruby and something that makes me cry every time I see it.

app/controllers/application_controller.rb
```diff
@@ -70,7 +70,11 @@ def sign_out
   end
 
   def signed_in?
-    current_user ? true : false
+    current_user?
```
afeld: Tests failing, since this isn't a method.

app/models/user.rb
afeld: I don't feel super strongly about this method in particular, but the extension of this argument is that you then need two methods for every method that returns a boolean, which seems excessive.

afeld:
Besides the test failure, looks good! :shipit:

jessieay:
@afeld it's passing now!

pkarman:
I'm +1 to merge once this is QA'd. I can do that.

pkarman:
Ok I tested on c2-dev. It works as written.

That said, I now wonder if rendering a blank page is very helpful UX. Should we instead be redirecting to `/` with a flash error about requiring sign-in?

jessieay:
That is what `authenticate_user!!` does... could re-use that behavior.

I was going to mention this earlier but didn't want to stall us too much: I think the best thing here might be to actually return a `404` -- similar to what github does when you visit a repo you are not authorized to view. This SO answer explains benefits of 403/404 better than I can: http://security.stackexchange.com/a/46568

pkarman:
404 is good for when you're in an obsecurity model. Since we're OSS, not sure that really matters much. I prefer the status to reflect reality but my biases stem from my API background.

Could we return the 401 and render the `authentication_error.html.haml` template?

jessieay:
@pkarman good point, hadn't thought about the fact that anyone could view GH and know we have an admin portal. Will update to render `authentication_error`

app/models/user.rb
jessieay: @afeld yeah, basically I was trying to get away from rubocop blowing up on me because of an `if !` (rubocop wants that to be an `unless`)...don't think all booleans should have two methods

jessieay:
@pkarman updated to use `authentication_error` template at 1e94e5c523bfc4ddffe0889c1ddb91ee5ef53fa9

pkarman:
@jessieay looking good. I just pushed one commit (5251ed6) to pretty it up a little. If you approve, I will merge.

jessieay:
@pkarman looks good! surprised that layout isn't used by default?

pkarman:
I'm not 100% certain but I think the layout uses the current controller (subclass) instead of the controller where the method is defined (parent class) so in this case, we are being explicit about the layout regardless of the current controller.

PR #700, opened 2015-10-15
One feature spec per file for NCR work order specs

jessieay:
* Was trying to see if * https://www.pivotaltracker.com/story/show/98774038
  was already implemented, but was having trouble navigating the 662
  line spec file
* Good rule of thumb is to have one featuer spec file per feature
* This split out will help me actually see what's going on
* Working on this refactor made it clearer why `before` and `let`
  statements be very problematic: splitting the specs into separate
  files was very hairy because they were nested and some of the parent
  spec's setup was required for nested specs and some wasn't. Figured it
  out through trial and error but would still be nice to move toward
  including all of the setup for each spec within the test itself (which
  can still mean putting shared logic outside of the specs, but as a
  method that needs to be called rather than a `before` or `let`)


pkarman:
+1 to more files ("take two, they're free")

I do agree with localizing test fixture data inside each spec; much more legible.

afeld:
Heh, I actually did something similar in https://github.com/18F/C2/commit/57fa81a, but split them up by the [controller] actions (edit, approve, etc.). I want to continue to play with the arrangement, but this makes a lot of sense as a first step. Because the merge conflicts on file splits get to be a nightmare, going to merge now, and we can iterate on it. Thanks for doing it!

PR #702, opened 2015-10-15
tweak CoffeeLint config

afeld:
Building on https://github.com/18F/C2/pull/695.

Easiest way for me to figure out which rules I do/don't agree with is to run it locally. My process:

1. Changed all of the `level`s to `error` (i.e. enabled all of the rules).
1. Ran `coffeelint app/assets/javascripts/`.
1. For lint errors I agreed with, made the appropriate fix(es). For those I didn't, changed the config value to `ignore`.
1. Committed as 65124c3
1. Removed the config options that matched the defaults (044a329)

`coffeelint app/assets/javascripts/` passes locally.

### DIFF
--- a/app/assets/javascripts/filter.js.coffee
+++ b/app/assets/javascripts/filter.js.coffee
@@ -18,20 +18,20 @@ class Filter
 
   filter: ($el) ->
     value = $el.val()
-    if !$el.is(":checked")
+    unless $el.is(":checked")
       value = "!" + value
     @$("[data-filter-key=#{ @key }]").each (idx, el) ->
-      hidden = el.getAttribute("data-filter-value") != value
+      hidden = el.getAttribute("data-filter-value") isnt value
       el.setAttribute("aria-hidden", hidden.toString())
 
   hideAll: () ->
     @$("[data-filter-key=#{ @key }]").attr("aria-hidden", true)
 
-  this.generateIn = ($scope) ->
+  @generateIn = ($scope) ->
     filters = {}
     $scope.find("[data-filter-control]").each (idx, el) ->
       key = el.getAttribute('data-filter-control')
-      if !filters.hasOwnProperty(key)
+      unless filters.hasOwnProperty(key)
         filters[key] = new Filter($scope, key)
     filters
 
--- a/app/assets/javascripts/filter.js.coffee
+++ b/app/assets/javascripts/filter.js.coffee
@@ -18,10 +18,10 @@ class Filter
 
   filter: ($el) ->
     value = $el.val()
-    unless $el.is(":checked")
+    if !$el.is(":checked")
       value = "!" + value
     @$("[data-filter-key=#{ @key }]").each (idx, el) ->
-      hidden = el.getAttribute("data-filter-value") isnt value
+      hidden = el.getAttribute("data-filter-value") != value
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/filter.js.coffee
```diff
@@ -18,20 +18,20 @@ class Filter
 
   filter: ($el) ->
     value = $el.val()
-    if !$el.is(":checked")
+    unless $el.is(":checked")
       value = "!" + value
     @$("[data-filter-key=#{ @key }]").each (idx, el) ->
-      hidden = el.getAttribute("data-filter-value") != value
```
jessieay: I prefer this! Find it easier to spot a symbol than a word. I think because there's always a chance `isnt`is a var or something

app/assets/javascripts/filter.js.coffee
```diff
@@ -18,20 +18,20 @@ class Filter
...
   hideAll: () ->
     @$("[data-filter-key=#{ @key }]").attr("aria-hidden", true)
 
-  this.generateIn = ($scope) ->
+  @generateIn = ($scope) ->
     filters = {}
     $scope.find("[data-filter-control]").each (idx, el) ->
       key = el.getAttribute('data-filter-control')
-      if !filters.hasOwnProperty(key)
+      unless filters.hasOwnProperty(key)
```
jessieay: always prefer `if` to `unless` - I find unless has a concept very hard to understand. Eg: "Unless Aidan has brown hair, do XYZ" -- I have to think really hard about whether we should do XYZ. If the sentence is "If Aidan does not have brown hair, do XYZ" then I am like "alright, don't need to worry about XYZ because I know Aidan DOES"

app/assets/javascripts/filter.js.coffee
afeld: Fair.

app/assets/javascripts/filter.js.coffee
afeld: I don't _necessarily_ agree, but will change it back, since it's beside the point of the PR.

afeld:
Updated!

coffeelint.json
```diff
@@ -1,135 +1,54 @@
...
-    "level": "ignore"
-  },
-  "no_throwing_strings": {
     "level": "error"
   },
-  "no_trailing_semicolons": {
+  "eol_last": {
     "level": "error"
-  },
-  "no_trailing_whitespace": {
```
jessieay: we want to allow trailing whitespace?

jessieay:
General questions about PRs that are branches of other PRs: does it make more sense to put all of my comments on here or should I also comment on https://github.com/18F/C2/pull/695 ? I have some comments over there that might be addressed by this PR, but would be a lot of manual work to cross check and see what is addressed by this PR and what isn't.

coffeelint.json
afeld: Ah, I didn't catch that one because we didn't have any. Added.

afeld:
@jessieay Definitely not worth putting them in both places...whatever works, I guess. Was thinking the secondary PR would make things easier by not putting in a million comments, but that might have been an incorrect guess.

pkarman:
I'm +1 to merge this to the original PR.

PR #703, opened 2015-10-15
[Delivers #89433396] add rake task for sending reminder emails

pkarman:
This PR refers to this story: https://www.pivotaltracker.com/n/projects/1149728/stories/89433396

It carves out a namespace for reminding users to take actions, starting with approvers of pending requests.

### DIFF
--- a/lib/tasks/reminder.rake
+++ b/lib/tasks/reminder.rake
@@ -1,0 +1,21 @@
+namespace :reminder do
+  desc "Send actionable approver reminder email for pending requests"
+  task actionable_approvers: :environment do
+    props = Proposal.pending.where(client_data_type: "Ncr::WorkOrder")
+    props.each do |proposal|
+      send_reminder_email(proposal)
+    end
+  end
+
+  private
+
+  def send_reminder_email(proposal)
+    user = proposal.client_data.current_approver or return
+    approval = proposal.individual_approvals.find_by(user: user)
+    if ENV['SEND_OK']
+      Dispatcher.on_approval_approved(approval)
+    else
+      puts "Remind #{user.email_address} #{proposal.id} #{proposal.created_at} #{approval.id} #{approval.created_at}"
+    end
+  end
+end
--- a/lib/tasks/reminder.rake
+++ b/lib/tasks/reminder.rake
@@ -11,7 +11,7 @@ namespace :reminder do
 
   def send_reminder_email(proposal)
     user = proposal.client_data.current_approver
-    if !user
+    unless user
       return # legit for there to be no user, but then no reminder needed.
     end
     approval = proposal.individual_approvals.find_by(user: user)
--- a/lib/tasks/reminder.rake
+++ b/lib/tasks/reminder.rake
@@ -1,8 +1,8 @@
 namespace :reminder do
   desc "Send actionable approver reminder email for pending requests"
   task actionable_approvers: :environment do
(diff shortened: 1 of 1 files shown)


lib/tasks/reminder.rake
```diff
@@ -0,0 +1,21 @@
...
+    props = Proposal.pending.where(client_data_type: "Ncr::WorkOrder")
+    props.each do |proposal|
+      send_reminder_email(proposal)
+    end
+  end
+
+  private
+
+  def send_reminder_email(proposal)
+    user = proposal.client_data.current_approver or return
```
jessieay: `proposal.client_data.current_approver or return` -- not sure that is super readable to me. What about wrapping this all in conditional to take care of the situation where there is not an approver?

lib/tasks/reminder.rake
```diff
@@ -0,0 +1,21 @@
+namespace :reminder do
```
jessieay: might make sense to have a more generic namespace, maybe something like `email` ?

lib/tasks/reminder.rake
```diff
@@ -0,0 +1,21 @@
...
+  desc "Send actionable approver reminder email for pending requests"
```
jessieay: I am not sure why there is a need for this because there is no linked tracker story or test. Can we get one or both of those?

lib/tasks/reminder.rake
```diff
@@ -0,0 +1,21 @@
...
+  task actionable_approvers: :environment do
+    props = Proposal.pending.where(client_data_type: "Ncr::WorkOrder")
```
jessieay: what about `proposals`? Whenever I see `props` I think it means `properties` :)

pkarman:
thanks for the review @jessieay. I have fixed code where appropriate and expanded the PR description for more context.

lib/tasks/reminder.rake
```diff
@@ -0,0 +1,24 @@
...
+    proposals.each do |proposal|
+      send_reminder_email(proposal)
+    end
+  end
+
+  private
+
+  def send_reminder_email(proposal)
+    user = proposal.client_data.current_approver
+    unless user
```
jessieay: for the record, I do not prefer `unless` to if! I love `if` and have some sort of weird mental block when it come to `unless`. Like, right now, I am thinking really hard about whether or not we will `return` here if there is a user.

jessieay:
I am still not sure I understand the context for this? Why are we setting up a manual rake task for something that seems like we would want it done regularly? (sorry, I know there was context on Slack that I missed)

lib/tasks/reminder.rake
pkarman: Take it up with rubocop. That's why I changed it.

I don't disagree with you; just requires a change to our rules. Maybe start a VOTE thread on slack or a PR to the `.rubocop.yml` ?

jessieay: PR it is! Didn't realize rubocop was such a punk

jessieay:
:ship: :ship: :ship:

pkarman:
this PR doesn't really address the original story.

lib/tasks/reminder.rake
phirefly: It would be also possible to do `return unless user = proposal.client_data.current_approver`, no?

PR #704, opened 2015-10-15
Disabled "Negated If" Rubocop style

jessieay:
* I think it is clearer to say `if !blah` than `unless blah`

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,6 +15,8 @@ Style/EmptyElse:
   Enabled: false
 Style/GuardClause:
   Enabled: false
+Style/NegatedIf:
+  Enabled: false
 Style/IfUnlessModifier:
   # inline conditionals don't get picked up by simplecov, so preferable to avoid them
   Enabled: false


pkarman:
could we split these changes into 2 PRs? the codeclimate results are going to bomb because of quote infractions, and it would be nice to limit the scope of that change with a code clean up at the same time.

jessieay:
That's weird, I thought it only ran rules on changed code, not all code. Happy to split.

jessieay:
@pkarman updated

pkarman:
+1 from me. 

I think we need some (lazy) consensus on style changes like this. Maybe leave open for 72 hours for any -1 comments?

jessieay:
@pkarman good idea!

afeld:
I'm `!in_favor` of this (ha) – I don't think `if !` is less confusing because it's easy to miss the `!` – but don't feel super strongly about it.

jessieay:
@afeld did you mean to say that you are not in favor of this unless others feel strongly about it? Because if you did, I have no idea what that means. :)

afeld:
Haha. I'm saying that I'm a -0.5 on this.

pkarman:
```
unless consensus_reached
  use_the_language_as_it_exists
end
```

jessieay:
@pkarman what do you mean by that? This isn't about Ruby it's about Rubocop.

pkarman:
it was mostly an (ill-made) attempt at humor.

I meant, the Ruby language has a built-in conditional `unless` and it yet it very common to see `if !` instead. Why?

Ruby gets its `unless` from Perl, so if you google `ruby unless` and `perl unless`  you'll get many of the same arguments, for and against. The word was originally included by Larry Wall (a linguist before he invented Perl) because it made certain code easier to read. Not all code, but certain code. The same is true of `until` as equivalent of `while !`.

My point is that having a strict rule that says "always use `unless`" or "always use `if !`" is going to miss some cases where the opposite actually makes for clearer code. It's a language after all, and the inventor thought there were valid reasons for creating a built-in keyword.

My takeaway from the links that follow is that `unless` is ok in simple cases, unless  you have an `else` in which case `unless` is not ok.

https://signalvnoise.com/posts/2699-making-sense-with-rubys-unless
http://www.perlmonks.org/?node_id=947
http://www.railstips.org/blog/archives/2008/12/01/unless-the-abused-ruby-conditional/

jessieay:
@pkarman agreed, there will be cases where `unless` makes sense. I think a better phrasing would be "avoid unless" rather than "never use unless" 

We've had many convos about `unless` and I still find it confusing and would rather we not have to discuss it on every few PRs - suggestions on how to proceed?

pkarman:
if the phrasing is "avoid" rather than "never" then I think it goes into a [as-yet-non-existent] style guide rather than rubocop, who is not as adept at parsing the difference.

jessieay:
@pkarman ok, will add this to a style guide markdown file and close this. thanks for discussing!

PR #705, opened 2015-10-15
Rubocop: enforce single quotes

jessieay:


jessieay:
this will probably fail code climate but want to get consensus before going through and changing all the quotes

pkarman:
TBH I'm somewhat mixed on this topic. I actually prefer double quotes for string literals because I find it easier later if I am refactoring and need to add some interpolated value to the string.

E.g. if this:
```
if foo == 'bar'
```
later needs to become this:
```
if foo == "bar #{color}"
```
it becomes quite possible that I will miss the single-vs-double quote and do this:
```
if foo == 'bar #{color}'
```
which will Fail.

This post shares my bias: http://anti-pattern.com/always-use-double-quoted-strings-in-ruby

OTOH, I understand all the valid reasons single quotes are preferred. This thread is illustrative of the (cultural) divide: https://github.com/bbatsov/ruby-style-guide/issues/96

jessieay:
@pkarman I also prefer double quotes for the reasons you list above! 

But what I prefer most of all is consistency, so I am happy to stick with single quotes if that is what *most* people want. It does appear that there are a lot of single quotes in this project, but there are a lot of dubs, too.

pkarman:
we are a caucus of 2! who is with us?!

I will confess to sometimes being inconsistent myself on this score. When defining constants, e.g., I will often use single quotes because I never expect a constant to require interpolation later. Likewise in some tests I am sometimes lazy. It's the SHIFT key after all, isn't it? ;)

afeld:
A couple data points:

* My style (which I admit is...unique) is to use double-quoted strings for things like phrases/sentences, and single quotes for things like hash keys.
* GitHub Inc. uses double quotes everywhere, for [the reasons above](#issuecomment-148484659).

I'm +0 on the change, but would favor merging this with Code Climate failing rather than going through and changing every occurrence, since that causes merge conflict headaches with all the other open PRs. Code Climate will run the checks for any new PRs going forward, so it can be changed over incrementally.

jessieay:
@afeld if we are all open to the idea of standardizing, and there is some agreement on double quotes, what about changing this rule to enforce double quotes everywhere? 

I am +1 to your idea of merging with CC failing

afeld:
@jessieay My attitude toward coding style is that I'm not completely opposed to standardizing across a project/team, but I also wouldn't ever want to hold up a bugfix or feature for something tiny like single vs. double quotes; the benefit doesn't outweigh the cost, in my opinion. You'll notice that we have disabled more and more `Style` cops in [the Rubocop config](https://github.com/18F/C2/blob/master/.rubocop.yml) over time, rather than the other way around.

All of that being said, I defer to the will of the team.

pkarman:
I am reluctant to add more rules, esp since this is really a (personal) style thing. It seems we're all on the same page that we use double quotes most of the time, and when we don't, we probably have a reason not to. I would like to avoid getting dinged by code climate when I consider my judgement superior on such a small thing.

jessieay:
@pkarman @afeld 

> My attitude toward coding style is that I'm not completely opposed to standardizing across a project/team, but I also wouldn't ever want to hold up a bugfix or feature for something tiny like single vs. double quotes

I also don't think it should hold it up. As a reviewer, I would never want to not merge a PR because of something small like single/double quotes. This is why automated tools are helpful - if we have a rule for this in rubocop, the PR author will address any style issues before the PR is reviewed.

> this is really a (personal) style thing

yes, not trying to determine which style is superior or anything like that. Just trying to standardize on one so that the app has a consistent style. 

The benefit of a consistent style is that it makes it easier to read and navigate a project if there a certain established norms. Single/double quotes is just one small thing, but it is part of a larger whole that we should also be working to make more consistent. If we all just write Ruby according to our own personal styles, this app will become more difficult to understand and then difficult to contribute to. 

####

Most importantly, I would like the enforcement of these rules to be programmatic. 

We are much too talented a crew to be spending our time doing pull request reviews with comments like "use double quotes here?" or "Stay away from ternary operators, please" or "Avoid !!" -

Rubocop should deal with all of those things so we can focus on things that a regex could never detect. Naming. Organization. Different implementation ideas. 

As I see it, the options are: 

1) Don't worry about consistent style (no Rubocop, no PR comments about style)
2) Worry about consistent style but allow for exceptions by not using Rubocop for enforcement
3) Use Rubocop to enforce consistent style

I think there are major problems with the first 2 for the reasons explain above. So I vote for option 3.

Interested to hear from @annalee and @yozlet  on this, too.

annalee:
I agree that standards are good and that, in general, automatic enforcement is better than copyediting PRs.

I also favor defaulting to double quotes for reasons already discussed.

yozlet:
What @annalee said.

pkarman:
To be clear: I like 3) Use Rubocop to enforce consistent style, in general.

For this particular issue, I don't want Rubocop to flag this as a violation:
```
 FOO = 'bar'
```
when I have very intentionally used single quotes in my constant definition because I know that a constant should (a) never need interpolation in its definition and (b) single quotes enforce the grokability of "this is a static string".

That said, if we all decide this is preferred:
```
FOO = "bar"
```
then so be it, I will defer.

jessieay:
@pkarman @yozlet @annalee @afeld 

Changed this to standardize on double quotes. Sounds like we have consensus on that.

pkarman:
So 'be' "it".

PR #706, opened 2015-10-16
improve availability of production

afeld:
I've been seeing the app hanging and `502 Bad Gateway` errors when trying to load the site a few times recently, and yesterday at 5pm ET the app was `down` for no particular reason. After looking at the top graph in New Relic, I believe the problem is a memory constraint during garbage collection.

screen shot 2015-10-15 at 5 37 56 pm

There may be other issues that are causing us to consume that much memory, but increasing the instance capacity and having a backup instance in production should at least keep it from crashing. See the individual commits for more explanation of the changes.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -18,7 +18,6 @@ MYUSA_SECRET: consumer_secret_key
 # DEFAULT_URL_SCHEME
 # FORCE_USER_ID=123456
 # GA_TRACKING_ID
-# MAX_THREADS
 # MYUSA_URL
 # NCR_BA61_TIER1_BUDGET_MAILBOX
 # NCR_BA61_TIER2_BUDGET_MAILBOX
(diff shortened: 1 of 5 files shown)


jessieay:
Since we can change this with ENV var value changes rather than code, is it worth trying to change the env var vals first and then hard-coding defaults later if it turns out to be a good change?

afeld:
@jessieay The issue this PR is (attempting to) solve is hitting the memory limit within individual instances. By increasing the number of `instances`, we are improving availability in case one of them _does_ crash. Two `instances` means we will have twice as many server threads available, but we already had more capacity than needed, so I reduced the number to keep our capacity about equivalent to what it was before. Does that make sense?

> is it worth trying to change the env var vals first and then hard-coding defaults

I mean, I'm kinda in favor of the "configuration as code" idea...this way we can see the history.

---

@dlapiduz Would love feedback in general!

pkarman:
+1 to config in code. Even if we are experimenting, I like the see the version history of the experiment.

jessieay:
@pkarman @afeld I am totally fine with that, but should we remove the `MAX_THREADS` env var since we don't appear to be using it?

afeld:
Updated!

jessieay:
@afeld :+1: :ship: just waiting for Code Climate to pass before merging

jessieay:
MERGING (doesn't appear that this needs QA)

PR #707, opened 2015-10-16
[Delivers #98483692] client_slug filter approver

pkarman:
Story: https://www.pivotaltracker.com/story/show/98483692

The real feature change is one line: 
```diff
app/helpers/ncr/work_orders_helper.rb
@@ -1,9 +1,8 @@
 module Ncr
   module WorkOrdersHelper
     def approver_options
-      # @todo should this list be limited by client/something else?
       # @todo is there a better order? maybe by current_user's use?
-      User.order(:email_address).pluck(:email_address)
+      User.where(client_slug: 'ncr').order(:email_address).pluck(:email_address)
     end
 
     def building_options
```

Everything else is updating the spec suite to make the client_slug explicit and propagated through all the User-related associations.


app/helpers/ncr/work_orders_helper.rb
```diff
@@ -1,9 +1,8 @@
 module Ncr
   module WorkOrdersHelper
     def approver_options
-      # @todo should this list be limited by client/something else?
       # @todo is there a better order? maybe by current_user's use?
-      User.order(:email_address).pluck(:email_address)
+      User.where(client_slug: 'ncr').order(:email_address).pluck(:email_address)
```
jessieay: perhaps create a method on `User` with this logic in it so we can have a unit test?

app/models/proposal.rb
```diff
@@ -48,6 +48,7 @@ class Proposal < ActiveRecord::Base
 
   validates :client_data_type, inclusion: {
     in: ->(_) { self.client_model_names },
+    message: "%{value} is not in client_model_names #{CLIENT_MODELS.inspect}",
```
jessieay: What about:

`"%{value} is not a valid client model type. Valid client model types are: #{CLIENT_MODELS.inspect}",`

spec/controllers/proposals_controller_spec.rb
```diff
@@ -73,7 +73,7 @@
 
       before do
         expect(Proposal).to receive(:client_model_names).and_return(['SomeCompany::SomethingApprovable'])
-        expect(Proposal).to receive(:client_slugs).and_return(%w(some_company some_other_company))
+        expect(Proposal).to receive(:client_slugs).at_least(:once).and_return(%w(some_company some_other_company ncr))
```
jessieay: why `at_least` once? Also, feels off to me to have an expectation in a `before` block -- thoughts on moving into individual specs?

spec/factories/proposals.rb
```diff
@@ -4,51 +4,52 @@
     status 'pending'
     association :requester, factory: :user
 
+    transient do
+      client_slug { ENV['CLIENT_SLUG_DEFAULT'] || 'ncr' }
```
jessieay: fancy!

spec/features/requester_edits_ncr_work_order_spec.rb
```diff
@@ -7,10 +7,11 @@
 
   let(:work_order) { create(:ncr_work_order, description: 'test') }
   let(:ncr_proposal) { work_order.proposal }
-  let!(:approver) { create(:user) }
+  let!(:approver) { create(:user, client_slug: 'ncr') }
```
jessieay: maybe we should make `ncr` a constant? Seems like we are using it all over the place and is worthy of constantization

spec/controllers/proposals_controller_spec.rb
pkarman: The `at_least` is there because the intent of that whole line is to override the Proposal.client_slugs method to avoid breaking the User validation, and apparently the `receive ... and_return` was defaulting to an implicit `.once` that started breaking when I added the explicit `client_slug` value.

Yes, if we remove the `expect` part of that stubbing action, then the problem would go away. I.e.
```
allow(Proposal).to receive(:client_slugs).and_return(...)
```
might be enough.

app/helpers/ncr/work_orders_helper.rb
pkarman: there is a unit test on the helper class itself in `spec/helpers/ncr/work_orders_helper_spec.rb` which has been changed to reflect this new feature.

I got the impression that we were trying to keep the model class limited to business-rule-specific code, and put presentation in decorators or helpers. This method feels like the latter and has test coverage.

That seem ok to you?

spec/features/requester_edits_ncr_work_order_spec.rb
pkarman: I dunno. Doesn't save much typing. It's not quite a constant use case. The string itself is user data (stored as a string in the db) and so doing comparisons might require a cast? Validation assumes string comparison? Maybe? Not obvious (to me) that it's a clear win.

pkarman:
thanks for the review @jessieay -- I have committed changes where I agreed and made comments where I had questions.

app/helpers/ncr/work_orders_helper.rb
jessieay: Seems OK to keep as-is.

To explain my thinking: If it were me, I would write a unit test for this because it is business-specific. It's a query on the `User` class. It happens to be used by a helper, but the query itself is an operation on `User` so it would be nice to just call `User.ordered_by_email_address` and know that the underlying logic is covered by a unit test.

Again, totally fine as-is, that's just the thinking on how I would handle it.

spec/features/requester_edits_ncr_work_order_spec.rb
jessieay: not so much about the typing, more about making sure we don't have typos. If we have a typo on the string, it won't throw an error. If we have a typo on a constant, it will throw an `undefined constant` erorr

pkarman: Actually, that behavior is why I don't think we want it to be a constant. A typo on the string *would* throw a validation exception because `User.client_slug` validates against `Proposal.client_slugs`.

I'm -0 on the change. I think keeping it a string keeps exercising the validation and doesn't mean needing to type-convert anywhere.

jessieay: That makes sense, re: string validation.

What do you mean by "type-convert"?

app/helpers/ncr/work_orders_helper.rb
pkarman: I tend to have the same thinking as you. Per this conversation with Aidan, however, I'm trying to keep things out of the model that are just view-based extensions: https://github.com/18F/C2/pull/681#discussion_r41535674

jessieay: Makes sense. Another option would be to create a query object encapsulate this logic and then use the query object in the helper, but this doesn't seem complicated enough to warrant that.

Good blog post on query objects http://blog.codeclimate.com/blog/2012/10/17/7-ways-to-decompose-fat-activerecord-models/

afeld: I'm fine with keeping this method as-is, since the query is relatively simple.

> Per this conversation with Aidan, however, I'm trying to keep things out of the model that are just view-based extensions

That specific case was that it was returning a placeholder string (essentially, `"user not found"`) when the query didn't return anything, which definitely feels like it falls under the realm of Helpers. I'm kind of indifferent about where queries [that return data] live, even if they are only being used for views.

> Another option would be to create a query object

Very :+1: on query objects in general! Especially for complex queries, or very purpose-specific ones. We have some under [`lib/query/`](https://github.com/18F/C2/tree/master/lib/query) already.

jessieay:
Code looks good! But there is a merge conflict. Mind updating, @pkarman ?

spec/features/requester_edits_ncr_work_order_spec.rb
pkarman: cast to a string from a symbol: `client_slug.to_s` etc

jessieay:
QAing now

jessieay:
Confirmed on staging that a User with the `18fgsa` slug does not show up in the approving official dropdown when creating an NCR request

screen shot 2015-10-20 at 3 43 40 pm

screen shot 2015-10-20 at 3 43 49 pm

PR #708, opened 2015-10-18
attempt to improve the model diagram

afeld:
PR on top of https://github.com/18F/C2/pull/697. /cc @jessieay

@cmc333333 [pointed out](https://18f.slack.com/archives/dev/p1445135628000121) that there are a lot of duplicated classes in the graph, which is a known issue, apparently: https://github.com/preston/railroady/issues/40. I also wanted to see if I could improve the layout at all – laid out my steps in a new doc. I still don't _love_ the output, but think it is easier to understand.


cmc333333:
Looks a lot less cluttered!  

Random thought: no doubt  difficult to implement, but it'd be interesting to model "through" relationships differently -- `Ncr::WorkOrder` definitely has a connection to `Comment`, but only by-way-of `Proposal`. I wonder, if all the "through" relationships were deleted, if it'd be more apparently that `Proposal` were the middleman

jessieay:
Looks good! I wasn't originally too worried about the model diagram because it's mostly part of the ATO documentation stuff, so more important to just have it than to have it be perfect..but this looks much better!!

In the future, feel free to give me feedback on PRs rather than branching off of them -- in which situations do you tend to branch rather than give feedback?

afeld:
> feel free to give me feedback on PRs rather than branching off of them -- in which situations do you tend to branch rather than give feedback?

I default to just giving feedback, but wasn't honestly wasn't sure what was possible through railroady/graphviz, so ended up playing with it myself, out of curiosity. I wouldn't have wanted to hold up your PR with a vague request like "improve the layout of the graph", so took that exploratory burden on myself. I'll also do PRs when it seems easier to show than to tell, a la https://github.com/18F/C2/pull/712.

_Hadn't really articulated that before...was nice to write it down!_

PR #709, opened 2015-10-18
use a default port, if not specified

afeld:


pkarman:
+1 for the change. (Aside: that Bash syntax is wacky. Cool, but wacky.)

QA'd on c2-staging.

PR #710, opened 2015-10-18
[Delivers #105339678] display the user's email if their first or last name are empty strings

afeld:


pkarman:
+1, QA'd on staging per reported problem URLs. Works as expected.

PR #711, opened 2015-10-19
[Delivers #105137620] use Climate Control gem for setting environment variables in specs

afeld:
Doesn't take out quite as much of our support code as I was hoping it would, but ah well.


afeld:
Gem suggested by @jessieay in https://github.com/18F/C2/pull/669#discussion_r41262157.

pkarman:
looks good. Merging w/o explicit QA since it only affects the rspec suite which is passing.

PR #713, opened 2015-10-19
[Delivers #106019438] CSV should respect proposal status when picking approver

pkarman:
Story: https://www.pivotaltracker.com/story/show/106019438

This makes the CSV logic match the in-line report email logic.

### DIFF
--- a/app/decorators/ncr/work_order_decorator.rb
+++ b/app/decorators/ncr/work_order_decorator.rb
@@ -13,6 +13,14 @@ module Ncr
       approver_email_address(final_approver)
     end
 
+    def status_aware_approver_email_address
+      if proposal.approved?
+        final_approver_email_address
+      else
+        current_approver_email_address
+      end
+    end
+
     private
 
     def approver_email_address(approver)
--- a/spec/decorators/ncr_work_order_decorator_spec.rb
+++ b/spec/decorators/ncr_work_order_decorator_spec.rb
@@ -19,10 +19,16 @@ describe Ncr::WorkOrderDecorator do
     expect(wo.current_approver_email_address).to eq(wo.approvers.first.email_address)
   end
 
-  it "returns approver email address based on proposal status" do
+  it "returns current approver email address when status is pending" do
     wo = create(:ncr_work_order, :with_approvers).decorate
+    expect(wo.pending?).to be_truthy
     expect(wo.status_aware_approver_email_address).to eq(wo.current_approver_email_address)
+  end
+
+  it "returns final approver when status is approved" do
+    wo = create(:ncr_work_order, :with_approvers).decorate
     wo.proposal.approve!
+    expect(wo.approved?).to be_truthy
     expect(wo.status_aware_approver_email_address).to eq(wo.final_approver_email_address)
   end
 end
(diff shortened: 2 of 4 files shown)


app/decorators/ncr/work_order_decorator.rb
```diff
@@ -13,6 +13,14 @@ def final_approver_email_address
       approver_email_address(final_approver)
     end
 
+    def status_aware_approver_email_address
```
jessieay: I know we have the `reporter` spec below, but what about a unit test for this to explain reasoning behind logic? If I were going to update this decorator, I would assume it wasn't covered by tests because there is no unit test.

pkarman:
@jessieay thanks, good call on more explicit tests. Added in a838d5c

spec/decorators/ncr_work_order_decorator_spec.rb
```diff
@@ -18,4 +18,11 @@
     wo = create(:ncr_work_order, :with_approvers).decorate
     expect(wo.current_approver_email_address).to eq(wo.approvers.first.email_address)
   end
+
+  it "returns approver email address based on proposal status" do
```
jessieay: don't want to be a nitpicker here, but it might be easier to understand the behavior with a separate test for each use case.

`it "returns current approver email address for pending proposals" do`

`it "returns final approver email address for approved proposals" do`

pkarman:
@jessieay test split in 6370566

spec/decorators/ncr_work_order_decorator_spec.rb
```diff
@@ -18,4 +18,17 @@
...
+  it "returns current approver email address when status is pending" do
+    wo = create(:ncr_work_order, :with_approvers).decorate
+    expect(wo.pending?).to be_truthy
+    expect(wo.status_aware_approver_email_address).to eq(wo.current_approver_email_address)
+  end
+
+  it "returns final approver when status is approved" do
+    wo = create(:ncr_work_order, :with_approvers).decorate
+    wo.proposal.approve!
+    expect(wo.approved?).to be_truthy
```
jessieay: Don't think this needs to change now, but for future I would say let's avoid `truthy` in favor or RSpec predicate matchers

`expect(wo).to be_approved`

Super long convo on this topic and explanations for why here https://github.com/thoughtbot/guides/pull/191

jessieay:
Code looks good to go! :ship: Waiting on CI to pass before QAing

jessieay:
yo @pkarman I am trying to QA this but I am not sure how. Could you perchance provide some steps on how to QA the csv feature?

pkarman:
@jessieay You can manually run the report and send yourself the csv files like this:
```
cf-ssh c2-staging
[lots of waiting]
% rails c
> ENV['BUDGET_REPORT_RECIPIENT'] = 'your.email'
> ReportMailer.budget_status
```

To really verify that the problem is fixed, might want to actually run that against `c2-prod` with `cf-ssh` so that you can generate the report against the same data that raised the bug report originally. We can slack about it.

jessieay:
thanks for the info! Will QA later this morning

jessieay:
grrrrr I am having trouble with `cf-ssh`

```
cf-ssh c2-staging
Deploying SSH container 'c2-dev-ssh'...
2015/10/20 15:12:17 Failed to run SSH container: exit status 1
```

pkarman:
@jessieay to expedite, I will run the QA process on prod and send the email to both of us for confirmation.

pkarman:
@jessieay I have generated a report and sent to you. You should have 2 budget reports in your inbox: one generated from the current master HEAD branch and one from this PR branch. The way to confirm the fix is to verify that the attached CSV file lists the same approver email address as the inline report does. Look at "Approved BA61 requests last week" for example, and check the FY16-1155 request.

jessieay:
@pkarman just checked out the email and the approvers in the CSV match. Thanks for the QA instructions, very helpful.

PR #714, opened 2015-10-19
[Delivers #104865396] Ensure adding observer with reason creates an update comment

yozlet:
Also, adding an observer _without_ specifying a reason should not create a comment.

### DIFF
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -268,7 +268,7 @@ class Proposal < ActiveRecord::Base
     self.observers(true)
     # when explicitly adding an observer using the form in the Proposal page...
     if adder
-      if reason and reason.present?
+      if reason && reason.present?
         add_observation_comment(user, adder, reason)
       end
 
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -268,7 +268,7 @@ class Proposal < ActiveRecord::Base
     self.observers(true)
     # when explicitly adding an observer using the form in the Proposal page...
     if adder
-      if reason
+      if reason and !reason.blank?
         add_observation_comment(user, adder, reason)
       end
 
@@ -286,6 +286,7 @@ class Proposal < ActiveRecord::Base
         observer: user.full_name,
         reason: reason
       ),
+      update_comment: true,
       user: adder
     )
   end
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -268,7 +268,7 @@ class Proposal < ActiveRecord::Base
     self.observers(true)
     # when explicitly adding an observer using the form in the Proposal page...
     if adder
-      if reason and !reason.blank?
+      if reason and reason.present?
         add_observation_comment(user, adder, reason)
       end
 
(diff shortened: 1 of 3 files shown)


app/models/proposal.rb
```diff
@@ -268,7 +268,7 @@ def create_new_observation(user, adder, reason)
     self.observers(true)
     # when explicitly adding an observer using the form in the Proposal page...
     if adder
-      if reason
+      if reason and !reason.blank?
```
jessieay: about about `if reason && reason.present?`

jessieay:
> adding an observer without specifying a reason should not create a comment.

do we have a test for this? If not, can we add one?

yozlet:
@jessieay There was a test for a `nil` reason but not a blank one, so I've added that. And I didn't know about `present?`, thank you!

spec/models/proposal_spec.rb
```diff
@@ -314,13 +314,24 @@
         end
       end
 
+      context "with a blank reason" do
+        let(:reason) { ' ' }
```
jessieay: what about just assigning `reason` within the `it` block!? (I see that would be inconsistent with the rest of this file...so not necessary, but certainly an organizational pattern I think we should be moving toward!!

app/models/proposal.rb
```diff
@@ -268,7 +268,7 @@ def create_new_observation(user, adder, reason)
     self.observers(true)
     # when explicitly adding an observer using the form in the Proposal page...
     if adder
-      if reason
+      if reason and reason.present?
```
jessieay: I think we're mostly using `&&` rather than `and` in this project!

spec/models/proposal_spec.rb
```diff
@@ -314,13 +314,24 @@
...
+
+        it 'does not add a comment' do
+          expect(proposal.comments).to be_empty
+          proposal.add_observer(observer_email, user, reason)
+          expect(proposal.comments).to be_empty
```
jessieay: niiiice

spec/models/proposal_spec.rb
```diff
@@ -314,13 +314,24 @@
         end
       end
 
+      context "with a blank reason" do
+        let(:reason) { ' ' }
+
+        it 'does not add a comment' do
```
jessieay: double quotes! We are enforcing this starting today :)

yozlet:
@jessieay Fixed!

jessieay:
I think you will fail code climate because we are now enforcing double quotes...but otherwise code looks :ship:  !

pkarman:
@yozlet I think codeclimate ignores spec files and will never finish on this PR.
Do you have some advice for how to QA this?

yozlet:
@pkarman Ah, that explains those non-finishing CC runs, thanks.

To QA, check that
* when adding an observer to a proposal with a reason
   * A formatted comment is added, with most of the comment italicised except the reason, which is in `monospace`
* when adding an observer to a proposal with a blank reason
   * no comment is added

pkarman:
Looks great.

PR #715, opened 2015-10-19
minor refactor to named method in WorkOrdersController

afeld:
Per suggestion in https://github.com/18F/C2/pull/656/files#r42406827.

/cc @jessieay

### DIFF
--- a/app/controllers/ncr/work_orders_controller.rb
+++ b/app/controllers/ncr/work_orders_controller.rb
@@ -29,7 +29,7 @@ module Ncr
 
       super
 
-      unless skip_state_update?
+      if @model_changing
         @model_instance.setup_approvals_and_observers(@approver_email)
         @model_instance.email_approvers
       end
@@ -73,10 +73,6 @@ module Ncr
       if self.errors.empty?
         @model_instance.setup_approvals_and_observers(@approver_email)
       end
-    end
-
-    def skip_state_update?
-      !@model_changing || @model_instance.emergency
     end
   end
 end
--- a/app/controllers/ncr/work_orders_controller.rb
+++ b/app/controllers/ncr/work_orders_controller.rb
@@ -29,7 +29,7 @@ module Ncr
 
       super
 
-      if @model_changing && !@model_instance.emergency # skip approvals if emergency
+      unless skip_state_update?
         @model_instance.setup_approvals_and_observers(@approver_email)
         @model_instance.email_approvers
       end
@@ -73,6 +73,10 @@ module Ncr
       if self.errors.empty?
         @model_instance.setup_approvals_and_observers(@approver_email)
       end
+    end
+
+    def skip_state_update?
+      !@model_changing || @model_instance.emergency
     end
   end
 end


app/controllers/ncr/work_orders_controller.rb
```diff
@@ -74,5 +74,9 @@ def add_approvals
         @model_instance.setup_approvals_and_observers(@approver_email)
       end
     end
+
+    def skip_state_update?
```
afeld: Very open to suggestions about what to call this – don't love it.

app/controllers/ncr/work_orders_controller.rb
```diff
@@ -29,7 +29,7 @@ def update
 
       super
 
-      if @model_changing && !@model_instance.emergency # skip approvals if emergency
+      unless skip_state_update?
```
jessieay: still not a fan of unless! can we change this to an `if` ?

jessieay: how about: `if model_changing_and_not_emergency` ? It's long and blah, but it does explain what's happening!

afeld: Ha. Well, that does describe (literally) what the method is doing, but I was going for a name that would convey why those things matter. Does that make sense?

jessieay: I guess I am not 100% sure why they matter, which is probably why this is hard for me. Perhaps you can explain? I looked into the controller to see the important for `model_changing` but I am still not clear

pkarman: This feels like one of those cases where some apt comments would explain the "why" and the method can just be the "what".

jessieay: Why can't the method name explain the why!?

pkarman: I'm really on +0 on `if model_changing_and_not_emergency?`

jessieay: @pkarman suggestions on something you would be +1 on would be helpful!

pkarman: how about
```
if subscribers_require_notification
```

jessieay: :heart:

jessieay: although I might add a `?` to the end of that to indicate its boolean-ness

afeld:
Updated!

jessieay:
this looks good! 

should we be concerned that this change required no spec changes? I would expect that L146 in spec/controllers/ncr/work_orders_controller_spec.rb would fail now because it deals with updates  and emergency status.

afeld:
> should we be concerned that this change required no spec changes?

No. `setup_approvals_and_observers` checks for `emergency`, and `email_approvers` does nothing, since emergency requests have none. Now this PR is just taking out a redundant check.

jessieay:
Got it, then looks good to :ship: 

How would someone QA this change? Or does it not need QAing?

afeld:
## Testing notes

* [ ] Update an emergency request, without changing the approver
* [ ] Update an emergency request, changing the approver

pkarman:
I will QA this.

pkarman:
Looks fine on staging.

PR #716, opened 2015-10-19
move NCR feature specs under single folder

afeld:
Makes the names less redundant, and makes it easier to run them all at once (`bin/rspec spec/features/ncr`).


jessieay:
so much better! Will wait for checks to pass but :ship: from me

jessieay:
mergin'

PR #717, opened 2015-10-19
Updates to Team Practices doc

jessieay:
* `script/deploy` failed for me, perhaps because of blue-green
* Ideally, we should indicate to others when we are QAing something


pkarman:
looks great.

PR #719, opened 2015-10-20
ruboconf directive for double_quotes fails on CodeClimate

pkarman:
CodeClimate appears unhappy with this directive. Maybe a version issue?
```
undefined method `each_key' for "EnforcedStyle:double_quotes":String
```

FWIW, Rubocop does not require double quotes by default, on purpose: https://github.com/bbatsov/rubocop/issues/1615

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -24,6 +24,8 @@ Style/RedundantSelf:
   Enabled: false
 Style/SignalException:
   Enabled: false
+Style/StringLiterals:
+  EnforcedStyle: double_quotes
 Style/TrailingComma:
   Enabled: false
 Style/AndOr:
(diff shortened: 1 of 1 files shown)


jessieay:
weird, still failing. Want me to take over and debug this since I am the originator of the error?

Perhaps we also need to add a rule for string literal in interpolation: https://github.com/thoughtbot/hound/blob/master/.ruby-style.yml#L352

pkarman:
yes please @jessieay

jessieay:
@pkarman just removed the config altogether (https://github.com/18F/C2/commit/052e1436d65d0d2e8be0fb28296a1bcea4ab7e48) and it is still failing - I think what's happening is that code climate is still running the code based on the config that's one master.

I think your original guess that it was a spacing issue is probably right. Going to revert back to that and perhaps we can try merging to see if having this new config on master works?

pkarman:
+1 to merge fix to master. I'm not sure that CC would be using that conf, but it shouldn't hurt anything.

jessieay:
@pkarman should be ready to merge

PR #720, opened 2015-10-20
[Delivers #106110788] manually trigger email for new comment

pkarman:
Story: https://www.pivotaltracker.com/story/show/106110788

Email was not being triggered for new comments.

### DIFF
--- a/lib/incoming_mail/handler.rb
+++ b/lib/incoming_mail/handler.rb
@@ -63,14 +63,17 @@ module IncomingMail
       proposal = find_proposal(find_public_id(msg)) or return
       comment_text = find_comment_text(msg)
       comment_user = find_comment_user(msg)
+      comment = nil
       if proposal.has_subscriber?(comment_user)
         # already in the loop, just add comment.
-        proposal.comments.create(comment_text: comment_text, user: comment_user)
+        comment = proposal.comments.create(comment_text: comment_text, user: comment_user)
       else
         # yes, user adds self as observer, which also generates comment
         proposal.add_observer(comment_user, comment_user, comment_text)
-        proposal.comments.last
+        comment = proposal.comments.last
       end
+      Dispatcher.on_comment_created(comment) # sends email
+      comment
     end
 
     def find_proposal(public_id)
--- a/spec/features/incoming_mail_handler_spec.rb
+++ b/spec/features/incoming_mail_handler_spec.rb
@@ -39,6 +39,7 @@ describe "Handles incoming email" do
     expect(resp.action).to eq(IncomingMail::Response::COMMENT)
     expect(my_approval.proposal.existing_observation_for(my_approval.user)).to be_truthy
     expect(my_approval.proposal.existing_approval_for(my_approval.user)).to be_truthy
+    expect(deliveries.length).to eq(2) # 1 each to requester and approver
   end
 
   it "should create comment for non-subscriber and add as observer" do
@@ -53,6 +54,7 @@ describe "Handles incoming email" do
     expect(resp.action).to eq(IncomingMail::Response::COMMENT)
     expect(resp.comment.proposal.existing_observation_for(user)).to be_truthy
     expect(resp.comment.proposal.existing_approval_for(user)).to be_falsey
+    expect(deliveries.length).to eq(4) # 2 each to requester and new observer
   end
 
   it "should parse proposal public_id from email headers" do
--- a/spec/features/incoming_mail_handler_spec.rb
+++ b/spec/features/incoming_mail_handler_spec.rb
@@ -44,6 +44,7 @@ describe "Handles incoming email" do
 
   it "should create comment for non-subscriber and add as observer" do
     my_approval = approval
+    my_proposal = my_approval.proposal
     user = create(:user)
     mandrill_event = mandrill_payload_from_message(mail)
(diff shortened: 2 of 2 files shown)


lib/incoming_mail/handler.rb
```diff
@@ -63,14 +63,17 @@ def create_comment(msg)
...
         # already in the loop, just add comment.
-        proposal.comments.create(comment_text: comment_text, user: comment_user)
+        comment = proposal.comments.create(comment_text: comment_text, user: comment_user)
       else
         # yes, user adds self as observer, which also generates comment
         proposal.add_observer(comment_user, comment_user, comment_text)
-        proposal.comments.last
+        comment = proposal.comments.last
       end
+      Dispatcher.on_comment_created(comment) # sends email
```
jessieay: ah yes! missed this one when I replaced the callback. Glad you caught it

if there is no proposal subscriber, do we care about emailing out about a comment being created? (eg: put L75 here into the `if` statement above)

spec/features/incoming_mail_handler_spec.rb
```diff
@@ -53,6 +54,7 @@
     expect(resp.action).to eq(IncomingMail::Response::COMMENT)
     expect(resp.comment.proposal.existing_observation_for(user)).to be_truthy
     expect(resp.comment.proposal.existing_approval_for(user)).to be_falsey
+    expect(deliveries.length).to eq(4) # 2 each to requester and new observer
```
jessieay: what are these 2 emails? would there me a way to write this expectation that would make it clearer why each person is receiving two emails? (ideally without a comment :) )

jessieay:
needs to pull in changes from master for CC to pass. will let you do it!

pkarman:
@jessieay assigning this to you. Has been updated from master but unclear to me whether CC will ever finish.

lib/incoming_mail/handler.rb
pkarman: I believe the Dispatcher logic will eliminate anyone who doesn't need to get the email notification, so I believe we are safe delegating that decision elsewhere. In any case, I don't think it should be up to this code to decide who gets notified, when.

jessieay:
@pkarman when CC hangs, you can manually re-run via the refresh button

screen shot 2015-10-20 at 1 12 29 pm

jessieay:
@pkarman so refreshing didn't work...you have ideas on why code climate is hanging?

afeld:
http://status.codeclimate.com/incidents/yghjqx6rh0pl

I wouldn't wait for it.

jessieay:
Thanks for the link, @afeld ! assumed this had something to do with our rubocop update from this morning. will move forward without worrying about CC

spec/features/incoming_mail_handler_spec.rb
```diff
@@ -53,6 +55,12 @@
     expect(resp.action).to eq(IncomingMail::Response::COMMENT)
     expect(resp.comment.proposal.existing_observation_for(user)).to be_truthy
     expect(resp.comment.proposal.existing_approval_for(user)).to be_falsey
+    expect(my_approval.user).to_not eq(my_proposal.individual_approvals.last.user)
+    expect(deliveries.length).to eq(4)
+    expect(deliveries.select{|m| m.to.first == user.email_address}.length).to eq(1)
```
jessieay: nice update! In the future I might wrap something like this so it reads like:

`expect_one_email_sent_to(user.email_address)`


```ruby
def expect_one_email_sent_to(email_address)
  expect(deliveries.select { |email| email.to.first == email_address }.length).to eq (1)
end
```

lib/incoming_mail/handler.rb
jessieay: I think I understand the source of my confusion - this `create_comment` method is doing quite a bit more than creating a comment. It finds a proposal, finds comment text, finds comment user, subscribes the user if they are not already a subscriber, and emails out about the comment. 

It feels like we might want to have a little service class for this. Having a single private method handle all of that means it can be hard (for me, at least!) to understand what's going on and hard to have good unit tests, which would ensure that the behavior doesn't change without us realizing it.

Do you want me to take a stab at a service class for this? I am happy to do it. Also happy for you to do it, whatever makes sense.

pkarman: If you want to optimize the code layout further, I have no objection. I would ask, though, that you make that a story in itself so we can move forward with this particular PR. Production is broken till this gets merged and deployed.

jessieay: I would say changing this is not just an optimization; the way it is now is confusing because of the amount of logic in that one method. Not pointing the finger at you or anyone on this -- but since reading and understanding this method was part of reviewing this pull request, my feedback is that it is difficult to parse as-is. 

That being said, I hear you on the "let's get this in because it is important" front. I didn't realize that this was fixing something that is broken on production. Maybe we should create a github tag to indicate when there is urgency of that nature in a PR?

:ship:  code!

jessieay:
:sheep: it!

jessieay:
I will QA this now

jessieay:
Having trouble w QA. Comments on the nature of my trouble are in the tracker story

jessieay:
confirmed at https://c2-dev.18f.gov/proposals/203 -- merging!

PR #722, opened 2015-10-20
[HOLD][Delivers #98295520] Add purchase step to 18F procurements

yozlet:
@annalee and I are doing this by generalising `Approval` to include other kinds of steps. As part of this, renaming `Approval` to `Step`.

 - [x] Rename `Approval` -> `Step`
 - [x] Rename `Approvals::*` -> `Steps::*`
 - [x] Rename `steps`/`approvals_user_proposal_idx` in schema
 - [x] Check that API output change is OK (`spec/requests/api/ncr/work_orders_spec.rb:99`) - if so, update API documentation
 - [x] Migration to change `Approvals::Serial` to `Steps::Serial` etc. in the `type` column of the (new) `steps` table (thanks, Aidan)
 - [x] If we're subclassing `Steps::Individual` to `Approval` and `Purchase`, then the above migration also needs to handle that renaming
 - [x] The actual feature addition we came here for (see below)

Not doing:
 - _Rename `Approval` factory and all usages_ - keeping this as `:approval`, but now it creates a `Steps::Approval`, not a `Steps::Individual`
 - _Rename `ApprovalDelegate` -> `StepDelegate`_ - Do this in a later PR

**User-facing changes:**
- In 18F procurements: Create a new Purchase step after initial Approval step
- Change button labeling for Purchase Step from "Approve" to "Mark As Purchased"


afeld:
* [ ] Migration to change `Approvals::Parallel` to `Steps::Parallel` etc. in the `type` column of the (new) `steps` table

afeld:
Also, API change is no big deal, since no one is using it at the moment.

app/controllers/concerns/token_auth.rb
```diff
@@ -34,7 +34,7 @@ def validate_access
       sign_in(token.user)
     end
     # expire tokens regardless of how user logged in
-    tokens = ApiToken.joins(:approval).where(approvals: {
+    tokens = ApiToken.joins(:step).where(steps: {
       user_id: current_user, proposal_id: self.proposal})
```
jessieay: I *think* you could just do `where( steps: { user: current_user, proposal: proposal })`

app/decorators/proposal_decorator.rb
```diff
@@ -40,7 +40,7 @@ def display_status
   end
 
   def generate_status_message
-    if object.approvals.non_pending.empty?
+    if object.steps.non_pending.empty?
```
jessieay: would it make sense to add a method on proposal call `approvals` that returns approval steps, so we could still cal `object.approvals.non_pending.empty?` here?

app/models/proposal.rb
```diff
@@ -62,8 +62,8 @@ class Proposal < ActiveRecord::Base
   after_create :update_public_id
 
   # @todo - this should probably be the only entry into the approval system
-  def root_approval
-    self.approvals.where(parent: nil).first
+  def root_step
```
jessieay: `root_approval_step` ?

jessieay:
Overall comment: I know this is an incremental step toward having "steps" which can either be approvals or purchases or other things (?). 

But since many of the methods we currently have in here are specific to approvals, does it make sense to keep the method names using "approval" in them even though the underlying ActiveRecord class is a `Step`?

That way, when different types of steps are added, you can change the logic within the `approvals` methods to select steps of the approval type rather than having to rename all of the `step` methods back to something more specifically approvall-y. 

Does that make sense? Def let me know if you want clarification on my thinking here.

yozlet:
@jessieay That's an excellent idea, though I'm not sure how many of those methods will _only_ involve Approvals versus other kinds of Steps. We're taking the renaming fairly slowly at the moment but it's a good effort-saver to keep in mind, thank you!

yozlet:
OK, @annalee and I think this is ready for review. We've shunted certain tasks off for doing later (if doing at all), mainly renaming of internals. There's also [a whole other story](https://www.pivotaltracker.com/story/show/106244512) around doing more purchase-step-related relabelling, and then [some UX additions to the dashboard](https://www.pivotaltracker.com/n/projects/1149728/stories/106895976).

**Main open question we'd like comment on:** As of this PR, nothing should be creating `Steps::Individual` records; they should all be `Steps::Approval` or `Steps::Purchase` (whether created by NCR or 18F clients). Should we make `Steps::Individual` an abstract class to prevent it being instantiated directly?

pkarman:
@yozlet @annalee alas, branch in conflict.

db/schema.rb
```diff
@@ -175,6 +160,21 @@
...
   add_index "roles", ["name"], name: "roles_name_idx", unique: true, using: :btree
 
+  create_table "steps", force: :cascade do |t|
+    t.integer  "user_id"
+    t.string   "status",              limit: 255
+    t.datetime "created_at"
+    t.datetime "updated_at"
+    t.integer  "position"
+    t.integer  "proposal_id"
+    t.datetime "approved_at"
```
pkarman: does it make sense to rename this column to `stepped_at` to align the naming logic?

db/schema.rb
```diff
@@ -175,6 +160,21 @@
 
   add_index "roles", ["name"], name: "roles_name_idx", unique: true, using: :btree
 
+  create_table "steps", force: :cascade do |t|
+    t.integer  "user_id"
```
pkarman: since this table is now not approval-specific, is this the time to also refactor it to use `proposal_roles`? I.e. how do we tell by looking at a `step` what flavor it is?

pkarman:
@yozlet @annalee wrote:
> Should we make Steps::Individual an abstract class to prevent it being instantiated directly?

If it shouldn't be instantiated directly, then I would say yes. Same principle as making methods private.

db/schema.rb
yozlet: There are still a load of methods and things using `Approval`/`approver` language in this PR; we're not changing it all in one go. And we still need to work out what all the new terminology should be. `completed_at`? `performed_at`?

db/schema.rb
yozlet: You look at the `type`, because `steps` is (still) using Single Table Inheritance. Individual steps will either be a `Steps::Approval` or `Steps::Purchase`.

pkarman:
@yozlet @annalee is this ready to merge? Code looks fine to me. Wasn't sure if you were still waiting on guidance for your abstract class question.

yozlet:
Not quite ready - we're doing the abstract class thing, and it's revealed some related issues that we're currently fixing. I've taken the _Review Needed_ label off and will restore it when we're ready. Thank you!

yozlet:
OK, abstract class thing _done_, ready for final review and merge.

pkarman:
Looks like https://github.com/18F/C2/pull/722#issuecomment-149728688 is complete so I checked that box.

The other changes look fine to me. Since this is such a big change, I will leave this open for other reviewers. Here's my +1 to :sheep: it.

app/models/step.rb
```diff
@@ -12,18 +12,18 @@ class Approval < ActiveRecord::Base
...
 
-  belongs_to :parent, class_name: 'Approval'
-  has_many :child_approvals, class_name: 'Approval', foreign_key: 'parent_id', dependent: :destroy
-
-  scope :individual, -> { where(type: 'Approvals::Individual') }
+  belongs_to :parent, class_name: "Step"
+  has_many :child_approvals, class_name: "Step", foreign_key: "parent_id", dependent: :destroy
 
+  # @TODO: Auto-generate list of subclasses
+  scope :individual, -> { where(type: ["Steps::Approval", "Steps::Purchase"]).order("position ASC") }
```
jessieay: I would add `order` as a separate scope. Could be overlooked when it is part of this scope. Either that, or rename to be something like `individial_in_position_order`

app/models/steps/approval.rb
```diff
@@ -0,0 +1,4 @@
+module Steps
```
jessieay: was wondering if this should be plural or singular...seems like plural is how Rails does it (eg: http://api.rubyonrails.org/classes/ActiveRecord.html)...good lesson for me, I default to singular! so :+1: on this naming

app/models/steps/parallel.rb
```diff
@@ -37,7 +37,7 @@ class Parallel < Approval
       end
     end
 
-    def on_actionable_entry(old_state, event)
+    def on_actionable_entry(_, _)
```
jessieay: two unused args -- intentional?

app/serializers/step_serializer.rb
```diff
@@ -1,4 +1,4 @@
-class ApprovalSerializer < ActiveModel::Serializer
+class StepSerializer < ActiveModel::Serializer
   # make sure to keep docs/api.md up-to-date
```
jessieay: did we follow the instructions here and update docs?

db/migrate/20151028205249_rename_approval_types_to_step_types.rb
```diff
@@ -0,0 +1,17 @@
+class RenameApprovalTypesToStepTypes < ActiveRecord::Migration
+  def up
+    execute <<-SQL
```
jessieay: nice!

app/serializers/step_serializer.rb
jessieay: read below and saw that you did - niiice

spec/factories/approvals.rb
```diff
@@ -1,10 +1,10 @@
 FactoryGirl.define do
```
jessieay: should we move this factory since we now have the exact same thing at `steps.rb` ?

spec/models/proposal_spec.rb
```diff
@@ -34,10 +34,10 @@
     it "gives only the first approver when linear" do
       proposal = create(:proposal, :with_serial_approvers)
       approver1, approver2 = proposal.approvers
-      expect(proposal.currently_awaiting_approvers).to eq([approver1])
+      expect(proposal.currently_awaiting_approvers).to contain_exactly(approver1)
 
       proposal.individual_approvals.first.approve!
-      expect(proposal.currently_awaiting_approvers).to eq([approver2])
+      expect(proposal.currently_awaiting_approvers).to contain_exactly(approver2)
```
jessieay: interesting - what are benefits of `contain_exactly` vs `eq` ?

app/models/gsa18f/procurement.rb
```diff
@@ -26,14 +26,14 @@ def self.purchase_amount_column_name
...
 
-    after_create :add_approvals, :add_observers
-
-    def add_approvals
-      self.proposal.approver = User.for_email(Gsa18f::Procurement.approver_email)
-    end
-
-    def add_observers
-      self.add_observer(Gsa18f::Procurement.purchaser_email)
+    after_create :add_steps
```
jessieay: this doesn't have to happen now, but do you have thoughts on a way we could do this without a callback? I've been working to remove as many callbacks as possible!

jessieay:
Left a few comments, but overall looks :sheep: to me as well!

app/models/step.rb
yozlet: Interesting - how would it work as a separate scope? Would I be able to combine the two?

app/models/steps/parallel.rb
yozlet: Don't know, I'm just cleaning up previous code. As far as I know it's still unused.

spec/factories/approvals.rb
yozlet: Good question! At present _nothing_ uses the new `Step` factory, everything's still using `Approval`. I'm fine with that for the moment, since you can't just create a `Step` without specifying what kind of step it is. In the future we should probably rearrange things so that you always create a `Step` with an associated trait, but we don't have to do that right now.

yozlet: Bearing that in mind, I'm going to delete `spec/factories/steps.rb` for the moment.

spec/models/proposal_spec.rb
yozlet: I switched this in case it won't equate an array with an ActiveRecord collection (which I should probably check), and it's more about the contents of the array/collection.

yozlet: OK, just checked, `eq` is fine. So I'm going to switch it back.

spec/factories/approvals.rb
yozlet: Oops, looks like we need it to make some associations in other factories work easier.

app/models/gsa18f/procurement.rb
yozlet: Callback removed! Step addition is now triggered by the controller.
I needed to add an `after(:create)` in the `Procurement` factory, but that's OK, right?

app/models/step.rb
jessieay: you can chain scopes!

app/models/gsa18f/procurement.rb
jessieay: you know, in that case I would keep as a callback for now. If we have the factories doing things that our model doesn't do, that might get us into a situation where our tests aren't reflecting reality the way they should be.

yozlet:
Basic test plan:

* This PR contains migrations that alter existing record data. In particular the `approvals` table is renamed to `steps`, and the values in the `type` field are switched from `Approval::*` to `Steps::*`. Also - and this is especially important - all `Approvals::Individual` become `Steps::Approval`
* The two user-visible changes are:
   1. Creating an 18F Procurement adds _two_ steps - one for the approver, one for the purchaser
   2. Because the second step is actually a Purchase step the big **Approve** button at the bottom of the Proposal detail page should instead read **Mark For Purchase** - but only for that _second_ step.

pkarman:
Tested migrations against a recent (2015-10-19) production snapshot in local dev env and all was fine.

Will now deploy to c2-dev and run through test plan.

pkarman:
Test on c2-dev looks good.
purchase-step

pkarman:
HOLD on merge until ready to deploy.

PR #724, opened 2015-10-21
fix email send regression

pkarman:
refactor test for clarity, fix regression where replying via email did not generate email to subscribers

### DIFF
--- a/lib/incoming_mail/handler.rb
+++ b/lib/incoming_mail/handler.rb
@@ -63,17 +63,23 @@ module IncomingMail
       proposal = find_proposal(find_public_id(msg)) or return
       comment_text = find_comment_text(msg)
       comment_user = find_comment_user(msg)
-      comment = nil
-      if proposal.has_subscriber?(comment_user)
-        # already in the loop, just add comment.
-        comment = proposal.comments.create(comment_text: comment_text, user: comment_user)
-      else
-        # yes, user adds self as observer, which also generates comment
-        proposal.add_observer(comment_user, comment_user, comment_text)
-        comment = proposal.comments.last
+      # IMPORTANT that we check/add as observer before we create comment,
+      # since comment will create as a user if not already,
+      # and we want the reason logged.
+      unless proposal.has_subscriber?(comment_user)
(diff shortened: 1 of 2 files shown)


lib/incoming_mail/handler.rb
```diff
@@ -63,19 +63,25 @@ def create_comment(msg)
...
       comment
     end
 
+    def add_user_as_observer(user, proposal)
+      observer_role = Role.find_or_create_by(name: 'observer')
+      observation = Observation.new(user_id: user.id, role_id: observer_role.id, proposal_id: proposal.id)
+      proposal.observations << observation
+      Dispatcher.on_observer_added(observation, "Added comment via email reply")
+      observation
+    end
```
yozlet: Doesn't this mostly replicate `Proposal.create_new_observation` (and the checking in `Proposal.add_observer`)?

jessieay:
@yozlet I also observed the similar logic and started a refactor here: https://github.com/18F/C2/pull/726

I wanted to throw it up as an idea before going through and writing unit tests / improving further. I also know  that this PR is needed to get prod green so I would be cool with a :ship: for now and then I can continue work on #726 tomorrow. Let us know how you'd like to proceed, Peter!

lib/incoming_mail/handler.rb
pkarman: It does, mostly. The key difference is that it does not create a comment with the `update_comment` flag set, which has the effect of making `Dispatcher.on_observer_added` a no-op (no email is sent).

I thought of refactoring `Proposal.create_new_observation` and/or adding a new method to `Proposal` but that felt pretty intrusive and I wasn't convinced would actually end up saving any lines of code.

I'm happy to see this refactored eventually. In the meantime, I am going to merge and deploy since it fixes the immediate problem.

pkarman: I was wrong. During my refactor, I went from creating 1 comment to creating 2 comments: 1 for the email reply, and one `update_comment` comment for the addition of the observer, instead of a single combined comment.

So yes, `Proposal.add_observer` is appropriate to re-use here.

PR #725, opened 2015-10-21
[Delivers #99273060] Send cancellation conf email to active observers

jessieay:
* Active observers are observer who are not approver or requester
* Previously, were not receiving an email
* Story: https://www.pivotaltracker.com/story/show/99273060

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -11,6 +11,7 @@ class User < ActiveRecord::Base
 
   has_many :approvals
   has_many :observations
+  has_many :observers, through: :observers, source: :user
   has_many :comments
 
   # we do not use rolify gem (e.g.) but declare relationship like any other.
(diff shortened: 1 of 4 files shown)


pkarman:
I will QA

pkarman:
I left QA comments on the PT story. Looks great.

app/models/user.rb
```diff
@@ -11,6 +11,7 @@ class User < ActiveRecord::Base
 
   has_many :approvals
   has_many :observations
+  has_many :observers, through: :observers, source: :user
   has_many :comments
```
pkarman: why this recursive relationship definition?

pkarman:
@jessieay - just one review comment, and then I'm ready to :sheep: it

app/models/user.rb
jessieay: oops! meant for this to be `through: :observations`...wonder how that was even working!?

jessieay:
@pkarman fixed!

PR #726, opened 2015-10-21
[spike] Refactor incoming mail handler

jessieay:


lib/incoming_mail/handler.rb
```diff
@@ -60,47 +60,24 @@ def forward_msg(msg)
...
+      comment = CommentCreator.new(parsed_email).run
+      Dispatcher.on_comment_created(comment) # sends email
+      comment
     end
 
-    def find_comment_user(msg)
-      from = msg['from_email']
-      User.find_by_email_address(from)
+    def add_user_as_observer(user, proposal)
+      proposal.add_observer(user, user, "Added comment via email reply")
```
jessieay: @pkarman does it make sense to add the user in as the "adder" here if they are being added as an observer via email reply...or should we create a different method in `Proposal` that adds an observer without an adder being present 

(right now, an observer added email is *not* sent if there is not an added, see: L271 of Proposal.rb)

lib/services/inbound_mail_parser.rb
```diff
@@ -0,0 +1,48 @@
+class InboundMailParser
```
jessieay: do we think `lib/services` is a good place to put this?

config/application.rb
```diff
@@ -38,6 +38,7 @@ class Application < Rails::Application
     config.autoload_paths << Rails.root.join('app', 'mailers', 'concerns')
     config.autoload_paths << Rails.root.join('app', 'policies', 'concerns')
     config.autoload_paths << Rails.root.join('lib')
+    config.autoload_paths << Rails.root.join("lib", "services")
```
jessieay: or should we just autoload everything in `lib` ?

lib/incoming_mail/handler.rb
```diff
@@ -60,47 +60,24 @@ def forward_msg(msg)
...
+      proposal = parsed_email.proposal
+      comment_user = parsed_email.comment_user
 
-      if references.match(ref_re)
-        references.match(ref_re)[1]
-      elsif msg['subject'].match(sbj_re)
-        msg['subject'].match(sbj_re)[1]
-      else
-        fail "Failed to find public_id in msg #{msg.inspect}"
+      unless proposal.has_subscriber?(comment_user)
```
jessieay: wonder if this logic should be added to the `add_observer` method within `Proposal`

lib/services/comment_creator.rb
```diff
@@ -0,0 +1,17 @@
+class CommentCreator
+  def initialize(parsed_email)
+    @parsed_email = parsed_email
+  end
+
+  def run
```
jessieay: on the one hand I like how little this is doing. On the other hand, `inbound_mail_parser` is doing most of the work here so maybe it's not as helpful as I originally thought

lib/incoming_mail/handler.rb
pkarman: My thinking was that the commenter (user) was adding themselves by virtue of creating a comment. In practice I don't expect this to happen often, since someone would have to respond to an email related to a request for which they were not already a subscriber. A chicken and egg dilemma.

lib/incoming_mail/handler.rb
pkarman: Probably. IIRC there is a unique constraint on the `proposal_roles` table that should prevent a duplicate observation from being created, but additional checks in code would allow us to present a more helpful error message (if necessary).

pkarman:
As I read this PR, it basically does these things:

* adds a new `Parser` class to encapsulate the relevant methods from `Handler`
* adds a new namespace in `lib/services` (presumably following the advice from https://blog.engineyard.com/2014/keeping-your-rails-controllers-dry-with-services)

If you think that makes the code easier to understand and maintain, then +1.

jessieay:
Hey @pkarman thanks for the review!

I think the question is, do *you* think this makes the code easier to understand and maintain? I certainly think that as the author of the PR, but the others' perspectives are very important in validating my hypothesis.

jessieay:
@pkarman also, going to write some unit tests for these service classes now so let's hold off on merging until then!

pkarman:
As the original author, I find the code equally clear either way. :smile: So I'm +1 to this change if it helps others.

jessieay:
@pkarman following up on https://github.com/18F/C2/pull/724#discussion_r42752732

I am re-using the method in `Proposal` for creating a new observation (per convo in Slack earlier). You mentioned in the above-linked comment that we should go back to creating a single comment rather than 1 for the email reply and 1 update comment about the observer being added.

What should we do about email? 

Right now, handler.rb calls `Dispatcher.on_comment_created(comment)` and the oberver method in proposal.rb calls `Dispatcher.on_observer_added(observation, reason)`

Do we want to create an email notification for this comment or mover over to an email notification for an observer being added? I am not sure if it matters but wanted to check ...

pkarman:
@jessieay I don't think we need to go back to 1 comment. 2 is fine. I was just trying to explain why I was wrong. I think the way you've got it now
```
proposal.add_observer(user, user, "Added comment via email reply")
```
is perfectly fine.

jessieay:
@pkarman thanks for the update. What about the second part of my last message re: email?

pkarman:
I don't think we want the commenter to get 2 emails (one for the comment, one for the notification). So doing it as it is now, the commented only gets one email, which is imo good.

jessieay:
closing this for now while I complete it

PR #727, opened 2015-10-22
Remove extra dash from beginning link item

mbland:
The extra dash made it look like two items, one without anchor text, and one without a URL.

cc: @afeld @pkarman @openbrian


pkarman:
thanks @mbland

PR #728, opened 2015-10-23
Add service classes to encapsulate mail handler logic

jessieay:
* Branch off of https://github.com/18F/C2/pull/724
* Also add service class to capture duplicate logic in `Proposal`

### DIFF
--- a/lib/incoming_mail/handler.rb
+++ b/lib/incoming_mail/handler.rb
@@ -60,53 +60,31 @@ module IncomingMail
     end
 
     def create_comment(msg)
-      proposal = find_proposal(find_public_id(msg)) or return
-      comment_text = find_comment_text(msg)
-      comment_user = find_comment_user(msg)
       # IMPORTANT that we check/add as observer before we create comment,
       # since comment will create as a user if not already,
       # and we want the reason logged.
-      unless proposal.has_subscriber?(comment_user)
-        add_user_as_observer(comment_user, proposal)
+      parsed_email = InboundMailParser.new(msg)
+      proposal = parsed_email.proposal
+      user = parsed_email.comment_user
+
+
+      unless proposal.existing_observation_for(user)
+        reason = "Added comment via email reply"
+        ObservationCreator.new(
+          observer: user,
+          proposal_id: proposal.id,
+          reason: reason,
+        ).run
       end
-      comment = proposal.comments.create(comment_text: comment_text, user: comment_user)
+
+      comment = Comment.create(
+        comment_text: parsed_email.comment_text,
+        user: parsed_email.comment_user,
+        proposal: parsed_email.proposal
+      )
+
       Dispatcher.on_comment_created(comment) # sends email
       comment
-    end
-
-    def add_user_as_observer(user, proposal)
-      observer_role = Role.find_or_create_by(name: 'observer')
-      observation = Observation.new(user_id: user.id, role_id: observer_role.id, proposal_id: proposal.id)
-      proposal.observations << observation
-      Dispatcher.on_observer_added(observation, "Added comment via email reply")
-      observation
-    end
-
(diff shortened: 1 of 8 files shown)


lib/services/inbound_mail_parser.rb
```diff
@@ -0,0 +1,48 @@
+class InboundMailParser
```
jessieay: this service class is covered by feature specs for `Handler.rb` - would be great to have unit tests too and then just stub this when testing `Handler.rb` but this PR felt like it was already getting pretty big

pkarman:
@afeld I did the review on the previous incarnation of this PR and it might benefit from your eyes this time.

spec/features/incoming_mail_handler_spec.rb
```diff
@@ -49,21 +50,25 @@
...
     resp = handler.handle(mandrill_event)
+
     expect(resp.action).to eq(IncomingMail::Response::COMMENT)
-    expect(resp.comment.proposal.existing_observation_for(user)).to be_truthy
-    expect(resp.comment.proposal.existing_approval_for(user)).to be_falsey
+    expect(resp.comment.proposal.existing_observation_for(user)).to be_present
+    expect(resp.comment.proposal.existing_approval_for(user)).not_to be_present
     expect(my_approval.user).to_not eq(my_proposal.individual_approvals.last.user)
-    recipients = [
-      user.email_address, # observer notification
```
jessieay: this was removed intentionally - we are sending a comment email to all of the other people watching the proposal. We are no longer sending an "observation notification" email because the way the logic is set up, an observer added email is only sent when there is an "adder" . But, this logic can be changed. Thoughts?

pkarman:
@jessieay this PR has merge conflicts now.

jessieay:
@pkarman updated :heavy_check_mark:

spec/features/incoming_mail_handler_spec.rb
pkarman: This feels like a UX issue to me. If, as a non-observer, I create a comment by replying to an email (admittedly an edge case), then as it is now, I will get zero emails confirming that I made the comment. Maybe that's ok (it is what GitHub does, e.g.). I don't know. I just wanted to point out that the current behavior (multiple emails) will change to zero emails.

pkarman:
The code changes look fine to me, as discussed on #726 .

I did note one UX issue, which I think is ok to punt on for now since it represents an edge case.

I will QA.

pkarman:
tested on c2-dev. I found some edge case regressions that were introduced in the refactor, for which I have patches and will write tests. Coming soon.

pkarman:
@jessieay this PR is all green and looks fine to :sheep: it. If you are ok with my recent changes, I will merge.

lib/incoming_mail/handler.rb
```diff
@@ -60,53 +65,32 @@ def forward_msg(msg)
...
-      observation = Observation.new(user_id: user.id, role_id: observer_role.id, proposal_id: proposal.id)
-      proposal.observations << observation
-      Dispatcher.on_observer_added(observation, "Added comment via email reply")
-      observation
-    end
-
-    def find_proposal(public_id)
-      Proposal.find_by_public_id(public_id) || Proposal.find(public_id)
-    end
+      return unless user  # cannot create comment for non-existent user
```
jessieay: can we add this to the statement on L 77 instead? Will make it easier to understand, I think.

pkarman: My line numbers might not match up with your comment. Feel free to refactor as you see fit. The important thing is to avoid all the comment/observer creation if we do not have a valid User in play.

jessieay:
Let's go ahead and merge this for now and refactor later - git gets hard with multiple people committing on the same branch :) don't want to accidentally delete anything

PR #729, opened 2015-10-23
[Delivers #98149712] authorization audit

pkarman:
Story: https://www.pivotaltracker.com/story/show/98149712

Implements documentation, tests and fixes for authorization audit, notably:

* responds with 403 and authorization error message rather than redirecting with flash message
* articulates authorization roles and rules for pending ATO application

### DIFF
--- a/app/models/concerns/proposal_delegate.rb
+++ b/app/models/concerns/proposal_delegate.rb
@@ -48,4 +48,8 @@ module ProposalDelegate
   def client
     self.class.client
   end
+
+  def slug_matches?(user)
+    user.client_slug == client
+  end
 end
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -274,7 +274,7 @@ class Proposal < ActiveRecord::Base
     if email_or_user.is_a?(User)
       email_or_user
     else
-      User.for_email(email_or_user)
+      User.for_email(email_or_user, client)
     end
   end
 end
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -153,7 +153,7 @@ class Proposal < ActiveRecord::Base
     # this authz check is here instead of in a Policy because the Policy classes
     # are applied to the current_user, not (as in this case) the user being acted upon.
     if client_data && !client_data.slug_matches?(user)
-      fail Pundit::NotAuthorizedError.new("May not add observer with different client.")
+      fail Pundit::NotAuthorizedError.new("May not add observer belonging to a different organization.")
     end
 
     unless existing_observation_for(user)
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -274,7 +274,7 @@ class Proposal < ActiveRecord::Base
     if email_or_user.is_a?(User)
       email_or_user
     else
-      User.for_email(email_or_user, client)
+      User.for_email_with_slug(email_or_user, client)
     end
   end
 end
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -159,6 +159,12 @@ class Proposal < ActiveRecord::Base
       user = User.for_email(email_or_user)
     end
 
+    # this authz check is here instead of in a Policy because the Policy classes
+    # are applied to the current_user, not (as in this case) the user being acted upon.
+    if client_data && !client_data.slug_matches?(user)
+      fail Pundit::NotAuthorizedError.new("May not add observer with different client.")
+    end
+
     create_new_observation(user, adder, reason) unless existing_observation_for(user)
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -91,8 +91,12 @@ class User < ActiveRecord::Base
     !admin?
   end
 
-  def self.for_email(email)
-    User.find_or_create_by(email_address: email.strip.downcase)
+  def self.for_email(email, client_slug = nil)
+    u = User.find_or_create_by(email_address: email.strip.downcase)
+    unless u.client_slug
+      u.client_slug = client_slug
+    end
+    u
   end
 
   def self.from_oauth_hash(auth_hash)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -91,8 +91,12 @@ class User < ActiveRecord::Base
     !admin?
   end
 
-  def self.for_email(email, client_slug = nil)
-    u = User.find_or_create_by(email_address: email.strip.downcase)
+  def self.for_email(email)
+    User.find_or_create_by(email_address: email.strip.downcase)
+  end
+
+  def self.for_email_with_slug(email, client_slug)
+    u = self.for_email(email)
     unless u.client_slug
       u.client_slug = client_slug
     end
--- a/doc/authorization.md
+++ b/doc/authorization.md
@@ -1,0 +1,85 @@
+# C2 Authorization
+
+This document describes the data managed by the C2 application,
+the various roles that users of the C2 application play vis-a-vis 
+that data, and the authorization rules in place to 
+ensure the security and integrity of the data.
+
+In this document the key words "MUST", "SHOULD" and "MAY" are used
+as defined in https://www.ietf.org/rfc/rfc2119.txt.
+
+## Terms
+
+A *User* represents a human being who accesses the application via the HTTP (web)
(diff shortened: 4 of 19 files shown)


pkarman:
@phirefly assigning this to you to make sure the rules get a review, before I start fixing tests.

app/models/concerns/proposal_delegate.rb
```diff
@@ -48,4 +48,8 @@ def self.client
   def client
     self.class.client
   end
+
+  def slug_matches?(user)
```
jessieay: should we write a little unit test for this so that if it changes we have a test failure?

app/models/proposal.rb
```diff
@@ -150,6 +150,12 @@ def existing_observation_for(user)
   def add_observer(email_or_user, adder=nil, reason=nil)
     user = find_user(email_or_user)
 
+    # this authz check is here instead of in a Policy because the Policy classes
+    # are applied to the current_user, not (as in this case) the user being acted upon.
+    if client_data && !client_data.slug_matches?(user)
+      fail Pundit::NotAuthorizedError.new("May not add observer with different client.")
```
jessieay: I myself am not exactly sure what a client is in this context, and I am wondering if the end user might have a similar confusion. Would it also make sense to say something like "May not add observer from a different organization." ?

app/models/user.rb
```diff
@@ -91,8 +91,12 @@ def not_admin?
     !admin?
   end
 
-  def self.for_email(email)
-    User.find_or_create_by(email_address: email.strip.downcase)
+  def self.for_email(email, client_slug = nil)
```
jessieay: `for_email_and_client_slug`?

spec/controllers/observations_controller_spec.rb
```diff
@@ -31,12 +31,12 @@
...
 
-    it "redirects with a warning if unsuccessful" do
+    it "responds with a warning if unsuccessful" do
       login_as(create(:user))
       post :destroy, proposal_id: proposal.id, id: observation.id
-      expect(response).to redirect_to(proposals_path)
+      expect(response.status).to eq(403)
       expect(flash[:success]).to be_nil
-      expect(flash[:alert]).not_to be_empty
+      expect(flash[:alert]).to be_nil
```
jessieay: :cool:

spec/controllers/proposals_controller_spec.rb
```diff
@@ -54,16 +54,16 @@
       it 'should allow the requester to see it' do
         proposal = create(:proposal, requester: user)
         get :show, id: proposal.id
+        expect(response.status).to eq(200)
         expect(response).not_to redirect_to("/proposals/")
         expect(flash[:alert]).not_to be_present
       end
 
       # might be flaky? -Aidan, 8/14/15
```
jessieay: remove this comment while we're here? (not sure what flaky means in this context)

spec/features/client_slug_authz_spec.rb
```diff
@@ -5,4 +5,93 @@
...
+    fill_in 'Vendor', with: 'Yoshi'
+    fill_in 'Amount', with: 123.45
+    select approver.email_address, from: "Approving official's email address"
+    fill_in 'Building number', with: Ncr::BUILDING_NUMBERS[0]
+    select Ncr::Organization.all[0], from: 'ncr_work_order_org_code'
+    find('input[name="commit"]').click
+  end
+
+  def add_as_observer(user)
+    #puts page.body
```
jessieay: remove comment?

spec/features/client_slug_authz_spec.rb
```diff
@@ -5,4 +5,93 @@
...
+    visit proposal_path
+    expect(page.status_code).to eq(200)
+    add_as_observer(gsa_user)
+    expect(page.status_code).to eq(403)
+    expect(page).to have_content("You are not allowed to add observers to that proposal.")
+  end
+
+  private
+
+  def submit_ba60_work_order(approver)
```
jessieay: yay :heart: this

doc/authorization.md
```diff
@@ -0,0 +1,115 @@
...
+for the National Capital Region (NCR) there are `WorkOrder` records, each of which is
+related to exactly one *Proposal*.
+
+*ClientData* records contain client-specific funding request details: building or vendor
+names, codes or numbers. Each Client may have different validation rules around their
+respective details: required, min/max ranges, etc. A full list of *ClientData* specifics
+is in the database schema document [`db/schema.rb`](../db/schema.rb) and the related Model
+classes.
+
+*ClientData* is not PII and is not considered legally sensitive; however *ClientData* is siloed
```
jessieay: what about `however different types of *ClientData*`

doc/authorization.md
```diff
@@ -0,0 +1,115 @@
...
+
+A User who MAY act on any Proposal or User in the system. Access to the *Admin* role
+may only be granted by another User with the *Admin* role (via the `/admin` UI) or by
+someone with console access (e.g. via `cf-ssh`).
+
+### Client Admin :anchor:
+
+A User who MAY view all Proposals with the same `client_slug` as part of their dashboard
+aggregation reporting. NOTE this aggregation feature is a UX convenience and does not
+reflect any data sensitivity, since any User MAY view any Proposal with a consisten `client_slug`.
```
jessieay: small typo `consisten` -> `consistent`

doc/authorization.md
```diff
@@ -0,0 +1,115 @@
...
+An umbrella term, referring to an Approver, Delegate, Observer or Requester.
+I.e. a User explicitly associated with a Proposal.
+
+### Admin :anchor:
+
+A User who MAY act on any Proposal or User in the system. Access to the *Admin* role
+may only be granted by another User with the *Admin* role (via the `/admin` UI) or by
+someone with console access (e.g. via `cf-ssh`).
+
+### Client Admin :anchor:
```
jessieay: I didn't know we had this role until now! :mouse:

doc/authorization.md
```diff
@@ -0,0 +1,115 @@
...
+1. Any User MAY create a new Proposal consistent with their `client_slug`. E.g.
+a User with a `client_slug` of **ncr** MAY only create a NCR `WorkOrder` and not any other
+kind of Proposal.
+
+1. Any Subscriber MAY add another User with the same `client_slug` as an Observer
+to the subscribed Proposal.
+
+1. Only the Requester of a Proposal MAY edit or cancel the Proposal.
+
+1. Any Subscriber MAY comment on a Proposal, which will add that User as
```
jessieay: This might be more specific: "Any subscriber for a Proposal MAY comment on that Proposal," -- not sure if being super specific matters here

doc/authorization.md
```diff
@@ -0,0 +1,115 @@
...
+an Observer on the Proposal.
+
+1. Only the Requester and the Observer MAY remove the Observer from a Proposal.
```
jessieay: "Only the Requester or the Observer" ?

jessieay:
I learned a lot about this app by reading that document! :clap:

app/models/user.rb
pkarman: I was going for backwards compatibility. 

The case here is the `or_create_by` behavior. We want `client_slug` to be explicit when it matters. I could create a new method with your proposed name and not-nil client_slug required. I just didn't want to change `for_email` since it is used in so many places as a short-cut for `User.find_by(email_address: ...)` without the `or_create_by` behavior.

jessieay: in that case, maybe it makes sense to have a separate method for this? `for_email-and_slug` that calls `for_email` and also does the slug stuff?

 I agree that we are using this method in a lot of places and so it doesn't make sense to change, but I also am worried that we are calling it in a lot of places and doing more than just finding or creating a user by email

doc/authorization.md
jessieay: after our meeting this morning I was wondering: who exactly are client admins? Does it make sense to indicate who we give client admin permissions in this doc?

jessieay:
Two final questions, then it looks good to :ship:

pkarman:
@jessieay thanks for the thorough review. I have added a new method in ab54ecb
 and added examples to the Client Admin documentation.

jessieay:
woo hoo! merging

PR #730, opened 2015-10-26
[WIP] local development with Docker

afeld:
I was running a workshop this weekend for new open source contributors, which made me very aware of how difficult it is to get a project running locally for development. I've heard promising things about Docker making this easier, and know that @dhcole has been experimenting it for Midas and Federalist, so figured I should give it a shot. Was able to get C2 running inside of it within a couple hours:

screen shot 2015-10-26 at 4 40 54 am

I still don't have a great sense of all the Docker components/configuration/commands, and not sure that it saves that much setup headache for our relatively simple architecture (you trade installing Postgres and Ruby for VirtualBox and Docker), but want to continue to play around with it.

/cc @jeremiak

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,4 +10,12 @@ WORKDIR $APP_HOME
 ADD Gemfile* $APP_HOME/
 RUN bundle install
 
+# https://github.com/18F/C2/pull/730#discussion_r43045350
+RUN apt-get clean
+RUN rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
 ADD . $APP_HOME
+
+ENV PORT 3000
+
+CMD $APP_HOME/bin/rake db:setup && $APP_HOME/script/start
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+# https://docs.docker.com/compose/rails/
+FROM ruby:2.2.3
+RUN apt-get update -qq && apt-get install -y build-essential libpq-dev
+
+# https://robots.thoughtbot.com/rails-on-docker
+ENV APP_HOME /c2
+RUN mkdir $APP_HOME
+WORKDIR $APP_HOME
+
+ADD Gemfile* $APP_HOME/
+RUN bundle install
+
+ADD . $APP_HOME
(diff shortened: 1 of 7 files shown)


jessieay:
this is cool! C2 was pretty easy to set up without Docker thanks to the great documentation / setup script but anything that makes it even easier is a win in my book! :books:

barakmich:
Looks (generally) Good To Me, with nits -- one thing I might suggest is adding a CMD directive, which would remove your dependence on docker-compose (though you could keep that file/option around for other deployment reasons) .  `docker run <your container>` would be the goal.

Ah, you say, but what about the database? A common pattern is to bind-mount your config dir (And, possibly, database data) into the container -- something like `-v ./config:/path/to/config`. The built-in option is SQLite or similar for development, and then compose or similar for postgres -- and hopefully similar final deployment. With different bind-mounts (like config-prod, config-dev, etc) you can quickly change it up.

Dockerfile
```diff
@@ -0,0 +1,13 @@
+# https://docs.docker.com/compose/rails/
+FROM ruby:2.2.3
+RUN apt-get update -qq && apt-get install -y build-essential libpq-dev
+
```
Cbeck527: Totally optional, but at WP we've added these lines to slim down the container size and remove some unnecessary cruft:
```docker
RUN apt-get clean
RUN rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
```

pkarman:
As a POC this is cool.

I'm not sure that, for development purposes, the extra dependencies are worth the trouble. As you noted elsewhere, Docker runs natively only on Linux, so unless you are developing under Linux, there are a couple more steps of indirection (some flavor of Linux VM) to take advantage of Docker. If you're on OS X (e.g.) and already have Ruby and PostgreSQL installed, this actually makes setup more complicated.

Would we really want to deploy production via Docker? How would that help us?

IME, there's a finite amount of bootstrapping pain for any project, and it doesn't matter if you sweep up lots of little piles or one big pile: the dust bin looks the same afterwards.

jessieay:
@afeld what are your thoughts on this? do you want to continue to keep it open indefinitely?

afeld:
I'll probably keep playing with it, but we can close the pull request for now. Thanks for the feedback all!

PR #731, opened 2015-10-26
[Delivers 98774038] Create one seed user for each approver type

jessieay:
* Mirrors production data more closely
* Add spec that ensures removed approvers are notified
* Story: https://www.pivotaltracker.com/story/show/98774038

### DIFF
--- a/lib/roles_conversion.rb
+++ b/lib/roles_conversion.rb
@@ -34,7 +34,7 @@ class RolesConversion
 
   def ool_ba80_budget_approver
     with_email_role_slug!(
-      ENV['NCR_OOL_BA80_BUDGET_MAILBOX'] || 'communicart.budget.approver+0ool_ba80@gmail.com',
+      ENV['NCR_OOL_BA80_BUDGET_MAILBOX'] || 'communicart.budget.approver+ool_ba80@gmail.com',
       'OOL_BA80_budget_approver',
       'ncr'
     )
--- a/lib/roles_conversion.rb
+++ b/lib/roles_conversion.rb
@@ -1,14 +1,49 @@
 class RolesConversion
-  def self.exists_with_role_slug?(role, slug)
-    the_role = role.is_a?(Role) ? role : Role.find_by_name(role)
-    return false unless the_role
-    the_role.users.exists?(client_slug: slug)
+  def ncr_budget_approvers
+    ba61_tier1_budget_approver
+    ba61_tier2_budget_approver
+    ba80_budget_approver
+    ool_ba80_budget_approver
+  end
+
+  private
+
+  def ba61_tier1_budget_approver
+    with_email_role_slug!(
+      ENV['NCR_BA61_TIER1_BUDGET_MAILBOX'] || 'communicart.budget.approver+ba61@gmail.com',
+      'BA61_tier1_budget_approver',
+      'ncr'
+    )
+  end
+
+  def ba61_tier2_budget_approver
+    with_email_role_slug!(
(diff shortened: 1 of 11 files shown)


jessieay:
Blargh there are still some failing specs because of the seeds change - will update but should be pretty small stuff (we have a lot of specs checking for the number of users in the DB)

lib/roles_conversion.rb
```diff
@@ -1,65 +1,67 @@
 class RolesConversion
```
pkarman: I'm fine with the changes here, just curious as to why you switched from class methods to instance methods?

pkarman:
looks good. Once all tests are passing, I can merge and QA.

lib/roles_conversion.rb
jessieay: I find it awkward to do private class methods in Ruby - I've never liked the `private_class_method` convention and here we were just keeping all of the class methods public when we were really needed one (and that method called the rest).

In my mind, methods that are only used for internal logic in a class should be private (keeps the public interface of the class small), and since it's easier to do private instance methods, I usually do that.

jessieay:
Hmmm failed on a poltergeist timeout - all passed locally, going to re-run circle and see if it was a fluke

jessieay:
@pkarman this is passing after re-running on Circle

pkarman:
I am QA'ing (really just deploy sanity).

lib/roles_conversion.rb
```diff
@@ -1,65 +1,67 @@
...
       'ncr'
     )
   end
 
-  def self.ool_ba80_budget_approver
-    self.with_email_role_slug!(
-      ENV['NCR_OOL_BA80_BUDGET_MAILBOX'] || 'communicart.budget.approver@gmail.com',
+  def ool_ba80_budget_approver
+    with_email_role_slug!(
+      ENV['NCR_OOL_BA80_BUDGET_MAILBOX'] || 'communicart.budget.approver+0ool_ba80@gmail.com',
```
pkarman: is that `0` a typo in `0ool_ba80` ?

jessieay: ha yup, will change

lib/roles_conversion.rb
```diff
@@ -1,65 +1,67 @@
...
       'ncr'
     )
   end
 
-  def self.ool_ba80_budget_approver
-    self.with_email_role_slug!(
-      ENV['NCR_OOL_BA80_BUDGET_MAILBOX'] || 'communicart.budget.approver@gmail.com',
+  def ool_ba80_budget_approver
+    with_email_role_slug!(
+      ENV['NCR_OOL_BA80_BUDGET_MAILBOX'] || 'communicart.budget.approver+ool_ba80@gmail.com',
```
jessieay: @pkarman typo fixed at 2292b7a97c77384d33c2a4aeff8559e2d553723f

spec/features/roles_conversion_spec.rb
```diff
@@ -1,17 +1,17 @@
...
-
-    expect(user2).to be_nil
-  end
-
-  it "should convert NCR budget approvers" do
-    expect(User.count).to eq(2) # via db/seeds
-    RolesConversion.ncr_budget_approvers
-    expect(User.count).to eq(2) # no change (idempotent)
+    it "should be idempotent based on client_slug+role" do
+      with_env_var('NCR_BA61_TIER1_BUDGET_MAILBOX', 'someoneelse@example.com') do
```
jessieay: nice!

PR #732, opened 2015-10-26
[Delivers #102133698] delayed job queue visibility

pkarman:
Story: https://www.pivotaltracker.com/story/show/102133698

Adds `delayed_jobs` to the `/admin` UI.

It seems like there could/should be a way to monitor queue length with New Relic, but I'm not sure how best to do that, since we are testing for a negative activity. See comments on https://www.pivotaltracker.com/story/show/103541696

**Testing**: since the intended non-dev behavior is for delayed jobs to be executed very quickly, it can be difficult to test this under our CF environments. If you wanted to QA this in dev environment, you could generate email and then check the dashboard to verify that the reporting SQL correctly finds the pending jobs. Otherwise, to QA in a CF environment, generate email in any common way and verify that there are no regressions. In practice the `/admin` UI will usually be empty for jobs unless you are very fast in your click-and-reload browser dexterity.

### DIFF
--- a/app/admin/dashboard.rb
+++ b/app/admin/dashboard.rb
@@ -7,15 +7,15 @@ ActiveAdmin.register_page "Dashboard" do
       now = Time.zone.now
       ul do
         li do
-          jobs = DelayedJob.where('failed_at is not null').count(:id)
+          jobs = Delayed::Job.where('failed_at is not null').count(:id)
           link_to "#{jobs} failing jobs", admin_jobs_path(q: {failed_at_is_not_null: true}), style: 'color: red'
         end
         li do
-          jobs = DelayedJob.where('run_at <= ?', now).count(:id)
+          jobs = Delayed::Job.where('run_at <= ?', now).count(:id)
           link_to "#{jobs} late jobs", admin_jobs_path(q: {run_at_lte: now.to_s(:db)}), style: 'color: hsl(40, 100%, 40%)'
         end
         li do
-          jobs = DelayedJob.where('run_at >= ?', now).count(:id)
+          jobs = Delayed::Job.where('run_at >= ?', now).count(:id)
           link_to "#{jobs} scheduled jobs", admin_jobs_path(q: {run_at_gte: now.to_s(:db)}), style: 'color: green'
         end
       end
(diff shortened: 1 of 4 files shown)


app/models/delayed_job.rb
```diff
@@ -0,0 +1,3 @@
+class DelayedJob < ActiveRecord::Base
```
jessieay: do we need this defined? I thought the gem gave us this

app/admin/job.rb
```diff
@@ -0,0 +1,68 @@
...
+      f.input :attempts,    input_html: {disabled: true}
+      f.input :failed_at,   input_html: {disabled: true}
+      f.input :last_error,  input_html: {disabled: true}
+      f.input :locked_at,   input_html: {disabled: true}
+      f.input :locked_by,   input_html: {disabled: true}
+    end
+    f.buttons
+  end
+
+  action_item :only => [:edit] do
```
jessieay: new hash syntax would be :+1:

app/admin/dashboard.rb
```diff
@@ -2,10 +2,22 @@
   menu priority: 1, label: proc { I18n.t("active_admin.dashboard") }
 
   content title: proc { I18n.t("active_admin.dashboard") } do
-    div class: "blank_slate_container", id: "dashboard_default_message" do
-      span class: "blank_slate" do
-        span I18n.t("active_admin.dashboard_welcome.welcome")
-        small I18n.t("active_admin.dashboard_welcome.call_to_action")
+
+    section "Background Jobs" do
```
jessieay: maybe we should create a little feature spec for this? like a smoke test that the admin dashboard is up and running when visited

pkarman: thanks for the testing note. We already have a feature spec for it; added an `expect` for the new delay jobs panel.

pkarman:
Thanks for the review @jessieay -- I think everything has been addressed.

jessieay:
looks good :ship:

jessieay:
I will QA this now

jessieay:
Tested in c2-dev by posting a comment -- briefly saw 4 jobs as late and then they ran and everything went back to 0. What does that mean exactly?

Merging

pkarman:
@jessieay thanks for testing. What you saw is expected. It means that 4 jobs got queued up and then they got run successfully and deleted.

PR #734, opened 2015-10-26
[Delivers #104723294] Admin can set a user as "inactive"

jessieay:
https://www.pivotaltracker.com/story/show/104723294


pkarman:
This looks great.

Do we need to add some tests around the soft delete feature? In particular I am wondering if we have any record retention/integrity issues for `Proposal` and its related client data. I.e. side effects for those models that now might appear to be orphaned in reports when their related User is soft deleted. Especially when deletes cascade, some relationships will now soft delete and others will hard delete. Looking over the code, I don't *think* that's an issue, but it might be worth writing some tests to verify that behavior for future.

jessieay:
@pkarman 

We can write tests around soft delete -- that is what I was trying to do by introducing shoulda matchers (basically make sure we had `dependent: delete` in there since it is a requirement for cascading soft-deletes with paranoia).

I now see that it would be problematic if a user is deleted and their proposals are still around. In that case, would we want to soft delete proposals as well?

Also, were you thinking about more of an integration test for this, or what did you have in mind exactly?

pkarman:
yes, more like an integration test. Whenever we "delete" something that has data retention issues. I just don't know how/if that effects any policies (implicit or not) we might have. Writing the test(s) would make the behavior explicit so if/when we ever have a written policy, we have some code to point to that (fails to) enforce it.

Enough equivocation in that paragraph above ^^ ? :stuck_out_tongue:

pkarman:
specifically, I would probably soft-delete proposals too, yes, as well as the client data.

jessieay:
Ok, I will write an integration test. Do you imagine it would be something like we create a proposal for a user, then delete the user, and make sure that visiting that proposal returns a 404? I am not sure what you mean by "policies" -- something at the level of FOIA or are you referring to pundit policies?

pkarman:
By "policy" I mean a business rule, maybe making it's way into the ATO documentation but more likely just an agreed upon rule for how the data in the db is treated. E.g. this is real money getting spent and having the audit trail seems important, so deleting a User or Proposal should not be done without some kind of trail.

cc/ @phirefly

pkarman:
re: the integration test, maybe not even necessary to do from the perspective of the user (feature test) but could even be done in a model test (i.e. peering behind the curtain). Just to verify what happens when a User is deleted -- does the Proposal also get soft-deleted? The comments? All the "stuff" associated with a human's actions in the system.

jessieay:
@pkarman good idea re: unit tests. writing now

jessieay:
@pkarman added specs for soft delete (and a few more softly deleted objects) here https://github.com/18F/C2/commit/fe24d51a0d888606d1ae9e25c240d85d804671d9

pkarman:
I would love to see a (failing or deferred) test that illustrates that not everything in a proposal tree is getting (soft) deleted. That's the kind of thing that a data retention audit would want revealed explicitly.

jessieay:
I will try to get approvals to soft delete via soft deleting a proposal. any other relations you think we should soft delete? I can't see any but I could certainly be overlooking something...

pkarman:
I would soft delete everything, frankly. The audit trail is key, and most (all?) things relate to a User somehow.

jessieay:
Ok, I can do that. What do you think about getting this PR merged first and then adding soft delete to everything else? I am for the plan but want to make sure this doesn't turn into a massive PR that takes forever to merge and causes many headaches because it contains a bunch of migrations. 

Also, what do you think abut me combining all of the migrations in this PR into one big one? I am fine with keeping as-is or merging into one file, I can see pros and cons to both approaches

pkarman:
TBH, I'm reluctant to merge this until we get some word from @phirefly on what our data retention approach is. If we merge, then we are implicitly saying "it's ok to destroy things because we really aren't deleting data" when we don't yet have enough test coverage (imo) to validate that assumption.

Re: migrations, I would just leave them as-is, unless it bothers you to have them in separate files.

jessieay:
Ok I am cool with waiting on Raphy. What other test coverage do you think we need for this?

pkarman:
Just this: https://github.com/18F/C2/pull/734#issuecomment-151647230

phirefly:
Here is the original context for this:
1. Someone recently retired will no longer be accessing the system. He was a Tier 1 approver.
2. Someone recently switched teams and will no longer be accessing the system. She is a reconciler.

In both cases, we do not want them to show up as a user in the approving official drop down box. Additionally, we do not expect them to login anymore.

We do want the requests that they approved or reconciled to continue existing in our system. For example, at the end of the year, we want to be able to pull up those requests and see the approval information -- who approved it and when. Deleting a user does not nullify a request. It just means that, moving forward, that user is not active in the system.

Does this help?

phirefly:
Out of curiosity, would there be an easy way to integrate this into the active admin panel?

jessieay:
@phirefly sure, but perhaps that should be a separate story since this one was specifically for a dev admin?

jessieay:
hmmm OK, Given that context, maybe what we should do is not add the ability to delete a user at all. Instead, we could just mark a user as "inactive" and not show inactive users in the drop down. Seems like this approach was doing more than we needed it to!

jessieay:
@pkarman updated with New and Improved and Very Different implementation

app/models/proposal.rb
```diff
@@ -30,12 +30,12 @@ class Proposal < ActiveRecord::Base
   has_many :individual_approvals, ->{ individual }, class_name: 'Approvals::Individual'
   has_many :approvers, through: :individual_approvals, source: :user
   has_many :api_tokens, through: :individual_approvals
-  has_many :attachments
+  has_many :attachments, dependent: :destroy
```
pkarman: Are these `dependent: destroy` attributes intended to implement cascading deletes? I had assumed they were added because of the `paranoia` gem, but since that is no longer part of this PR, are they here as a best practice?

pkarman:
@jessieay this looks great. I like the simpler, more direct way of solving the original story problem.

Just one question about association definitions. Otherwise, I am ready to QA.

pkarman:
QA results:

Trying to access a User account via the `/admin/users/:id/edit` link generates a 5xx error
```
ActionView::Template::Error (undefined method `active' for #<User:0x007ff87108c1d8>)
```
I don't see a db migration file for this. Any chance it's sitting uncommitted in your dev env @jessieay ?

jessieay:
@pkarman ah yes, forgot a few files. Added at 046642e600c4df6491940770ed3eee3cc121c9d0

app/models/proposal.rb
jessieay: I generally think of `dependent: :destroy

pkarman: We added foreign key constraints in `db/migrate//20150921183423_add_db_constraints.rb` for the same reasons outlined in that article (a nice piece btw). I had intentionally *not* included `ON DELETE CASCADE` to the FK definitions because my assumption was that we did not ever want to delete anything from the db, and that the FK definitions should help make it really hard to delete anything. If we circumvent that db constraint by having AR cascade the deletes at the code level, that undoes some of the data retention guards the db constraint provides.

That said, if we really cared about data retention as a business policy, I think we should be far more intentional about it, rather than abusing a happy side effect of the db constraints. So I would be inclined to (a) not include the `dependent: destroy` and (b) take a more systematic approach to the data retention issue, similar to what we've done with the authorization audit. Story created in https://www.pivotaltracker.com/story/show/106870524

pkarman:
QA on c2-dev looks good.

app/models/proposal.rb
jessieay: great - thanks for the background.

PR #736, opened 2015-10-26
simplify assignment of WorkOrder#approving_official_email

afeld:
I feel like I'm losing my mind here...can anyone explain to me why [the previous commit](https://github.com/18F/C2/commit/176a03c43768484ce48b243ba58246d578c5a0c2) is [passing](https://circleci.com/gh/18F/C2/2555), but this change makes it fail?

### DIFF
--- a/app/models/ncr/work_order.rb
+++ b/app/models/ncr/work_order.rb
@@ -60,10 +60,7 @@ module Ncr
     }
 
     def set_defaults
-      # not sure why the latter condition is necessary...was getting some weird errors from the tests without it. -AF 10/5/2015
-      if !self.approving_official_email && self.approvers.any?
-        self.approving_official_email = self.approvers.first.try(:email_address)
-      end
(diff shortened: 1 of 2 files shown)


jessieay:
I believe the specs would pass if you did

`self.approving_official_email ||= self.approvers.first.try(:email_address)`

(the difference between what you have no and this is that it does not assume the existence of approvers)

jessieay:
hmmm I tried that and it does not work, so I was wrong.

I think it would be nice to get to the bottom of this. It would also be really nice to get this out of a callback since I always find those difficult to disentangle. Don't think it makes sense to remove all callbacks at once, but as we need to change callbacks it would be amazing to actually move the logic in them outside of a callback. That way, it's much easier to test, too!

pkarman:
@afeld seems like the `.approvers` is cached in the object before the `.approvals` are set, so the association is returning empty instead of re-querying the db. Fix in ea08eae.

PR #737, opened 2015-10-26
various spec cleanup

afeld:
Extracted from #660. See commits.

### DIFF
--- a/spec/models/ncr/work_order_spec.rb
+++ b/spec/models/ncr/work_order_spec.rb
@@ -1,4 +1,6 @@
 describe Ncr::WorkOrder do
+  include ProposalSpecHelper
+
   describe '#relevant_fields' do
     it "shows BA61 fields" do
       wo = Ncr::WorkOrder.new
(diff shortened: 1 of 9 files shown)


spec/models/ncr/work_order_spec.rb
```diff
@@ -1,4 +1,6 @@
 describe Ncr::WorkOrder do
+  include ProposalSpecHelper
```
jessieay: I think we are already including all spec helpers in all specs with https://github.com/18F/C2/blob/master/spec/rails_helper.rb#L7

spec/support/proposal_spec_helper.rb
```diff
@@ -0,0 +1,10 @@
+module ProposalSpecHelper
+  def fully_approve(proposal)
```
jessieay: I am not sure what the difference between approving and full approving is...

jessieay:
lgtm :ship:

spec/models/ncr/work_order_spec.rb
afeld: That's just `require`ing the file...I'm only `include`ing the mixin in certain places.

spec/support/proposal_spec_helper.rb
afeld: I guess I just wanted to indicate that it would approve all of the Approvals...totally open to a different name for it, though.

afeld: Happy to do so in a subsequent PR, if you like.

PR #738, opened 2015-10-27
Address "todos" for ApiToken

jessieay:
* Add validations for expired_at and access_token
* Use `has_secure_token`, which will be part of Rails 5
* Add uniqueness validation for access_token
* Add unique index for access_token
* Add specs for untested methods
* Refactor ApiToken policy

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20151013224625) do
+ActiveRecord::Schema.define(version: 20151027010707) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -40,6 +40,8 @@ ActiveRecord::Schema.define(version: 20151013224625) do
     t.integer  "approval_id"
   end
 
+  add_index "api_tokens", ["access_token"], name: "index_api_tokens_on_access_token", unique: true, using: :btree
+
   create_table "approval_delegates", force: :cascade do |t|
     t.integer  "assigner_id"
     t.integer  "assignee_id"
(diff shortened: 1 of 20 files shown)


jessieay:
Story: https://www.pivotaltracker.com/story/show/106711436

jessieay:
Since this is a refactor, I think QA would just need to be a sanity check that everything isn't broken when deployed.

db/schema.rb
```diff
@@ -58,8 +60,10 @@
     t.string   "type"
     t.integer  "parent_id"
     t.integer  "min_children_needed"
+    t.datetime "deleted_at"
   end
 
+  add_index "approvals", ["deleted_at"], name: "index_approvals_on_deleted_at", using: :btree
```
pkarman: some of the schema changes in this PR make it seem like this was branched from #734 which might have implications for the order in these PRs get merged? That's a yellow flag for me but mostly just about mechanics not substance.

pkarman:
There's a lot in here that's simply syntax style changes, which look fine to me. One note about merge mechanics. 

I am happy to QA.

pkarman:
deployed to c2-dev, logged in/out in a few different combinations. Seems fine to me.

pkarman:
Going to wait to merge this until the `db/schema.rb` issue is sorted because I don't want to mess up #734

db/schema.rb
jessieay: ah yes, darn I always forget to roll back migrations when I have two ongoing PRs with migrations. good catch!!

jessieay:
In addition to addressing a few todos from the code, this is a small proof of concept for future spec refactors. Changes include:

- moving away from `before` and `let` statements in favor of private methods
- using factories in tests rather than initializing objects
- separating test phases by newlines (setup, exercise, expectations)
- including test coverage for associations / validations
- including method tests in `spec/models` rather than in feature specs
- removing callbacks where possible in favor of explicitly called methods

jessieay:
@pkarman schema updated at 1498687b5da397e1db430112862588d5465eda3d

PR #739, opened 2015-10-27
simplify assignment of WorkOrder#approving_official_email

afeld:
Moved from #736, to have a target of `master`.

### DIFF
--- a/app/controllers/ncr/work_orders_controller.rb
+++ b/app/controllers/ncr/work_orders_controller.rb
@@ -34,8 +34,7 @@ module Ncr
     protected
 
     def attribute_changes?
-      # TODO remove need for passing in the approving_official_email
-      super || @model_instance.approver_changed?(@model_instance.approving_official_email)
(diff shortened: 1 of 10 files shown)


pkarman:
@afeld seems like `wo.reload` is "fixing" the failing tests, but that seems to indicate that there is some check for `.approvers` in the creation process that is caching the empty array. Fixing that initial caching is probably what is best. I think that's what the original `approvers.any?` line had the side effect of doing.

afeld:
Btw, I rebased this branch when creating the new pull request and that seems to have made duplicate commits...let's squash before merging. I'll try and look into this deeper at some point today, unless you want to keep playing with it. Thanks for the help!

pkarman:
If it were up to me, I would just clarify the comment since we now know why the `.approvers.any?` check is there, and/or add an explicit `.approvers(true)` inside `set_defaults`.

afeld:
Not sure that this is worth spending that much time on. Closing for now.

PR #741, opened 2015-10-27
[Delivers #105956396] slim WorkOrder model

afeld:
Follow-up to #490.

This is simply a refactoring. The biggest change is that the business logic methods that set up the Approvals on WorkOrders are now in a standalone class, and not mixed in with the other methods of the model...the WorkOrder model and associated spec file are now much easier to reason about.


app/models/ncr.rb
```diff
@@ -0,0 +1,9 @@
+module Ncr
+  # Make sure all table names use 'ncr_XXX'
+  def self.table_name_prefix
+    'ncr_'
+  end
+
+  EXPENSE_TYPES = %w(BA60 BA61 BA80)
```
jessieay: wondering why these are in the module and not the class itself? (I know this isn't a change, but I've thought this every time I visited `work_order` so thought I'd ask)

lib/ncr/approval_manager.rb
```diff
@@ -0,0 +1,99 @@
+module Ncr
+  class ApprovalManager
```
jessieay: as long as we are moving stuff around, what about breaking this approval stuff into even smaller bits? This class is certainly more digestible than how it was before, but it's still doing several things.

app/models/ncr.rb
afeld: I was going back-and-forth about moving them with this PR...debatable whether it's NCR-wide information or if it's specific to work orders. Strong feelings?

lib/ncr/approval_manager.rb
afeld: Moved the mailbox logic to it's own module.

afeld:
Updated!

pkarman:
:+1: from me. @jessieay you want to review as well since you were part of slack discussion?

lib/ncr/mailboxes.rb
```diff
@@ -0,0 +1,29 @@
...
+      self.approver_with_role("BA80_budget_approver")
+    end
+
+    def self.ool_ba80_budget
+      self.approver_with_role("OOL_BA80_budget_approver")
+    end
+
+    protected
+
+    def self.approver_with_role(role_name)
```
jessieay: protected class methods are tricky: http://blog.jayfields.com/2006/11/ruby-protected-class-methods.html

what about making the methods in here instance methods instead and then we can use `private` for `approver_with_role` ?

lib/ncr/mailboxes.rb
```diff
@@ -0,0 +1,29 @@
+module Ncr
+  module Mailboxes
```
jessieay: thoughts on making this a class vs module?

lib/ncr/approval_manager.rb
```diff
@@ -0,0 +1,90 @@
...
+    attr_reader :work_order
```
jessieay: are we using this `attr_reader` outside of this class? If not, about about making it private?

lib/ncr/approval_manager.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    def setup_approvals_and_observers
+      if self.work_order.emergency
+        setup_emergency
+      else
+        setup_non_emergency
+      end
+    end
+
+    protected
```
jessieay: private instead?

jessieay:
a few more questions - this is looking great :clap:

lib/ncr/mailboxes.rb
afeld: > protected class methods are tricky

Not sure that I fully grok the "virtual class" class thing, but is it really a problem? Especially since Ruby's `protected`/`private` behavior is...unique...I just use them to indicate to other developers "you probably don't want to call this from outside the method/module.

> what about making the methods in here instance methods

I personally think a singleton makes sense here, since conceptually there's no reason to create multiple `Ncr::Mailboxes` instances. Sound reasonable?

lib/ncr/mailboxes.rb
afeld: See comment above.

lib/ncr/approval_manager.rb
afeld: I don't really see the value of `private` over `protected`, based on the reason [above](#discussion_r44192717). Strong feelings?

lib/ncr/mailboxes.rb
jessieay: When I see `protected` as a developer, I think "something that inherits from this class uses these methods` when I see `private`, I think: this class itself uses these methods. Agreed that it is an indication to other developers, but there are also real differences in behavior so I think it is worth using private/protected as intended. See http://tenderlovemaking.com/2012/09/07/protected-methods-and-ruby-2-0.html

Interesting that you see this module as a singleton...I usually think of modules in ruby as a set of behaviors to be mixed in multiple other classes. Since we are just using this in one other class, that is not how we are using this set of methods. 

Agreed that there is not a conceptual reason for us to have multiple instances. I would still LOVE to change up a lot of the underling behavior with these mailboxes since our seeds dependency / slow test suite is really cramping my TDD style...so having these methods extracted into a separate module will make it easier to do further changes down the road, I hope!

lib/ncr/approval_manager.rb
jessieay: the value is that the methods can't be called by objects that inherit from this class. Private has a small scope. if the scope is larger, I assume it's for a reason. Which I think it isn't here?

lib/ncr/mailboxes.rb
afeld: > Interesting that you see this module as a singleton...I usually think of modules in ruby as a set of behaviors to be mixed in multiple other classes

They can be either!

http://dalibornasevic.com/posts/9-ruby-singleton-pattern

 Thanks Ruby :unamused:

lib/ncr/approval_manager.rb
afeld: I'm not sure that I've ever seen `attr_reader` (or `attr_accessor`) made private...I can, but I like them specified at the top of the class for visibility reasons. Strong feelings?

afeld:
Updated!

lib/ncr/approval_manager.rb
jessieay: I've been putting them all over this project! :) I don't have strong feelings per se but I do think it generally makes sense to keep the public API of an object as small as possible and making this private would do that.

afeld: Updated.

jessieay:
needs a rebase -- but looks good to :ship:

afeld:
Updated!

pkarman:
QA'd on c2-dev

PR #742, opened 2015-10-28
[Delivers #106855394] close authorization backdoor for non-subscriber email reply

pkarman:
Story: https://www.pivotaltracker.com/story/show/106855394

Removes authorization backdoor where a non-subscriber could add themselves as an observer by replying to an email mentioning a request. The rules were clarified in a24118d9012c2bd53af40997907fa9dbd9fc88ea

**Testing**: this can be tested by sending an email to the `*inbox@c2.18f.gov` address from a non-subscriber. The email subject should have the same Request ID pattern as any legit email from the system. The most straightforward way to do this would be:

* create new request
* reply to the generated announcement email, cc'ing a non-subscriber who is a user in the system
* reply from the non-subscriber's email account

### DIFF
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -100,6 +100,8 @@ class Proposal < ActiveRecord::Base
     results.compact.uniq
   end
 
+  alias_method :subscribers, :users
+
   def root_approval=(root)
     old_approvals = self.approvals.to_a
 
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -143,6 +143,10 @@ class Proposal < ActiveRecord::Base
     end
   end
 
+  def has_subscriber?(user)
+    users.include?(user)
+  end
+
   def existing_observation_for(user)
     observations.find_by(user: user)
   end
--- a/lib/incoming_mail/handler.rb
+++ b/lib/incoming_mail/handler.rb
@@ -26,7 +26,7 @@ module IncomingMail
       resp = Response.new(type: identify_mail_type(payload))
       case resp.type
       when REQUEST
-        resp.comment = create_comment(payload['msg'])
+        resp.comment = transform_msg_to_comment(payload['msg'])
         if resp.comment
           resp.action = Response::COMMENT
         else
@@ -64,7 +64,7 @@ module IncomingMail
       CommunicartMailer.resend(msg).deliver_later
     end
 
-    def create_comment(msg)
+    def transform_msg_to_comment(msg)
       # IMPORTANT that we check/add as observer before we create comment,
       # since comment will create as a user if not already,
       # and we want the reason logged.
@@ -77,6 +77,13 @@ module IncomingMail
       unless user && proposal.has_subscriber?(user)
         return
       end
+
+      create_comment(parsed_email)
+    end
+
+    def create_comment(parsed_email)
+      proposal = parsed_email.proposal
+      user = parsed_email.comment_user
 
       unless proposal.existing_observation_for(user)
         reason = "Added comment via email reply"
--- a/lib/incoming_mail/handler.rb
+++ b/lib/incoming_mail/handler.rb
@@ -72,7 +72,11 @@ module IncomingMail
       proposal = parsed_email.proposal
       user = parsed_email.comment_user
 
-      return unless user  # cannot create comment for non-existent user
+      # cannot create comment for non-existent user, or
+      # for user who is not already a subscriber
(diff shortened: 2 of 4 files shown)


spec/features/incoming_mail_handler_spec.rb
```diff
@@ -82,19 +82,9 @@
...
-      my_proposal.requester.email_address, # comment
-      my_proposal.individual_approvals.last.user.email_address, # comment
-      my_approval.user.email_address, # comment
-    ]
-
-    expect(deliveries.length).to eq(comment_recipients.size)
-    expect(deliveries.map{|m| m.to.first}.sort).to eq(comment_recipients.sort)
+    expect(resp.action).to eq(IncomingMail::Response::FORWARDED)
+    expect(deliveries.length).to eq(1)
+    expect(my_proposal.has_subscriber?(user)).to be_falsey
```
jessieay: what do you think about writing this expectation as `expect(my_proposal).not_to have_subscriber(user)` ? 

I aways find `falsey` a bit weird because I can't remember if it is the same as `eq false` or not and also spelling (both `falsey` and `falsy` work in RSpec which is bizarre IMO)...so I try to avoid it when I can

lib/incoming_mail/handler.rb
```diff
@@ -72,7 +72,11 @@ def create_comment(msg)
       proposal = parsed_email.proposal
       user = parsed_email.comment_user
 
-      return unless user  # cannot create comment for non-existent user
+      # cannot create comment for non-existent user, or
+      # for user who is not already a subscriber
+      unless user && proposal.has_subscriber?(user)
```
jessieay: feels like the logic in this `create_comment` method is getting pretty complicated (~30 lines long). Any ideas on how we could simplify it?

app/models/proposal.rb
```diff
@@ -143,6 +143,10 @@ def reset_status()
     end
   end
 
+  def has_subscriber?(user)
```
jessieay: I like this method -- wonder if we should rename `users` to `subscribers` ?

pkarman: I like both. `users` has parity with other method names in the model space and clearly returns an array of `User` objects. The term `subscriber` is shorthand for a business rule specific to the `Proposal` model.

Maybe `alias_method :users, :subscribers` ?

lib/incoming_mail/handler.rb
pkarman: Maybe rename `create_comment(msg)` to `transform_message(msg)` and add a new method `create_method(user, proposal, text)` to handle the mechanics of actually creating the message? I can try that.

app/models/proposal.rb
jessieay: yeah alias would work, or a `subscribers` method that just calls `users`...either way. I noticed in a mailer class we have in this project that there are a bunch of aliased methods and that can get confusing, but just one wouldn't be too bad.

lib/incoming_mail/handler.rb
jessieay: sure!!

app/models/proposal.rb
pkarman: I used an alias in 3c8adc1

pkarman:
@jessieay thanks for the review. All comments addressed.

jessieay:
@pkarman this appears to be failing now :(

pkarman:
@jessieay I rebuilt the CircleCI tests with no code change and it passes now. The original failure in question seemed to be for a test that relies on randomization, and I am suspicious that it might generate false positives somehow? This is the first time I've seen it, and I don't pretend to understand what/how it is supposed to be testing. Inclined to :see_no_evil: until it happens again.

jessieay:
Updates look awesome :ship: 

will QA now

jessieay:
OK, I created a request and then logged into one of the approver email accounts and fowarded the email to my personal gmail. I then replied to the forwarded email and sent it to dev-inbox@c2.18f.gov -- and the account with my personal email address was not added as an observer.

Seem like sufficient QA to you, Peter? Wasn't sure if I was doing the "reply to forwarded email" part right...

jessieay:
If sounds good, I will merge

pkarman:
@jessieay perfect QA. Your email showed up in the communicart.sender@gmail inbox with this header:
`X-C2-Original-From: "Jessie A. Young" <jessie.a.young@gmail.com>` which is exactly right.

jessieay:
nice! merging

PR #743, opened 2015-10-28
Add info on blocked stories to Team Practices doc

jessieay:
* Stories that are not actionable should be moved to icebox
* Reference: https://18f.slack.com/archives/cap/p1446062132000076

### DIFF
--- a/doc/team_practices.md
+++ b/doc/team_practices.md
@@ -14,6 +14,10 @@ and to add more detail into stories before IPM. Team members are encouraged to a
 
 Team members self-select story ownership.
 
+If a story in backlog does not have enough detail to be actionable, add a
+comment asking for the required information, tag the story with the `blocked`
+tag and assign the product manager as the owner to determine next actions.
+
 Production deployments may happen one to several times a week, so keep master in production-ready shape.
 
 In the workflow below, there are multiple roles to be played:


jessieay:
Want to make sure everyone sees this bc it seems like a helpful clarification for Tracker management. cc @afeld @pkarman @yozlet @annalee

pkarman:
Do we really want to move it to the icebox straight away? Seems like a judgement call the product manager should make, since presumably they are the one who prioritized it in the first place.

Can I amend to "tag with *blocked* and assign to the product owner" as sufficient? Presumably our stand-ups are the place to bring this up, and moving it to icebox means it'll just "disappear" from the current sprint.

jessieay:
@pkarman I created this PR based on Slack convo linked to with Raphy, so I will let him respond to your questions. Either way works for me, I just want to make sure we have the "blocked story" process documented.

cc @phirefly

phirefly:
@pkarman I concur, partly because detail may immediately be added that will make the story actionable. How about "tag the story with `blocked` tag and assign the product manager as the owner to determine next actions"

jessieay:
@phirefly @pkarman updated!

pkarman:
great, thanks.

PR #744, opened 2015-10-28
[Delivers #105572578] Update the Expense Type label to include 'Program'

phirefly:


app/views/ncr/work_orders/form.html.haml
```diff
@@ -8,7 +8,7 @@
   = simple_form_for @model_instance, html: { multipart: true } do |f|
     = f.input :project_title
     = f.input :description
-    = field_set_tag "Expense type", class: 'required' do
+    = field_set_tag "Expense type / Program", class: 'required' do
```
jessieay: I am guessing that this field name will break some integration tests

pkarman: tests passing -- guess our suite is missing that particular integration test.

PR #745, opened 2015-10-28
[Delivers #99713258] Indicate significance of asterisk in form. 

phirefly:
Also Removed asterisk from the Work Orde field since it is not required

screen shot 2015-10-28 at 3 49 50 pm

### DIFF
--- a/app/views/gsa18f/procurements/form.html.haml
+++ b/app/views/gsa18f/procurements/form.html.haml
@@ -6,6 +6,9 @@
     Purchase Request - Supplies
   %p
     Eg; ergonomic keyboards, foamcore boards, ergonomic mice, monitor stands, post-its, sharpies, etc.
+  %p
+    %em
+      * Indicates a required field
   = simple_form_for @model_instance do |f|
     = f.input :office, collection: Gsa18f::Procurement::OFFICES
     = f.input :product_name_and_description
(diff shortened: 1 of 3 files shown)


jessieay:
Code looks good. 

Are there any other fields that are required in the form but not required via ActiveRecord that we should address while we are here?

jessieay:
ps - love the screenshot in the PR description. So helpful!

pkarman:
Given the scope of this story, I'd be inclined to merge as-is and if we discover other required/not-required fields later, open separate stories for them. +1 to merge

pkarman:
@phirefly thoughts on scope?

afeld:
> I'd be inclined to merge as-is and if we discover other required/not-required fields later, open separate stories for them

:+1:

afeld:
https://www.pivotaltracker.com/story/show/82418236

afeld:
* [ ] Add same note to Procurement form (maybe make a partial?)

phirefly:
Regarding scope, this was a small thing that I noticed when doing a walkthrough with new folks. Someone asked me if the field was required because of the asterisk and realized it shouldn't be there since not everyone has a work order. Afaik, this is the only field that has this discrepancy.

Happy to have this merge and adding a task for the Procurement form here: https://www.pivotaltracker.com/story/show/99713258 (great catch, @afeld!)

PR #746, opened 2015-10-28
Do not run seeds before each test run

jessieay:
* Speeds up test suite from 3 mins, 30 seconds to 2 mins 20 seconds


pkarman:
this is a really nice optimization. great work!

PR #747, opened 2015-10-28
Add defaults to DB for boolean NCR::WorkOrder fields

jessieay:
* Add DB constraint rather than setting in a callback


pkarman:
I always like making the db smarter. +1

PR #733, opened 2015-10-26
[Delivers #105300436] Add test for annual report mailer

jessieay:
* Story: https://www.pivotaltracker.com/story/show/105300436

PR #749, opened 2015-10-29
[Delivers #104767786] pending proposal clean up

pkarman:
Story: https://www.pivotaltracker.com/story/show/104767786

* add rake task for reporting pending proposals from previous fiscal years


lib/tasks/report.rake
```diff
@@ -0,0 +1,29 @@
+namespace :report do
+  desc "Show non-cancelled old fiscal year proposals"
+  task old_fiscal_year: :environment do
+    ok_to_act = ENV['OK_TO_ACT'] ? true : false
```
jessieay: what about putting this logic in a small class so it is easy to test? seems like a similar situation to the annual report, which I just backfilled tests for in https://github.com/18F/C2/pull/733

jessieay:
hey @pkarman - anything I can do to help get this unblocked?

pkarman:
@jessieay : nope, all in @phirefly 's hands.

pkarman:
The lone rubocop failure is `Class has too many lines. [105/100]` which seems harsh to me. We have lots of classes with more than 100 lines, and 100 lines feels arbitrarily small to me.
Thoughts?

jessieay:
@pkarman yeah, I think ideally in the future we could keep classes as small as possible (eg: I think `CommunicartMailer` could definitely benefit from being broken into a few smaller mailers) but I am fine with upping the rubocop class length limit since this project already has a lot of large classes

lib/vacuum.rb
```diff
@@ -0,0 +1,53 @@
+class Vacuum
```
jessieay: so I like this name, but it is also general enough that it isn't really clear to me what it does without opening it. Thoughts on more descriptive names?

pkarman: I had intentionally picked a general name (h/t feature of postgresql) as a placeholder for clean up of various flavors. I imagine a future where:

```
vacuum = Vacuum.new(Time.zone.parse('2016-09-30'))
vacuum.old_users
vacuum.old_proposals
```
is a regular maintenance task of some kind.

I guess we could make a `CleanerOfExpiredRecords` class instead?

jessieay: could use a `CleanExpiredRecords` class that uses `Vacuum` ? :cactus:

jessieay: (I like your idea of having a general use for the class)

pkarman: so what would usage look like?
```
cleaner = ExpiredRecordCleaner.new()
cleaner.vacuum_old_proposals(Time.zone.parse('2016-09-30'))
```
?

jessieay: yeah! something like that sounds good

pkarman:
@jessieay class renamed in 583537b

spec/lib/expired_record_cleaner_spec.rb
```diff
@@ -0,0 +1,33 @@
+describe ExpiredRecordCleaner do
+  describe "fiscal year" do
```
jessieay: since the only method we are using in the Rake task is `vacuum_old_proposals`, what do you think about just writing a test for that and making these other methods private? I think we could still test which proposals are deleted etc within the context of that spec. 

The reason I think it would be helpful to have the other methods be private is that having them private would indicate that they are used within the class only.

Also, while we are testing `vacuum_old_proposals`, what do you think about adding an expectation around the mailer it sends out? I almost missed that part until I saw the mailer spec.

pkarman: The `vaccum_proposal` method is public because I wanted to be able to use it while manually testing in a live environment. I.e. try it on a single proposal before sending email to 100s of people.

There are `expect` statements already around the mailer deliveries for each method. Were you wanting to see something more explicit about the kind of mail, or the content of the mail?

jessieay: I am not sure I agree w the logic on making the method public so it can be tested manually -- if we are not confident in the behavior/loigic of a non-public method, we should either test its usage via more rigorous via automated tests for the necessary public methods or it is sufficiently complicated that we should break it out into its own class.

Did not see the deliveries specs before - noted!

pkarman: I should clarify: the word "test" is probably overloaded here. I meant to say something more like "trial" -- that is, I find it prudent to run a task against a single proposal to verify it all works as expected in production, before running it on hundreds of proposals and generating email to lots of users.

I have added a rake task in 8961fdf to illustrate what I mean.

lib/expired_record_cleaner.rb
```diff
@@ -0,0 +1,53 @@
...
+  def initialize(dtim, ok_to_act = false, verbose = false)
+    @fiscal_year_start = self.class.fiscal_year_start(dtim)
+    @ok_to_act = ok_to_act
+    @verbose = verbose
+  end
+
+  def vacuum_old_proposals
+    proposals = Proposal.pending.where("created_at < '#{@fiscal_year_start}'")
+    ids = []
+    proposals.each do |proposal|
```
jessieay: what about doing `map` here and not worrying about returning the ids? Since we are already outputting the `"#{proposal.public_id} -> #{proposal.requester.email_address}"` below, seems like we have some level of logs documenting what was deleted.

lib/expired_record_cleaner.rb
```diff
@@ -0,0 +1,53 @@
+class ExpiredRecordCleaner
+  attr_reader :verbose, :fiscal_year_start
+
+  def initialize(dtim, ok_to_act = false, verbose = false)
+    @fiscal_year_start = self.class.fiscal_year_start(dtim)
+    @ok_to_act = ok_to_act
```
jessieay: smart -- in what case would we want this to be `false` ? Or is this just a safeguard? 

also: thoughts on named args?

app/views/communicart_mailer/proposal_reminder.text.erb
```diff
@@ -0,0 +1,4 @@
+All unapproved FY15 credit card funding requests are being cancelled.
+If the need still exists, please submit a new FY16 funding request.
+
+<%= url_for(controller: 'proposals', action: 'show', id: @proposal.id) %>
```
jessieay: why isn't rubocop failing for single quotes?

jessieay: This could also be written as `proposal_url(@proposal)`

app/views/communicart_mailer/proposal_reminder.text.erb
```diff
@@ -0,0 +1,4 @@
+All unapproved FY15 credit card funding requests are being cancelled.
```
jessieay: given the content of this email, might it make sense to name it something like `proposal_cancellation_notification` instead?

lib/expired_record_cleaner.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def send_reminder_email(proposal)
+    CommunicartMailer.proposal_reminder(proposal).deliver_later
+  end
+
+  def handle_no_requester(proposal)
+    @verbose and STDERR.puts "#{proposal.id} <= no Requester defined"
+    if @ok_to_act
+      proposal.destroy
```
jessieay: thoughts on adding soft delete via paranoia or something like that?

spec/lib/expired_record_cleaner_spec.rb
jessieay: ah ok, if we are using via a rake task that makes total sense :+1:

lib/expired_record_cleaner.rb
pkarman: I return the list of ids in order to make it easier to test:

https://github.com/18F/C2/pull/749/files#diff-f27118fd70df5786f0be909423181f14R16

jessieay: what about using this type of syntax to make sure the action removes one proposal and then also checking that the proposal created in the test no longer exists? 

https://github.com/18F/C2/blob/a91e65046e0295b13423a26edf716040446678c8/spec/features/roles_conversion_spec.rb#L11

lib/expired_record_cleaner.rb
pkarman: it's a safeguard. assumes that by default it will be a no-op.

I typically do named args when I get more than 3 arguments, or if they are all required. That's just a personal preference (you're talking to an old C programmer).

lib/expired_record_cleaner.rb
jessieay: got it

lib/expired_record_cleaner.rb
```diff
@@ -0,0 +1,53 @@
...
+    year = (dtim.month >= 10) ? dtim.year : dtim.year - 1
+    Time.zone.parse("#{year}-10-01")
+  end
+
+  def send_reminder_email(proposal)
+    CommunicartMailer.proposal_cancellation(proposal).deliver_later
+  end
+
+  def handle_no_requester(proposal)
+    @verbose and STDERR.puts "#{proposal.id} <= no Requester defined"
```
jessieay: I think what this is doing (if I understand it right) would be clearer if there were an `if`

```
if @verbose
  STDERR.puts "#{proposal.id} <= no Requester defined"
end
```

same for case below

lib/expired_record_cleaner.rb
```diff
@@ -0,0 +1,53 @@
+class ExpiredRecordCleaner
+  attr_reader :verbose, :fiscal_year_start
+
+  def initialize(dtim, ok_to_act = false, verbose = false)
```
jessieay: what about `datetime` instead of `dtim`?

lib/expired_record_cleaner.rb
```diff
@@ -0,0 +1,53 @@
...
+    if !proposal.requester
+      handle_no_requester(proposal)
+    else
+      handle_cancellation(proposal)
+    end
+  end
+
+  private
+
+  def self.fiscal_year_start(dtim)
```
jessieay: why is this a class method rather than instance?

lib/expired_record_cleaner.rb
```diff
@@ -0,0 +1,53 @@
...
+    year = (dtim.month >= 10) ? dtim.year : dtim.year - 1
+    Time.zone.parse("#{year}-10-01")
+  end
+
+  def send_reminder_email(proposal)
```
jessieay: rename method to match mailer method?

lib/tasks/vacuum.rake
```diff
@@ -0,0 +1,20 @@
+namespace :vacuum do
+  desc "Clean up old fiscal year proposals"
+  task old_proposals: :environment do
+    ok_to_act = ENV["OK_TO_ACT"] ? true : false
```
jessieay: Since env vars are always strings, If `ENV['OK_TO_ACT']` were set to "false", I think this would still evaluate to `true` because if any string is present, `if [ANYSTRING]` will return `true` 

Perhaps we should pass an arg to this rake task instead of depending on an env var to set something to true or false?

pkarman:
@jessieay I believe I've caught all the things you've noted.

lib/tasks/vacuum.rake
```diff
@@ -0,0 +1,34 @@
+namespace :vacuum do
+  desc "Clean up old fiscal year proposals"
+  task old_proposals: :environment do
+    ok_to_act = false
+    if ENV["OK_TO_ACT"] && ENV["OK_TO_ACT"].match(/^[yt]/i)
```
jessieay: I am not sure what the regex value should be in order for this to return `true` -- what about `if ENV["OK_TO_ACT"] -- "true"` ?

jessieay:
one FINAL comment about the regex -- I think requiring a certain string could help regex challenged peeps like me figure out how to use this. But feel free to tell me that I should just learn more about regexes

pkarman:
I was trying to follow the "be generous in what you accept" philosophy. https://en.wikipedia.org/wiki/Robustness_principle

Learning some regex stuff is also good. ;)

jessieay:
OK I surrender. Will QA -- should I just run this rake task to QA?

pkarman:
I would run the rake task w/o `OK_TO_ACT` because you don't want to send email in a dev env. I would do:

```
% rake vacuum:old_proposals VERBOSE=true
```

and verify the results. Then I would pick a request that you requested and call:

```
% rake vacuum:old_proposal PROPOSAL_ID=theId VERBOSE=true
```

If you don't have a request from FY15 to test with, maybe pick one of @phirefly  or @afeld just so you can verify the email arrives.

lib/tasks/vacuum.rake
```diff
@@ -0,0 +1,34 @@
...
+      ok_to_act = true
+    end
+    verbose = false
+    if ENV["VERBOSE"] && ENV["VERBOSE"].match(/^[yt]/i)
+      verbose = true
+    end
+    cleaner = ExpiredRecordCleaner.new(Time.zone.now, ok_to_act: ok_to_act, verbose: verbose)
+    if verbose
+      STDERR.puts "Pending proposals created before #{vacuum.fiscal_year_start}"
```
jessieay: when I ran this on c2-dev, I am getting `undefined local variable or method `vacuum' for main:Object` on this line

jessieay:
I got the singular task to work on c2-dev:

`rake vacuum:old_proposal PROPOSAL_ID=6 VERBOSE=true OK_TO_ACT=true`

email:

screen shot 2015-11-17 at 10 44 06 am


One thing I will say is that it was confusing when it did not actually change the status with

` rake vacuum:old_proposal PROPOSAL_ID=theId VERBOSE=true`

because the output of that was 

`FY15-6 -> jessica.young@gsa.gov`

(needed "ok to act")

rather than telling me that it was not cancelled....wondering if we should have output that more clearly states what happened when those env vars are not set in a way that actually deletes an email?

pkarman:
@jessieay I have added more logging in 4362e9e

jessieay:
Nice! working on c2-dev now. MERGING

PR #751, opened 2015-10-30
[Delivers #99135662] improve HashDiff for empty original values

pkarman:
Story: https://www.pivotaltracker.com/story/show/99135662

* render nil and empty strings explicitly in HashDiff so it is clear what the original value was

### DIFF
--- a/app/decorators/hash_diff_decorator/base.rb
+++ b/app/decorators/hash_diff_decorator/base.rb
@@ -17,7 +17,11 @@ module HashDiffDecorator
     protected
 
     def diff_val(val)
-      if val.is_a?(Numeric)
+      if val.nil?
+        "[nil]"
+      elsif val.empty?
+        "[empty]"
+      elsif val.is_a?(Numeric)
         format('%.2f', val)
       else
         val.inspect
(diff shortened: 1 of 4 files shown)


jessieay:
code looks great! How can this be QA'd! :ship:

pkarman:
@jessieay Great question. While I was writing up instructions, I realized I hadn't actually fixed the problem. Doh!

Actual fix in 0d1dddb

To QA, create a new NCR WorkOrder, leaving the Description field empty. Save the new WorkOrder. Then Modify the WorkOrder and add a Description. There should now be a comment saying `Description changed to ...` where `...` is the text you entered.

jessieay:
Thanks for the update - will QA now

jessieay:
I just modified an 18F procurement at https://c2-dev.18f.gov/proposals/211 and no comment was left about the change -- is that expected? I know you said to QA with a work order but figured it would work for either...

jessieay:
Just change a description of a work order from something to nil --
screen shot 2015-10-30 at 1 17 33 pm

Should we indicate a value after the word "to"?

pkarman:
The paper_trail stuff doesn't work for gsa18f procurement, afaik. Only NCR WorkOrder. That's the only code I found anyway.

Yes, you found a new bug for changing from `'foo'` to `''` -- I will fix.

pkarman:
@jessieay good catch on QA. Fix in 35515b7

PR #752, opened 2015-10-30
Add service class for proposal creation

jessieay:
* Add public_id in class rather than callback
* Story: https://www.pivotaltracker.com/story/show/98376564

This is a refactor so QA should be that creating a new NCR Work order works as expected

### DIFF
--- a/lib/services/client_data_creator.rb
+++ b/lib/services/client_data_creator.rb
@@ -28,11 +28,7 @@ class ClientDataCreator
   end
 
   def public_id
-    if client_data.respond_to?(:fiscal_year)
-      "FY" + client_data.fiscal_year.to_s.rjust(2, "0") + "-#{proposal.id}"
-    else
-      "##{proposal.id}"
-    end
+    client_data.public_identifier
   end
 
   def proposal
--- a/lib/services/client_data_creator.rb
+++ b/lib/services/client_data_creator.rb
@@ -1,0 +1,41 @@
+class ClientDataCreator
+  def initialize(client_data, user, attachment_data = [])
+    @client_data = client_data
+    @user = user
+    @attachment_data = attachment_data
+  end
+
+  def run
+    client_data.build_proposal(flow: 'linear', requester: user)
+    client_data.save
+    create_attachments
+    add_public_id
+    proposal
+  end
+
+  private
+
+  attr_reader :client_data, :user, :attachment_data
+
+  def create_attachments
+    attachment_data.each do |file|
+      proposal.attachments.create(user: user, file: file)
+    end
+  end
+
+  def add_public_id
+    proposal.update(public_id: public_id)
+  end
+
+  def public_id
+    if client_data.respond_to?(:fiscal_year)
+      "FY" + client_data.fiscal_year.to_s.rjust(2, "0") + "-#{proposal.id}"
+    else
+      "##{proposal.id}"
+    end
+  end
+
+  def proposal
+    @proposal ||= client_data.proposal
+  end
+end
--- a/app/controllers/proposals_controller.rb
+++ b/app/controllers/proposals_controller.rb
@@ -59,7 +59,7 @@ class ProposalsController < ApplicationController
     end
 
     approval.approve!
-    flash[:success] = "You have approved #{proposal.public_identifier}."
+    flash[:success] = "You have approved #{proposal.public_id}."
     redirect_to proposal
   end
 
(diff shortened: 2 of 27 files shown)


lib/services/client_data_creator.rb
```diff
@@ -0,0 +1,41 @@
+class ClientDataCreator
+  def initialize(client_data, user, attachment_data = [])
+    @client_data = client_data
+    @user = user
+    @attachment_data = attachment_data
+  end
+
+  def run
+    client_data.build_proposal(flow: 'linear', requester: user)
```
jessieay: this is also being done in the controller because there is a validation on client_data for existence of a proposal. But I also like doing this here because otherwise in my test setup I need to build the proposal for the client data object, which seems like a smell. What do you think?

lib/services/client_data_creator.rb
```diff
@@ -0,0 +1,41 @@
...
+    proposal.update(public_id: public_id)
+  end
+
+  def public_id
+    if client_data.respond_to?(:fiscal_year)
+      "FY" + client_data.fiscal_year.to_s.rjust(2, "0") + "-#{proposal.id}"
+    else
+      "##{proposal.id}"
+    end
+  end
```
pkarman: I like the idea of encapsulating the proposal creation like this. This `public_id` method is going to get hairy though as we scale out to multiple clients.

Maybe instead delegate to a `build_public_id` method that each `client_data` class could (optionally) implement?

lib/services/client_data_creator.rb
pkarman: I don't see where in which controller "this" is done? Maybe some URLs and/or file+linenumbers would help me.

pkarman:
There's a chicken-and-egg problem with proposals and client_data creation inherent in the app that you're trying to solve here, and this seems like a decent way to approach it. I want to make sure I understand all of what's going on, though, since this PR seems bigger than the scope of what I read in the story.

The original architecture seems to have been that `Proposal.public_identifier` was the public (!) interface for the Proposal's GUID, and that `public_id` was the internal attribute where that value was cached after it was created. Either the Proposal or the client_data class could create the `public_identifier`. That was an earlier attempt at trying to provide a uniform interface to a Request (`Proposal` + `client_data`).

What you seem to be doing here is eliminating the difference between the public interface and the cached attribute, and just making the attribute (`public_id`) public and delegating all the setup and construction to the new service class.

That seems fine. I just wanted to make sure I understand the architectural shift, since it feels more seismic than what the story suggested it might be.

Am I close?

lib/services/client_data_creator.rb
jessieay: happens here: https://github.com/18F/C2/blob/master/app/controllers/use_case_controller.rb#L73 (this controller is shared for NCR work orders and 18F procurements)

lib/services/client_data_creator.rb
jessieay: good idea, will add method back to each client_data class

jessieay:
You are close in your assessment.

- We used to use `proposal.public_identifier` everywhere, which looked at the `public_identifier` method in the `client_data` classes (not at a database field) -- this seemed weird to me, I am not sure why we did this because this method was returning the same value as `public_id` on the proposal
- We used to set the `proposal.public_id` in a callback in the `proposal` class. this was an `after_create` because it used the `public_identifier` methods from the `client_data` classes, which referenced the `proposal.id`
- This PR moves the logic out of a callback and into a service class. While I was making this change, I realized the weirdness of setting `public_id` in a callback but using `public_identifier` all over the place when it was the same thing...so I changed that. I think on reason that was happening was that the system started with the method but then we wanted to search by this field so it started being stored in the  DB. If we weren't using it in search, I would have kept it as a method, but since we are using it I decided to keep in the DB and eliminate the method.

lib/services/client_data_creator.rb
jessieay: moved at b3cdc6a7a4b0757c224bd8f9bb897d8528968322

lib/services/client_data_creator.rb
pkarman: ok, understood.

Does it make any sense to check first for proposal?
```
unless client_data.proposal?
  client_data.build_proposal(flow: 'linear', requester: user)
end
```

?

pkarman:
This looks great. I will QA.

pkarman:
looks good on c2-dev

PR #755, opened 2015-11-02
Move `form` to partial

jessieay:
* Use default view partials rather than telling controller where to look


pkarman:
looks good!

PR #756, opened 2015-11-02
[WIP] Refactor: Remove `UseCaseController`

jessieay:
* Sharing logic is more DRY but controllers were different enough that
  using a shared controller  was causing a lot of confusion / workarounds
* Part I of fixing https://www.pivotaltracker.com/story/show/106987202


pkarman:
The code looks fine. I like that you've customized method names to match the use case (procurement vs work_order).

I feel some regret that our use cases have diverged so much after only 2 of them that we can't share more code, especially since we anticipating adding more use cases. I'd be interested to see if there was some mixin or other shared-code approach to use at all, or if we are really going to need to copy/paste a lot when adding a new use case.

jessieay:
@pkarman agreed it would be nice if they shared more. 

I felt like here we were putting in a lot of effort to make the logic shared that it was making things more complicated than they needed to be -- it would be one thing if we called `super` for each controller action, but it seems like we were doing that plus adding logic before and after `super`. 

Any ideas on how we could share more logic without going entirely back to the whole system of forcing things to have similar behavior through using a shared controller?

pkarman:
I typically approach abstract class construction cases like this by trying to create `pre_*` and `post_*` "hook" methods that are intended to be overridden in a subclass.

So for example, in `ncr/work_orders_controller.rb`, instead of this:
```ruby
    def edit
      if self.proposal.approved?
        flash[:warning] = "You are about to modify a fully approved request. Changes will be logged and sent to approvers but this request will not require re-approval."
      end

      super
    end
```

I might have:
```ruby
def pre_edit_hook
  if self.proposal.approved?
    flash[:warning] = "You are about to modify a fully approved request. Changes will be logged and sent to approvers but this request will not require re-approval."
  end
end
```

and the `UseCaseController` would have:
```ruby
def edit
  pre_edit_hook
end

def pre_edit_hook
end
```

Maybe that's too overly engineered? Too complicated? If anything I find the current layout not abstracted *enough* rather than overly abstracted.

jessieay:
closing for now as I am going to take an approach discussed on g hangout with P Karman

afeld:
Very much agree with the hook approach :+1:

PR #757, opened 2015-11-02
[Delivers #105675934] Send annual NCR report weekly

jessieay:
* Story: https://www.pivotaltracker.com/story/show/105675934
* Report includes all approved work orders for current fiscal year
* Send on Monday at 3:30am
* Use clockwork to send to all budget recipients
* Switch annual report to be for approved work orders per https://www.pivotaltracker.com/story/show/105300436


jessieay:
I tested that running `ClockTasks.send_annual_ncr_budget_report` sends the correct recipients a report for FY 2016 locally...is there something else I should to to test this? We don't tests for the other clockwork logic as far as I can see, but always makes me a little nervous to create something like this and leave it untested...

pkarman:
Why are we calling it an annual report if we are sending it weekly? Is it year-to-date data?

I don't think we need to test that ClockTasks work per se, as long as the report itself is tested (which it appears to be).

jessieay:
@pkarman it sends info for the current fiscal year but on a weekly basis; I can rename, just using the same logic as the annual report mailer we already had so didn't really think abut it. What about `weekly_fiscal_year_report` ?

pkarman:
`weekly_fiscal_year_report` is perfect.

jessieay:
@pkarman updated at c8aa5c2a2e60ef4fd3d5d69f1c54696f76dcac22

also removed rake task for previous use case since now we are sending this weekly rather than on-demand .

pkarman:
great naming choices. +1

I will QA.

pkarman:
No errors or regressions on c2-dev. 

Since there's no 2016 data on c2-dev, I also ran on c2-prod via cf-ssh only. Through that, I found a typo existed in the current master code, so I fixed that in 2486005eb086dc3b69b9f2d96f735f591b66467e and QA was successful on c2-prod.

PR #758, opened 2015-11-02
fix DOM structure where then are zero completed proposals

pkarman:
A small HTML nit: fix the empty list DOM structure to be consistent for all tables.


jessieay:
looks good! screenshots would be helpful for this type of change in future :ship:

pkarman:
div-close-before
div-close-after

jessieay:
yay screenshots - looks much better! what do you say we merge this puppy without QA?

pkarman:
:+1:

jessieay:
Feels so good to be bad - merging!

PR #759, opened 2015-11-03
Revert "Do not run seeds before each test run"

pkarman:
Reverts 18F/C2#746

This works great on a clean db, but on an existing db, we need to look up the User with the appropriate role, not rely on the env vars.

### DIFF
--- a/app/models/ncr/work_order.rb
+++ b/app/models/ncr/work_order.rb
@@ -10,11 +10,6 @@ module Ncr
   BUILDING_NUMBERS = YAML.load_file("#{Rails.root}/config/data/ncr/building_numbers.yml")
 
   class WorkOrder < ActiveRecord::Base
-      NCR_BA61_TIER1_BUDGET_APPROVER_MAILBOX = ENV['NCR_BA61_TIER1_BUDGET_MAILBOX'] || 'communicart.budget.approver+ba61@gmail.com'
-      NCR_BA61_TIER2_BUDGET_APPROVER_MAILBOX = ENV['NCR_BA61_TIER2_BUDGET_MAILBOX'] || 'communicart.ofm.approver@gmail.com'
-      NCR_BA80_BUDGET_APPROVER_MAILBOX = ENV['NCR_BA80_BUDGET_MAILBOX'] || 'communicart.budget.approver+ba80@gmail.com'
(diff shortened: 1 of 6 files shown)


jessieay:
What exactly was this breaking? Would be great to have a test case for that in the code so it does not get broken in the future

app/models/ncr/work_order.rb
```diff
@@ -10,11 +10,6 @@ def self.table_name_prefix
   BUILDING_NUMBERS = YAML.load_file("#{Rails.root}/config/data/ncr/building_numbers.yml")
 
   class WorkOrder < ActiveRecord::Base
-      NCR_BA61_TIER1_BUDGET_APPROVER_MAILBOX = ENV['NCR_BA61_TIER1_BUDGET_MAILBOX'] || 'communicart.budget.approver+ba61@gmail.com'
-      NCR_BA61_TIER2_BUDGET_APPROVER_MAILBOX = ENV['NCR_BA61_TIER2_BUDGET_MAILBOX'] || 'communicart.ofm.approver@gmail.com'
-      NCR_BA80_BUDGET_APPROVER_MAILBOX = ENV['NCR_BA80_BUDGET_MAILBOX'] || 'communicart.budget.approver+ba80@gmail.com'
-      OOL_BA80_BUDGET_APPROVER_MAILBOX = ENV['NCR_OOL_BA80_BUDGET_MAILBOX'] || 'communicart.budget.approver+ool_ba80@gmail.com'
-
```
pkarman: These env vars are not set in any CF environment, on purpose, because we are moving away from using env vars to store role-based information and instead using the database. So in a CF environment, the wrong emails get used (the defaults, rather than what is in the db).

jessieay: should we remove references to these env vars, then?

pkarman:
The only reference to those env vars is in the conversion code, which was originally intended to be run only once (as in a migration) but because seeds are run on each rspec run and on deploy, we consolidated that migration code into a single place in `lib/roles_conversion.rb`.

Bottom line: the Users with those email addresses must exist in a functioning baseline database. SInce all CF envs already have functioning baseline databases, we must use those existing Users and ignore env variables. On a new db initiation (as with rspec or any new dev env) the Users must be created in the database, and we use the env vars as the values. 

We could simply hardcode the values rather than using the env vars, but that's why the env var references exist as they do.

PR #760, opened 2015-11-03
[Delivers #107239306] improve OOO comment coverage

pkarman:
Story: https://www.pivotaltracker.com/story/show/107239306

This makes the reply-to address of each email include the proposal `public_id` so that OOO auto-replies are unique-per-request rather than unique-per-requester.

To test, trigger any email and check that the reply-to address contains the request public_id.

### DIFF
--- a/app/mailers/concerns/proposal_conversation_threading.rb
+++ b/app/mailers/concerns/proposal_conversation_threading.rb
@@ -59,10 +59,13 @@ module ProposalConversationThreading
     self.assign_threading_headers(proposal)
     subject = ProposalConversationThreading.subject(proposal)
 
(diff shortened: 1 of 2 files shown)


jessieay:
Looks neat-o. :ship:

jessieay:
QA ing this now

jessieay:
looks good on c2-dev! 
screen shot 2015-11-04 at 11 51 02 am

merging

quick Q: does this mean we are going to be searching proposals on `public_id` ? If so, we should add an index on that field in the db!

pkarman:
No action is taken on the public_id in the email address. It's just there for the sake of the email sender (Gmail), so that it sees the address as unique.

jessieay:
@pkarman :+1:

PR #762, opened 2015-11-03
[Delivers #89851836] drop Properties table

pkarman:
Story: https://www.pivotaltracker.com/story/show/89851836

Final clean up for deprecated Properties table. Related code was removed several months ago in dd5b0b5c56b28b44fa5d5effed7967f44294dde1

DB snapshots since then have backed up data.

### DIFF
--- a/db/migrate/20151103223107_drop_properties.rb
+++ b/db/migrate/20151103223107_drop_properties.rb
@@ -1,0 +1,5 @@
+class DropProperties < ActiveRecord::Migration
+  def change
+    drop_table :properties
+  end
+end
--- a/db/migrate/20151103223107_drop_properties.rb
+++ b/db/migrate/20151103223107_drop_properties.rb
@@ -1,5 +1,9 @@
 class DropProperties < ActiveRecord::Migration
-  def change
+  def up
     drop_table :properties
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
   end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20151028234110) do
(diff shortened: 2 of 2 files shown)


jessieay:
YEAH love this :heart:

db/migrate/20151103223107_drop_properties.rb
```diff
@@ -0,0 +1,5 @@
+class DropProperties < ActiveRecord::Migration
+  def change
```
jessieay: should we mark this migration is irreverisble? http://jacopretorius.net/2014/07/creating-irreversible-migrations-in-rails.html

jessieay:
One question on migration, otherwise looks :ship:

db/migrate/20151103223107_drop_properties.rb
pkarman: There's no precedent for doing that in our db/migrations so far, so I don't think we need it.

jessieay: should we make it reversible, then? I can't think of reasons why we would want to rollback, but I know it's always a good idea to make migrations reversible...

pkarman: This is a clean up step, right? I was following the previous precedents in db/migrate, which just `drop_table` and move on. We don't want to roll this back, ever, and couldn't if we wanted to, since the data is gone once the table is dropped. Restoring an empty table isn't really a reversible action.

jessieay: Yeah understood that we don't want to restore the table. I guess the reason I originally suggested marking is irreversible is to reveal intent. See: http://maxwellholder.com/blog/rails-migration-checklist

I can see that there's no precedent for doing that, but do you think we should start? In most cases, reversible migrations are the way to go. This is obviously an exception so perhaps we should indicate it as such?

pkarman:
As you wish: 548d34b

jessieay:
Let's :sheep: :it:

PR #763, opened 2015-11-03
Add service object for WorkOrder data normalization

jessieay:
* Add services to `app` because that is autoloaded and we want this autoloaded
* Remove normalization from callback because callbacks are bad news


pkarman:
I like this. Very clean.

pkarman:
QA ok c2-staging

jessieay:
@pkarman :tophat:

PR #764, opened 2015-11-03
[Delivers #106987202] Do not show system approver emails as approver …

jessieay:
…options

* Cannot have two approvals for some approver bc of DB constraint, so
  was returning a 500
* Adding a uniqueness validation did not improve UX because approval is
  created after work order is created so validation came to late
* Short term fix is to not show these user emails as options
* Longer term fix is probably to wrap creation of all objects on a
  transaction per http://blog.endpoint.com/2011/10/rails-controllers-and-transactions.html
* Story: https://www.pivotaltracker.com/story/show/106987202

### DIFF
--- a/spec/factories/approvals.rb
+++ b/spec/factories/approvals.rb
@@ -3,8 +3,11 @@ FactoryGirl.define do
     proposal
     user
     status 'pending'
-  end
 
-  factory :parallel_approval, class: Approvals::Parallel do
+    factory :parallel_approval, class: Approvals::Parallel do
+    end
+
+    factory :individual_approval, class: Approvals::Individual do
+    end
   end
 end
--- a/spec/features/ncr/work_orders/create_spec.rb
+++ b/spec/features/ncr/work_orders/create_spec.rb
@@ -107,13 +107,12 @@ feature 'Creating an NCR work order' do
       visit "/ncr/work_orders/new"
       within(".ncr_work_order_approving_official_email") do
         find(".selectize-control").click
+        expect(page).to have_content(approving_official.email_address)
+        expect(page).not_to have_content(Ncr::WorkOrder.ba61_tier1_budget_mailbox)
+        expect(page).not_to have_content(Ncr::WorkOrder.ba61_tier2_budget_mailbox)
+        expect(page).not_to have_content(Ncr::WorkOrder.ba80_budget_mailbox)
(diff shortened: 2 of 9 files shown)


jessieay:
Please check out the PR description re: why I am not 100% pleased with this solution. I am open to doing it The Right Way now, but feels like the setup approvals stuff is too big a beast to tackle as part of this (but I can easily be convinced otherwise!)

spec/features/ncr/work_orders/create_spec.rb
```diff
@@ -104,6 +100,22 @@
...
+      visit "/ncr/work_orders/new"
+      within(".ncr_work_order_approving_official_email") do
+        find(".selectize-control").click
+      end
+
+      expect(page).to have_content(approving_official.email_address)
+      expect(page).not_to have_content(Ncr::WorkOrder.ba61_tier1_budget_mailbox)
+      expect(page).not_to have_content(Ncr::WorkOrder.ba61_tier2_budget_mailbox)
+      expect(page).not_to have_content(Ncr::WorkOrder.ba80_budget_mailbox)
+      expect(page).not_to have_content(Ncr::WorkOrder.ool_ba80_budget_mailbox)
```
yozlet: Can these assertions focus on the select control itself rather than the whole page?

yozlet:
Having the feature test focus on the select box would be an improvement, but either way :+1:

spec/features/ncr/work_orders/create_spec.rb
jessieay: sure - great idea, will update

jessieay:
@yozlet updated at 58624f5b7c09327c2ae2f831b13ece376fed8a63 -- take another peek? :pray:

yozlet:
@jessieay Yep! (Actually took me a moment to spot the `within` block, but it's fine)

I'll QA it now.

spec/factories/approvals.rb
```diff
@@ -3,8 +3,11 @@
...
     user
     status 'pending'
-  end
 
-  factory :parallel_approval, class: Approvals::Parallel do
+    factory :parallel_approval, class: Approvals::Parallel do
+    end
+
+    factory :individual_approval, class: Approvals::Individual do
+    end
```
yozlet: This will need to be tweaked if you merge after our Purchase Step work merges.

yozlet:
QA'ed it, looks good! :sheep:

PR #765, opened 2015-11-04
Move `record_changes` to service object

jessieay:
* Call explicitly in controller rather than in callback


app/controllers/ncr/work_orders_controller.rb
```diff
@@ -21,9 +21,26 @@ def edit
...
+      @model_instance.assign_attributes(permitted_params)
       @model_instance.modifier = current_user
 
-      super
+      @model_changing = false
+      @model_instance.validate
+
+      if self.errors.empty?
+        if self.attribute_changes?
+          ProposalUpdateRecorder.new(@model_instance).run
```
jessieay: it's situations like these when the `UseCaseController` seems to be more of a barrier than a help...that is, unless we want to have this same behavior for `Gsa18f::Procurement` as well...(behavior being creating a comment for updates to a proposal). cc @phirefly

jessieay:
going to re-paste earlier comment here: 

It's situations like these when the UseCaseController seems to be more of a barrier than a help...that is, unless we want to have this same behavior for `Gsa18f::Procurement` as well...(behavior being creating a comment for updates to a proposal). cc @phirefly

app/controllers/ncr/work_orders_controller.rb
```diff
@@ -21,18 +21,32 @@ def edit
...
+      if attribute_changes?
+        ProposalUpdateRecorder.new(@model_instance).run
+        @model_instance.save
         @model_instance.setup_approvals_and_observers
         @model_instance.email_approvers
+        flash[:success] = "Successfully modified!"
+      else
+        flash[:error] = "No changes were made to the request"
       end
     end
```
afeld: Not sure that I agree with the changes in this file, since you're copying in the functionality from the `UseCaseController`... Would you mind if we decouple the `ProposalUpdateRecorder` change from the removal of the `super` call in the `update` action above and do them in separate PRs?

/cc https://github.com/18F/C2/pull/756

afeld:
Big :100: :+1: to moving the `update_comment` functionality from the `WorkOrder` class, though I wouldn't spend _too_ much more time on it, since it should probably be removed altogether.

https://www.pivotaltracker.com/story/show/105014352

> unless we want to have this same behavior for `Gsa18f::Procurement` as well

I actually think it was an oversight that it didn't apply to all use case models in the first place. In other words, yes, I think we do! Up to you about whether we care to do it in this PR or not, though switching to using Paper Trail (see story above) would give us that behavior anyway.

app/controllers/ncr/work_orders_controller.rb
jessieay: What I am finding as I make various updates is that there is a lot of different functionality required in the individual client_data controllers vs `UseCaseController` 

eg: here, we want to call `ProposalUpdateRecorder.new(@model_instance).run` only if there are no errors and there are changes for an `Ncr::WorkOrder`...so I am not sure how to do that for this controller only while sharing functionality with the `update` method in `UseCaseController`...suggestions?

jessieay:
@afeld Interesting, so you think I should close this without merging if we are moving away from this method of leaving comments? That story you linked to is in icebox, so not sure when it will be prioritized. I just made this change because I noticed the gnarly method in `WorkOrder` and the comment there said the next step was to move out of that class...

app/controllers/ncr/work_orders_controller.rb
afeld: Pretty sure that's what `@model_changing` corresponds to, no? If it makes it easier to enable the functionality for all use case models, feel free to do that.

afeld:
Totally fine with merging this work (once we work out the discussion in the inline comment) – I just meant I wouldn't bother spending _too_ much more time on that part of the code.

app/controllers/ncr/work_orders_controller.rb
jessieay: ah yeah -- I found that instance var to be a bit confusing because you usually only see instance vars for populating the view, not for passing around logic in controllers. will see if I can find another way around this

jessieay: ok @afeld updated at d1508741f9b7c2dcd9bb43f68e4f2b28ad96ccd5 -- what say ye?

afeld:
QA'd (after merging – whoops!) on dev.

PR #766, opened 2015-11-04
change concessions org code

pkarman:
See https://18f.uservoice.com/admin/tickets/35/


jessieay:
should we also add a migration or rake task for updating org code for existing work orders?

pkarman:
premature -- the story is still being fleshed out on that ticket.

PR #768, opened 2015-11-05
Use "client_data" instead of "client"

jessieay:
* Part of https://www.pivotaltracker.com/n/projects/1149728/stories/101710542
* But certainly not all of it


pkarman:
Now that I see the diffs here, I can understand why it was called `client` rather than `client_data`.

As I read the before/after:

* `client` refers to the customer in general
* `client_data` refers to the proposal values specific to the customer
* `client_slug` refers to the short string used to identify the customer

Given those definitions, what you are calling `client_data_prefix` is the same as the `client_slug` which introduces another inconsistency of naming.

In other words, there seem to be legitimate reasons to use the word `client` in some places, where `client_data` is less appropriate as a replacement.

app/helpers/client_data_helper.rb
```diff
@@ -0,0 +1,26 @@
+module ClientDataHelper
+  def client_data_partial(client_name, path, args = {})
+    client_name ||= "default"
```
pkarman: instead of `client_name` it seems like `client_slug` is more appropriate

app/helpers/client_data_helper.rb
```diff
@@ -0,0 +1,26 @@
+module ClientDataHelper
```
pkarman: `ClientHelper` actually seems more appropriate here, since these methods really have little to do with the proposal extension values themselves.

app/models/concerns/proposal_delegate.rb
```diff
@@ -40,16 +40,20 @@ module ProposalDelegate
 
     ###################################################
 
-    def self.client
```
pkarman: I would rather see `def self.client_slug` for consistency with what it returns: a short string identifying the customer namespace.

app/models/concerns/proposal_delegate.rb
```diff
@@ -40,16 +40,20 @@ module ProposalDelegate
...
+    def self.client_data_prefix
       self.to_s.deconstantize.downcase
     end
   end
 
-  def client
-    self.class.client
+  def client_data_prefix
```
pkarman: same here: `def client_slug`

app/models/proposal.rb
```diff
@@ -45,7 +45,7 @@ class Proposal < ActiveRecord::Base
   # :fields_for_display
   # :version
   # Note: clients should also implement :version
-  delegate :client, to: :client_data, allow_nil: true
+  delegate :client_data_prefix, to: :client_data, allow_nil: true
```
pkarman: same here: `client_slug` instead of `client_data_prefix`

jessieay:
@pkarman I hear what you're saying about `client` referring to the customer in general whereas "client_data" is the objects created for clients. 

One reason I found `client` confusing was that when I saw it I was wondering if it was referring to some type of library or something, eg: https://github.com/dblock/slack-ruby-client

I have similar thoughts are the use of `delegate` in this project: `delegate` makes total sense in the context of the project, but since it's also a keyword in Ruby, it can be confusing seeing it around as both an object name and a method. 

How do you think we should proceed?

pkarman:
http://peknet.com/blog/projects/two-hard-things

Unless we want to radically change our terminology away from `client` (maybe `partner`?) I think we should focus on disambiguating `client` and `client_slug` in the `ProposalDelegate` concern, and leave everything pretty much as it is.

app/models/concerns/proposal_delegate.rb
jessieay: I like that!

jessieay:
Gonna close this and re-open with a subset of these changes

PR #769, opened 2015-11-05
[Delivers #95549162] delegates redirected to approval after login

pkarman:
Story: https://www.pivotaltracker.com/story/show/95549162

This changes the ReturnTo helper to use the session to temporarily store the original request path so that the post-login callback can redirect after successful authentication.

QA: You can test this feature:

* add yourself (Y) as a delegate to an existing User (X)
* create a new proposal listing X as the approver
* using the approval email sent to X, copy the Approve button link (should contain the **cch** token param)
* using a browser *not logged in to MyUSA or the app*, paste the link. 
* login as Y (the delegate)
* on redirect back to the app, your browser should redirect and result in the proposal being approved.


app/controllers/auth_controller.rb
```diff
@@ -1,18 +1,18 @@
...
 
   def oauth_callback
-    auth = request.env['omniauth.auth']
-    return_to = self.return_to.try(:path)
+    auth = request.env["omniauth.auth"]
 
-    sign_out
-    user = User.from_oauth_hash(auth)
-    sign_in(user)
+    # fetch the redirect path (if any) before the authn sign in/out
```
jessieay: what do you think about writing a feature-level test for this general use case rather than having a comment? (test in this PR covers the delegate use-case, but was thinking about a more general situation with a redirect path)

pkarman: @jessieay good point. I hadn't considered that `return_to` would now persist for **any** request across the login. Test added in 1687353

spec/features/return_to_spec.rb
```diff
@@ -30,9 +29,22 @@
...
 
   it 'send back to main if not a valid sig (name)' do
+    login_as(create(:user))
     different_path = return_to.merge(path: 'other-key-here')
     visit query_proposals_path(return_to: different_path)
     expect(page).to have_content('Back to main portal')
     expect(page).not_to have_content(return_to[:name])
   end
+
+  it "persists the original request URL over login and redirects after" do
```
jessieay: yay! this is great

app/controllers/auth_controller.rb
jessieay: nice! test looks great. think we can remove these comments now since they are covered by tests?

pkarman: comments zapped in 22182df

jessieay:
QAing now

jessieay:
Pushed this to staging and it is giving 500 errors, Raphy and I are discussing in Slack now

jessieay:
perhaps this needs to be rebased off of master to include the most recent migrations?

jessieay:
this works! just tested on staging. Only unexpected behavior is that the error that says "please sign in" on home page is very small and easy to miss. Perhaps we could bring someone right to "sign in" page instead of showing that flash message? Seems like a separate feature, merging this for now

PR #770, opened 2015-11-05
[Delivers #107393700] /history endpoint returns 500 error

pkarman:
Story: https://www.pivotaltracker.com/story/show/107393700

This fixes a regression introduced in #751 where the `empty?` method was being called on an object w/o that method.

### DIFF
--- a/spec/decorators/hash_diff_decorator_spec.rb
+++ b/spec/decorators/hash_diff_decorator_spec.rb
@@ -24,5 +24,10 @@ describe HashDiffDecorator do
       output = HashDiffDecorator.html_for(['~', 'foo', '', 'bar'])
       expect(output).to eq("<code>foo</code> was changed from <code>[empty]</code> to <code>&quot;bar&quot;</code>")
     end
+
+    it "renders numeric events" do
+      output = HashDiffDecorator.html_for(['~', 'a_number', 456, 123])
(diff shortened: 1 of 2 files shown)


spec/decorators/hash_diff_decorator_spec.rb
```diff
@@ -24,5 +24,10 @@
       output = HashDiffDecorator.html_for(['~', 'foo', '', 'bar'])
       expect(output).to eq("<code>foo</code> was changed from <code>[empty]</code> to <code>&quot;bar&quot;</code>")
     end
+
+    it "renders numeric events" do
+      output = HashDiffDecorator.html_for(['~', 'a_number', 456, 123])
+      expect(output).to eq("<code>a_number</code> was changed from <code>456.00</code> to <code>123.00</code>")
```
jessieay: is this test for the case above? seems to cover a numeric rather than empty val?

pkarman: There is already a test for the empty value. There wasn't previously a test for anything numeric, which is why the regression slipped in. This rectifies that problem.

jessieay:
QA instructions? (not sure of  the conditions that led to a 500 on history page)

pkarman:
QA: You should be able to view any history page without error. The 500 was very consistent on master, so any history URL should work. If it helps to before/after, you can deploy master, trigger the 500 error, then deploy this branch and reload the page.

PR #771, opened 2015-11-05
[Delivers #107501786] hashdiff numerics

pkarman:
Story: https://www.pivotaltracker.com/story/show/107501786

Treat `Fixnum` numeric objects like integers, not floats.

You should see this in action on the `/proposals/:id/history` endpoint. Id values and other integers should display as `123` (integers) not as `123.00` (floats).

### DIFF
--- a/app/decorators/hash_diff_decorator/base.rb
+++ b/app/decorators/hash_diff_decorator/base.rb
@@ -19,14 +19,20 @@ module HashDiffDecorator
     def diff_val(val)
       if val.nil?
         "[nil]"
-      elsif val.is_a?(Fixnum)
-        val.to_s
       elsif val.is_a?(Numeric)
-        format('%.2f', val)
+        diff_numeric(val)
       elsif val.empty?
         "[empty]"
       else
         val.inspect
+      end
+    end
+
+    def diff_numeric(val)
+      if val.is_a?(Fixnum)
+        val.to_s
+      elsif val.is_a?(Numeric)
+        format("%.2f", val)
       end
     end
   end
--- a/app/decorators/hash_diff_decorator/base.rb
+++ b/app/decorators/hash_diff_decorator/base.rb
@@ -19,6 +19,8 @@ module HashDiffDecorator
     def diff_val(val)
       if val.nil?
         "[nil]"
+      elsif val.is_a?(Fixnum)
+        val.to_s
       elsif val.is_a?(Numeric)
         format('%.2f', val)
       elsif val.empty?
--- a/app/decorators/hash_diff_decorator/base.rb
+++ b/app/decorators/hash_diff_decorator/base.rb
@@ -21,7 +21,7 @@ module HashDiffDecorator
         "[nil]"
       elsif val.is_a?(Numeric)
         diff_numeric(val)
-      elsif val.empty?
+      elsif val.try(:empty?)
         "[empty]"
       else
         val.inspect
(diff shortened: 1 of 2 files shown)


app/decorators/hash_diff_decorator/base.rb
```diff
@@ -19,13 +19,21 @@ def to_html
...
-        format('%.2f', val)
       else
         val.inspect
       end
     end
+
+    def diff_numeric(val)
+      if val.is_a?(Fixnum)
+        val.to_s
+      elsif val.is_a?(Numeric)
```
jessieay: what this be a float? If so, what about just checking for that? (just looking up `Numeric` in Ruby docs since I don't see this class used too often)

app/decorators/hash_diff_decorator/base.rb
```diff
@@ -19,13 +19,21 @@ def to_html
...
+        format("%.2f", val)
```
jessieay: is this is always a float, perhaps just call `.round(2)` ? Not sure if that would be better, just throwing it out there

jessieay:
2 tiny questions, code looks good to :sheep:

app/decorators/hash_diff_decorator/base.rb
pkarman: I wasn't sure about the original motivation for using `Numeric` so I preserved that as the default, and just checked explicitly for the (more common?) case of `Fixnum` which includes all `id` values.

I'm inclined to leave the current behavior alone until we get a specific example of where it does not work (as was the case here).

app/decorators/hash_diff_decorator/base.rb
pkarman: I believe `round` returns another `Numeric` (integer or float) and doesn't force 2 digits pass the decimal, which isn't what we want. We want a String always formatted like currency. 

You might expect `round(2).to_s` to work, but:
```
2.2.3 :001 > n = 123.1
 => 123.1 
2.2.3 :002 > n.round(2).to_s
 => "123.1"
```

jessieay:
QAing now

jessieay:
I am not seeing any integers or floats on history pages -- further tips on QA setup?

screen shot 2015-11-05 at 2 02 45 pm

jessieay:
just changed the cost per unit on https://c2-staging.18f.gov/proposals/448 and am getting a 500 in the history page

pkarman:
You might need to deploy #770 and this PR as a single branch to QA both together.

jessieay:
I combined and pushed to staging ,still seeing a 500 at https://c2-staging.18f.gov/proposals/448/history

jessieay:
combined branch is here https://github.com/18F/C2/compare/jy-combined-history-branches?expand=1 -- think I did it right?

pkarman:
you did it right.

here's the logs
```
2015-11-05T21:20:38.17-0600 [App/0]      OUT 03:20:38 web.1    | Completed 500 Internal Server Error in 358ms (ActiveRecord: 78.5ms)
2015-11-05T21:20:38.18-0600 [App/0]      OUT 03:20:38 web.1    | 
2015-11-05T21:20:38.18-0600 [App/0]      OUT 03:20:38 web.1    | ActionView::Template::Error (undefined method `empty?' for 2015-10-22 07:00:00 UTC:Time):
2015-11-05T21:20:38.18-0600 [App/0]      OUT 03:20:38 web.1    |     1: = C2VersionDecorator.new(object).to_html
2015-11-05T21:20:38.18-0600 [App/0]      OUT 03:20:38 web.1    |   app/decorators/hash_diff_decorator/base.rb:24:in `diff_val'
```

looks like more objects that do not know how to call `.empty?`

I will refactor to test for that method first.

pkarman:
@jessieay should be fixed in ed01ea0 -- I deployed to stage and tested your proposal there.

jessieay:
woo hoo, it works! merging

PR #772, opened 2015-11-05
[Delivers #106050582] avoid duplicate observer

pkarman:
Story: https://www.pivotaltracker.com/story/show/106050582

Related story: https://www.pivotaltracker.com/story/show/107438934

This PR does 2 things:

1. only users with matching client_slug, and who are not already observers, are available in the Observer dropdown list
1. if a duplicate Observer is attempted, a graceful error response is given rather than a 5xx error

You can test (1) by trying to add the same user multiple times as an Observer.

You can test (2) by opening 2 browser tabs for the same page, selecting the same observer for both tabs, submitting the first tab (should be successful), submitting the second tab (should give nice error message).


app/controllers/observations_controller.rb
```diff
@@ -6,7 +6,7 @@ class ObservationsController < ApplicationController
 
   def create
     obs = @proposal.add_observer(observer_email, current_user, params[:observation][:reason])
-    flash[:success] = "#{obs.user.full_name} has been added as an observer"
+    prep_create_response_msg(obs)
```
jessieay: what about `observer` as a var name here (and below) rather than `obs` ? -- would be clearer, imo!

app/models/proposal.rb
```diff
@@ -148,6 +148,14 @@ def existing_observation_for(user)
     observations.find_by(user: user)
   end
 
+  def eligible_observers
+    if observations.count > 0
+      User.where(client_slug: client).where('id not in (?)', observations.pluck('user_id'))
```
jessieay: couldn't we run this query whether there are observations or not?

spec/features/client_slug_authz_spec.rb
```diff
@@ -69,9 +69,7 @@
     proposal_path = current_path
     visit proposal_path
     expect(page.status_code).to eq(200)
-    add_as_observer(gsa_user)
-    expect(page.status_code).to eq(403)
-    expect(page).to have_content("You are not allowed to add observers to that proposal.")
+    expect_to_not_select('observation_user_email_address', gsa_user.email_address)
```
jessieay: perhaps `expect_to_not_see_select_option` ? when I first saw this, I thought it was for whether someone was selecting something or not, which was a big confusing given that it's an expectation.

spec/features/observer_spec.rb
```diff
@@ -3,7 +3,7 @@
     expect(CommunicartMailer).to receive_message_chain(:on_observer_added, :deliver_later)
 
     work_order = create(:ncr_work_order)
-    observer = create(:user)
+    observer = create(:user, client_slug: 'ncr')
```
jessieay: maybe we should create a factory called `ncr_user` or add `ncr` as a trait on `user` that sets the client_slug so we don't have to set this in various tests?

spec/features/observer_spec.rb
```diff
@@ -39,7 +39,7 @@
   it "allows a user to add a reason when adding an observer" do
     reason = "is the archbishop of banterbury"
     proposal = create(:proposal)
-    observer = create(:user)
+    observer = create(:user, client_slug: nil)
```
jessieay: why are we setting `client_slug` to nil in these tests?

spec/models/ncr/work_order_spec.rb
```diff
@@ -167,6 +167,13 @@
       user = create(:user, client_slug: 'ncr')
       expect(wo.slug_matches?(user)).to eq(true)
     end
+
+    it "identifies eligible observers" do
```
jessieay: should we have a similar test for gsa18f procurements?

spec/features/observer_spec.rb
pkarman: I could have clarified with a comment in the test... ;)

The reason is that in order for `Proposal.client` to match, both values must be null. Since the Proposal has no attached ClientData, its value is null. So User must use explicitly identical value.

spec/features/observer_spec.rb
pkarman: I kinda like the explicit attribute-setting, so i don't have to guess what other magic something like `:ncr` might be applying.

spec/features/observer_spec.rb
jessieay: oh I see, that's interesting. I think that updating the description of the below test from "identifies eligible observers"  to "identifies eligible observers based on the client slug" would be :+1: 

One reason it feels a little unexpected that there are so many records here with a nil client_slug is that we don't have actual users creating proposals / adding observers who have a `client_slug` of nil, right?

pkarman: that's probably correct. there's lots of tests for `Proposal` w/ no client_data extension. That's good, imo, because its core functionality that applies to all clients.

fwiw the User factory does this:
```
client_slug { ENV['CLIENT_SLUG_DEFAULT'] || 'ncr' }
```

jessieay: ah!! let's get rid of that (not necessarily now)! Should be set explicitly, whether in specs or through trait etc

app/models/proposal.rb
jessieay: ping ping

spec/features/client_slug_authz_spec.rb
jessieay: thoughts?

jessieay:
hey @pkarman I assigned to you because there were a few questions I left before that weren't addressed...

app/models/proposal.rb
pkarman: no, we can't. If there are zero observations, the generated SQL looks like:
```sql
select * from users where client_slug="foo" and id not in (null)
```
which will fail.

spec/features/client_slug_authz_spec.rb
pkarman: maybe `expect_to_not_find_amongst_select_tag_options` ?

jessieay: sure, sounds good

spec/features/observer_spec.rb
pkarman: I opened https://www.pivotaltracker.com/story/show/107603900

jessieay:
Code looks good to :ship: -- will QA after my meeting!

jessieay:
QA'd - looks good. Added a card for improving the UX when a duplicate email is added: https://www.pivotaltracker.com/story/show/107744544

Also, the flash error message is still super hard to see. I know there is a card for that somewhere in the icebox.

Anyway, new functionality is :bomb: - MERGING

PR #773, opened 2015-11-05
[Delivers #107421592] Use HAML instead of markdown in view partial

jessieay:
To QA: visit proposals index and see that the "status" column is HTML, not non-interpreted markdown.

Story: https://www.pivotaltracker.com/story/show/107421592


pkarman:
looks great. :sheep: :it:

PR #775, opened 2015-11-06
[Delivers #104744396] allow Observer to delete themselves from a Proposal

pkarman:
Story: https://www.pivotaltracker.com/story/show/104744396

Enables **Delete** button for Observers to remove themselves from a Proposal.

Testing:

1. Login as User A
1. Create Proposal
1. Add User B as an Observer
1. Login as User B
1. Navigate to the Proposal
1. Click Delete next to User B's name
1. Verify success message says "Deleted Observation for {Proposal.public_id}"

### DIFF
--- a/app/controllers/observations_controller.rb
+++ b/app/controllers/observations_controller.rb
@@ -13,7 +13,7 @@ class ObservationsController < ApplicationController
   def destroy
     proposal = observation.proposal
     if current_user == observation.user
-      redirect_path = "/proposals"
+      redirect_path = proposals_path
     else
       redirect_path = proposal_path(proposal)
     end
--- a/app/controllers/observations_controller.rb
+++ b/app/controllers/observations_controller.rb
@@ -18,7 +18,7 @@ class ObservationsController < ApplicationController
       redirect_path = proposal_path(proposal)
     end
     observation.destroy
-    flash[:success] = "Deleted Observation for #{proposal.public_id}"
+    flash[:success] = "Removed Observation for #{proposal.public_id}"
     redirect_to redirect_path
   end
 
--- a/app/controllers/observations_controller.rb
+++ b/app/controllers/observations_controller.rb
@@ -11,9 +11,15 @@ class ObservationsController < ApplicationController
   end
 
   def destroy
-    self.observation.destroy
-    flash[:success] = "Deleted Observation"
-    redirect_to proposal_path(self.observation.proposal_id)
+    proposal = observation.proposal
+    if current_user == observation.user
+      redirect_path = "/proposals"
+    else
+      redirect_path = proposal_path(proposal)
+    end
+    observation.destroy
+    flash[:success] = "Deleted Observation for #{proposal.public_id}"
+    redirect_to redirect_path
   end
 
   protected
--- a/app/policies/observation_policy.rb
+++ b/app/policies/observation_policy.rb
@@ -11,7 +11,7 @@ class ObservationPolicy
   end
 
   def can_destroy!
-    self.can_show_proposal!
+    self.can_show_proposal! || user_is_observer!
   end
 
   protected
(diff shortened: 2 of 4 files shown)


app/controllers/observations_controller.rb
```diff
@@ -11,9 +11,15 @@ def create
   end
 
   def destroy
-    self.observation.destroy
-    flash[:success] = "Deleted Observation"
-    redirect_to proposal_path(self.observation.proposal_id)
+    proposal = observation.proposal
+    if current_user == observation.user
+      redirect_path = "/proposals"
```
jessieay: used named path (`proposals_path`) instead?

app/policies/observation_policy.rb
```diff
@@ -20,4 +20,8 @@ def can_show_proposal!
     policy = PolicyFinder.policy_for(@user, @observation.proposal)
     policy.can_show!
   end
+
+  def user_is_observer!
```
jessieay: should this be a `?` method rather than `!` -- usually expect to see `?` for methods that return true/false and `!` for methods that modify the objects they are called on

spec/features/observer_spec.rb
```diff
@@ -77,6 +77,22 @@
     expect(submit_button).to_not be_disabled
   end
 
+  it "observer can delete themselves as observer" do
+    proposal = FactoryGirl.create(:proposal)
```
jessieay: don't need `FactoryGirl` :) :woman:

spec/features/observer_spec.rb
```diff
@@ -77,6 +77,22 @@
...
+    login_as(proposal.requester)
+    visit "/proposals/#{proposal.id}"
+    select observer.email_address, from: 'observation_user_email_address'
+    click_on 'Add an Observer'
+
+    login_as(observer)
+    visit "/proposals/#{proposal.id}"
+    delete_button = find('table.observers .button_to input[value="Delete"]')
+    delete_button.click
+    expect(page).to have_content("Deleted Observation for ")
```
jessieay: nice feature-level spec! hoping we can move this project to where all feature specs are testing at this level rather than inspecting the db or other backend stuff in expectations

app/policies/observation_policy.rb
jessieay: oh wait, this is a policy. NM! still getting used to pundit

spec/features/observer_spec.rb
```diff
@@ -77,6 +77,22 @@
     expect(submit_button).to_not be_disabled
   end
 
+  it "observer can delete themselves as observer" do
```
jessieay: should we have a test somewhere that an observer cannot delete another observer? maybe a controller-level spec?

pkarman: we already do: https://github.com/18F/C2/blob/master/spec/controllers/observations_controller_spec.rb#L34

pkarman:
thanks for the review @jessieay -- fixes made and pushed.

spec/features/observer_spec.rb
jessieay: oh cool!

jessieay:
will QA as soon as CI passes :game_die:

jessieay:
Now that I am QAing, wonder if we should have more explicit copy. 

As I see it on staging, looks like I might be deleting myself as a user (unclear because link just says "delete") -- what about something like "Remove as observer" ? 

screen shot 2015-11-06 at 11 17 19 am

pkarman:
I agree that "Delete" can be scary. How about just "Remove" instead?

jessieay:
"Remove" definitely sounds less scary :+1:

pkarman:
@jessieay language updated

jessieay:
cool, will merge once CI passes

PR #776, opened 2015-11-06
add cssh script

pkarman:


jessieay:
should we include something about this / how to use this in the docs?

pkarman:
@jessieay : docs updated to change `cf-ssh` mentions to `script/cssh`

jessieay:
sweet, merging

PR #777, opened 2015-11-06
[Delivers #87947800] myusa first last

pkarman:
Story: https://www.pivotaltracker.com/story/show/87947800

If the User has configured MyUSA to share first and last name, the app will now use those values to populate the local User record.

Testing:

1. Logout of app
1. Logout of https://alpha.my.usa.gov/
1. Login to app
1. On MyUSA confirmation page, select **yes** for first and last names.
1. Confirm name is populated on app login

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -106,7 +106,9 @@ class User < ActiveRecord::Base
 
   def self.from_oauth_hash(auth_hash)
     user_data = auth_hash.extra.raw_info.to_hash
-    self.find_or_create_by(email_address: user_data['email'])
+    user = self.for_email(user_data['email'])
+    user.update_attributes(first_name: user_data['first_name'], last_name: user_data['last_name'])
+    user
   end
 
   def role_on(proposal)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -107,7 +107,9 @@ class User < ActiveRecord::Base
   def self.from_oauth_hash(auth_hash)
     user_data = auth_hash.extra.raw_info.to_hash
     user = self.for_email(user_data['email'])
-    user.update_attributes(first_name: user_data['first_name'], last_name: user_data['last_name'])
+    if user_data['first_name'].present? && user_data['last_name'].present?
+      user.update_attributes(first_name: user_data['first_name'], last_name: user_data['last_name'])
+    end
     user
   end
 
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -4,7 +4,7 @@ MYUSA_URL = ENV['MYUSA_URL'] || 'https://alpha.my.usa.gov'
 
 Rails.application.config.middleware.use OmniAuth::Builder do
   provider :myusa, MYUSA_KEY, MYUSA_SECRET, {
-    scope: 'profile.email',
+    scope: 'profile.email profile.first_name profile.last_name',
     client_options: {
       site: MYUSA_URL,
       token_url: "/oauth/authorize"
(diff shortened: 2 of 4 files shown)


app/models/user.rb
```diff
@@ -106,7 +106,11 @@ def self.for_email_with_slug(email, client_slug)
 
   def self.from_oauth_hash(auth_hash)
     user_data = auth_hash.extra.raw_info.to_hash
-    self.find_or_create_by(email_address: user_data['email'])
+    user = self.for_email(user_data['email'])
```
jessieay: I'm surprised rubocop isn't picking up singe quotes?

spec/requests/set_user_after_login_spec.rb
```diff
@@ -11,7 +11,10 @@
       get '/auth/myusa/callback'
     }.to change { User.count }.by(1)
 
-    expect(User.last.email_address).to eq('george-test@example.com')
+    new_user = User.last
+    expect(new_user.email_address).to eq('george-test@example.com')
+    expect(new_user.first_name).to eq("Georgie")
```
jessieay: should we create a fixture without first name and last name to make sure nothing errors out when they are not present?

config/initializers/omniauth.rb
```diff
@@ -4,7 +4,7 @@
 
 Rails.application.config.middleware.use OmniAuth::Builder do
   provider :myusa, MYUSA_KEY, MYUSA_SECRET, {
-    scope: 'profile.email',
+    scope: 'profile.email profile.first_name profile.last_name',
```
jessieay: That Was Easy:tm:

jessieay:
One testing question; once that is addressed looks good to :ship:

config/initializers/omniauth.rb
afeld: Note that this may(?) require all users to re-authorize, which is a bummer.

pkarman: IME, yes, it will trigger a re-authorization.

spec/requests/set_user_after_login_spec.rb
pkarman: I believe the next test in that same file does that.

jessieay: I guess what I was thinking was set up something like this that does not include fname / lname and log in with that: https://github.com/18F/C2/blob/b25faa52f4c75b8d1e56127251a805ea38e055f0/spec/support/integration_spec_helper.rb

pkarman:
@jessieay test added in 0c19ca0

jessieay:
woo hoo! will QA now

jessieay:
Just followed QA instructions and when I logged back in w myUSA I was not prompted to say yes to first and last names and when I re-logged in my name was not saved. Any ideas on why?

pkarman:
hmmm.

Try deleting the cf-c2-dev application my https://alpha.my.usa.gov/ ?

Though when I try that, and try to log in to c2-dev.18f.gov, I am only prompted for allowing my email.

afeld:
Hmm, I already had all of these selected for each environment on the MyUSA section:

screen shot 2015-11-09 at 3 02 31 pm

Paging @yozlet!

jessieay:
OK, works on staging where I deployed (oops!). but my email address is not showing up in upper righthand of staging:

screen shot 2015-11-09 at 12 34 04 pm

is that expected?

pkarman:
Yes that is expected (alas). `app/views/layouts/_header_nav.html.erb` uses `<%= current_user.email_address %>`

pkarman:
You can confirm your name is populated by looking at your User record in /admin

jessieay:
OK, I don't mind keeping as-is -- clearer for people w  multiple accts. One final question before merge: what *is* the purpose of pulling this info? are there other stories around using firstname and lastname?

jessieay:
@pkarman yes, confirmed in admin. It is there! :tada:

pkarman:
I believe the first/last name will appear in the Approver/Observer/Requester info on a proposal, if present.

PR #779, opened 2015-11-07
clean up AuthController

afeld:
Just a minor refactoring, removing things that are no longer used.


pkarman:
these methods are no longer used because we use the omniauth gem?

code looks fine to me.

pkarman:
tested on c2-dev -- looks good

afeld:
> these methods are no longer used because we use the omniauth gem?

Not clear what they were ever used for. Dunno!

PR #780, opened 2015-11-09
[Delivers #107603900] User factory client_slug now defaults to nil

pkarman:
Story: https://www.pivotaltracker.com/story/show/107603900

Internal refactor to make User factory default to `client_slug: nil`

This makes our tests not assume `ncr` by default.

Opened in response to https://github.com/18F/C2/pull/772#discussion_r44179509

### DIFF
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :user do
-    client_slug { ENV['CLIENT_SLUG_DEFAULT'] || 'ncr' }
+    client_slug { ENV['CLIENT_SLUG_DEFAULT'] || nil }
     sequence(:email_address) {|n| "user#{n}@example.com" }
     sequence(:first_name) {|n| "FirstName#{n}" }
     sequence(:last_name) {|n| "LastName#{n}" }
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :user do
-    client_slug { ENV['CLIENT_SLUG_DEFAULT'] || nil }
+    client_slug { nil }
     sequence(:email_address) {|n| "user#{n}@example.com" }
     sequence(:first_name) {|n| "FirstName#{n}" }
     sequence(:last_name) {|n| "LastName#{n}" }
--- a/spec/features/inactive_users_spec.rb
+++ b/spec/features/inactive_users_spec.rb
@@ -1,8 +1,8 @@
 feature "Inactive users" do
   scenario "not included in approving official dropdown" do
-    inactive_approving_official = create(:user, :inactive)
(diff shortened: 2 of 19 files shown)


spec/factories/users.rb
```diff
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :user do
-    client_slug { ENV['CLIENT_SLUG_DEFAULT'] || 'ncr' }
+    client_slug { ENV['CLIENT_SLUG_DEFAULT'] || nil }
```
jessieay: maybe we should just set to `nil` rather than using env var? I haven't ever set it, so not sure what the use case for env var is.

spec/features/inactive_users_spec.rb
```diff
@@ -1,8 +1,8 @@
 feature "Inactive users" do
   scenario "not included in approving official dropdown" do
-    inactive_approving_official = create(:user, :inactive)
-    active_approving_official = create(:user, :active)
-    user = create(:user)
+    inactive_approving_official = create(:user, :inactive, client_slug: 'ncr')
```
jessieay: rubocop should be raising on single quotes?

jessieay:
Two small comments, otherwise looks :fire: :ship:

spec/features/inactive_users_spec.rb
pkarman: methinks rubocop ignores everything in `spec/`

jessieay: huh! I never knew that

spec/factories/users.rb
pkarman: fair point. Removing the env var required a bit more clean up in the `Proposal` factory as well, since it was also referenced there.

spec/features/inactive_users_spec.rb
jessieay: do you think we should still try to be consistent w rubocop styles in tests?

spec/factories/users.rb
jessieay: oh really? I grepped for it and only saw it referenced in tests. 

Up to you if we should remove - how much more cleanup would it require?

pkarman:
`CLIENT_SLUG_DEFAULT` removed in 9bdf894 @jessieay

spec/features/inactive_users_spec.rb
pkarman: there's a tremendous weight of history in `spec/` that uses single quotes. I'm happy to change these that I have introduced. Only 3k+ more to change.

pkarman:
@jessieay sorry clicked "Finish" in PT not remembering that Chores automatically get accepted when I do that.

jessieay:
@pkarman no prob! yeah I hear you on changing specs. We also have a lot of non-spec code that doesn't conform with our rubocop setup, but changing as we go has been working so far IMO.

Will merge once circle passes!

PR #781, opened 2015-11-09
[Delivers #107609788] exclude Delegates from attachment notifications

pkarman:
Story: https://www.pivotaltracker.com/story/show/107609788

Do not send attachment notification emails to proposal delegates, even if they relate to the currently actionable approval/step.

Testing guidelines are listed in the Story (link above).

### DIFF
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -99,6 +99,10 @@ class Proposal < ActiveRecord::Base
 
   alias_method :subscribers, :users
 
+  def users_except_delegates
+    users - delegates
+  end
+
   def root_step=(root)
     old_steps = self.steps.to_a
     step_list = root.pre_order_tree_traversal
--- a/spec/models/proposal_spec.rb
+++ b/spec/models/proposal_spec.rb
@@ -112,6 +112,15 @@ describe Proposal do
       proposal.add_observer(observer)
       expect(proposal.users).to eq [observer]
     end
+
+    it "can exclude delegates" do
+      delegate_one = create(:user)
+      delegate_two = create(:user)
+      proposal = create(:proposal, :with_approver)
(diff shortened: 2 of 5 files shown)


app/models/proposal.rb
```diff
@@ -99,6 +99,10 @@ def users
 
   alias_method :subscribers, :users
 
+  def users_except_delegates
```
jessieay: unit test for this?

pkarman: test added in 5ca40f4

spec/models/proposal_spec.rb
```diff
@@ -112,6 +112,15 @@
...
       expect(proposal.users).to eq [observer]
     end
+
+    it "can exclude delegates" do
+      delegate_one = create(:user)
+      delegate_two = create(:user)
+      proposal = create(:proposal, :with_approver)
+      proposal.approvers.first.add_delegate(delegate_one)
+      proposal.approvers.first.add_delegate(delegate_two)
+      expect(proposal.users).to match_array(proposal.users_except_delegates + [delegate_one, delegate_two])
```
jessieay: not trying to be a nit picker here, but if I were looking for a test for this method, I would expect it to be within a `describe "#method_name` block as opposed to embedded in a test for another method

pkarman: nits picked in 4c73fbd

jessieay:
code looks good! :ship:

jessieay:
QA-ing now

jessieay:
Ok, needed shared mailbox login to QA so am holding off for now but left a comment on the tracker story

jessieay:
Passing this off to @phirefly for QA

phirefly:
Validated on dev

PR #783, opened 2015-11-09
Rename `Proposal#client` to `Proposal#client_slug`

jessieay:
* Pulled from https://github.com/18F/C2/pull/768


app/helpers/client_helper.rb
```diff
@@ -1,33 +1,27 @@
...
     default_check = "default/" + path
+
     if lookup_context.template_exists?(to_check, [], true)
       args[:partial] = to_check
       render(args)
     elsif lookup_context.template_exists?(default_check, [], true)
       args[:partial] = default_check
       render(args)
-    else
-      ""
```
afeld: Hmm, could this break things?

jessieay: yeah it could. right now, no tests are breaking so I am assuming it is not breaking anything at the moment. 

I think the benefit if returning nothing if there is not a partial is that things *will* break when there is not a `client_partial` for something. my worry with the method as it was is that it would work whether a partial existed or not, but we always had partials for the cases where we were using this method.

If there ends up being a case where we want to render nothing, I would put in an empty template at the expected path. does that make sense?

afeld: It does. At the rate we are adding client models, I suppose I'm not _too_ concerned about missing a partial somewhere. Fine for now.

afeld:
QA'd on dev.

PR #784, opened 2015-11-10
set ClassLength max to 250 lines (default is 100)

pkarman:


jessieay:
Works for me as long as we try to keep classes on the small side! Interested in what others think

pkarman:
feedback requested. cc/ @yozlet @annalee @afeld

annalee:
This is fine by me! I've never been terribly fussy about class length, though.

yozlet:
I agree with @jessieay, especially since some of our classes are longer than they should be (which we're fixing with concerns, etc.) But our classes tend to have other stuff that inflates them, especially those big `workflow` blocks. So :+1:

afeld:
+0 on this.

PR #785, opened 2015-11-10
[Delivers #99459308] Proposal tagging

pkarman:
Story: https://www.pivotaltracker.com/story/show/99459308

Adds backend tagging to `Proposal` class. No UI/UX work here.


db/migrate/20151110181710_change_collation_for_tag_names.acts_as_taggable_on_engine.rb
```diff
@@ -0,0 +1,10 @@
+# This migration comes from acts_as_taggable_on_engine (originally 5)
+# This migration is added to circumvent issue #623 and have special characters
+# work properly
+class ChangeCollationForTagNames < ActiveRecord::Migration
+  def up
+    if ActsAsTaggableOn::Utils.using_mysql?
```
jessieay: since we're not using mysql, do we need this migration?

Gemfile
```diff
@@ -4,6 +4,7 @@ ruby '2.2.3' # this should match `.ruby-version` and docs/setup.md
 gem 'active_model_serializers'
 gem 'activeadmin', github: 'activeadmin'
 gem 'acts_as_list'
+gem 'acts-as-taggable-on', '~> 3.4'
```
jessieay: is the version number important, or can we leave it out?

config/initializers/acts_as_taggable.rb
```diff
@@ -0,0 +1 @@
+ActsAsTaggableOn.force_lowercase = true
```
jessieay: this is cool

spec/models/proposal_spec.rb
```diff
@@ -332,4 +332,13 @@
       expect(observation_creator_double).to have_received(:run)
     end
   end
+
+  describe "#tags" do
+    it "can add case-insensitive tags" do
```
jessieay: I think it's also possible to use space separators rather than comma with this lib -- maybe while we're writing this test we can also add that so the description is "adds case-insensitive tags via comma separated strings" ?

jessieay:
Few teeny questions to address, then looks ready to :ship: :it:

db/migrate/20151110181710_change_collation_for_tag_names.acts_as_taggable_on_engine.rb
pkarman: it was auto-generated per the acts-as-taggle-on docs. In our case it's a no-op.

AFAIK we have no specific Postgresql db code present in the app, so keeping it RDBMS-neutral seems like a virtue, yes?

Gemfile
pkarman: that's a cargo-cult line from the acts-as-taggable README. Since it was important to the author, I assumed it should be important to us?

spec/models/proposal_spec.rb
pkarman: The README claims:

If you want to change the default delimiter (it defaults to ','). You can also pass in an array of delimiters such as ([',', '|']):
```
ActsAsTaggableOn.delimiter = ','
```

There's no mention of whitespace, and we haven't deviated from the default, so do we need to test that feature?

jessieay: fine w me to leave untested!

jessieay:
What do you think the best way to QA this would be?

pkarman:
Since there is no UI-visible work, QA can just be deploying via `cf push` and verifying migrations run ok.

db/migrate/20151110181710_change_collation_for_tag_names.acts_as_taggable_on_engine.rb
jessieay: we do have setu specific to postgres in in `config/databse.yml` but keeping this wont hurt, not a big deal

Gemfile
jessieay: I'd prefer to keep it out of we can. If there is an exact version #, I usually assume a gem should not be upgraded....but your call!

jessieay:
gonna QA this now

Gemfile
pkarman: afaik, the `~>` syntax is significant in terms of API/ABI compatability. In this case, at least version 3.4 but nothing 4.0 or higher.

jessieay: Yup, pessimistic versioning! https://robots.thoughtbot.com/a-healthy-bundle

jessieay:
deployed and all is not broken! merging :shipit:

PR #786, opened 2015-11-10
client -> client_slug in RolePicker

jessieay:
* Fixes https://rpm.newrelic.com/accounts/921394/applications/5480870/traced_errors/a28401-527aa035-87db-11e5-a540-f8bc12425498

### DIFF
--- a/lib/role_picker.rb
+++ b/lib/role_picker.rb
@@ -5,7 +5,7 @@ class RolePicker
     active_approver: ->(u, p) { p.is_active_approver? u },
     observer: ->(u, p) { p.observers.include? u },
     approver: ->(u, p) { p.approvers.include? u },
-    client_admin: ->(u, p) { u.client_admin? && p.client == u.client_slug }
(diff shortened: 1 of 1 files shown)


lib/role_picker.rb
```diff
@@ -5,7 +5,7 @@ class RolePicker
     active_approver: ->(u, p) { p.is_active_approver? u },
     observer: ->(u, p) { p.observers.include? u },
     approver: ->(u, p) { p.approvers.include? u },
-    client_admin: ->(u, p) { u.client_admin? && p.client == u.client_slug }
+    client_admin: ->(u, p) { u.client_admin? && p.client_slug == u.client_slug }
```
pkarman: should probably do instead:
```
p.slug_matches?(u)
```
now that `slug_matches` exists.

jessieay: good call, will change when I write specs for this class (next on my list)

PR #787, opened 2015-11-10
[Delivers #96028254] mange user profile

pkarman:
Story: https://www.pivotaltracker.com/story/show/96028254

Expands the `/me` endpoint to allow users to manage their own names. Will keep a header banner present until the first and last name is set in their profiles.

Test:

1. Login and click on your email address in the header nav.
1. Verify your first and last names are set. If not, enter and click *Update profile*


app/controllers/home_controller.rb
```diff
@@ -10,6 +10,17 @@ def index
   def me
   end
 
+  def edit_me
```
jessieay: what about making a `ProfilesController` or `UserProfilesController` and having this be a `show` action instead? That would be more Railsy (although that can be considered a compliment or a dis, depending on who you are :hamburger: )

app/views/layouts/_communicart_header.html.erb
```diff
@@ -1,6 +1,9 @@
 <% unless ENV['DISABLE_SANDBOX_WARNING'] == 'true' %>
   <%= render partial: 'shared/sandbox_warning' %>
 <% end %>
+<% if !current_page?(me_path) && current_user && current_user.requires_profile_attention? %>
```
jessieay: perhaps we should encapsulate this logic in a helper method w a test?

app/views/shared/_user_profile_warning.html.haml
```diff
@@ -0,0 +1,4 @@
+%div#profile-alert
+  %span
+    Your profile is missing some important information.
+    %a{href: "/me"}Click here to update your profile.
```
jessieay: use rails path helper instead of href?

app/models/user.rb
```diff
@@ -116,4 +119,8 @@ def self.from_oauth_hash(auth_hash)
   def role_on(proposal)
     RolePicker.new(self, proposal)
   end
+
+  def requires_profile_attention?
```
jessieay: niiiiice. This might also work well on a `UserDecorator` if we have / want to use one of those, since it's being used purely for view logic

spec/features/edit_me_spec.rb
```diff
@@ -0,0 +1,11 @@
+describe "User can manage their own profile" do
```
jessieay: perhaps we should test that users cannot edit the profiles of others?

spec/features/edit_me_spec.rb
```diff
@@ -0,0 +1,11 @@
...
+  it "allows editing of name" do
+    user = create(:user)
+    login_as(user)
+    visit "/me"
+    fill_in :first_name, with: "Some"
+    fill_in :last_name, with: "Body"
+    click_button "Update profile"
+    expect(page).to have_content("Welcome, Some Body <#{user.email_address}>")
```
jessieay: nice feature spec :cherries:

app/models/user.rb
pkarman: It's used only on view logic now, but imo it represents business logic that could be used anywhere. E.g. writing a rask task to report on or notify all users who require some attention to their profile.

spec/features/edit_me_spec.rb
pkarman: I hear you. I thought about to do that, but profile editing always acts upon the `current_user`. 

How would I test that `current_user` could edit the profile of someone else?

app/views/profile/show.html.haml
```diff
@@ -0,0 +1,9 @@
+.inset
+  %h2= "Welcome, #{current_user.full_name} <#{current_user.email_address}>"
+  = form_tag(me_path(current_user), method: :post) do
+    %label.string.control-label First Name
+    %input.form-control{type: 'text', name: 'first_name', value: current_user.first_name}
```
jessieay: since there are inputs here, should this be `profile/edit` ?

config/routes.rb
```diff
@@ -2,7 +2,8 @@
   ActiveAdmin.routes(self)
   root :to => 'home#index'
   get '/error' => 'home#error'
-  get '/me'    => 'home#me'
+  get '/me'    => 'profile#show'
```
jessieay: why use these rather than regular named paths?

spec/features/edit_me_spec.rb
jessieay: I was assuming we'd be using RESTful routes and this action would be `profiles/:id/edit` -- if we are sticking with `/me` then it would not be testable

app/controllers/profile_controller.rb
```diff
@@ -0,0 +1,15 @@
+class ProfileController < ApplicationController
```
jessieay: should we have a `before_filter` for auth here?

jessieay:
2 final questions, then looks ready to :shaved_ice:

app/views/profile/show.html.haml
pkarman: I am using the pattern:
```
POST /me  => update
GET /me   => show
```

If that's confusing, we can discuss.

config/routes.rb
pkarman: Sorry, not sure what you mean by "regular"? How would you construct the routes differently?

app/controllers/profile_controller.rb
pkarman: ah good call. Forgot it wasn't in `ApplicationController`. Fixed in 48c0a61

app/views/profile/show.html.haml
jessieay: ok! yeah let's discuss. As a developer on this project, I'd find it easier to understand the default PATHs rather than `me`, but if there is a good user-facing reason to not use those I am all :ear:

config/routes.rb
jessieay: ah yeah sorry about vagueness -- so in here, what I meant by regular is that we'd have `resources :profiles, only: [:edit, :update]` per the Rails default way of defining routes http://guides.rubyonrails.org/routing.html

app/views/profile/show.html.haml
pkarman: I guess this is an aesthetic choice, not a technical one.

I chose `/me` because it was short, pithy and unambiguous. It represents the `current_user` publicly to the user, regardless of how that current user is represented privately to the application. Like `/proposals` returns a list of proposals related to the `current_user`, `/me` returns a list of attributes that describe the `current_user`.

It feels somehow more personal than simply a resource to be managed.

The URL is the same for all users, like Google does with https://aboutme.google.com/

If it would be more clear internally, we could rename the controller to `MeController` or rename the route to `/profile`, if that's helpful.

jessieay: I think renaming to `profile` is a nice middle ground! Agree that the default path feels impersonal. I think as a user `me` makes sense but I can imagine a future where we have a bunch of named routes that don't match the controller name and that could get confusing so `profile` would be :tada:

pkarman:
@jessieay route re-named

jessieay:
cool will QA shortly

jessieay:
OK, confirmed that I can update my first name and last name and that doing so makes the banner go away.

One Q: do we want to replace the email address in the header with first and last name? if so, we can do that separately, just wondering...

screen shot 2015-11-17 at 11 48 17 am

PR #789, opened 2015-11-10
More Approval->Step refactoring

yozlet:
@annalee & I have been doing more bits here and there to both change terminology (from `Approval` to `Step`, though for the moment we're hanging onto `Approver`) and move a couple of step-related methods from `Proposal` into `StepManager`.


app/decorators/proposal_decorator.rb
```diff
@@ -25,7 +25,7 @@ def approvals_by_status
 
   def approvals_in_list_order
     if object.flow == 'linear'
-      object.individual_approvals.with_users
+      object.individual_steps.with_users
```
pkarman: see general comment about naming.

pkarman:
The code looks fine.

More generally about naming, I'm wondering if it makes sense to preserve the `proposal.individual_approvers` method syntax, just because the business context is more clear (to me) when the particular individual step(s) in question are of type `Steps::Approval`. I.e., "in this context I want the steps that are approval steps, not any other kind."

Could do this by adding an `approval` scope on the `Step` class, and then a has_many in Proposal like:
```
has_many :individual_approvals, ->{ individual.approval }, class_name: 'Steps::Individual'
```

Thoughts?

yozlet:
@pkarman: Makes sense, but I don't think we have a current need for that kind of narrowed scope anywhere; right now everything using `individual_steps` wants all the steps. (Admittedly, in NCR all the steps are Approvals, but that's not something we want to explicitly filter on)

I'd rather we do the renaming and explicitly change all the references to the old `individual_approvals` so that we don't then assume (yet) that it's just talking about Approvals rather than all Steps. Does that make sense?

pkarman:
@yozlet points taken. Looks like the convenience methods around `steps` still use `approver` so it's still clear what business action is taking place.

I will QA.

pkarman:
QA'd on c2-dev

PR #790, opened 2015-11-11
Add specs for RolePicker

jessieay:
* Follow up to https://github.com/18F/C2/pull/786

### DIFF
--- a/lib/role_picker.rb
+++ b/lib/role_picker.rb
@@ -1,35 +1,32 @@
 class RolePicker
-  attr_accessor :user
-  GUARDS = {
-    requester: ->(u, p) { u == p.requester },
-    active_approver: ->(u, p) { p.is_active_approver? u },
-    observer: ->(u, p) { p.observers.include? u },
-    approver: ->(u, p) { p.approvers.include? u },
-    client_admin: ->(u, p) { u.client_admin? && p.client_slug == u.client_slug }
-  }
+  attr_reader :user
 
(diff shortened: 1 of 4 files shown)


lib/role_picker.rb
```diff
@@ -1,35 +1,32 @@
...
+  end
+
+  def approver?
+    proposal.approvers.include?(user)
+  end
+
+  private
+
+  attr_reader :proposal
 end
```
pkarman: I like that you've split that auto-generating code up into explicit method definitions. Much easier to grok for first-timer.

lib/subscriber_list.rb
```diff
@@ -7,24 +7,23 @@ def initialize(proposal)
...
   # Returns triplets of (user, role name, observation)
   def triples
-    requesters, approvers, observers = self.partitioned_roles
-    requesters = requesters.map { |r| [r.user, "Requester", nil] }
-    approvers = approvers.map { |r| [r.user, "Approver", nil] }
-    observers = observers.map { |r| [r.user, nil, self.observation_for(r.user)] }
+    requesters, approvers, observers = partitioned_roles
+    requesters = requesters.map { |role| [role.user, "Requester", nil] }
+    approvers = approvers.map { |role| [role.user, "Approver", nil] }
+    observers = observers.map { |role| [role.user, nil, observation_for(role.user)] }
```
pkarman: +1 for better variable naming.

PR #791, opened 2015-11-12
create a #work_order method for use in that controller

afeld:
Just a minor refactor to make it more clear.


pkarman:
+1 for clarity. when this goes green I will merge.

PR #793, opened 2015-11-12
[Delivers #106300186] Revert "use the best-guess app URL on Cloud Foundry"

afeld:
Reverts 18F/C2#669.

* [x] `cf push` to `c2-dev` and `c2-staging`


pkarman:
+1 looks good.

PR #795, opened 2015-11-16
[Delivers #108012382] fix persistent flash message

pkarman:
Story: https://www.pivotaltracker.com/story/show/108012382
Story: https://www.pivotaltracker.com/story/show/97035674

use `flash.now` to make sure warning appears only on the current page

To test this, attempt to modify an already fully-approved NCR request. On the edit page, there should be a warning about attempting to modify the request. When you click "Discard Changes" at the bottom of the page, the flash warning should not appear on the request view page.

### DIFF
--- a/app/controllers/ncr/work_orders_controller.rb
+++ b/app/controllers/ncr/work_orders_controller.rb
@@ -15,7 +15,7 @@ module Ncr
 
     def edit
       if self.proposal.approved?
-        flash[:warning] = "You are about to modify a fully approved request. Changes will be logged and sent to approvers, and this request may require re-approval, depending on the change."
+        flash.now[:warning] = "You are about to modify a fully approved request. Changes will be logged and sent to approvers, and this request may require re-approval, depending on the change."
       end
 
       super
--- a/app/controllers/use_case_controller.rb
+++ b/app/controllers/use_case_controller.rb
@@ -21,7 +21,7 @@ class UseCaseController < ApplicationController
       flash[:success] = "Proposal submitted!"
       redirect_to proposal
     else
-      flash[:error] = errors
+      flash.now[:error] = errors
       render :new
     end
   end
--- a/spec/features/ncr/work_orders/create_spec.rb
+++ b/spec/features/ncr/work_orders/create_spec.rb
@@ -92,6 +92,26 @@ feature 'Creating an NCR work order' do
       ]
     end
 
+    scenario "flash message on error does not persist" do
+      approver = create(:user, client_slug: "ncr")
+      login_as(requester)
+
+      visit '/ncr/work_orders/new'
+      fill_in 'Project title', with: "blue shells"
+      fill_in 'Description', with: "desc content"
+      choose 'BA60'
+      fill_in 'Vendor', with: 'Yoshi'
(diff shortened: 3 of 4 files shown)


app/controllers/ncr/work_orders_controller.rb
```diff
@@ -15,7 +15,7 @@ def create
 
     def edit
       if self.proposal.approved?
-        flash[:warning] = "You are about to modify a fully approved request. Changes will be logged and sent to approvers, and this request may require re-approval, depending on the change."
+        flash.now[:warning] = "You are about to modify a fully approved request. Changes will be logged and sent to approvers, and this request may require re-approval, depending on the change."
```
jessieay: I believe this was happening for several different flash messages - should we add `now` to all flash messages? (there might be a downside to doing that, but I am not sure what it would be)

pkarman: that might be the fix in those cases, but this particular story's scope is just this one.

jessieay: there are 2 other stories cc'd in the tracker story: 

https://www.pivotaltracker.com/n/projects/1149728/stories/98398508
https://www.pivotaltracker.com/n/projects/1149728/stories/97035674

since this PR is pretty small, do you think it would make sense to address all while we are here? Also, persistence of other flash messages hasn't been reported per se, but it probably makes sense to not have persisting flash messages everywhere, right?

pkarman:
fix for story https://www.pivotaltracker.com/story/show/97035674 in a170aea

jessieay:
Cool - what are your thoughts on using `flash.now` everywhere? seems like there aren't any situations where we want to flash message to persist for multiple actions. This (very old) blog post was helpful when I just went to look up the exact difference http://travisonrails.com/2008/08/17/working-with-the-flash-hash

pkarman:
I don't think we should use it everywhere.

We only need `flash.now` when we are using it to display a message on the currently rendered page. That's really an abuse of the feature, imo, because it's really for short-lived messages that you want to persist after a redirect. On a redirect, you don't need the `.now` and that's mostly how we use it. If we used `flash.now` everywhere, it wouldn't persist after the redirect, afaik, which would defeat the purpose.

jessieay:
@pkarman makes sense! will QA now

jessieay:
Ok, trying to QA by creating a new work order per https://www.pivotaltracker.com/n/projects/1149728/stories/97035674 but I cannot actually click "submit" unless required fields are filled out because of client side validations :

screen shot 2015-11-16 at 1 44 00 pm

@pkarman were you able to reproduce? And if so, which browser were you using?

pkarman:
@jessieay I had the same issue. I had to access the work_order via rails console, fully approve all its individual_steps, and then I could follow the steps in the description above.

pkarman:
@jessieay also note that all fields with js client-side validation will be required, except for "Approver" which you can leave blank and that will generate a flash message on server response.

jessieay:
@pkarman thanks for the tip! will have to QA this after I get ssh to work..unless you'd like to live dangerously

jessieay:
woo hoo! got ssh to work. QAing now

PR #797, opened 2015-11-16
[Delivers #107777486] disable Delete User action in /admin/users

pkarman:
Story: https://www.pivotaltracker.com/story/show/107777486

`/admin/` section should not have a Delete User button.

### DIFF
--- a/app/admin/user.rb
+++ b/app/admin/user.rb
@@ -1,4 +1,6 @@
 ActiveAdmin.register User do
+  actions :all, :except => [:destroy]
+
   # See permitted parameters documentation:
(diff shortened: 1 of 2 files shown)


app/admin/user.rb
```diff
@@ -1,4 +1,6 @@
 ActiveAdmin.register User do
+  actions :all, except: [:destroy]
```
jessieay: about about explicitly setting which actions we want rather than which one we don't want? that is the style I usually use for `before_action`s or `config/routes` -- the result is the same but it's nice to see which actions are enabled explicitly!

jessieay:
woot woot will QA now

jessieay:
looks good on DEV! merging

PR #800, opened 2015-11-17
Add shared example for client data objects

jessieay:
* Shared examples have tests for shared behavior that was previously mostly untested

### DIFF
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -115,7 +115,7 @@ class Proposal < ActiveRecord::Base
   end
 
   def eligible_observers
-    if observations.any?
+    if observations.count > 0
       User.where(client_slug: client_slug).where('id not in (?)', observations.pluck('user_id'))
     else
       User.where(client_slug: client_slug)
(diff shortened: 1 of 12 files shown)


app/models/proposal.rb
```diff
@@ -119,7 +115,7 @@ def existing_observation_for(user)
   end
 
   def eligible_observers
-    if observations.count > 0
+    if observations.any?
```
pkarman: afaik this results is different sql and may be impacted by object caching.

`observations.count` will execute a SQL `select count(*) from ....` while `observations.any?` loads the objects (if not already) and counts them.

The former is more accurate and efficient, imo.

pkarman:
@jessieay I like the unified testing of client code. Just one comment about a refactor that has subtly different effects.

app/models/proposal.rb
jessieay: ah ok! Will change back

jessieay: updated at 09e3d4b935a6eda4382f865522c05c76433487b5

pkarman:
:+1: 

assuming this is all regression testing, so I will QA when it goes green.

jessieay:
I re-ran the build since it was an unrelated failure and now it is green. worrisome? I haven't seen too many intermittently failing specs...

Yes, regression testing indeed.

pkarman:
QA looks good on c2-dev

PR #792, opened 2015-11-12
[Delivers #97242150] disallow assigning requester as approver and vice-versa

pkarman:
Story: https://www.pivotaltracker.com/story/show/97242150

Testing:

* create new NCR work order
* verify you are not listed amongst the possible approvers

PR #801, opened 2015-11-17
[Delivers #106953230] no inactive users in approvers

pkarman:
Story: https://www.pivotaltracker.com/story/show/106953230


pkarman:
depends on #792 -- merge that first.

jessieay:
793 was merged - do you mind pulling in changes from master so that the changes specific to this PR are more obvious?

pkarman:
@jessieay done

spec/features/ncr/work_orders/create_spec.rb
```diff
@@ -112,6 +112,17 @@
       expect(page).to_not have_content("Approving official email can't be blank")
     end
 
+    scenario "inactive users do not appear as potential approvers", :js do
```
jessieay: there is a test for this here: https://github.com/18F/C2/pull/734/files#diff-9965b424c2f291874f17f5d65d09fc9dR1

if that was not a good / obvious place for it, should we perhaps remove it alongside adding this?

pkarman: I don't think it hurts to have more tests?

The one I added does use the `:js` feature and clickety stuff to make sure DOM is not hiding it. Not sure if that matters.

jessieay: Doesn't hurt to have more, but these tests are pretty much duplicates in different locations. I think it makes sense to have one with js!

pkarman:
If you think they're duplicates, let's just close this PR and the story and call it a day. I obviously didn't look closely enough before adding the test.

pkarman:
Ok, I have removed the inactive users spec file, since it only had one test, now duplicated with `:js` support in NCR-specific file.

jessieay:
looks great! merging

PR #802, opened 2015-11-17
[Delivers #107518164] Add commenter full name to comment email

jessieay:
* Also move comment email into its own mailer
* To QA, create a comment and confirm that the comment email containers
  your full name or email (depending on if you have a name in the db)
* Story: https://www.pivotaltracker.com/story/show/107518164


pkarman:
@jessieay looks good. I will QA.

PR #804, opened 2015-11-17
Delivers #108022488] gsa18f roles

pkarman:
Story: https://www.pivotaltracker.com/story/show/108022488

No functional change, just switches from environment variables to roles, similar to what we did with NCR a month+ ago.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -7,3 +7,4 @@ Role.find_or_create_by(name: 'admin')
 # it should be safe to run multiple times in the same environment, and to remove
 # the relevant ENV vars after it has run at least once.
 RolesConversion.new.ncr_budget_approvers
+RolesConversion.new.gsa18f_approvers
--- a/.env.example
+++ b/.env.example
@@ -2,8 +2,6 @@
 
 # required
 #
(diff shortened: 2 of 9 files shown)


db/seeds.rb
```diff
@@ -7,3 +7,4 @@
 # it should be safe to run multiple times in the same environment, and to remove
 # the relevant ENV vars after it has run at least once.
 RolesConversion.new.ncr_budget_approvers
+RolesConversion.new.gsa18f_approvers
```
jessieay: I am afraid that adding this is going to make our slow test suite even slower -- did working on this PR give you any good ideas on how we could move away from needing to run seeds before each test? I haven't put much thought into it recently but I am planning on doing so soon!

pkarman: @jessieay 2cbbf07 should improve overall test suite runtime.

jessieay: ok, that's a good start! I am still *really* hoping to find a way to not run seeds, but baby steps :baby: :chart_with_upwards_trend:

jessieay:
Looks :bomb: -- mergin!

PR #803, opened 2015-11-17
[Delivers #1072212000] Require auth in controllers by default

jessieay:
* Require controllers to opt out of auth rather than in
* Standardize on "before_action" rather than "before_filter"
* This is a refactor, only behavior change is slight logic change for
  admin -- redirect to home page w 302 when not logged in and trying
  to access admin rather than showing error and giving 401 status
* Story: https://www.pivotaltracker.com/story/show/107221200

PR #807, opened 2015-11-17
[Delivers #107212902] Add delegates via admin

jessieay:
* Admin can add delegates
* Admin can also delete delegates
* To test, visit '/admin/approval_delegates'
* Story: https://www.pivotaltracker.com/story/show/107212902


app/admin/approval_delegate.rb
```diff
@@ -0,0 +1,34 @@
+ActiveAdmin.register ApprovalDelegate do
+  actions :index, :new, :show, :create, :destroy
+
+  permit_params :assigner_id, :assignee_id
+
+  form do |f|
+    f.inputs 'Delegate' do
+      f.input :assigner, as: :select, collection: User.all.map { |user| [ user.email_address, user.id] }
```
jessieay: I had to do this in order to get email address to show up rather than `user.full_name` -- any ideas on why that was the default behavior? Only thing I could find via googling was that ActiveAdmin uses `to_s` but we haven't monkeypatched that AFAIK

jessieay:
Didn't add any feature-level specs for this -- should I? On the one hand, I think I should so we can make sure this isn't broken. On the other hand, our feature spec coverage for admin in general is pretty light. What level of coverage makes sense?

pkarman:
Since active_admin is mostly just configuration, not code we have to write, I haven't bothered testing anything that should Just Work Out of the Box.

Since you have configured the `actions` to explicitly exclude `:update` you might want to write a test to confirm that feature is not available.

I.e. I tend to write tests for the exception, not the rule, for gem-based features.

Code looks fine otherwise. I am assuming you excluded `:update` because add+delete are all the practical actions we want to take.

jessieay:
@pkarman makes sense re: description of testing strategy for active admin. Will write a test for exclusion of `update` now!

jessieay:
@pkarman feature spec added at bccdcbd74329d2d0e0f4e9806bf00917d587c3cd -- put in a new file that is introduced in https://github.com/18F/C2/pull/803 -- hoping that putting it there will minimize merge conflicts

pkarman:
@jessieay merge conflicts, alas.

test looks great.

jessieay:
@pkarman merge conflict fixed / re-pushed :tada:

pkarman:
cool. /me will QA

pkarman:
:+1: on c2-dev

PR #813, opened 2015-11-19
Clean up params methods for client data objs

jessieay:
* Was using `name` attr which is not an actual db attr for either obj

### DIFF
--- a/app/models/gsa18f/procurement.rb
+++ b/app/models/gsa18f/procurement.rb
@@ -26,32 +26,33 @@ module Gsa18f
     validates :product_name_and_description, presence: true
     validates :recurring_interval, presence: true, if: :recurring
 
+    def self.relevant_fields(recurring)
+      fields = self.default_fields
+
+      if recurring
+        fields += [:recurring_interval, :recurring_length]
+      end
+
+      fields
+    end
+
+    def self.default_fields
+      fields = self.column_names.map(&:to_sym) + [:approving_official_email]
+      fields - [:recurring_interval, :recurring_length, :created_at, :updated_at, :id]
+    end
+
+    def fields_for_display
+      attributes = self.class.relevant_fields(recurring)
+      attributes.map! {|key| [Procurement.human_attribute_name(key), self[key]]}
+      attributes.push(["Total Price", total_price])
+    end
+
     def add_steps
       steps = [
         Steps::Approval.new(user: User.for_email(Gsa18f::Procurement.approver_email)),
         Steps::Purchase.new(user: User.for_email(Gsa18f::Procurement.purchaser_email)),
       ]
       proposal.add_initial_steps(steps)
-    end
-
-    def self.relevant_fields(recurring)
-      fields = [:office, :justification, :link_to_product, :quantity,
-        :date_requested, :urgency, :additional_info, :cost_per_unit,
-        :product_name_and_description, :recurring]
-      if recurring
-        fields += [:recurring_interval, :recurring_length]
-      end
-      fields
-    end
-
-    def relevant_fields
-      Gsa18f::Procurement.relevant_fields(recurring)
-    end
-
-    def fields_for_display
-      attributes = self.relevant_fields
-      attributes.map! {|key| [Procurement.human_attribute_name(key), self[key]]}
-      attributes.push(["Total Price", total_price])
     end
 
     def total_price
@@ -64,7 +65,7 @@ module Gsa18f
     end
 
     def name
-      self.product_name_and_description
+      product_name_and_description
     end
 
     def editable?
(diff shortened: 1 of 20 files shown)


jessieay:
hey @afeld is it expected that this PR is failing hakiri for a security issue that was not introduced in this PR?

afeld:
Whoa, the Hakiri integration worked!

> is it expected that this PR is failing hakiri for a security issue that was not introduced in this PR?

Don't worry about it. [Storied](https://www.pivotaltracker.com/story/show/108578654).

app/models/gsa18f/procurement.rb
```diff
@@ -26,34 +26,35 @@ def self.purchase_amount_column_name
...
       if recurring
         fields += [:recurring_interval, :recurring_length]
       end
+
       fields
     end
 
-    def relevant_fields
-      Gsa18f::Procurement.relevant_fields(recurring)
+    def self.default_fields
```
jessieay: I would really like to clean this up further with a Form object, but found myself going down a rabbit hole so decided to move onto other things for the time being

spec/controllers/ncr/work_orders_controller_spec.rb
```diff
@@ -115,6 +115,19 @@
       end
     end
 
+    it "allows the project title to be edited" do
```
jessieay: this is a key test for this PR. Before the changes in this PR, this did not pass (we were using `name` param instead)

spec/features/ncr/work_orders/requester_edit_spec.rb
```diff
@@ -7,7 +7,7 @@
     end
   end
 
-  let(:work_order) { create(:ncr_work_order, description: 'test') }
+  let(:work_order) { create(:ncr_work_order, org_code: Ncr::Organization.all[0].to_s,  description: 'test') }
```
pkarman: would `Ncr::Organization.first.to_s` be more efficient?

spec/features/requester_switches_work_order_to_whsc_spec.rb
```diff
@@ -5,7 +5,7 @@
     end
   end
 
-  let(:work_order) { create(:ncr_work_order, description: 'test') }
+  let(:work_order) { create(:ncr_work_order, org_code: Ncr::Organization.all[0].to_s, description: 'test') }
```
pkarman: same comment as above.

spec/features/requester_switches_work_order_to_whsc_spec.rb
```diff
@@ -57,7 +57,7 @@
 
   context "as a BA80" do
     scenario "reassigns the approvers properly" do
-      work_order = create(:ba80_ncr_work_order)
+      work_order = create(:ba80_ncr_work_order, org_code: Ncr::Organization.all[0].to_s)
```
pkarman: same comment as above.

spec/lib/ncr/reporter_spec.rb
```diff
@@ -31,7 +31,7 @@
       )
       whs_work_order.setup_approvals_and_observers
 
-      approved_work_order = create(:ncr_work_order, :with_approvers)
+      approved_work_order = create(:ncr_work_order, :with_approvers, org_code: Ncr::Organization.all[0].to_s )
```
pkarman: same comment as above.

pkarman:
I like the clean up and clarified naming here. Just a comment minor comments. Otherwise, I will QA.

spec/features/ncr/work_orders/requester_edit_spec.rb
jessieay: sure! I also like the idea of having a factory for this non-AR class since `all` loads all ~900 records into memory, which seems like overkill. will look into doing that and report back...

pkarman:
Should we be thinking about moving these Orgs to the db itself rather than a CSV file, so we can get proper referential integrity, etc? Are we at the point in the maturity of the project where that makes sense?

jessieay:
@pkarman that might be a great idea. Actually, I think it definitely is. 

Could then create a task that imports / updates records via CSV. Thoughts from @afeld and @phirefly would be great, too!

spec/features/ncr/work_orders/requester_edit_spec.rb
jessieay: Ok so I just tried that route and it was kind of a pain because this is a non-AR class. But if we are going to make this an AR class (via https://github.com/18F/C2/pull/814) then it will become much easier. So I am gonna work on that and return to this PR afterward. Thanks for the suggestion!

jessieay:
@pkarman spoke w Raphy about this. He says there is no reason we shouldn't move in that direction.

jessieay:
@pkarman Going to move Org logic in a different PR. For now, this should be ready to move forward since it isn't doing a lot w `Ncr::Organization`

pkarman:
great I will QA

pkarman:
no regressions observed.

PR #814, opened 2015-11-20
Add `Ncr::OrganizationCollection` class

jessieay:
* Previously, Ncr::Organization was handling methods for single and
  collection methods
* Simplify (and test) collection methods via adding collection class
* Added more specs around WHSC SQL because I wasn't sure it was working
  before (since we save both code and name for organizations)

### DIFF
--- a/lib/ncr/organization.rb
+++ b/lib/ncr/organization.rb
@@ -1,69 +1,37 @@
 module Ncr
   class Organization
-    BY_CODE = {} # populated later
-    OOL_CODES = Set.new # populated later
-    WHSC_CODE = 'P1122021'
+    WHSC_CODE = "P1122021"
+    OOL_CODES = [
+      "P1171001",
+      "P1172001",
+      "P1173001"
+    ]
 
-    attr_accessor :code, :name
+    attr_reader :code, :name
 
     def initialize(code:, name:)
-      self.code = code
-      self.name = name
-    end
-
-    def ==(other)
-      other.code == self.code
+      @code = code
+      @name = name
     end
 
     def to_s
-      "#{self.code} #{self.name}"
+      "#{code} #{name}"
     end
 
-    # Office of Leasing
     def ool?
-      OOL_CODES.include?(self.code)
+      OOL_CODES.include?(code)
     end
 
-    # White House Service Center
     def whsc?
-      self.code == WHSC_CODE
+      code == WHSC_CODE
     end
 
     def self.all
-      BY_CODE.values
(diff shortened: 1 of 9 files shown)


lib/ncr/organization.rb
```diff
@@ -1,69 +1,37 @@
 module Ncr
   class Organization
-    BY_CODE = {} # populated later
-    OOL_CODES = Set.new # populated later
-    WHSC_CODE = 'P1122021'
+    WHSC_CODE = "P1122021"
+    OOL_CODES = [
```
jessieay: not that these are added as an array here rather than via the yaml file. That yaml file was added for this tracker story: https://www.pivotaltracker.com/n/projects/1149728/stories/100956358 -- but those org codes all appear to be in the main csv of organizations now. perhaps @phirefly can clarify whether we need to treat OOL orgs differently? (the only difference of this PR is we are no longer using the names in `config/data/ncr/ool_org_codes.yml` but are rather using the default names

pkarman:
The new class makes sense to me.

There are subtleties here with renaming of some variables like `code` and `org_code` that I am really not qualified to review. Maybe someone with more history with this product feature should also take a look?

Otherwise +1 from me.

jessieay:
@pkarman Thanks for the review!

Which renaming are you specifically referring to? I renamed many local vars but don't remember renaming anything "public"

I did, however, create a card for renaming `code` on the DB since we have both `org_code` and `code` on work orders and that was confusing. Waiting on Raphy for suggestions before moving forward, though: https://www.pivotaltracker.com/story/show/108581470

pkarman:
yes, it was the variable renaming that I was referring to. Hard for me to know if that introduces any (subtle) regressions since I don't know that part of the code history.

Sounds like @phirefly is the person to know.

jessieay:
OK, happy to wait on Raphy for review. As far as I know, the only behavior change for this is referenced here: https://github.com/18F/C2/pull/814/files#r45428888

Basically, we used to read the org codes from a separate YAML file and then replace the names for those org codes that are in the CSV with the names in YAML. 

So formerly we had

Real Estate Administration (REA)
Lease Project Management (LPMD)
Lease Execution (LED)

which is now

REAL ESTATE ADMINISTRATION DIVISION
LEASE PROJECT MANAGEMENT DIVISION
REAL ESTATE DELIVERY DIVISION

The org codes are the same, so this is not problematic if people search by org code, but I will wait on Raphy for feedback!

jessieay:
Closing this for now since we are going to move forward by putting Ncr Orgs into the database

PR #816, opened 2015-11-20
Dev admin can view proposals

jessieay:
* To test:
  visit admin
  click on proposals
  view-only options should be present (no edit/delete/create)

screen shot 2015-11-20 at 2 39 07 pm

### DIFF
--- a/app/admin/proposal.rb
+++ b/app/admin/proposal.rb
@@ -1,0 +1,14 @@
+ActiveAdmin.register Proposal do
+  actions :index, :show
+
+  permit_params :status
+
+  index do
+    column :id
+    column :status
+    column :name
+    column :public_id
+    column :requester
+    actions
+  end
+end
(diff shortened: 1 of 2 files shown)


jessieay:
@phirefly this logic implements https://github.com/18F/C2/pull/816

But it does not have a place for "reason" or send out cancellation emails like cancelling an email through the regular UI does. Do we need that functionality? Not sure what the exact use case for this PR is

pkarman:
@jessieay seems like maybe adding an automatic comment like "cancelled by _user_ via admin" or something?

jessieay:
@pkarman just caught up w Raphy about this. He meant for dev admins to be able to cancel via regular UI. Going to change this to be a read-only for proposals in admin and do the editing stuff in a different PR.

jessieay:
@pkarman ready for re-review (with new scope for story) :railway_car:

pkarman:
looks great.

PR #815, opened 2015-11-20
[Delivers #108372360] 18f approver can cancel

pkarman:
Story: https://www.pivotaltracker.com/story/show/108372360

This PR adds a couple methods to distinguish a *purchaser* from another kind of approver, and amends the gsa18f policy to allow non-purchaser approver on gsa18f procurement to cancel the proposal.

Testing:

* Verify who the gsa18f purchaser user is. That is controlled via Roles so you should be able to manage/view it from the `/admin` UI. Verify you will be able to login as that user.
* Create a gsa18f procurement
* Login as the approver and verify that you can cancel the procurement

PR #817, opened 2015-11-20
[Delivers #108485468] User with admin role can cancel any request

jessieay:
* Via regular UI
* To test:
  log in as admin
  visit proposal that you are not a requester or approver for (will have
  to find via rails console since it will not be in your dashboard)
  you should see option to cancel
  confirm that you can cancel

- story https://www.pivotaltracker.com/story/show/108485468

### DIFF
--- a/app/policies/proposal_policy.rb
+++ b/app/policies/proposal_policy.rb
@@ -11,7 +11,7 @@ class ProposalPolicy
   end
 
   def can_edit!
-    (@user.admin? || requester!) && not_approved! && not_cancelled!
+    (admin? || requester!) && not_approved! && not_cancelled!
   end
   alias_method :can_update!, :can_edit!
 
@@ -26,7 +26,7 @@ class ProposalPolicy
   alias_method :can_new!, :can_create!
 
   def can_cancel!
-    (@user.admin? || requester!) && not_cancelled!
+    (admin? || requester!) && not_cancelled!
   end
   alias_method :can_cancel_form!, :can_cancel!
 
@@ -54,8 +54,10 @@ class ProposalPolicy
   end
 
   def not_approved!
-    check(!@proposal.approved?,
-          "That proposal's already approved. New proposal?")
+    check(
+      !@proposal.approved?,
+      "That proposal's already approved. New proposal?"
+    )
   end
 
   def not_cancelled!
@@ -70,9 +72,15 @@ class ProposalPolicy
     @proposal.delegate?(@user)
   end
 
+  def admin?
+    @user.admin?
+  end
+
   def approver!
-    check(self.approver? || self.delegate?,
-          "Sorry, you're not an approver on this proposal")
+    check(
+      self.approver? || self.delegate?,
+      "Sorry, you're not an approver on this proposal"
+    )
   end
 
   def observer!
--- a/app/policies/proposal_policy.rb
+++ b/app/policies/proposal_policy.rb
@@ -11,7 +11,7 @@ class ProposalPolicy
   end
 
   def can_edit!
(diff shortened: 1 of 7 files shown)


jessieay:
Hey @pkarman this seems to have some overlap with https://github.com/18F/C2/pull/815

I marked as "WIP" because right now I found via QAing locally that I got a weird flash message:

"May not add observer belonging to a different organization."

It did cancel an 18F request when I was logged in as a dev admin with an NCR slug, but getting this flash message was confusing. 

Would it be crazy to remove that restriction?

app/policies/proposal_policy.rb
```diff
@@ -26,7 +26,7 @@ def can_create!
   alias_method :can_new!, :can_create!
 
   def can_cancel!
-    requester! && not_cancelled!
+    @user.admin? || requester! && not_cancelled!
```
pkarman: I think you need to group your logical booleans here:
```
(@user.admin? || requester!) && not_cancelled!
```

app/policies/proposal_policy.rb
```diff
@@ -11,7 +11,7 @@ def can_approve!
   end
 
   def can_edit!
-    requester! && not_approved! && not_cancelled!
+    @user.admin? || requester! && not_approved! && not_cancelled!
```
pkarman: need to group booleans here too.

pkarman:
I would be reluctant to remove that restriction, since I think it's there on purpose to prevent authz rules from breaking. See docs/authorization.md

pkarman:
Yes this does overlap with #815 . Might want to finish testing that one, merge and then rebase this because it will probably conflict.

jessieay:
@pkarman ok will do. I don't think the code itself will conflict but the logic will since `can_cancel` in the `gsa18f` policy will override `can_cancel` in `proposal` policy for GSA18F procurements...

pkarman:
@jessieay ah yes I see now what you mean. Let's circle back once this PR is complete and up-to-date with master. If we have gaps in our test coverage, this would be a good place to find them. It might be we need to add some tests to the `gsa18f` use cases.

jessieay:
ok @pkarman added a failing spec here to show that an 18F procurement cannot be cancelled by an admin right now: dae3c4ec43cca78798f998e12c859937f25a6a0b

Didn't fix it yet because want to think through how we can prevent this situation from sneaking in in the future.

How do you think we should handle this? Make sure we call `super` in all client_data pundit calls? Other ideas?

pkarman:
I don't think a `super` call will quite do it. The logic in the base class might throw an exception so unless we do something like:
```
begin
  super
rescue => err
   # ...
end
```

which just seems onerous.

I think the way we prevent this from slipping is having good test coverage that fails when we introduce regressions. Which you've now done! :)

jessieay:
@pkarman ok, updated code according to specs! In the future, I will try to remember to remind others to add test coverage for all use cases when updating the policy for one type of client data. I imagine this may become onerous as we get more clients, but with just 2 it should be fine

pkarman:
I will QA.

I wonder if we should just skip *all* policy checks for dev admins? I mean, since dev admins could do anything from the console anyway.

pkarman:
QA'd on c2-dev for https://c2-dev.18f.gov/proposals/237

I still get that flash message "May not add observer belonging to a different organization." (screenshot attached). I'd like to fix that before we merge.
screen shot 2015-11-24 at 8 23 19 am

pkarman:
flash bug fixed in 680d759

pkarman:
Ok I was able to fix the flash bug with a small patch + tests. QA is fine now against https://c2-dev.18f.gov/proposals/215

screen shot 2015-11-24 at 8 55 02 am

PR #818, opened 2015-11-20
optimize /admin/users endpoint to not load filters for all relationships

pkarman:
because `/admin/users` endpoints create filters for all relationships by default, page load time was 10+ seconds every time. This gets it down in the 1-2 second range.


jessieay:
looks good! will QA now

jessieay:
Looks good on dev! Much snappier. Guess we should do this for all of our admin models. :)

PR #820, opened 2015-11-23
rename elk services to workaround blue-green deploy bug

pkarman:
there's a bug in the cf-blue-green deploy that gets a false positive match based on the current ELK naming convention. I have re-named all our ELK services to workaround that bug.


jessieay:
wish there were a good way to write tests for this type of thing...

but LGTM. merging.

PR #821, opened 2015-11-24
[Delivers #108105962] Add "purchase type" to 18F procurements

jessieay:
* Required field
* Using ActiveRecord enum

* To test:
  log in as a user with 18f client slug and create a new request
  "purchase type" field should be required
  "purchase type" should show up on Proposals#show and in email body

* Story: https://www.pivotaltracker.com/story/show/108105962


pkarman:
I'm a big fan of enum, so +1 here.

Will QA tomorrow.

pkarman:
@jessieay tests failing on this branch.

jessieay:
@pkarman tests fixed at c9af5577e1b65dbcbdf33284445f90675a2181c0

Failures were related to ENV  var changes made in recent PR. We were showing slightly different button text in envs when that was set to TRUE. I removed the branching that caused this difference because it doesn't seem it was mission critical to me.

pkarman:
The button text was introduced in https://github.com/18F/C2/pull/617 via story https://www.pivotaltracker.com/n/projects/1149728/stories/98729742 so before we change it back, I'd like to sanity check that with the original people involved. 

cc / @afeld @phirefly

jessieay:
@pkarman  good idea -- and just to be clear, we are still include "TEST" in the subject line, this just removes the word "test" from the button when creating a request in sandbox mode

app/models/gsa18f/procurement.rb
```diff
@@ -10,6 +10,13 @@ class Procurement < ActiveRecord::Base
     URGENCY = DATA['URGENCY']
     OFFICES = DATA['OFFICES']
     RECURRENCE = DATA['RECURRENCE']
+    PURCHASE_TYPES = {
+      "Software" => 0,
+      "Training/Event" => 1,
+      "Office Supply/Miscellaneous" => 2
+    }
```
afeld: You might add a comment to the effect of "careful changing these numbers, as they will still be present in the database".

db/migrate/20151123235600_add_purchase_type_to_gsa18f_procurements.rb
```diff
@@ -0,0 +1,11 @@
+class AddPurchaseTypeToGsa18fProcurements < ActiveRecord::Migration
+  def change
+    add_column :gsa18f_procurements, :purchase_type, :integer
```
afeld: Thoughts about using Postgres's [`enum`](http://www.postgresql.org/docs/9.3/static/datatype-enum.html) data type, rather than an integer? Seems like it might make querying a bit more straightforward (since it would use the string rather than an arbitrary integer). Also makes it easier to avoid the issue of the names/numbers accidentally getting changed.

app/models/gsa18f/procurement.rb
jessieay: Since this is for activerecord enums and is using standard enum functionality, I would assume that people adding to the codebase would know that these integers are in the db / should not be changed if they know how AR enums work. Do you think that is not the case?

db/migrate/20151123235600_add_purchase_type_to_gsa18f_procurements.rb
```diff
@@ -0,0 +1,11 @@
...
+
+    execute <<-SQL
+      UPDATE gsa18f_procurements SET purchase_type = 0 WHERE purchase_type IS NULL;
+      SQL
```
afeld: Seems that this is done within `change_column_null`, so don't think it's necessary.

http://apidock.com/rails/ActiveRecord/ConnectionAdapters/PostgreSQL/SchemaStatements/change_column_null

Ditto for the migration above.

app/models/gsa18f/procurement.rb
```diff
@@ -24,6 +31,7 @@ def self.purchase_amount_column_name
       greater_than_or_equal_to: 1
     }, presence: true
     validates :product_name_and_description, presence: true
+    validates :purchase_type, presence: true
     validates :recurring_interval, presence: true, if: :recurring
 
     def self.relevant_fields(recurring)
```
afeld: Do we need to add the `purchase_type` to this method so it gets displayed on the `show` page and in the emails?

db/migrate/20151123235600_add_purchase_type_to_gsa18f_procurements.rb
jessieay: You know, I haven't used it before! It looks like it is nice but since the behavior would be different than that standard AR enum behavior (which by now most people in the Rails world are familiar with), I am a little hesitant to introduce it here. The only good blog post I found on using postgres enum with activerecord enum is here: http://www.sitepoint.com/enumerated-types-with-activerecord-and-postgresql/

Another consideration, which I spoke with Peter about yesterday, is the behavior of enums in admin. I've dealt with crazy stuff when trying to use Rails admin with enums in the past. I am wondering if ActiveAdmin has good support for Postgres enums...

afeld:
> I'd like to sanity check that with the original people involved

I defer to @phirefly on that one. My gut instinct was that we went a little overkill in terms of the app shouting "THIS ISN'T A REAL REQUEST" all over the place, but clearly people were getting confused without it. +0 on that change.

db/migrate/20151123235600_add_purchase_type_to_gsa18f_procurements.rb
jessieay: I thought it would, too! but when I ran this migration locally it did not set the default for existing records.

app/models/gsa18f/procurement.rb
jessieay: see `spec/features/gsa18f/procurements/show_spec.rb` :)

app/models/gsa18f/procurement.rb
afeld: Changing them without thinking about it would be an easy mistake to make, and would be a nasty data inconsistency problem after the fact, so I'd err on the side of caution.

db/migrate/20151123235600_add_purchase_type_to_gsa18f_procurements.rb
afeld: `/shrug`

db/migrate/20151123235600_add_purchase_type_to_gsa18f_procurements.rb
afeld: Hmmmm, that method really needs some documentation :unamused: I'd simplify all of this to:

```ruby
add_column :gsa18f_procurements, :purchase_type, :integer, default: 0, null: false
```

app/models/gsa18f/procurement.rb
afeld: :+1:

phirefly:
I'd like to see reports in Mandrill on mail sent from c2-staging before to 1) make sure that we it hasn't happened in a while and 2) have a quick way of verifying in the future. Since moving away from gmail, i haven't been checking

phirefly:
@pkarman Go ahead and update the button text. I found out how to search in mandrill. TY!

db/migrate/20151123235600_add_purchase_type_to_gsa18f_procurements.rb
jessieay: I tried that too and it didn't work because there were existing records with a `null` value and the default was not being applied to them :( :(

pkarman:
I will QA

db/migrate/20151123235600_add_purchase_type_to_gsa18f_procurements.rb
afeld: Ah, gotcha. Maybe this then?

```ruby
change_column_null :gsa18f_procurements, :purchase_type, false, 0
```

Again, ditto for the migration above.

app/models/gsa18f/procurement.rb
jessieay: ok, added tests for documentation here: 6d2481ae2ce0e1667a485a4d9cc7dd4961d17467 

I think this is a good way to ensure people don't change this order because changing would also require a change in specs -- what do you think?

db/migrate/20151123235600_add_purchase_type_to_gsa18f_procurements.rb
jessieay: I think I tried that too. I really did try a lot of things. Do you think the migration as-is is really in need of shortening?

afeld: Would just be making it more concise – no worries.

app/models/gsa18f/procurement.rb
afeld: :+1:

jessieay:
@pkarman let me know if you need anything else in order to QA

pkarman:
- [x] client-side validation works (Purchase Type required)
- [x] Purchase Type appears on Proposal page and in email

PR #822, opened 2015-11-24
Disable sandbox warning in test env

jessieay:
* Turn on when needed


pkarman:
I like this approach. Will QA tomorrow.

pkarman:
tests passing so :ship:

PR #823, opened 2015-11-24
[Delivers #108771232] tier one delegate authz

pkarman:
Story: https://www.pivotaltracker.com/story/show/108771232

This PR creates a new column `completed_by_id` on the `Step` model.

Testing:

* create a proposal with an approver that has two or more delegates.
* approve the proposal as a delegate
* verify you can still view the proposal as one of the other delegates
* verify that the approval records the delegate as the person who approved


app/models/step.rb
```diff
@@ -10,6 +10,7 @@ class Step < ActiveRecord::Base
 
   belongs_to :user
   belongs_to :proposal
+  belongs_to :completed_by, class_name: "User"
```
jessieay: what about `belongs_to :completer` ? That seems more consistent with the other AR relationships we have in this class and elsewhere in the project

db/migrate/20151123211837_add_completed_by_to_steps.rb
```diff
@@ -0,0 +1,13 @@
+class AddCompletedByToSteps < ActiveRecord::Migration
+  def up
+    add_column :steps, :completed_by_id, :integer
```
jessieay: per comment above, wonder if this should be `completer_id` ?

app/models/step.rb
```diff
@@ -33,6 +34,10 @@ def pre_order_tree_traversal
     [self] + self.child_approvals.flat_map(&:pre_order_tree_traversal)
   end
 
+  def completer
```
jessieay: oh interesting, since we already have a `user` relation, what about just adding a `completed_at` timestamp to `step` and setting that when it is completed? Seems redundant to have 2 foreign keys going to the same person, as far as I can tell...

spec/features/ncr/work_orders/approve_spec.rb
```diff
@@ -21,6 +21,21 @@
     end
   end
 
+  scenario "allows delegate to approve work order" do
+    Timecop.freeze do
```
jessieay: why do we need Timecop here?

jessieay:
@pkarman let me know what you think re: my feedback on making this a timestamp rather than another foreign key. I wonder if I am missing something...

app/models/step.rb
pkarman: We already have a `approved_at` timestamp column. AFAIK the plan is to rename it to `completed_at` in some future naming cleanup as part of the whole `Approval` => `Step` conversion.

Our problem is not detecting when/if a step is completed. Our problem is that 2/3 of NCR approvals are assigned to non-human mailbox Users, and then actually completed via real-human delegate Users, and we want to preserve the original non-human link as well as the actual real-human link so that authorization is preserved.

I.e., for 100% of Tier 1 and Tier 2 approvals, the 2 foreign keys will go to 2 different User records.

spec/features/ncr/work_orders/approve_spec.rb
pkarman: not sure. cargo-culted that layout from the test above it. The `begin` block sets the time 10 hours in the past so presumably there is some time flow control going on. My guess would be timezone-related checks, since I waded through that mess a few months ago.

app/models/step.rb
pkarman: I'm not attached to the naming convention I used. I just wanted to have a method where the user who completed the step could be accessed via a standard method name that was backwards compatible. So I started with the `completer` method and worked backwards in terms of naming.

app/models/step.rb
jessieay: AH yes! Makes sense re: delegate. In that case, let's do with `completer_id` so we can have a `belongs_to: :completer` relation. That seems good!

spec/features/ncr/work_orders/approve_spec.rb
jessieay: do you think it's worth it to try and remove while we are here? It is not clear to me from reading these tests why they would need to set up approvals and observers 10 hours ago.

pkarman: Here's the commit when those tests went in https://github.com/18F/C2/commit/73a4d386 which contains a comment with a link that hints at why the Timecop stuff is there.

I would leave them as is. It's lamentable there are not better comments as to why, but it obviously had some purpose at the time or @nickbristow wouldn't have gone to the trouble.

jessieay: Fine to keep in other specs, I guess my question is: does this test pass without freezing time? And if so, should we write it without time freezing?

pkarman:
@jessieay association has been renamed.

spec/features/ncr/work_orders/approve_spec.rb
jessieay: @pkarman did you see this comment? thoughts on trying w/out timecop?

pkarman:
@jessieay I have removed the Timecop wrappers.

spec/features/ncr/work_orders/approve_spec.rb
```diff
@@ -21,6 +21,37 @@
...
+    approver.save!
+
+    login_as(delegate)
+    visit proposal_path(ncr_proposal)
+    click_on('Approve')
+    expect(current_path).to eq(proposal_path(ncr_proposal))
+
+    login_as(delegate_two)
+    visit proposal_path(ncr_proposal)
+    expect(page.status_code).to eq(200)
```
jessieay: can we move this spec to be a `request` spec per description of spec types in https://www.pivotaltracker.com/story/show/105780488

app/models/user.rb
```diff
@@ -20,6 +20,11 @@ class User < ActiveRecord::Base
   has_many :outgoing_delegations, class_name: 'ApprovalDelegate', foreign_key: 'assigner_id'
   has_many :outgoing_delegates, through: :outgoing_delegations, source: :assignee
 
+  has_many :incoming_delegations, class_name: 'ApprovalDelegate', foreign_key: 'assignee_id'
```
jessieay: surprised code climate isn't yelling about single quotes. Seems like that integration is a bit flaky these days...

app/models/user.rb
```diff
@@ -20,6 +20,11 @@ class User < ActiveRecord::Base
...
+  has_many :incoming_delegates, through: :incoming_delegations, source: :assigner
```
jessieay: why are we defining these now? do not appear to be used anywhere else in this PR, as far as I can tell

app/models/step.rb
```diff
@@ -33,6 +34,10 @@ def pre_order_tree_traversal
     [self] + self.child_approvals.flat_map(&:pre_order_tree_traversal)
   end
 
+  def completed_by
```
jessieay: unit test for this?

jessieay: oh wait, there is one, just didn't see it because it is not wrapped in a `describe "#completed_by"` block -- can we add that ?

app/models/user.rb
pkarman: I used them in order to diagnose the original story description. Since they seemed to be missing as symmetrical associations for `outgoing_delegations` I left them there as useful later on (as when exposed via /admin e.g.).

spec/features/ncr/work_orders/approve_spec.rb
pkarman: I added some more expectations to the spec so that now it qualifies as a feature spec.

jessieay: ok sounds good! in the future I think we are going to try to move away from having status code in feature spec expectations since that isn't a "user-facing feature" but I am cool with keeping this here for now

pkarman:
fwiw codeclimate appears to find no errors but still showing in-progress

jessieay:
looks great! will QA Now

jessieay:
Ok, created https://c2-dev.18f.gov/proposals/243

assigned user with 2 delegates as the approver
approved as one delegate
confirmed that I can view the proposal as the other delegate even after approval
confirmed via ssh that the completer_id is the user id of the delegate who approved the propsoal

the only edge case I discovered is that if I click "approve" as the second delegate (after the first delegate has already approved the proposal), I get an authorization error. Would it make sense to create a separate card to deal with this? I would expect that if I try to approve a proposal that is already approved it would just tell me "Hey, this is already approved!"

pkarman:
that's a good QA catch, @jessieay 

I think we should disable the Approve button if it is already approved. I'll add that.

jessieay:
@pkarman this was via the "approve" button in the email, FYI. Sorry for not being more specific

pkarman:
ah that is different. That's the original problem report, too.

I think we should handle the case more gracefully when using an approval token > 2x when the proposal is already approved. Let's make a separate story for that though.

pkarman:
I will resolve conflicts on this shortly.

jessieay:
@pkarman what is the QA step for when to confirm behavior "When I try to approve an already approved as a delegate..."

PR #824, opened 2015-11-24
[Delivers #108578654] fix Hakiri XSS warning

pkarman:
Story: https://www.pivotaltracker.com/story/show/108578654

### DIFF
--- a/app/views/shared/_table.html.haml
+++ b/app/views/shared/_table.html.haml
@@ -3,12 +3,14 @@
   %thead
     - # @todo: "header" class shouldn't be required (we're in thead)
     %tr.header
+      - # untaint user input
+      - req_query_params = request.query_parameters.slice("tables")
       - container.columns.each do |column|
         %th{class: column.sort_dir}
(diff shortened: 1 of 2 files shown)


jessieay:
THANK YOU for fixing this

One question about the XSS fix: is this actually fixing the issue, or is it tricking hakiri into not recognizing this XSS issue? If the latter, is there some way to mark this error as "acknowledged" within Hakiri rather than doing a workaround within the code? I know that when Code Climate finds XSS errors you can mark them as "not a concern" ...

pkarman:
presumably we could use `.brakeman.ignore` to whitelist this particular use case. Maintaining that kind of file feels onerous to me.

FWIW, I tried to escape the input in a variety of ways per the many recommendations I found via google. In the end it was easier just to rewrite it to avoid `link_to` altogether, which seems to be the trigger for Brakeman. The warning really is spurious: we are simply passing through the GET parameters, which automatically get URL-encoded anyway to avoid XSS issues.

jessieay:
OK sounds good. MERGING

PR #825, opened 2015-11-24
add proposals panel to /admin/users/:id page

pkarman:
small developer enhancement

PR #826, opened 2015-11-24
Roles conversion test

pkarman:


jessieay:
this looks good! but it also contains the changes I was testing in https://github.com/18F/C2/pull/825 -- should we close that and I will just QA this?

pkarman:
no I branched off the wrong base. closing.

PR #828, opened 2015-11-25
[Delivers #106953324] Do not allow inactivated users to log in

jessieay:
* Redirect inactivated users to root path with flash message
* Flash message contains link but that is currently not working

* Story: https://www.pivotaltracker.com/story/show/106953324

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -80,7 +80,7 @@ class ApplicationController < ActionController::Base
       return_to_param = make_return_to("Previous", request.fullpath)
       session[:return_to] = return_to_param
       redirect_to root_url(return_to: return_to_param)
-    elsif current_user.inactivated?
+    elsif current_user.deactivated?
       redirect_to feedback_path
     end
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -76,12 +76,13 @@ class ApplicationController < ActionController::Base
 
   def authenticate_user!
     if not_signed_in?
-      flash[:error] = 'You need to sign in for access to this page.'
+      flash[:error] = "You need to sign in for access to this page."
       return_to_param = make_return_to("Previous", request.fullpath)
       session[:return_to] = return_to_param
       redirect_to root_url(return_to: return_to_param)
-    elsif session[:return_to]
-      session.delete(:return_to)
+    elsif current_user.inactivated?
+      redirect_to root_url
+      flash[:error] = "You are not allowed to login because your account has been inactivated. Please <a href='/feedback'>contact an administrator</a>."
     end
   end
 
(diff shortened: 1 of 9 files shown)


app/controllers/application_controller.rb
```diff
@@ -76,12 +76,13 @@ def signed_in?
 
   def authenticate_user!
     if not_signed_in?
-      flash[:error] = 'You need to sign in for access to this page.'
+      flash[:error] = "You need to sign in for access to this page."
       return_to_param = make_return_to("Previous", request.fullpath)
       session[:return_to] = return_to_param
       redirect_to root_url(return_to: return_to_param)
-    elsif session[:return_to]
```
jessieay: @pkarman I removed this because I wasn't sure why it was here - thoughts? Happy to re-introduce and write a test for it if we need it :)

pkarman: @jessieay I opened #829 to show that your change is ok. IIRC it was vestigal from the work that spawned the rspec I modified in #829.

app/controllers/application_controller.rb
```diff
@@ -76,12 +76,13 @@ def signed_in?
...
-      session.delete(:return_to)
+    elsif current_user.inactivated?
+      redirect_to root_url
```
pkarman: Maybe instead of redirecting to the root url, redirect immediately to the feedback URL?  That way no link is needed.

config/routes.rb
```diff
@@ -1,6 +1,6 @@
 C2::Application.routes.draw do
   ActiveAdmin.routes(self)
-  root :to => 'home#index'
+  root to: "home#index"
```
pkarman: If you're going to change that one, why not change them all and get it done with?

app/controllers/application_controller.rb
jessieay: I like that idea!

jessieay: updated at 6fb4814dcdf20548d9c5f11dc3814313fc7dd87d

config/routes.rb
jessieay: if you insist... :imp: 83f2a93ccf43c9d72dd8b207a2690b8360db6b1f

jessieay:
@pkarman updates made based on your feedback!

jessieay:
@pkarman this is ready for re-review

pkarman:
looks good. I will QA.

pkarman:
QA looks great.

One wordsmithing nit: I think our messaging should be "your account has been deactivated" rather than "inactivated". We might say "your account is inactive" but "inactivated" feels a little awkward.

See http://www.answers.com/Q/What_is_the_difference_between_the_words_inactivate_vs_deactivate

jessieay:
@pkarman I like it! Found "inactivated" awkward as I was writing this code.

Changed in 0df65a0c486896a1036c18db5fbb69f118f57a27

pkarman:
great, thanks. Looks like just one missed test.

jessieay:
@pkarman fixed ;)

PR #829, opened 2015-11-25
add test for session[:return_to] to verify it is cleaned up on use

pkarman:


jessieay:
this looks good - OK if I just add these changes to https://github.com/18F/C2/pull/828 ?

pkarman:
I opened this as a PR against #828 so you should have able to just use the github green buttons.

PR #830, opened 2015-11-27
[Delivers #98247028] work around required form field bug

pkarman:
Story: https://www.pivotaltracker.com/story/show/98247028

Rather than fixing orphaned attachment files, prevent invalid form from being submitted in the first place.

selectize.js has known bug with form fields marked as required being ignored.
https://github.com/brianreavis/selectize.js/issues/733

This workaround patches vendor copy in-place per https://github.com/brianreavis/selectize.js/issues/733#issuecomment-148835703

To test:
- [ ] GET /ncr/work_orders/new
- [ ] fill out each field marked required, one at a time, and after each click `Submit for approval` button. You should be prevented from submitting the form till all required fields are complete.


jessieay:
would it be possible to write a feature spec that tests this change?

pkarman:
I tried, but I am failing to get the form validation to trigger. Here's what I tried:
```ruby
    scenario "requires an approver via client-side validation", :js do
      login_as(requester)
      visit new_ncr_work_order_path
      fill_in "Project title", with: "buying stuff"
      choose "BA60"
      find("input[aria-label='Building number']").native.send_keys("BillDing")
      find("input[aria-label='Vendor']").native.send_keys("ACME")
      fill_in "Amount", with: "123"
      click_on "Submit for approval"
      expect(current_path).to eq(new_ncr_work_order_path)
      expect(page).to have_content("Please fill out this field")
    end 
```

jessieay:
I test manually and I can see why this might not work - not able to inspect element and see that text anywhere (doesn't help that it disappears as soon as you click anywhere)

Rather than testing that ""Please fill out this field"", why not just leave it at testing for the current path, which you have aready

pkarman:
Ah, I was under the impression that we were somehow using JS for the form validation, but we are instead relying on HTML5 form validation. Some searching tells me that PhantomJS + Poltergeist does not use HTML5 form validation, so we cannot write rspec tests for that feature (it is browser-specific).

jessieay:
I didn't realize we were using HTML 5 form validations either! 

Do you know if we have some type of SLA around supporting different browsers? Perhaps @afeld would know...

Looks like HTML 5 form validation is not supported on IE 8 or 9: http://caniuse.com/#feat=form-validation

And this is a graph of our users by browser via New Relic -- we def have some IE 8 and 9 users, but by no means a large %:

screen shot 2015-11-30 at 3 09 52 pm

Since we are already using HTML 5 form validation, I see no reason to hold up this PR. Just thought I'd raise the issue while it is top of mind.

Will QA noq

vendor/assets/javascripts/selectize.js
```diff
@@ -2652,7 +2652,8 @@
 			var invalid, self = this;
 			if (self.isRequired) {
 				if (self.items.length) self.isInvalid = false;
-				self.$control_input.prop('required', invalid);
+				self.$control_input.prop('required', self.$input.val() == '');
+				self.$input.prop('required', false);
```
afeld: Is this upgrading Selectize, or a manual tweak? Would strongly suggest avoiding modifying 3rd-party libraries, and if we do, put a note at the top with the corresponding pull request to the original, or at least explaining the change.

afeld: Oh sorry, just read the rest of the PR. Please add a comment at the top with the link to that comment!

afeld:
> PhantomJS + Poltergeist does not use HTML5 form validation

That would surprise me, since PhantomJS uses a recent version of WebKit.

> I didn't realize we were using HTML 5 form validations either!

:rocket:

> Do you know if we have some type of SLA around supporting different browsers?

No, but standard (industry) practice is to support any browser that has >= 3-5% of visitors. We're lucky, in that GSA PCs have Chrome installed by default (and double-clicking the "Mail" shortcut on the desktop opens it), and are auto-upgraded to IE 11. The stats from Google Analytics:

screen shot 2015-11-30 at 5 50 05 pm

IE breakdown:

screen shot 2015-11-30 at 5 50 36 pm

[*source*](https://analytics.google.com/analytics/web/?authuser=1#report/visitors-browser/a48605964w99418896p103376476/)

> HTML 5 form validation is not supported on IE 8 or 9

Right – I knew that going into it, though maybe didn't communicate it. We could add a polyfill for client-side validation in old browsers, but since

* it's such a small share of browsers without support, and
* it's just a nice-to-have (since the corresponding server-side validation should be in place)

I wouldn't spend too much time on worrying about it.

jessieay:
@pkarman I QA'd this on c2-dev --works like a charm!

you think we should try to get an automated spec in per @afeld 's comments on phantom JS using webkit? Will defer to you!

vendor/assets/javascripts/selectize.js
pkarman: Thanks, @afeld -- comment added in 08972b3

pkarman:
I am basing my claim on HTML5 form validation not working with our current phantomjs set up on my failing test, and this http://stackoverflow.com/questions/29795743/html5-form-validation-with-capybara-phantomjs-and-poltergeist

pkarman:
test showing phantomjs does not do html5 form validation added in 29e97f8

If QA works, I would merge and move on.

jessieay:
there are some merge conflicts now -- happy to merge once those are fixed!

pkarman:
@jessieay updated now

PR #832, opened 2015-11-27
[Delivers #108595344] ba60 building optional

pkarman:
Story: https://www.pivotaltracker.com/story/show/108595344

### DIFF
--- a/app/views/ncr/work_orders/_form.html.haml
+++ b/app/views/ncr/work_orders/_form.html.haml
@@ -36,21 +36,3 @@
     = f.submit class: "form-button"
     - if @model_instance.persisted?
       = link_to "Discard Changes", proposal_url(@model_instance.proposal)
-
-:javascript
-  // toggle "required" flag on Building depending on Expense Type
-  $(document).ready( function() {
-    $('.expense-type .radio input').click(function() {
-      var radio = $(this);
-      var buildingPicker = $('.form-group.ncr_work_order_building_number');
-      if (radio.attr("id").match(/_ba60$/)) {
-        buildingPicker.removeClass("required");
-        buildingPicker.find("*").removeClass("required");
-      }
-      else {
-        buildingPicker.addClass("required");
-        buildingPicker.find("*").addClass("required");
-      }
-    })
-  });
-
--- a/app/views/ncr/work_orders/_form.html.haml
+++ b/app/views/ncr/work_orders/_form.html.haml
@@ -12,7 +12,7 @@
   = simple_form_for @model_instance, html: { multipart: true } do |f|
     = f.input :project_title
     = f.input :description
-    = field_set_tag "Expense type / Program", class: 'required' do
+    = field_set_tag "Expense type / Program", class: 'expense-type required' do
       = expense_type_radio_button(f, 'BA60')
       = expense_type_radio_button(f, 'BA61')
       = f.input :emergency, disabled: @model_instance.persisted?, wrapper_html: { data: { filter_key: 'expense-type', filter_value: 'BA61' } }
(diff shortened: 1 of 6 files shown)


app/views/ncr/work_orders/_form.html.haml
```diff
@@ -36,3 +36,21 @@
     = f.submit class: "form-button"
     - if @model_instance.persisted?
       = link_to "Discard Changes", proposal_url(@model_instance.proposal)
+
+:javascript
```
jessieay: Might make sense to put this js in a `.js` file and just require it here -- that is where I would expect to find javascript, might unintuitive for others to look for it in here, too

db/schema.rb
```diff
@@ -141,7 +141,7 @@
     t.integer  "client_data_id"
     t.string   "client_data_type", limit: 255
     t.integer  "requester_id"
-    t.string   "public_id"
+    t.string   "public_id",        limit: 255
```
jessieay: git checkout since this is unrelated to this PR? (running migrations also changes this for me -- not sure why it keeps going back and forth

spec/features/ncr/work_orders/create_spec.rb
```diff
@@ -159,6 +159,19 @@
       expect(page).to have_content('$2,000 for construction')
     end
 
+    scenario "selecting Expense Type toggles Building required flag", :js do
+      login_as(requester)
+      visit "/ncr/work_orders/new"
```
jessieay: `new_ncr_work_order_path` ? (I've been slowly moving specs over to that way of calling paths)

spec/models/ncr/work_order_spec.rb
```diff
@@ -206,6 +206,13 @@
       wo = build(:ncr_work_order, building_number: nil)
       expect(wo.building_id).to be_nil
     end
+
+    it "requires unless expense_type == BA60" do
```
jessieay: do you think we should break this out into two specs? I generally aim for one expectation per spec for readability sake (eg: there is one spec description per scenario)

spec/features/ncr/work_orders/create_spec.rb
```diff
@@ -159,6 +159,19 @@
       expect(page).to have_content('$2,000 for construction')
     end
 
+    scenario "selecting Expense Type toggles Building required flag", :js do
```
jessieay: It seems like this one test is testing a few different things. If it were me, I might break out out like this:
```
context "expense type is BA60" do
  scenario "building id is not required" do
     select_expense_type("ba60")
    find("#ncr_work_order_expense_type_ba60").click
    expect_building_id_not_to_be_required
  end
end

context "expense type is not BA60" do
  scenario "building id is required" do
    ["ba61", "ba80"].each do |expense_type|
      select_expense_type(expense_type)
      expect_building_id_to_be_required
    end
  end
end

context "selects BA60 and then unselects BA60" do
  scenario "building id is required" do
    select_expense_type("ba60")
    select_expense_type("ba80")
     expect_building_id_to_be_required
  end
end

def select_expense_type(expense_type)
  find("#ncr_work_order_expense_type_#{expense_type}").click
end

def expect_building_id_not_to_be_required
  expect(find('.ncr_work_order_building_number input')['class']).to_not match('required')
end

def expect_building_id_to_be_required
  expect(find('.ncr_work_order_building_number input')['class']).to match('required')
end
```

what do you think?

pkarman:
Thanks @jessieay I have made all your suggested changes.

jessieay:
looks beautiful -- will QA now!

jessieay:
Just tested on c2-dev -- worked like a charm! Merging

PR #833, opened 2015-11-28
Refactor to prep for moving organizations to DB

jessieay:
* Been working on moving NCR orgs to the database. In the process, I
  decided that the relationship betwen `ncr_work_order`s and
  `ncr_organization`s should actually be named `ncr_organization` rather
  than `organization`. By making this change now, the move to AR will be
  smaller

*  In the process of that PR, I also found that the use of I18n for
  dealing with email subject logic was confusing. This clarifies that
  logic and puts it under test at the unit level


pkarman:
No objections to the rename or the other clean up here.

+1 on the code.

PR #834, opened 2015-11-30
[Delivers #101710542] Unify naming of client-specific code

jessieay:
* Previously, used `UseCase`, `model_instnace`, and `proposal_delegate`
* Move to variations of `client_data` everywhere
* Story: https://www.pivotaltracker.com/story/show/101710542

### DIFF
--- a/app/models/concerns/client_data_type.rb
+++ b/app/models/concerns/client_data_type.rb
@@ -4,11 +4,11 @@ module ClientDataType
   included do
     Proposal::CLIENT_MODELS << self
 
-    has_paper_trail class_name: 'C2Version'
+    has_paper_trail class_name: "C2Version"
 
     has_one :proposal, as: :client_data
     has_many :steps, through: :proposal
-    has_many :individual_steps, ->{ individual }, class_name: 'Steps::Individual', through: :proposal
+    has_many :individual_steps, -> { individual }, class_name: "Steps::Individual", through: :proposal
     has_many :approvers, through: :individual_steps, source: :user
     has_many :observations, through: :proposal
     has_many :observers, through: :observations, source: :user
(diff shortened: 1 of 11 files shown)


app/models/concerns/client_data_type.rb
```diff
@@ -1,14 +1,14 @@
-module ProposalDelegate
+module ClientDataType
```
jessieay: This is the rename I struggled with most. thoughts?

pkarman: `ClientDataType` suggests it is an instantiate-able class. How about `ClientDataMixin` or `ClientDataConcern` or `ClientDataExtensions` ?

pkarman:
I agree that `ClientDataType` feels awkward. All the other renames seem sensible.

app/models/concerns/client_data_type.rb
jessieay: I feel like `Mixin` and `Concern` are good at communicating the purpose of this module, but it's also a bit repetitive - like naming a class `WorkOrderClass`. The example I see for ActiveModel Concerns is always `Taggable` -- trying to think of names that are in that spirit

What about just calling it `ClientData` ?

pkarman: Simply `ClientData` is misleading. It isn't client data. It's the basic association definitions and validation rules that ClientData-type model classes should implement. It's like an abstract class for client data models, but as a concern.

`Mixin` has precedent in `PurchaseCardMixin` and doesn't repeat anything -- there is no `mixin` keyword in Ruby afaik.

jessieay: Ok, going with`ClientDataMizin

PR #835, opened 2015-12-02
[WIP] Remove parallel approvals

jessieay:
When working through all of the possible user flows today, I realized it was impossible to document how parallel approvals work because we are not using them. I realize there is a potential future use for them, but for now this is all completely unused code that I suspect is becoming less and less useful over time because we are building a system that works for serial approvals.

This is a WIP because I didn't want to spend more than 30 mins on this before discussing further with team. To fully remove, would also remove `flow` field from DB and move logic from `Steps::Serial` to `Step`


pkarman:
@cmc333333 can you give some insight on the inspiration for the parallel approval paradigm? I know you just wrote all this a few months ago.

cmc333333:
The parallel approval type allowed a flow that 18f's approvals desired - "alice OR bob". For more history, C2's original client ("navigator") wanted approvals to be "alice AND bob" but in any order. Using approval chains was a mechanism to replace the "flow" field but maintain its functionality.

See https://docs.google.com/a/gsa.gov/presentation/d/1aK1GX9sSmbIpOmXC5MJYbPSOGRM9byMP4VifTQjFbX4/edit?usp=sharing for a presentation about approval chains (be sure to check the speaker notes) :)

jessieay:
@cmc333333 do you happen to remember who from 18F requested this? It is not a feature that we've used so far for 18F or NCR.

jessieay:
@pkarman from what I can see, it seems that the concept of "flow" (linear vs parallel) has been in this app for over a year.

jessieay:
@afeld thoughts on this?

pkarman:
All the stories I found in PT assume the presence of the "parallel" flow type, but I could not find the original story, which suggests to me that its origins are in the misty past.

@jessieay I was referring to @cmc333333's recent rewrite of all the approval flows in https://www.pivotaltracker.com/story/show/98024644

jessieay:
@pkarman ah interesting, I didn't see that card. The card says that "we're running into more complicated systems" -- wondering if anyone remembers where those systems were?

cmc333333:
Here's one about an `OR`: https://www.pivotaltracker.com/story/show/95710992 . I don't recall the specific use case for a multitiered approval; but I vaguely recall the desire to have multiple approvers play the same "role". As in any of these three people can be the first approver, any of these four can be the second, etc.

pkarman:
I'm :-1: on this PR. Let's keep the feature in place, especially since we haven't yet started a national rollout and discovered whether it really isn't necessary.

jessieay:
@pkarman ok will close. I know this stuff was built before our time, but in the future I think we should wait on adding features like this until we are actually going to use them for user-facing features. 

Chatted with Raphy about this yesterday on Slack and he said that 18F *does* want parallel approvers, but since we don't have the UI fleshed out for the parallel use case, we are just going to have them use delegation. 

I wouldn't be surprised if it is quite a bit longer until we use parallel steps. And when that time comes, it may be the case that serial steps have become so ingrained in the way the app works that we will have to do a major overhaul to get parallel steps working (despite the fact that they already exist as a concept in the code)

afeld:
Yeeeeeah, not sure what the right answer is here. The linear approvals work in _theory_...the missing features might be notifications/warnings like "someone else already approved this, so you don't need to." +0 on the change I guess, especially until we have more concerted effort on the 18F use case.

> would also remove `flow` field

I think we can do this regardless.

cmc333333:
> I think we should wait on adding features like this until we are actually going to use them for user-facing features.

Food for thought: this is a chicken and egg problem. Arguably (and I do argue this :) ), if we had spent more time adding generalized features like this one, we'd have had more clients. Instead, we didn't follow through; there was no effort put into a user facing component and before too long, 18F stopped using C2 internally.

jessieay:
@cmc333333 I think that if we had added this as a part of creating that feature for 18F, that would have been ideal. 

I am not sure why we created the egg and then didn't get to raising the chicken -- my guess is  shifting priorities. 

Agreed that we need to create backend support for features before we can create frontend, but when there is a several month delay between creating a feature and using it, that feels off to me.

PR #836, opened 2015-12-02
Upgrade Nokogiri

jessieay:


afeld:
Any particular reason for only upgrading Nokogiri?

Anyway, :shipit:

jessieay:
@afeld I had it upgraded on another project and using different versions was creating a bit of headache for my vim-rspec setup. Figured I could do a fancy workaround or just upgrade :)

PR #839, opened 2015-12-02
Rename CommunicartMailer -> Mailer

jessieay:
* Part of general move from Communicart to C2
* Move cancellation mailer methods into their own class
* Remove env vars from RolesConversion


app/mailers/mailer.rb
```diff
@@ -1,11 +1,8 @@
-class CommunicartMailer < ApplicationMailer
+class Mailer < ApplicationMailer
```
afeld: What about `ProposalActivityMailer` or something?

afeld: This name feels a bit too generic.

afeld: Slash maybe we should split it up more.

jessieay: It is generic. Right now it is really just the base mailer for this project. In rails apps, I usually name that `Mailer`. 

I agree that we should move more methods out of this and into smaller mail classes but was trying to keep this PR under 400 changes per my usual rule of thumb. 

What do you think about keeping this name for now as the base mailer? 

I am not sure that a name like `ProposalActivityMailer` would improve the clarity of purpose of this class given the generic nature of `Activity` -- would rather stick with a generic name and then create more specific classes like we have with `CancellationMailer` and `ReportMailer`

pkarman:
I'm :+1: for this as-is.

app/mailers/mailer.rb
JuanitoFatas: `ApplicationMailer` would be a good choice for the base mailer. Matches with `ApplicationController`, `ApplicationJob`, and `ApplicationHelper`. :blush:

jessieay: Hi @JuanitoFatas -- we do have an `ApplicationMailer` in this project.

Our other mailers inherit from it and it basically just contains helper methods at the moment. 

I'd rather rename this mailer to `Mailer` and then move stuff out of it into more specific mailers than put everything in ApplicationMailer since it's nice to be able to easily navigate that file in order to find helper the methods.

Thank you for weighing in!

JuanitoFatas: :+1:

pkarman:
@jessieay I will merge when conflicts are resolved.

jessieay:
@pkarman rebased! tests still running

app/mailers/mailer.rb
afeld: > It is generic

It's generic, but only for notifications around Proposals. The `ApplicationMailer` is the generic base mailer class, which is what all of our mailers (including ones that aren't Proposal-centric, like the `FeedbackMailer`) have as an ancestor. The fact that `Mailer` is a subclass of `ApplicationMailer` here is strange to me.

> I agree that we should move more methods out of this and into smaller mail classes but was trying to keep this PR under 400 changes

:+1:

PR #840, opened 2015-12-02
[Delivers #108876756] client summary

pkarman:
Story: https://www.pivotaltracker.com/story/show/108876756

This creates a new endpoint `/summary` that contains fiscal year summary of all proposals broken down by status. The endpoint is limited to proposals matching the current_user `client_slug` and is only available to admin and client_admin users.


app/assets/stylesheets/common/communicarts.scss
```diff
@@ -80,7 +80,19 @@ table td span.tag-values {
   font-weight: normal;
 }
 
+td.amount {
```
jessieay: rename this file while we're here? (there's a story somewhere re: moving away from communicart naming)

app/controllers/summary_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    @client_summarizer = ClientSummarizer.new(
+      client_namespace: client_namespace,
+      fiscal_year: params[:fiscal_year]
+    )
+    @summary = @client_summarizer.run
+  end
+
+  private
+
+  def authorize
```
jessieay: in what cases do we define this here vs writing a pundit policy? Don't mind this being here, just wondering

app/models/concerns/fiscal_year_mixin.rb
```diff
@@ -0,0 +1,13 @@
+module FiscalYearMixin
```
jessieay: YES much better.

app/models/ncr/work_order.rb
```diff
@@ -23,8 +23,6 @@ def self.purchase_amount_column_name
     # This is a hack to be able to attribute changes to the correct user. This attribute needs to be set explicitly, then the update comment will use them as the "commenter". Defaults to the requester.
     attr_accessor :modifier
 
-    after_initialize :set_defaults
```
jessieay: :joy: << tears of joy

app/services/ncr/work_order_value_normalizer.rb
```diff
@@ -41,6 +42,12 @@ def normalize_soc_code
       end
     end
 
+    def normalize_approving_official_email
+      if !work_order.approving_official_email.present? && work_order.approvers.any?
```
jessieay: this might be a little more readable: `if work_order.approving_official_email.blank? && work_order.approvers.any?`

app/services/ncr/work_order_value_normalizer.rb
```diff
@@ -41,6 +42,12 @@ def normalize_soc_code
...
+        work_order.approving_official_email = work_order.approvers.first.try(:email_address)
```
jessieay: I wonder why this step is needed -- isn't this being set in the form?

app/views/summary/index.html.haml
```diff
@@ -0,0 +1,43 @@
+%div.inset
+  %table.table.table-striped.table-bordered.table-responsive
+    %caption
+      %span.namespace= @summary.client_namespace
+      Requests: Fiscal Year
+      = @summary.fiscal_year
```
jessieay: if I am understanding the code correctly, of someone visited this page at `/summary`it would show results for the current year but `summary/2015

app/assets/stylesheets/common/communicarts.scss
pkarman: what new name would you suggest?

app/controllers/summary_controller.rb
pkarman: a policy requires an object + user to act upon. Here, we just have a user.

app/assets/stylesheets/common/communicarts.scss
jessieay: `c2`? or `base` or something? I am not really clear on the purpose of this file but based on current name I would assume it's kind of a scss junk drawer?

app/controllers/summary_controller.rb
pkarman: I guess you could argue that the `@summary` could have a policy, but it's an ephemeral, read-only object, and we want to avoid creating it unless the request is authorized (a catch-22).

app/services/ncr/work_order_value_normalizer.rb
pkarman: ah good! I will have to remember all these fun rails-isms.

lib/client_summarizer.rb
```diff
@@ -0,0 +1,31 @@
+class ClientSummarizer
+  include FiscalYearMixin
+
+  attr_reader :client_namespace, :fiscal_year
+
+  def initialize(args)
+    @client_namespace = args[:client_namespace]
+    now = Time.zone.now
+    @fiscal_year = (args[:fiscal_year] || self.class.which_fiscal_year(now.year, now.month)).to_i
+    @_records = build_records
```
jessieay: since we are using this below, I'd probs name it `@records` -- I've been using underscored ivars to indicate that an ivar is for memoization only (eg: your use in `run`)

lib/client_summarizer.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def run
+    @_summary ||= build_summary
+  end
+
+  private
+
+  def build_records
+    Proposal.for_fiscal_year(fiscal_year).where("client_data_type LIKE ?", "#{client_namespace}::%")
```
jessieay: I might memoize this and just call `build_records` directly in `build_summary` rather than in `initialize` -- what do you think?

app/services/ncr/work_order_value_normalizer.rb
pkarman: same reason it was originally needed in the `after_initialize` callback, I guess. All I know is lots of tests start failing w/o it.

lib/client_summarizer.rb
```diff
@@ -0,0 +1,31 @@
+class ClientSummarizer
+  include FiscalYearMixin
+
+  attr_reader :client_namespace, :fiscal_year
```
jessieay: do we need these? since we return a `ClientSummary` instance from `run`, we need an `attr_reader` on `ClientSummary` in order to return `fiscal_year` and `client_namespace`, but I don't see that we are using those attributes on an instance of `ClientSummarizer` directly

lib/client_summarizer.rb
pkarman: interesting. I typically use underscore more generally to indicate the "privateness" of a variable. I realize there's no attr_reader for records, so it's already "private" in that sense. Perl habit I guess (where there is no sense of private/public for methods except by convention).

Is underscore-for-memoization-onl a more common Ruby convention?

app/views/summary/index.html.haml
pkarman: Since this endpoint is not (yet) linked to from anywhere, on purpose, I expected we would get some user feedback before we made it more "formal".

I added the `:fiscal_year` option in order to make it easier to rule-of-thumb benchmark the effect of profiling an entire year's worth of proposals in one request, since there is a SQL execution for each one.

spec/lib/client_summarizer_spec.rb
```diff
@@ -0,0 +1,12 @@
+describe ClientSummarizer do
+  describe "Ncr::WorkOrder" do
+    it "builds summary" do
+      wo = create(:ncr_work_order, amount: 123)
```
jessieay: you think we should create a second record to setup to make sure addition stuff is working? Also, what do you think about testing this for 18F procurement?

spec/controllers/summary_controller_spec.rb
```diff
@@ -0,0 +1,29 @@
+describe SummaryController do
+  describe "#index" do
```
jessieay: should we test the case where someone visits `index/:last_year` since that is an option?

lib/client_summary.rb
```diff
@@ -0,0 +1,45 @@
+class ClientSummary
```
jessieay: unit tests for this?

lib/client_summarizer.rb
jessieay: I've been following this discussion re: underscore for memoization: https://github.com/thoughtbot/guides/pull/373

historically I only see underscored variables in Ruby in order to indicate that a variable is unused. This is enforced via rubocop. example:

```ruby
{ 1: "hi", 2: "hello" }.each do |_key, value|
  puts value
end
```

I am fine with doing underscores for memoized ivars in this project. In cases like this one, I'd probably not use an underscore.

app/services/ncr/work_order_value_normalizer.rb
jessieay: maybe we should create a tech debt card for investigating this, then? feels icky

app/views/summary/index.html.haml
jessieay: Not sure I understand the benchmarking comment -- since `summary` does the current year, doesn't that provide a good benchmark for a year? 

I'd move to remove the `fiscal_year` use case until we are using it / get feedback from users that they want it

pkarman: Sorry, I wasn't clear. I wanted to understand the performance implications of the design I had chosen, to see if summarizing via Ruby vs summarizing via SQL was fast enough to give a reasonable user experience. Since the current fiscal year (2016) only has a couple months' worth of data, I wanted to be able to summarize an entire year for comparison. Ergo, needed to be able to easily select a year via the URL, defaulting to the current year.

So I am using it during development, and since it provides helpful historical comparison, year-over-year, I would like to keep it.

This whole summary feature is intentionally undocumented and under-scoped till we can get some user feedback. I've put out a call for that feedback so we can move forward.

app/services/ncr/work_order_value_normalizer.rb
pkarman: It appears to be necessary in order for `def approver_changed?` to be effective, which is called from the controller to signal that some model attributes have changed. It seems to essentially be a shorthand for indicating that some association has been correctly configured.

If you feel that merits a tech debt card, feel free to open one.

lib/client_summary.rb
pkarman: thanks, added in c2dc19e

spec/controllers/summary_controller_spec.rb
pkarman: thanks, test added in 9ea621d

spec/lib/client_summarizer_spec.rb
pkarman: all the math should now be tested via ClientSummary unit tests.

added Gsa18f::Procurement test in 5a2459d

lib/client_summarizer.rb
pkarman: Mostly they are defined per my idea of good public api construction. Since fiscal year, in particular, might be internally derived if not explicitly set, I wanted to make sure it was possible to interrogate the object to discover what was used. I have added tests for that in 588f093

lib/client_summarizer.rb
pkarman: good idea.

app/views/summary/index.html.haml
pkarman: slack conversation: https://18f.slack.com/archives/cap/p1449155873001851

seems like it is :+1: for keeping this feature, and I have added tests.

app/assets/stylesheets/common/communicarts.scss
jessieay: thoughts on this, pkarman?

app/controllers/summary_controller.rb
jessieay: ok, sounds good to me!

app/views/summary/index.html.haml
jessieay: ok sounds good

jessieay:
code looks good, will QA now

app/assets/stylesheets/common/communicarts.scss
pkarman: yes I renamed to `base`

jessieay:
argh merge conflicts - mind updating @pkarman ?

jessieay:
Looks good on staging

app/assets/stylesheets/common/communicarts.scss
jessieay: sometimes github PR view is weird-- shows this file as existing in overview but I see in file view that it has beeb renamed. thanks for the update!

pkarman:
@jessieay master reconciled. probably good to wait for all green lights before merging.

jessieay:
will do! thanks pkarman :green_apple:

PR #841, opened 2015-12-02
[Delivers #109396552] treat step.completer as an approver in proposal policy

pkarman:
Story: https://www.pivotaltracker.com/story/show/109396552

To test:

* create a NCR Work Order
* login as a Tier 1 approver delegate (see #823 for testing detail tips)
* after approving the request, the delegate user should be able to modify the proposal (has a big blue Modify Request button on the /proposals/:id page)

### DIFF
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -33,6 +33,7 @@ class Proposal < ActiveRecord::Base
   has_many :steps
   has_many :individual_steps, ->{ individual }, class_name: 'Steps::Individual'
   has_many :approvers, through: :individual_steps, source: :user
+  has_many :completers, through: :individual_steps, source: :completer
   has_many :api_tokens, through: :individual_steps
   has_many :attachments, dependent: :destroy
   has_many :approval_delegates, through: :approvers, source: :outgoing_delegations
--- a/app/models/concerns/client_data_mixin.rb
+++ b/app/models/concerns/client_data_mixin.rb
@@ -10,6 +10,7 @@ module ClientDataMixin
     has_many :steps, through: :proposal
     has_many :individual_steps, -> { individual }, class_name: "Steps::Individual", through: :proposal
     has_many :approvers, through: :individual_steps, source: :user
+    has_many :completers, through: :individual_steps, source: :completer
     has_many :observations, through: :proposal
     has_many :observers, through: :observations, source: :user
     has_many :comments, through: :proposal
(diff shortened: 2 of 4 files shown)


app/models/proposal.rb
```diff
@@ -33,6 +33,7 @@ class Proposal < ActiveRecord::Base
   has_many :steps
   has_many :individual_steps, ->{ individual }, class_name: 'Steps::Individual'
   has_many :approvers, through: :individual_steps, source: :user
+  has_many :completers, through: :individual_steps, source: :completer
```
jessieay: I worry that this association name is a but confusing. Yoz was asking me about the concept of completion yesterday, which is why I think of it. It wasn't clear to him exactly what it was.

what about calling this a `step_completer` ? we know that a proposal has many steps, so I think that might be clearer. in the future, we also might want the rename the `approver` relation because a step is not always an approval these days.

jessieay:
will QA as-is -- one comment re: naming

jessieay:
confirmed that this works on c2-dev

app/models/proposal.rb
pkarman: Why not `step_approvers` then?

Right now we are in flux from a code world where there were `Approvals` and `Approvers` to a world of `Steps` and ...? what? `Steppers`?

I think architecturally it was absolutely correct to rename `approvals` to `steps` in order to reflect that a proposal evolves through various milestones in its workflow/lifetime, not all of which are "approval". Each of those steps gets assigned to someone initially. In the case of that "someone" being a generic mailbox (as all Tier 1 and Tier 2 NCR approvers are), there are then delegate users who actually perform the step. We want the audit trail (and the authorization delegation) that goes along with distinguishing between who was assigned the step and who actually performed the step. `StepPerformer`? `StepCompleter`? We landed on completer for the time being.

Naming things is the hardest thing in this job.

I feel strongly that the code should reflect, as much as possible, the realities of the business problems it is trying to solve. That helps ground the programmer in the business reality at the same time they are grokking the code itself.

How about we leave `completers` alone for the time being, until such time as we completely get rid of approvers. At that point perhaps a more perfect naming scheme will have presented itself. I like the simple poetic parity between `approver` and `completer`. In 66% of NCR steps, they are different Users, so it's not a trivial distinction.

jessieay: I would probably go with `step_user` and `step_completer` -- but I am fine with keeping as-is for now and reivisiting later

PR #842, opened 2015-12-03
[Delivers #108106044] Show purchase type on "My requests" page for 18F

jessieay:
* When logged in as a user with an 18F client slug, you should see the
  purchase type for requests on the "my requests" page (/proposals(
* Proposals can be sorted by this attribute
* Story: https://www.pivotaltracker.com/story/show/108106044

screen shot 2015-12-02 at 6 11 49 pm


jessieay:
note that this addition breaks the table a bit. Do we care? This breaks for 18F only (not NCR)

pkarman:
@jessieay "breaks the table" how? Design?

jessieay:
@pkarman if you see the screenshot in the PR, the far right "purchase type" header is going beyond the grey background. 

I can try to fix, but i also feel like we just need a more flexible front end / grid so not sure if it worth the effort to put a bandaid on what we have now. what do you think?

pkarman:
agreed about design. code looks :+1: 

I will QA.

pkarman:
column presence and sorting work fine in QA on c2-dev.

The width bug you mentioned is really bad IME, since it shunts everything to the right and leaves a big left margin. Snap attached:

screen shot 2015-12-03 at 12 56 16 pm

pkarman:
Could fix that layout bug with this patch:
```diff
diff --git a/app/assets/stylesheets/webapp/webviews.scss b/app/assets/stylesheets/webapp/webviews.scss
index 04d38ea..2f8abc9 100644
--- a/app/assets/stylesheets/webapp/webviews.scss
+++ b/app/assets/stylesheets/webapp/webviews.scss
@@ -5,7 +5,7 @@
 }
 
 .inset {
-  max-width: $web-content-width;
+  //max-width: $web-content-width;
   margin: 0 auto;
   padding: 0 25px;
   p {
```

not sure how that will break other things though.

jessieay:
ok! i also see that the table headers are not the color they should be on dev. will work on that and the layout stuff after my next mtg!

jessieay:
@pkarman removing that line looks MUCH better on a wider screen, but as soon as I get down to smaller than full width on a 13" screen it looks like this:

screen shot 2015-12-03 at 3 37 57 pm

jessieay:
@pkarman ok pushed to c2-dev to see how it looked. this is full screen on 13" monitor:

screen shot 2015-12-03 at 4 06 50 pm

this is half screen on a 13" monitor:

screen shot 2015-12-03 at 4 07 00 pm

it seems like `submitted` was cut off on a small screen even before this change. I am not sure how to proceed without making some major UI changes. Suggestions?

pkarman:
ok, I think we live with it as-is (no CSS change). Obviously we need a UX/UI re-think of this page, based on what actual users are trying to do there. My suspicion is that the main users of this page are approvers, and that sort-ability may not be an actual user need. More research needed.

pkarman:
@jessieay not sure if you have more work to do here (you mentioned colors), so re-assigning to you. I'm happy to merge and/or re-test if/when you're done.

jessieay:
@pkarman oh yeah, I saw that when you QA'd the colors were not right, but when I QA'd they showed up as expected so I would say we can merge as-is if all looks good to you

PR #844, opened 2015-12-03
[Delivers #106870524] add data retention policy per GSA audit guide

pkarman:
Story: https://www.pivotaltracker.com/story/show/106870524

Documentation only.

See also https://github.com/18F/DevOps/issues/512

### DIFF
--- a/doc/data_retention.md
+++ b/doc/data_retention.md
@@ -1,0 +1,19 @@
+# C2 Data Retention
+
+This document describes at a high level the data retention rules to which the C2 production
+instance must adhere.
+
+## Database
+
+For audit purposes, real (i.e. non-test) purchase card pre-authorization requests must be stored for at least
+6 years and 3 months. This timeline is for parity with the Pegasus system data retention policy.
+See [_IT Security Procedural Guide: Audit and Accountability (AU) CIO-IT Security-01-08_]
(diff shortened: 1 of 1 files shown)


yozlet:
Should this be linked somewhere from the app?

pkarman:
@yozlet that's a good question.

This story was not initiated on behalf of the users; I initiated it a few weeks ago because I didn't know what the rules where. In some ways, the users shouldn't have to worry about the issue at all -- their data will be available in conformance with the Rules as Long As It Needs To Be. Now we just have a definition for "Long".

I'm happy to add a link, I just don't know where. The target audience for this doc is really us (developers and system administrators).

jessieay:
maybe we should have a docs index page (a markdown file, most likely) that links to all of our docs / gives a brief explanation of each. I know that for myself that I never know if the info i need will be in `setup` or `developer-onboarding`, so that type of page would be very helpful for me!

pkarman:
@jessieay that sounds great, but outside the scope of this story, which is just to identify and document what the policy is. Organizing the docs for the sake of developers is fine, if that's an itch you want to scratch. For me, I just use `grep`.

jessieay:
ok we can do it separately. should Raphy read this before we merge? looks good to me

pkarman:
@phirefly please review before we merge

jessieay:
ping ping @phirefly

phirefly:
:shipit:

PR #845, opened 2015-12-04
Move some NCR specs to be more generic

yozlet:
These specs were under `features/ncr/work_orders/` but apply to all Proposals generally. (There are other specs still in that folder that theoretically should be general, but in practice they use the `edit`/`create` views, which are client-specific)


pkarman:
I'm sure @jessieay will want to comment on the use of `let` and `begin` but I am happy to merge as-is, precedent being a reasonable justification imo.

+1 to the idea of testing `proposal` more generally.

jessieay:
yes! this looks awesome. +2 on genericizing specs for all use cases

BUT, as Peter said, would love to see us inlining `let` and `before` statements in new and existing specs as we touch them. You cool with doing that, Yoz? If you'd like to have a longer convo about it, I am down to do that too.

spec/features/approval_delegates_spec.rb
```diff
@@ -0,0 +1,35 @@
+describe "Approval Delegates" do
+  let(:proposal) { create(:proposal, :with_approver) }
+
+  it "allows delegate to approve a proposal" do
+    delegate = create(:user)
+    approver = proposal.approvers.first
+    approver.add_delegate(delegate)
+    approver.save!
```
jessieay: let's also add newlines between test phases (setup / exercise / expectations)

jessieay:
hey @yozlet -- we reassigned this PR to you so you can update. Let us know when it's ready for re-review by either assigning one of us or just unassigning yourself. :)

jessieay:
@yozlet do you have time to update this? If not, I am happy to help out. Want to make sure we get on it before there are merge conflicts from other spec refactorings in the pipeline

yozlet:
Gaah, sorry for leaving this so long!

@jessieay and I obviously have disagreements about what specs should look like, and I'm wondering if this is the best place to have it out. I _think_ it's all about how much we think should be in the `it` block:

* I want it to be as short as possible, with all repeated setup happening outside to avoid error-prone duplication (and subtle, easily missed differences in a big list of near-identical 10-line specs) and make it easy to overload on a contextual basis. 
* @jessieay wants as much setup to happen in the `it` block as possible, so that the reader just needs to read that one block to see what's going on rather than chase upward through nested blocks to work out which `let` is the one that applies to this spec.

_Anyway_, having typed that out, and despite being totally up for having that argument **WITH RIGHTEOUS TESTING FURY**, I will tweak the specs the way that Jessie wants them for the moment so we can just merge the bloody PR because for god's sake how long have I let this sit here I mean really

jessieay:
looks awesome! merging

PR #846, opened 2015-12-04
[Delivers #109530778] fix new relic report tracking

pkarman:
Story: https://www.pivotaltracker.com/story/show/109530778

This doesn't fix the problem, but it should fix the reporting of the problem as step zero.

Story should stay open (un-finished) even after this is merged.


jessieay:
looks great!

PR #847, opened 2015-12-04
Update omniauth-myusa

jessieay:
* Gem now locks at pre- v. 1.4.0, so we don't have to
* https://github.com/18F/omniauth-myusa/pull/10


pkarman:
great. I should have done that originally instead of just patching C2.

PR #848, opened 2015-12-04
[Delivers #108469926] add User.display_name method for active_admin to use

pkarman:
Story: https://www.pivotaltracker.com/story/show/108469926

Testing:

 * visit `/admin`
 * any time a user record is displayed, it should use the name+email if both are defined, otherwise just email

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -71,6 +71,14 @@ class User < ActiveRecord::Base
     end
   end
 
+  def display_name
+    if full_name == email_address
+      email_address
+    else
+      "#{full_name} <#{email_address}>"
+    end
+  end
+
   def last_requested_proposal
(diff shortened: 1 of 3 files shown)


jessieay:
code looks :bomb: , will QA right :cat:

jessieay:
this looks great! only thing you might consider updating is that I did some weird stuff in order to get email address to show up for assignee / assigner in admin for approval delegates: https://c2-dev.18f.gov/admin/approval_delegates 

code: https://github.com/18F/C2/blob/master/app/admin/approval_delegate.rb

now that we have `display_name` defined, we could probably remove some of that setup. think we should do that now or in a separate PR?

pkarman:
let's do it separately. easier to track discrete changes.

PR #851, opened 2015-12-04
use json data with same value that selector uses (org.code_and_name)

pkarman:


jessieay:
@pkarman this does not actually handle the issue of parsing code and name on the server, right? will merge, but just want to be clear on the fact that there is more to be done re: assigning organization via NCR form

pkarman:
I found that this change meant I didn't have to parse the code out on the server. I had written that server code, then played with the js a little and found it was sending the id as you originally desired.

jessieay:
@pkarman you are correct!

PR #852, opened 2015-12-04
unmangle ncr_organization label name for when creating new work_orders

pkarman:


jessieay:
oh hey pkarman thanks for closing, forgot to comment here that I used this code but had since rebased my branch so merging via github was not gonna work. thanks for your help!

PR #853, opened 2015-12-05
[Delivers #109559252] Do not trigger reapproval when delegate updates…

jessieay:
… work order

* Previously were checking if a "budget approver" was updating when deciding
  whether to require reapproval
* A budget approver is defined as someone other than the approving
  official
* If one of the approver mailboxes had a delegate that updated a
  request, it should not have been triggering a reapproval
* This was a new bug because we recently introduced the concept of
  "compeleted" where an approval is completed by a delegate even though
  the approver for the step remains the original approver (mailbox)

Fix:

* Update `Ncr::Workrder#budget_approvers` to include completers
* Update specs in `WorkOrderReapprovalChecker` that were passing for the wrong reasons

* Story: https://www.pivotaltracker.com/story/show/109559252

To QA:

* Create a work order
* Fully approve work order
* Set completer of one of the budget approval steps as a user
* Log in as that user and modify a protected field (eg: function code)
* Make sure that the field was previously set. Setting val from nil
  never triggers reapproval
* Ensure that the field update has not triggered a reapproval flow (see
  emails sent out)

### DIFF
--- a/lib/ncr/work_order_reapproval_checker.rb
+++ b/lib/ncr/work_order_reapproval_checker.rb
@@ -6,13 +6,21 @@ module Ncr
       @work_order = work_order
     end
 
-    def current_user
-      work_order.modifier
+    def requires_budget_reapproval?
+      work_order.approved? &&
+        work_order.requires_approval? &&
+        !budget_approver? &&
+        (amount_increased? || protected_fields_changed?)
     end
 
-    def previous_val(key)
-      changes = work_order.previous_changes || {}
-      changes[key.to_s].try(:first)
+    private
+
+    def budget_approver?
+      work_order.budget_approvers.include?(current_user)
+    end
+
+    def current_user
+      work_order.modifier
     end
 
     def amount_increased?
@@ -21,24 +29,13 @@ module Ncr
     end
 
     def protected_fields_changed?
-      self.class.protected_fields.any? do |field|
-        previous_val(field).present?
+      protected_fields.any? do |field|
+        previous_val(field) &&
+          previous_val(field) != current_val(field)
       end
     end
 
-    def budget_approver?
-      work_order.budget_approvers.include?(current_user)
-    end
-
-    def requires_budget_reapproval?
-      work_order.approved? &&
-        work_order.requires_approval? &&
-        !budget_approver? && (
-          amount_increased? || protected_fields_changed?
-        )
-    end
-
-    def self.protected_fields
+    def protected_fields
       [
         :building_number,
         :function_code,
@@ -46,6 +43,17 @@ module Ncr
         :rwa_number,
         :soc_code
       ]
+    end
+
+    def previous_val(key)
+      changes = work_order.previous_changes
+      if changes[key.to_s]
+        changes[key.to_s][0]
+      end
+    end
+
+    def current_val(key)
+      work_order.public_send(key)
     end
   end
 end
--- a/spec/lib/ncr/work_order_reapproval_checker_spec.rb
+++ b/spec/lib/ncr/work_order_reapproval_checker_spec.rb
@@ -1,74 +1,53 @@
 describe Ncr::WorkOrderReapprovalChecker do
   include ProposalSpecHelper
 
-  describe '#protected_fields_changed?' do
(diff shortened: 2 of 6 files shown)


pkarman:
this looks really great: clean and good test coverage. I will QA first thing Monday morning.

spec/lib/ncr/work_order_reapproval_checker_spec.rb
```diff
@@ -76,20 +55,29 @@
...
+      create(:approval_delegate, assigner: budget_approver, assignee: delegate_user)
+      fully_approve(work_order.proposal, delegate_user)
+      work_order.reload
+
+      work_order.modifier = delegate_user
+      work_order.update!(function_code: "PG789")
+      checker = Ncr::WorkOrderReapprovalChecker.new(work_order)
+
+      expect(checker.requires_budget_reapproval?).to eq(false)
     end
```
afeld: Mind verifying that this test would fail without the other changes?

lib/ncr/work_order_reapproval_checker.rb
```diff
@@ -6,13 +6,21 @@ def initialize(work_order)
...
     end
 
-    def previous_val(key)
-      changes = work_order.previous_changes || {}
-      changes[key.to_s].try(:first)
+    private
+
+    def budget_approver?
+      work_order.budget_approvers.include?(current_user)
+    end
```
afeld: I'm not positive that the behavior has actually been changed here. I think this method should roughly read:

```ruby
def budget_approver?
  Ncr.budget_mailboxes_and_their_delegates.include?(current_user)
end
```

afeld:
My comments sort of hint at this, but my hypothesis is that this still won't fix the issue. As I understand the problem:

1. A Proposal makes it to the Tier 2 mailbox
1. The email gets forwarded to one of the Tier 2 budget folks (who is a `delegate`)
1. The budget person modifies the request (without/before approving it)
1. Since they aren't (yet) associated with the Proposal as one of the `approvers`:
    1. They are not returned in the list from `WorkOrder#budget_approvers`
    1. `WorkOrderReapprovalChecker#budget_approver?` returns `false`
    1. `WorkOrderReapprovalChecker#requires_budget_reapproval?` returns `true`

The TL;DR here is that "we want don't want a Proposal to require re-approval if modified by _any_ budget person, regardless of whether they have been associated with the Proposal directly (as an `approver`) or not." @phirefly Does that sound correct?

lib/ncr/work_order_reapproval_checker.rb
jessieay: no, the method itself (in NCR) is what changed. the refactors I did in this class were just to more clearly understand the code and what it does

spec/lib/ncr/work_order_reapproval_checker_spec.rb
jessieay: I did that while writing it via TDD :)

lib/ncr/work_order_reapproval_checker.rb
afeld: Fair enough, though from what that still doesn't cover the edge case (or similar ones).

pkarman: @afeld are you talking about the edge case where a delegate approves, then modifies? That is covered by the new definition of `completed_by` -- that could be the mailbox (`user`) or the delegate (`completer`) of the `step`.

That's a subset of `budget_mailxboxes_and_their_delegates` limited to the delegates who **actually clicked Approve**.

pkarman:
@afeld The scenario you lay out isn't the one that sparked this PR. The current problem report is slightly different. In the current problem, the proposal is **fully** approved (the budget approver has already acted) and then modified. This was a simple regression introduced in https://github.com/18F/C2/commit/5f40b02c7263b622bc3d3f88fd9d24f8b806a676#diff-6ad0fd187c45599fe82093fcbd18e639

I.e. we haven't encountered a new business scenario we had not encountered before; we just introduced a bug based on the old scenarios.

pkarman:
QA'd ok on c2-dev in https://c2-dev.18f.gov/proposals/254

PR #855, opened 2015-12-05
Add suggestions in troubleshooting section of setup for when user has…

rachelylim:
… rbenv and foreman command is not found.

### DIFF
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -39,6 +39,12 @@ Per [the Twelve-Factor guidelines](http://12factor.net/config), all necessary co
 * Check that PostgreSQL is running
 * Set the `DATABASE_URL` variable in [`.env`](../.env.example) to match your setup
 
+#### If 'foreman' command not found, you may be using rbenv. If so, run the following...
+```bash
+rbenv rehash
+gem install foreman
+```
+
 ## Starting the application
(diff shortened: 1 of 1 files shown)


jessieay:
Hey @rachelylim ! 

Thank you for the contribution. I think these are great tips generally - but I am not sure that they belong in this codebase since not everyone working on this will be using rbenv. 

My reasoning is this: If we start adding setup instructions for all varieties of dev setups, our setup instructions could get quite long / detailed. In general, I try to keep anything that is specific to my dev environment out of the codebase. 

All that being said, if this is as true pain point for a large # of people setting up C2, then it probably does make sense to include this information in the README. So I will keep this PR open for others to weigh in on!

pkarman:
I'm inclined to accept this, as a helpful hint. I don't use rbenv, but I don't mind seeing this here, and it might help the next person.

If/when the doc gets too unwieldy with lots of dev-tool-specific notes, we could start a FAQ doc. We aren't there yet, imo.

+1 to merge

pkarman:
Thanks @rachelylim !

rachelylim:
@jessieay @pkarman Thanks for considering my suggested contribution! Glad it was somewhat useful!

PR #860, opened 2015-12-08
Turn on JS for feature specs using selectize

jessieay:
* Previously, we turned JS on for some and off for others
* The problem with this was that we were not mirroring reality --
  selectize changes the values for certain fields, so params were being
  set differently when javascript was turned off
* This PR moves on JS for all feature specs for creating a work order
* Adds helper methods for setting and viewing values for selectize fields

### DIFF
--- a/spec/features/attachment_spec.rb
+++ b/spec/features/attachment_spec.rb
@@ -13,7 +13,7 @@ describe "Add attachments" do
     expect(page).to have_content(attachment.file_file_name)
   end
 
-  it "disables 'add attachment' button if no attachment is selected", js: true do
+  it "disables 'add attachment' button if no attachment is selected", :js do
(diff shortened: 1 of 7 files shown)


pkarman:
@jessieay all your good conventions must be wearing off on me: I have a local branch with a `fill_in_selectized` method that does essentially the same thing!

This looks great.

pkarman:
fwiw, I found this helpful: http://climber2002.github.io/blog/2014/09/22/capybara-integration-tests-with-jquery-selectize/

jessieay:
@pkarman that blog post is great! Did not come up in my googling, I used some SO posts to generate the helper methods added in this PR. Glad we have some more streamlined selectize usage in specs now, going to be super helpful to finally finish up that observations branch. :)

PR #862, opened 2015-12-08
[Delivers #108973804] delete cascade steps

pkarman:
Story: https://www.pivotaltracker.com/story/show/108973804

Adds db constraint to clean up Steps when a Proposal is deleted (see `lib/expired_record_cleaner.rb`).

Note that unlike the story description, this does not contain a `dependent: :destroy` or `:delete_all` declaration because the re-assignment of complex step combinations in `.root_step=` can reveal circular dependencies that Active Record cannot cleanly untangle and clean up, while the db can.


app/models/proposal.rb
```diff
@@ -30,6 +30,8 @@ class Proposal < ActiveRecord::Base
 
   acts_as_taggable
 
+  # we do not have "dependent:" declaration on "steps" because there can be circular references
```
jessieay: would it be possible to write a test that illustrates how the cascade works rather than having this comment here? 

I have some idea of what it means, but I am not exactly sure what the "circular references" you refer to are, so a test would help me to understand more clearly.

pkarman: Writing a test to illustrate why *not* to define something seems hard (impossible?) to me. It would require 2 different association definitions to compare the different behavior, yes?

How about we just eliminate the comment?

jessieay: I do think that the difference between doing this in the DB vs with AR is relevant. 

I can imagine my future self adding `dependent: :destroy` and removing the DB constraint (silly future me!) and not realizing the side effects of doing that. 

Would it be possible to write a unit test for proposals that destroys one and confirms that all of its steps were destroyed? (ideally, in a way that `dependent: :destroy` would not do so that we have a failing test in the case that future me ever tries to make the change mentioned above)

jessieay:
added another comment, but it got buried: https://github.com/18F/C2/pull/862#discussion_r46985772

let me know what you think!

pkarman:
@jessieay test illustrating database constraints in 5af897c

jessieay:
Created a proposal on c2-dev, deleted it via SSH and then conformed that its steps were also deleted. Anything else I should do to QA, @pkarman ? 

May wait to merge this until https://github.com/18F/C2/pull/864 is merged, since the timestamps on this migration are later than the migrations in that PR

pkarman:
@jessieay that covers it for QA.

Fine to wait till after #864

jessieay:
mind resolving merge conflicts? will re-push to c2-dev to make sure we're good once that's done and then we can :shipit:

pkarman:
@jessieay updated

jessieay:
looks good on dev! merging

PR #863, opened 2015-12-08
[Delivers #109547738] catch policy exception for already-approved proposal

pkarman:
Story: https://www.pivotaltracker.com/story/show/109547738

Test:

* Create NCR work order with approver that delegates to 2 or more other users
* Using the Approve link in the email that is sent, approve the work order as one of the delegates. The Proposal should approve as usual.
* Using the same Approve link in a different browser, try to use the link as the other delegate. Instead of an authorization error, you should be redirected to the proposals/:id link and get an error message that "A response has already been logged for this proposal"
screen shot 2015-12-08 at 10 54 17 am

### DIFF
--- a/app/policies/proposal_policy.rb
+++ b/app/policies/proposal_policy.rb
@@ -98,7 +98,7 @@ class ProposalPolicy
 
   def pending_approval!
     check(self.pending_approver? || self.pending_delegate?,
-          "A response has already been logged a response for this proposal")
+          "A response has already been logged for this proposal")
   end
 
   def visible_proposals
--- a/spec/controllers/proposals_controller_spec.rb
+++ b/spec/controllers/proposals_controller_spec.rb
@@ -300,7 +300,7 @@ describe ProposalsController do
       post :approve, id: proposal.id
 
       expect(response).to redirect_to(proposal_path(proposal))
-      expect(flash[:error]).not_to be_nil
+      expect(flash[:error]).to eq "A response has already been logged for this proposal"
     end
 
     it "allows a delegate to approve via the web UI" do
--- a/spec/controllers/proposals_controller_spec.rb
+++ b/spec/controllers/proposals_controller_spec.rb
@@ -218,7 +218,7 @@ describe ProposalsController do
       approval = proposal.individual_steps.first
       token = create(:api_token, step: approval)
 
-      post :approve, id: proposal.id, cch: token.access_token
+      get :approve, id: proposal.id, cch: token.access_token
 
       expect(controller.send(:current_user)).to eq(approval.user)
     end
--- a/spec/controllers/proposals_controller_spec.rb
+++ b/spec/controllers/proposals_controller_spec.rb
@@ -229,9 +229,8 @@ describe ProposalsController do
       token = create(:api_token, step: approval)
       approval.user.add_delegate(create(:user))
 
-      post :approve, id: proposal.id, cch: token.access_token
+      get :approve, id: proposal.id, cch: token.access_token
 
-      # TODO simplify this check
       expect(response).to redirect_to(root_path(return_to: self.make_return_to("Previous", request.fullpath)))
     end
 
@@ -279,7 +278,8 @@ describe ProposalsController do
       flash.clear
       post :approve, id: proposal.id
 
-      expect(response.status).to eq(403)
+      expect(response).to redirect_to(proposal_path(proposal))
+      expect(flash[:error]).not_to be_nil
     end
 
     it "won't allow different delegates to approve" do
@@ -299,7 +299,8 @@ describe ProposalsController do
       login_as(delegate2)
       post :approve, id: proposal.id
 
-      expect(response.status).to eq(403)
+      expect(response).to redirect_to(proposal_path(proposal))
+      expect(flash[:error]).not_to be_nil
     end
 
     it "allows a delegate to approve via the web UI" do
--- a/spec/controllers/proposals_controller_spec.rb
+++ b/spec/controllers/proposals_controller_spec.rb
@@ -279,7 +279,7 @@ describe ProposalsController do
       post :approve, id: proposal.id
 
       expect(response).to redirect_to(proposal_path(proposal))
-      expect(flash[:error]).not_to be_nil
+      expect(flash[:error]).to eq "A response has already been logged for this proposal"
     end
 
     it "won't allow different delegates to approve" do
--- a/app/controllers/concerns/token_auth.rb
+++ b/app/controllers/concerns/token_auth.rb
@@ -45,7 +45,12 @@ module TokenAuth
         redirect_to root_path(return_to: return_to_param), alert: "Please sign in to complete this action."
       end
     when Proposal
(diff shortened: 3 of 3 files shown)


app/policies/proposal_policy.rb
```diff
@@ -98,7 +98,7 @@ def pending_delegate?
 
   def pending_approval!
     check(self.pending_approver? || self.pending_delegate?,
-          "A response has already been logged a response for this proposal")
+          "A response has already been logged for this proposal")
```
jessieay: would it be possible to do something like`"This proposal has already been approved by #{approver.email_address}"` ? are there situations other than an already approved proposal where someone would see this error msg?

spec/controllers/proposals_controller_spec.rb
```diff
@@ -279,7 +278,8 @@
       flash.clear
       post :approve, id: proposal.id
 
-      expect(response.status).to eq(403)
+      expect(response).to redirect_to(proposal_path(proposal))
+      expect(flash[:error]).not_to be_nil
```
jessieay: what about checking for the exact flash error text?

spec/controllers/proposals_controller_spec.rb
```diff
@@ -229,9 +229,8 @@
       token = create(:api_token, step: approval)
       approval.user.add_delegate(create(:user))
 
-      post :approve, id: proposal.id, cch: token.access_token
+      get :approve, id: proposal.id, cch: token.access_token
```
jessieay: seems like we have `get` in some parts of this spec file and `post` in others-- should they all be one or the other?

app/policies/proposal_policy.rb
pkarman: Since a request goes through multiple approvals in the course of its lifetime, and thus multiple approvers, and since the complete approval history is already available in the /proposals/:id and email summary views, adding it to the error message could be both misleading and redundant.

spec/controllers/proposals_controller_spec.rb
pkarman: if you read over the other tests in that file, you'll see that `get` will only work with an access_token, while `post` is required otherwise.

spec/controllers/proposals_controller_spec.rb
pkarman: sure thing. I was trying to not tie the test too closely to the exact error wording, but I guess the controller action already depends on an exact string match (which I find unfortunate but couldn't see an easy way around it).

app/policies/proposal_policy.rb
jessieay: I take your point on the fact that there are multiple approvals. I am imagining receiving this message when trying to approve an already approved proposal, and I imagine that this precise message would not tell me why my approval did not go through. Ideas on other wording? Ok to keep as-is, I just don't know if our users will know what "response" means in the context of C2

spec/controllers/proposals_controller_spec.rb
jessieay: looks like you changed other specs but forgot this one

spec/controllers/proposals_controller_spec.rb
jessieay: oh interesting! I think the source of my confusion is that we have one `post` spec at the top of the approval specs (which uses a token) and then a bunch of `get`s and then a bunch of `post`s. Do you think it makes sense to group them?

app/policies/proposal_policy.rb
pkarman: Let's leave it as is, and see what the users say?

spec/controllers/proposals_controller_spec.rb
pkarman: ah thanks, fixed in 8483b3d

spec/controllers/proposals_controller_spec.rb
pkarman: I changed the `post` to a `get` in order to more closely represent the actual use case, which is a delegate clicking on a link in an email.

The HTTP method used in each spec should be specific to the particular use case being tested. I'm not sure what kind of grouping you have in mind? Maybe we could defer that as outside the scope of this change?

app/policies/proposal_policy.rb
jessieay: sure, although I am not sure how we would get direct user feedback about this type of thing -- perhaps ping Amber and ask her to include it in her user interviews?

spec/controllers/proposals_controller_spec.rb
jessieay: I meant group the tests with `post` vs `get` -- find to keep as-is

jessieay:
I created a work order: https://c2-dev.18f.gov/proposals/256

clicked "approve" in email and logged in as jessie.a.young@gmail.com

logged out

clicked "approve" as jessica.young+ncr@gsa.gov

saw following error:

screen shot 2015-12-08 at 10 28 23 pm

this seems like it was not the expected behavior?

pkarman:
When you logged out, did you also log out of MyUSA? I ask because the test instructions indicated using different browsers to more closely approximate 2 different human beings clicking the Approve link.

app/policies/proposal_policy.rb
pkarman: I was imagining waiting to hear via the support queue, which is where we learned about this original problem (the authorization error on multiple approval attempts).

jessieay:
Yes, I also logged out of MyUSA! Could try different browsers but I tried
this several times in incognito so I wouldn't expect the result to be
different. Are you able to reproduce?

pkarman:
@jessieay I think I probably failed to be specific enough in my test instructions. Based on what I can see in c2-dev, the steps you took don't quite match the real-world situation.

You have a user "jessie.a.young@gmail.com" with two delegates: "jessica.young+ncr@gsa.gov" and "jessica.young+18fpurchaser@gsa.gov".

You assigned the proposal to have "jessie.a.young@gmail.com" as the first approver, which is fine.

However, you used that same account "jessie.a.young@gmail.com" to actually approve the proposal the first time, instead of using one of the delegates. You must do all the approving as a delegate, not as the assigned User.

Does that make sense?

In practice, no one ever signs in as the assigned User. It's just a generic mailbox. People only ever sign in under their own accounts, which are delegates of the assigned User.

jessieay:
@pkarman that makes sense. I will QA again.

One question: would it make sense to make the way I QA'd this story working the same as the correct way of QAing? I know that nobody is using delegation in that way (where a mailbox user signs in and approves things), but wondering if it makes sense to make the system more flexible w/r/t approvals  and delegates

pkarman:
I think it would make sense to create a story for that, possibly in relation to this https://www.pivotaltracker.com/story/show/89827406

The use case is similar but not identical. This use case is: we need one approval from amongst a group of potential approvers, so we send a single email to a shared account.

The use case you've identified is: we need one approval from a user, who might designate one or more proxy approvers.

jessieay:
Testing on staging - looks great! 

https://c2-staging.18f.gov/proposals/496

screen shot 2015-12-10 at 5 10 52 pm

PR #864, opened 2015-12-08
[Delivers #108795040] Move NCR Organizations to database

jessieay:
* Were being read from CSV and loaded into memory
* This is a refactor, no behavioral change
* Story: https://www.pivotaltracker.com/story/show/108795040


spec/db/chores/roles_conversion_spec.rb
```diff
@@ -1,3 +1,5 @@
+require_relative "../../../db/chores/roles_conversion"
```
pkarman: this seems brittle. what happens when we (inevitably) reorganize the file structure of the tests?

spec/db/chores/ncr_organization_importer_spec.rb
```diff
@@ -0,0 +1,19 @@
+require_relative "../../../db/chores/ncr_organization_importer"
```
pkarman: see previous comment about brittle relative path names.

spec/db/chores/ncr_work_order_organization_reassigner_spec.rb
```diff
@@ -0,0 +1,53 @@
+require_relative "../../../db/chores/ncr_work_order_organization_reassigner"
```
pkarman: see previous comment about brittle relative path names.

pkarman:
@jessieay looks great. Just a comment about using relative paths in test files -- maybe use `Rails.root` instead?

jessieay:
@pkarman good idea! Updated relative paths at 02ca38a3329c2b9ad0791291f86e71d2db35eec1

jessieay:
Something to look for when QAing: I noticed a few times that when I ran the migration it removed all of the approver names from my proposal. I am not sure why this would happen and it seems to correct itself, but there is something weird there.

pkarman:
great I will QA

pkarman:
I QA'd on c2-dev.

I found a weird bug similar to the one you described. If I change the NCR Organization by modifying the proposal after I have created it, then it re-orders the Approval Steps and drops one.
screen shot 2015-12-09 at 10 48 16 am
screen shot 2015-12-09 at 10 51 26 am
screen shot 2015-12-09 at 10 49 10 am
screen shot 2015-12-09 at 10 49 21 am

jessieay:
Closing this in favor of https://github.com/18F/C2/pull/870

870 is similar but a lighter touch - still adds NCR orgs to the database but keeps `org_code` as a string on `work_order`. This is to prevent the migration from doing the weird thing where it removes approvers from work orders. We should eventually move to put a foreign key on work orders rather than storing a string, but this is a good first step

jessieay:
pushed to c2-dev and rolled back, so dev is no longer broken :)

jessieay:
oh wait, it is broken. somehow the migration rollback via ssh didn't work. re-pushed master to c2-dev and it is broken again. HALP!

PR #866, opened 2015-12-10
[Delivers #109818478] disable client slugs

pkarman:
Story: https://www.pivotaltracker.com/story/show/109818478

This PR implements a `DISABLE_CLIENT_SLUGS` environment variable, which can be set to a comma-delimited list of `client_slug` values. Any user that matches one of the client_slug values will be dis-allowed from viewing, editing or creating proposals.


app/controllers/application_controller.rb
```diff
@@ -31,6 +32,21 @@ def authorize(record, query = nil, user = nil)
     end
   end
 
+  def check_disabled_client
+    if client_disabled?
+      ex = NotAuthorizedError.new("Client is disabled")
```
jessieay: what does `ex` mean here? Perhaps a longer var nam would be clearer?

spec/requests/ncr_work_order_spec.rb
```diff
@@ -0,0 +1,20 @@
+describe 'proposals' do
+  include ReturnToHelper
+
+  describe "DISABLE_CLIENT_SLUGS" do
+    with_env_var("DISABLE_CLIENT_SLUGS", "ncr") do
+      it "disallows any request for disabled client_slug" do
+        user = create(:user, client_slug: "ncr")
```
jessieay: would it be possible to write this spec in a generalized way eg: set to slug to "foobar" ?

app/views/ncr/_disabled.html.haml
```diff
@@ -0,0 +1 @@
+Notice to C2 users from the National Capital Region: Until further notice, access is disabled. Please speak to your management for further clarification.
```
jessieay: should we link to feedback form?

app/controllers/proposals_controller.rb
```diff
@@ -2,6 +2,7 @@ class ProposalsController < ApplicationController
   include TokenAuth
 
   skip_before_action :authenticate_user!, only: [:approve]
+  before_action :check_disabled_client
```
jessieay: I think we should only need to do this for `:approve` since we are using `authenticate_user!

app/controllers/concerns/token_auth.rb
```diff
@@ -36,17 +36,29 @@ def validate_access
...
+      flash[:error] = exception.message
+      render "authentication_error", status: 403
+    else
+      return_to_param = make_return_to("Previous", request.fullpath)
+      session[:return_to] = return_to_param
+      redirect_to root_path(return_to: return_to_param), alert: "Please sign in to complete this action."
+    end
+  end
+
+  def render_unidentified_exception(exception)
```
jessieay: maybe `render_other_exception`  since disabled client seems to be an "identified" exception?

app/controllers/client_data_controller.rb
```diff
@@ -94,12 +94,20 @@ def auth_errors(exception)
...
-      render "authorization_error", status: 403, locals: { msg: exception.message }
+      render_auth_errors(exception)
     else
       redirect_to path, alert: exception.message
     end
   end
 
+  def render_auth_errors(exception)
+    if exception.message == "Client is disabled"
+      render_disabled_client_message(exception.message)
```
jessieay: since this method rescues and returns an `authorization_error`, do we also need to return it here? Or maybe we should not rescue in `render_disabled_client_message`? Seems like there is something being duplicated somewhere, not sure which should stay...

app/controllers/application_controller.rb
```diff
@@ -31,6 +32,21 @@ def authorize(record, query = nil, user = nil)
...
+      fail ex
+    end
+  end
+
+  def render_disabled_client_message(message)
+    begin
+      render "#{current_user.client_slug}/_disabled", status: 403
+    rescue
```
jessieay: would it be possible to rescue a specific type of error? http://stackoverflow.com/questions/10048173/why-is-it-bad-style-to-rescue-exception-e-in-ruby

app/controllers/application_controller.rb
```diff
@@ -52,16 +68,20 @@ def current_user
...
   end
 
+  def client_disabled?
+    current_user && (ENV["DISABLE_CLIENT_SLUGS"] || "").split(/,/).include?(current_user.client_slug)
+  end
+
   def sign_in(user)
     session[:user] ||= {}
-    session[:user]['email'] = user.email_address
+    session[:user]["email"] = user.email_address
```
jessieay: :blush:

spec/requests/ncr_work_order_spec.rb
jessieay: oh haha I just saw the one below. AND YOU USED ALMOST EXACTLY THE SAME SLUG. guess we are mind melding

 I wonder if we don't need this specific spec, then? do you think there is a benefit to having both?

doc/envvars.md
```diff
@@ -15,6 +15,12 @@ and how they affect app behavior.
 ## DEFAULT_URL_PORT
 ## DEFAULT_URL_SCHEME
 
+## DISABLE_CLIENT_SLUGS
```
jessieay: niiiice

spec/requests/ncr_work_order_spec.rb
pkarman: I wanted to test both in the abstract (`foo`) and in the concrete (`ncr`) because the story is specifically for NCR.

app/views/ncr/_disabled.html.haml
pkarman: We don't want feedback. We want users to contact their management, not us.

app/controllers/proposals_controller.rb
pkarman: it's a trivial operation in terms of cpu cycles, and it future-proofs the feature for when the inevitable additional methods are added that do not use `authenticate_user!` and someone has to remember to add them to a whitelist of actions to check.

app/controllers/concerns/token_auth.rb
pkarman: it's un-identified in the context of the case/when block (it's the default fall-through action). But I take your point. English befuddles.

app/controllers/client_data_controller.rb
pkarman: If there is an exception in the `client_data_controller` that is *not* a disabled-client error, we want to render the 403 authz error page. That is the existing behavior.

I'm not sure exactly what you think is being duplicated? I've basically just refactored all the rendering code in the various controllers into its own methods so that rubocop likes me.

spec/requests/ncr_work_order_spec.rb
jessieay: makes sense.

app/controllers/client_data_controller.rb
jessieay: Because there are several sets of `if/else` if was a bit tricky to follow, here is my thinking in English:

ok, so IF the exception message is "Client is disabled" then `render_client_message` ... goes to that method and it calls `  render "#{current_user.client_slug}/_disabled", status: 403` but then rescues that to return `render "authorization_error", status: 403, locals: { msg: exception.message }` which is the SAME as the `else` here. so if we just called `render_disabled_client_message(exception.message)` here, wouldn't that yield the same result?

spec/requests/ncr_work_order_spec.rb
```diff
@@ -0,0 +1,20 @@
+describe 'proposals' do
```
jessieay: this should probs be changed :)

jessieay:
One comment https://github.com/18F/C2/pull/866/files#r47298563 on some (possibly) duplicated logic. Otherwise looks ready to QA :last_quarter_moon_with_face:

app/controllers/client_data_controller.rb
pkarman: yes, it would currently. but it would break encapsulation to assume that an exception-specific rendering method (`render_disabled_client_message`) would always use the same template to render as a generic auth exception.

It's a subtle difference, I realize, but consider this scenario. Six months from now we have another need for a disabled client, only this time we have a UI paradigm that has a specific "disabled" template we want to use as a fallback for all clients.

So six-months-from-now you says, oh, I'll just change the default fallback template and all will be well:
```diff
  def render_disabled_client_message(message)
    begin
      render "#{current_user.client_slug}/_disabled", status: 403
    rescue ActionView::MissingTemplate => _error
-      render "authorization_error", status: 403, locals: { msg: message }
+     render "custom_disabled_template", status: 403, locals: { msg: message }
     end
   end
```

and lo! you've now introduced a regression where every auth error renders using that "custom_disabled_template".

That's why I wrote it like I did: encapsulating the template-picking logic within the method based on what the exception is, not what the template is.

jessieay: it would introduce a regression *if* there weren't a test for the expected behavior. but we do have a test! I see your point that it is coincidental that they return the same response, makes sense to keep as-is.

jessieay:
What do you suggest for QAing this? Set disabled slug to `ncr` on dev and make sure I can't log in?

pkarman:
That QA plan sounds great.

jessieay:
OK, testing on c2-dev. I am seeing the correct template at `/proposals/ and `ncr/work_orders/:id` but I am able to see the https://c2-dev.18f.gov/ncr/dashboard -- is that problematic?

pkarman:
It wasn't listed explicitly in the story, but you are probably right: it should also be disabled. I will fix.

jessieay:
cool @pkarman, let me know when that's done :sun_with_face:

app/controllers/gsa18f/dashboard_controller.rb
```diff
@@ -1,5 +1,7 @@
 module Gsa18f
   class DashboardController < ApplicationController
+    before_action :check_disabled_client
```
jessieay: seems like this `check_disabled_client` check should be in `application_controller` and just selectively skipped? (I think the only page we want disabled clients to be able to see is the home page, right?) That would ensure that future controllers also use the check rather than needing to explicitly add it to each controller

jessieay:
woot woot, merging!

PR #867, opened 2015-12-10
[Delivers #109219130] Preserve approving official email on Ncr::WorkO…

jessieay:
…rder#edit

* To QA:
  - create an ncr work order
  - modify work order
  - confirm that previously selected approving official email is
    populated in form

* Story: https://www.pivotaltracker.com/story/show/109219130

### DIFF
--- a/app/controllers/ncr/work_orders_controller.rb
+++ b/app/controllers/ncr/work_orders_controller.rb
@@ -14,7 +14,7 @@ module Ncr
     end
 
     def edit
-      @client_data_instance.approving_official_email = @client_data_instance.approvers.first.email_address
+      @client_data_instance.approving_official_email = @client_data_instance.approvers.first.try(:email_address)
 
       if proposal.approved?
         flash.now[:warning] = "You are about to modify a fully approved request. Changes will be logged and sent to approvers, and this request may require re-approval, depending on the change."
--- a/app/controllers/ncr/work_orders_controller.rb
+++ b/app/controllers/ncr/work_orders_controller.rb
@@ -14,7 +14,9 @@ module Ncr
     end
 
     def edit
-      if self.proposal.approved?
+       @client_data_instance.approving_official_email = @client_data_instance.approvers.first.email_address
+
+      if proposal.approved?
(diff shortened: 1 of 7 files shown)


app/controllers/ncr/work_orders_controller.rb
```diff
@@ -14,7 +14,9 @@ def create
     end
 
     def edit
-      if self.proposal.approved?
+       @client_data_instance.approving_official_email = @client_data_instance.approvers.first.email_address
```
pkarman: Why not `Ncr::WorkOrderValueNormalizer.new(@client_data_instance).run` which already has the approving_official_email logic (amongst others)?

pkarman:
My initial comment got squashed: https://github.com/18F/C2/pull/867#discussion_r47278930

We can now see why @afeld did this the way he did originally, in the callback, so that the value is always populated when the object is initialized. Were it not for me trying to save on some SQL executions in the /summary feature, I would probably say let's go back to that.

In anycase, see my comment. Otherwise, this seems legit as is.

pkarman:
Oh, and if we are moving to Elasticsearch, then I will probably refactor the /summary feature to use ES directly, thus making those extra SQL calls moot.

app/controllers/ncr/work_orders_controller.rb
jessieay: the reason I didn't do that is that the value of the normalizer is taking raw values and converting them to formatted values -- like formatting strings etc. So running that whole class feels like overkill when all we need to do it set the one attribute.

I also feel like the `approving_official_email` stuff in general is a little wonky. Wish we would remove it altogether and just treat it like an association. I am sure there are good reason that it is the way that it is today, but in the future I'd love to see how we could not need to set that attribute manually and just let the form handle it.

jessieay:
@pkarman replied to your comment. Let me know what you think! Also updated code so specs should pass now :chart_with_upwards_trend:

jessieay:
@pkarman in terms of comment on callback: even though moving back to a callback would fix this issue, callbacks in general are problematic so I would prefer to not use callbacks whenever possible.

pkarman:
:+1: , will QA.

pkarman:
QA'd locally, looks awesome. :sheep:

PR #868, opened 2015-12-10
Refactor query objects

jessieay:
* Privatize methods that are unused outside of class to minimize API of
  objects
* Prep for https://www.pivotaltracker.com/story/show/105014352 (uses
  `Versions` query object, which I was having a hard time wrapping my
  head around with previous organization)


pkarman:
Looks good. I will QA.

pkarman:
tested on c2-dev -- :sheep:

PR #869, opened 2015-12-11
Add more code climate checks

jessieay:
* List of engines: http://docs.codeclimate.com/article/296-engines-available-engines

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,12 +1,4 @@
-# For an overview of the Code Climate Platform, see here:
-# http://docs.codeclimate.com/article/300-the-codeclimate-platform
-
-# For more details, see here:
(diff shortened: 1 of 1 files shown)


jessieay:
cc @afeld :) 

not sure what to do with the fact that code climate is failing now -- it is failing because the new checks are bringing up errors that the old checks didn't.

afeld:
I say we merge, and worry about fixing existing warnings separately. 


:shipit:

jessieay:
What do you think, @pkarman ?

pkarman:
The definition of "similar code" might need some thinking about, but I am ok to merge this and see what happens.

jessieay:
ok, works for me @pkarman !

PR #870, opened 2015-12-12
[Delivers #108795040] Move NCR Organizations to database

jessieay:
* Were being read from CSV and loaded into memory
* This is a refactor, no behavioral change
* Story: https://www.pivotaltracker.com/story/show/108795040


pkarman:
it's disappointing not to get the foreign key constraint working, but I understand the need to expedite.

I will QA

jessieay:
@pkarman Since we've already worked on it elsewhere, I think it would be good / easy to do it next / separately. The issue I was having, just FYI, is that setting `data-initial`to a single record was raising a JavaScript error, since `data-initial` expects an array. But when I set in array, it did not populate the existing value on edit. Grrr

pkarman:
QA on c2-dev looks fine: https://c2-dev.18f.gov/ncr/work_orders/218/edit

PR #874, opened 2015-12-15
Remove `flow` from proposals

jessieay:
* We do not use this concept for serial or parallel steps
* Steps of type `Serial` or `Parallel` are added to a proposal, instead
* Removing at suggestion of https://github.com/18F/C2/pull/835#issuecomment-161836392

### DIFF
--- a/app/decorators/proposal_decorator.rb
+++ b/app/decorators/proposal_decorator.rb
@@ -9,26 +9,8 @@ class ProposalDecorator < Draper::Decorator
     object.individual_steps.count
   end
 
-  def steps_by_status
-    # Override default scope
-    object.individual_steps.with_users.reorder(
-      # http://stackoverflow.com/a/6332081/358804
-      <<-SQL
-        CASE steps.status
-        WHEN 'approved' THEN 1
-        WHEN 'actionable' THEN 2
-        ELSE 3
-        END
-      SQL
-    )
-  end
-
   def steps_in_list_order
-    if object.flow == 'linear'
-      object.individual_steps.with_users
-    else
-      steps_by_status
-    end
+    object.individual_steps.with_users
   end
 
   def display_status
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -7,7 +7,6 @@ class Proposal < ActiveRecord::Base
   has_paper_trail class_name: 'C2Version'
 
   CLIENT_MODELS = []  # this gets populated later
-  FLOWS = %w(parallel linear).freeze
 
   workflow do
     state :pending do
@@ -53,7 +52,6 @@ class Proposal < ActiveRecord::Base
     message: "%{value} is not a valid client model type. Valid client model types are: #{CLIENT_MODELS.inspect}",
     allow_blank: true
   }
-  validates :flow, presence: true, inclusion: {in: FLOWS}
   validates :requester_id, presence: true
   validates :public_id, uniqueness: true, allow_nil: true
 
@@ -65,14 +63,6 @@ class Proposal < ActiveRecord::Base
 
   def root_step
     steps.where(parent: nil).first
-  end
-
-  def parallel?
-    flow == "parallel"
-  end
-
-  def linear?
-    flow == "linear"
   end
 
   def delegate?(user)
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -65,6 +65,14 @@ class Proposal < ActiveRecord::Base
     steps.where(parent: nil).first
   end
 
+  def parallel?
+    root_step.type == "Steps::Parallel"
+  end
+
+  def serial?
+    root_step.type == "Steps::Serial"
+  end
+
   def delegate?(user)
     approval_delegates.exists?(assignee_id: user.id)
   end
--- a/app/views/proposals/show.html.haml
+++ b/app/views/proposals/show.html.haml
@@ -19,15 +19,9 @@
           %strong
             = date_with_tooltip(@proposal.created_at, true)
 
-    - if @proposal.linear?
-      .col-md-12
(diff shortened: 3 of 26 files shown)


jessieay:
Trello card: https://trello.com/c/0BLjWz1K/56-remove-flow-from-proposals

spec/features/ncr/work_orders/requester_edit_spec.rb
```diff
@@ -141,14 +141,6 @@
...
   end
 
-  scenario "has 'Discard Changes' link" do
-    visit edit_ncr_work_order_path(work_order)
-
-    click_link "Discard Changes"
-
-    expect(current_path).to eq(proposal_path(ncr_proposal))
-  end
-
```
pkarman: Why is this test deleted?

spec/lib/dispatcher_spec.rb
```diff
@@ -1,9 +1,8 @@
 describe Dispatcher do
   let(:proposal) { create(:proposal) }
 
-  describe '.deliver_new_proposal_emails' do
-    it "uses the LinearDispatcher for linear approvals" do
-      proposal.flow = 'linear'
+  describe ".deliver_new_proposal_emails" do
+    it "uses the LinearDispatcher for non-NCR approvals" do
```
pkarman: What happens when we add a client that isn't NCR that does not use LinearDispatcher?

app/decorators/proposal_decorator.rb
```diff
@@ -9,26 +9,8 @@ def total_approvers
...
-        CASE steps.status
-        WHEN 'approved' THEN 1
-        WHEN 'actionable' THEN 2
-        ELSE 3
-        END
-      SQL
-    )
-  end
-
   def steps_in_list_order
```
pkarman: Even though we are not dropping the `flow` column, a Proposal still has a a "flow" flavor, indicated by the root step type. Would a proposal with a Parallel root step still need `steps_by_status`?

app/models/proposal.rb
```diff
@@ -74,14 +72,6 @@ def root_step
...
   end
 
-  def parallel?
-    flow == "parallel"
-  end
-
-  def linear?
-    flow == "linear"
-  end
-
```
pkarman: `parallel?` and `linear?` still seem like useful methods to have? Maybe they should just look at the root step instead?

pkarman:
Is it really true that "We do not use this concept for serial or parallel steps" ? Isn't `flow == 'linear'` the same thing as `proposal.individual_steps.first.type == "Steps::Serial"` ?

I guess what I'm saying is that even if we do not store the proposal step flavor as an explicit string in the database, we still have the concept of a proposal having a step flavor, as indicated by its root step. If we are going to keep Parallel step logic around (as I think we should, per conversation in #835), then I think we should keep all the view logic in place, and maybe just create a custom read-only `Proposal.flow` method that checks the root step type instead of a column value.

spec/features/ncr/work_orders/requester_edit_spec.rb
jessieay: hah WOOPS! accident. Was failing so I was editing it but must have accidentally deleted- will restore!

spec/lib/dispatcher_spec.rb
jessieay: I am not sure. One of the challenges of having the concept of parallel steps without using them is that we don't have test cases on the integration level for the situation where we are using them.

app/models/proposal.rb
jessieay: what if a proposal has some linear steps and some parallel?

app/views/proposals/show.html.haml
```diff
@@ -75,31 +69,6 @@
     %li
       = link_to "View history", history_proposal_path(@proposal)
 
-- # @todo: there are multiple approval-status ids
-- if @proposal.parallel?
```
jessieay: this is what a parallel proposal looks like when there are no approvals and when all parallel steps are approved:

screen shot 2015-12-16 at 10 05 21 am

When I update the status on the proposal itself to "approved", this is what it looks like:

screen shot 2015-12-16 at 10 06 56 am

app/views/shared/_email_status.html.erb
```diff
@@ -28,7 +28,7 @@
                   </td>
                   <td>
                     <%- last_approver = index == @proposal.steps_in_list_order.count - 1 %>
-                    <%= status_icon_tag(display_status, @proposal.linear?, last_approver) %>
+                    <%= status_icon_tag(display_status, last_approver) %>
```
jessieay: this is what the email looks like for parallel approvals before this was removed

screen shot 2015-12-16 at 10 13 12 am

app/decorators/proposal_decorator.rb
jessieay: `steps_in_list_order` is used in the emails. I just created a proposal on master with a parallel flow and 2 steps, this is what the email looks like:

screen shot 2015-12-16 at 10 13 12 am

seems like this method is not working

app/models/proposal.rb
pkarman: root step dictates flavor, imo

pkarman:
agreed, those screenshots really illustrate the problem. I'm fine to drop all the view-related code as part of this PR.

Let's resolve #879 first, then return to this PR since a new `flow` method would depend upon #879 to work.

jessieay:
`serial?` and `parallel?` methods added at e6c1a4591274e24f6a5d65fa481e210d8261faa3

Root step PR is completed.

This is ready for re-review

pkarman:
Perfect, thanks @jessieay. I will QA.

pkarman:
no regressions found in c2-dev

PR #876, opened 2015-12-16
Move `for_fiscal_year` method into mixin

jessieay:
* Part of moving NCR-specific logic out of Ncr::WorkOrder
* Also removes duplication

https://trello.com/c/smBumRFX

### DIFF
--- a/app/models/concerns/fiscal_year_mixin.rb
+++ b/app/models/concerns/fiscal_year_mixin.rb
@@ -4,13 +4,10 @@ module FiscalYearMixin
   FISCAL_YEAR_START_MONTH = 10 # 1-based
 
   included do
-    def self.which_fiscal_year(year, month)
-      if month >= FISCAL_YEAR_START_MONTH
-        year + 1
-      else
-        year
-      end
-    end
+    scope :for_fiscal_year, ->(year) {
+      range = range_for_fiscal_year(year)
+      where(created_at: range[:start_time]...range[:end_time])
+    }
 
     def self.range_for_fiscal_year(year)
       start_time = Time.zone.local(year - 1, FISCAL_YEAR_START_MONTH, 1)
--- a/app/models/concerns/fiscal_year_mixin.rb
+++ b/app/models/concerns/fiscal_year_mixin.rb
@@ -1,13 +1,21 @@
 module FiscalYearMixin
   extend ActiveSupport::Concern
 
+  FISCAL_YEAR_START_MONTH = 10 # 1-based
+
   included do
     def self.which_fiscal_year(year, month)
-      if month >= 10
+      if month >= FISCAL_YEAR_START_MONTH
         year + 1
       else
         year
       end
+    end
+
+    def self.range_for_fiscal_year(year)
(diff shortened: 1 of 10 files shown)


spec/lib/client_summarizer_spec.rb
```diff
@@ -16,7 +16,7 @@
 
   describe "Ncr::WorkOrder" do
     it "builds summary" do
-      wo = create(:ncr_work_order, amount: 123)
+      create(:ncr_work_order, amount: 123)
```
pkarman: why throw away the variable assignment? and why in this PR?

app/models/concerns/fiscal_year_mixin.rb
```diff
@@ -9,5 +11,11 @@ def self.which_fiscal_year(year, month)
         year
       end
     end
+
+    def self.for_fiscal_year(year)
```
pkarman: This was a `scope` and now a class method? Are they, strictly speaking, the same thing?

http://blog.plataformatec.com.br/2013/02/active-record-scopes-vs-class-methods/ was illuminating to me.

pkarman:
I like the sentiment here, of reducing duplication and using the existing mixin.

I am not clear on scope-vs-class-method, otherwise +1.

spec/lib/client_summarizer_spec.rb
jessieay: They are unused variables. I had the file open and my syntastic plugin was
highlighting them as unused so I thought I would get rid of them while I
was there. Would you prefer them removed in a separate PR? I could go
either way on that

app/models/concerns/fiscal_year_mixin.rb
jessieay: Since we use this mixin in one non-AR class, scope did not work because
that is part of AR (returns "undefined method scope").

 I did have a weird feeling about this mixin change because the mixin is
now used 3 times: twice in AR classes and once in a regular Ruby class. And
in each instance only one method or the other is used. But I did like the
idea of having these methods together since they are logically related.

Ideas on improvement?

pkarman: How about a refactor so that the class method part is in the mixin and the scope definition is still in the AR classes themselves? The date math can be part of the mixin.

jessieay: I am not sure I exactly understand what you mean -- more detail?

pkarman: see #878

jessieay:
@pkarman this is updated, incorporating your changes! 

I really liked the way you extracted the logic in the scope into another method. I ended up putting the scope and the method into the mixin, though, because the scopes for `ClientDataType` and `Proposal` were exactly the same.

BUT I did keep it as a scope in the mixin. I then moved the `which_fiscal_year` logic into its own service class,  since working on this made me realize that `which_fiscal_year` is entirely independent from any one model, so doesn't really need to be a method available on, say, `Proposal`. Having a service class that finds the current fiscal year seemed like a nice way to go. 

Please re-review and let me know what you think :christmas_tree:

pkarman:
`FiscalYearFinder` ftw

PR #878, opened 2015-12-16
keep scope in AR classes, date math in mixin

pkarman:


app/models/ncr/work_order.rb
```diff
@@ -47,6 +47,11 @@ def self.purchase_amount_column_name
       message: "must be three letters or numbers"
     }, allow_blank: true
 
+    scope :for_fiscal_year, lambda { |year|
```
jessieay: since the logic here and in `Proposal` is exactly the same, do you think it makes sense for us to include it elsewhere? 

I am not opposed to duplicated code when it makes sense, but the reason I first identified this was that I was looking for code in NCR::WorkOrder that was not specific to Work Orders. 

Seems like fiscal year logic might be helpful for reporting for other clients. In that case, we'd want this scope on those client data classes as well. Might be wrong to include this in the FiscalYear mixin, but it does seem like it should be outside of Proposal and WorkOrder. Does that make sense?

pkarman:
I moved the client_data scope to the `ClientDataMixin`.

It seems that we search for Proposal or ClientData depending on whichever is easiest to join against for whatever other search logic is necessary. So having the same scope defined on those 2 models doesn't feel bad to me, any more than having a scope called `approved` on both `Step` and `Proposal` doesn't feel bad.

The important part, imo, is putting the date math all in the same place, which it is now, in the FiscalYearMixin.

jessieay:
cool, merging into mine, will finish up over there!

PR #879, opened 2015-12-16
Root step is first step

jessieay:
* `where` does not work on steps
* see docs at http://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-find

### DIFF
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -70,7 +70,7 @@ class Proposal < ActiveRecord::Base
   }
 
   def root_step
-    steps.where(parent_id: nil).first
+    steps.where(parent: nil).first
   end
 
   def parallel?
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -69,9 +69,10 @@ class Proposal < ActiveRecord::Base
     where(created_at: start_time...end_time)
   }
 
-  # @todo - this should probably be the only entry into the approval system
   def root_step
-    steps.where(parent: nil).first
+    if steps.first.parent_id == nil
+      steps.first
+    end
   end
 
   def parallel?
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -70,9 +70,7 @@ class Proposal < ActiveRecord::Base
   }
 
   def root_step
-    if steps.first.parent_id == nil
-      steps.first
-    end
+    steps.where(parent_id: nil).first
   end
 
   def parallel?
--- a/app/models/steps/individual.rb
+++ b/app/models/steps/individual.rb
@@ -54,7 +54,7 @@ module Steps
     end
 
     def user_is_not_requester
-      if user_id && user_id == proposal.requester_id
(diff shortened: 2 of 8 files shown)


app/models/proposal.rb
```diff
@@ -69,9 +69,10 @@ class Proposal < ActiveRecord::Base
     where(created_at: start_time...end_time)
   }
 
-  # @todo - this should probably be the only entry into the approval system
   def root_step
-    steps.where(parent: nil).first
```
jessieay: I am not sure why this didn't raise before. But it always returned nil

pkarman:
interesting. and even moreso that this now causes other tests to fail.

jessieay:
@pkarman this should pass now. Did not need to define `root_step` differently, in the end. The issue was my spec setup used the `step` factory, which was actually an approval step and approval steps don't work as root steps. I redefined step factory names to make the instantiated class more obvious.

app/models/steps/individual.rb
```diff
@@ -54,7 +54,7 @@ def restart
     end
 
     def user_is_not_requester
-      if user_id && user_id == proposal.requester_id
+      if user && user == proposal.requester
```
pkarman: my guess is that the use of `user_id` instead of `user` and `requester_id` instead of `requester` is to minimize the amount of SQL statements generated. By using the AR objects instead, it adds 2 more statements. 

Is that a misguided, false laziness (preferring present integers over possibly-not-yet-loaded AR objects)?

spec/factories/api_tokens.rb
```diff
@@ -3,7 +3,7 @@
 
   factory :api_token do
     access_token
-    step
+    step { create(:approval_step) }
```
pkarman: this is the actual fix?

pkarman:
@jessieay it seems like the crucial changes in this PR are an additional test, and one line in the api_token factory. Is that right?

jessieay:
@pkarman yes, other changes are what I saw as I tried to figure out the root (LOL) of the issue!

app/models/steps/individual.rb
jessieay: One benefit of using objects vs foreign keys here is that it checks that the record actually exists. eg; there could be a `user_id` but no user for that `id` (perhaps user was deleted?)

 I generally follow a similar rule in validations "Validate the associated belongs_to object (user), not the database column (user_id)."

I am not too worried about the performance implications, but can remove if you think this is unnecessary.

spec/factories/api_tokens.rb
jessieay: yes and no: I had to change this because previously the `step` factory was an approval step, which was what lead to the failure in the `root_step` spec I created (root step needs to be parallel or serial). So  I changed the `step` spec definition, which required me to change this because api_token specs failed when this factory used a vanilla step rather than an approval or purchase step (api tokens can be related to either)

PR #880, opened 2015-12-17
remove superflous usage of self. within classes

pkarman:
This is a style change only. No ego here.

For posterity, this was a few Perl one-liners:
```bash
perl -pi -e '!m/def self\./ && !m/self\..+ =/ && !m/self\.class/ && s/self\.//g' app/*/*rb
```

and repeat for each path. A couple manual clean up cases, but that regex sequence gets most of it.

### DIFF
--- a/app/controllers/api/users_controller.rb
+++ b/app/controllers/api/users_controller.rb
@@ -10,7 +10,7 @@ module Api
         users = users.offset(params[:offset].to_i)
       end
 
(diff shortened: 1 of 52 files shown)


pkarman:
looks like codeclimate picks up some unrelated changes for "duplicate code" -- I would prefer to ignore those.

jessieay:
fine w me to ignore code climate. YEAH love this. :shaved_ice:

PR #883, opened 2015-12-18
Change status text in requests table display for purchase steps

yozlet:
Implements https://trello.com/c/XM2HDJ32/24-as-an-18f-purchaser-i-should-see-approved-items-on-my-dashboard-under-ready-for-purchase-and-purchased-items-under-recently-comp

Also added some classes to templates so as to make page objects easier to create and use.

PR #884, opened 2015-12-19
Hide "needs review" section when no proposals need review

jessieay:
* For https://trello.com/c/XM2HDJ32

### DIFF
--- a/app/views/proposals/index.html.erb
+++ b/app/views/proposals/index.html.erb
@@ -1,54 +1,53 @@
 <% content_for :title, "My Requests" %>
 <div class="inset">
   <%= render partial: 'search_ui' %>
+
   <div id="proposals-pending-review" class="row pending-pr first">
-    <div class="col-md-12">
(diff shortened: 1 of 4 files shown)


jessieay:
There is definitely some overlap between this and https://github.com/18F/C2/pull/883/files

pkarman:
@jessieay should we wait till #883 is merged to see if there are conflicts?

pkarman:
code looks good to me -- I like that we are giving elements ID attributes and using those rather than page placement for identification.

jessieay:
now that I re-review #883, I see that none of the specs Yoz added are in the spec file modified in this PR (`sort_index_spec.rb`) -- perhaps some of the specs I wrote here should be moved to `spec/features/proposals/index_spec.rb`, which he did modify. I will talk with him today re: his availability to update #883. If he can't do it today, we should probably merge this first and then merge his later.

pkarman:
ok on c2-dev

PR #885, opened 2015-12-21
Small Tabular Data refactors

jessieay:
* Move `config_for_client` method into its own class


jessieay:
hmmm trying to make the method smaller to fix CC failure. Tough! but working on it. otherwise, this is ready for review

jessieay:
ok code climate should be happy now

pkarman:
+1 looks good.

pkarman:
QA'd on c2-dev. No regressions observed.

PR #886, opened 2015-12-21
Clean up duplicate logic in Dashboard controllers

jessieay:
* Use query objects and decorators instead of controller methods
* This should help us generalize in the future


jessieay:
screen shot 2015-12-21 at 3 46 53 pm

:blush:

pkarman:
sweet!

PR #887, opened 2015-12-22
Single -> Double quotes in config files

jessieay:


config/environments/test.rb
```diff
@@ -1,36 +1,14 @@
...
-  config.action_controller.allow_forgery_protection = false
-
-  # Accumulates sent emails in the ActionMailer::Base.deliveries array.
+  config.action_mailer.default_url_options = { host: "localhost" }
   config.action_mailer.delivery_method = :test
-
-  # Print deprecation notices to the stderr.
   config.active_support.deprecation = :stderr
-  config.action_mailer.default_url_options = { :host => 'localhost' }
+  config.cache_classes = true
```
pkarman: this is also re-ordering of config lines? alphabetizing?

config/environments/production.rb
```diff
@@ -1,29 +1,27 @@
...
+    password: ENV.fetch("SMTP_PASSWORD"),
+    authentication: "login",
     enable_starttls_auto: true
   }
 
-  config.roadie.url_options = config.action_mailer.default_url_options
-
-  config.i18n.fallbacks = true
   config.active_support.deprecation = :notify
+  config.assets.compile = true
```
pkarman: this is also re-ordering of config lines? alphabetizing?

pkarman:
+1 to getting this kind of syntactic shoveling out of the way.

is this more than quotes? Looks like you are also zapping the boilerplate comments and reordering config lines? It's hard to tell if there are any config changes with the churn in some files.

jessieay:
ah yes also alpha ordering config files and removing comments that make config files hard to read - apologies for not adding that bit in PR description, was on autopilot as I did this. You think we should move those changes to a separate PR?

pkarman:
My concern is the errant copy/paste in re-ordering the config options, just to make sure we haven't missed any. Because the diff here is hard to read with both re-order and zapping comments, maybe breaking those into separate PRs would make it easier, yes.

jessieay:
ok, removed the comment-removed and reordered files from this PR so it's more focused. Ready to re-review!

PR #871, opened 2015-12-15
Elasticsearch

pkarman:
Switch our search infrastructure over to using Elasticsearch instead of hand-rolled Postgres SQL.
See discussion in #865 .

Related Trello cards:

https://trello.com/c/65PQIEkb
https://trello.com/c/2GTK6XHm
https://trello.com/c/dMrA8G2h
https://trello.com/c/rMDPRY6t
https://trello.com/c/i4Koni9u

This PR also implements a new model `Test::ClientRequest` purely for writing `client_data` tests against a generic (non-business-specific) client implementation.

Deploying this to cloud.gov environments requires further set up; see https://docs.cloud.gov/ops/accessing-v1-managed-services/
When it is time to test this branch, we will need to provision ES at cloud.gov.

CircleCI docs: https://circleci.com/docs/installing-elasticsearch

PR #888, opened 2015-12-22
[Delivers #62] advanced search UI

pkarman:
Depends on #871 merge to master first.

Story: https://trello.com/c/krQvAtfw


pkarman:
screen shot 2015-12-28 at 3 08 17 pm

pkarman:
I will re-base here to simplify diff

app/controllers/application_controller.rb
```diff
@@ -120,6 +122,10 @@ def authenticate_user!
     end
   end
 
+  def set_default_view_variables
+    @adv_search = Query::Proposal::FieldedSearch.new({})
```
jessieay: do we want to make search available on *every* page? 

Right now, for example, search is not available on Proposals#show but it is available on Proposals#show with this PR. Not bad to have it there, just not sure if that was intended. 

Placement on proposals#show does not make it super obvious that the search is for any proposal rather than within the current proposal:

screen shot 2016-02-01 at 10 55 12 am

app/models/ncr/work_order.rb
```diff
@@ -197,6 +197,10 @@ def restart_budget_approvals
       proposal.root_step.initialize!
     end
 
+    def self.expense_type_options
+      EXPENSE_TYPES.map { |et| [et, et] }
```
jessieay: `expense_type` instead of `et`? longer, but I spent a few seconds thinking about :alien: when I first saw this method

app/models/user.rb
```diff
@@ -58,6 +58,14 @@ def self.from_oauth_hash(auth_hash)
     user
   end
 
+  def client_model
+    Proposal.client_model_for(self)
+  end
+
+  def client_model_slug
+    client_model.to_s.underscore.tr("/", "_")
```
jessieay: how is this different from the `client_slug` method already available on a `user` ?

spec/features/search_spec.rb
```diff
@@ -17,20 +17,28 @@
...
     expect(field.value).to eq('foo')
   end
+
+  it "does not show search UI for user without client_slug" do
+    no_client_user = create(:user)
+    login_as(no_client_user)
+
+    visit proposals_path
+    expect(page).not_to have_button("Search")
+  end
```
jessieay: should we add a feature test specifically for advanced search? eg: filling in several fields.

lib/query/proposal/search_dsl.rb
```diff
@@ -28,19 +28,39 @@ def apply_authz?
...
-        params[client_data_type.underscore.tr("/", "_").to_sym]
+      def munge_fielded_params(fielded)
+        if fielded[:created_at].present? && fielded[:created_within].present?
+          munge_created_at_field(fielded)
+        end
+        # do not calculate more than once, or when created_at is null
+        fielded.delete(:created_within)
+      end
+
+      def munge_created_at_field(fielded)
```
jessieay: wondering if there might be a more descriptive word than `munge` -- what exactly is this method doing? (based on my reading, it creates a range based on user input)

app/models/gsa18f/procurement.rb
```diff
@@ -122,5 +122,9 @@ def self.user_with_role(role_name)
 
       users.first
     end
+
+    def self.expense_type_options
```
jessieay: thoughts on defining this in `client_data_mixin` rather than here?  if there is an option to define a method as an empty array, maybe that should be the default rather than requiring all client models to define this?

app/models/user.rb
jessieay: seems like we are using this method all over in this PR. Interested in the thinking behind this rather than using the string available on a user record.

jessieay:
@pkarman left a round of feedback! let me know what you think

app/controllers/application_controller.rb
pkarman: it was definitely intended to make search available on every page.

If we need to re-think the UI layout to make it clear what is being searched, my 2c is let's do that.

cc: @amberwreed @phirefly

app/models/user.rb
pkarman: an example:
```
[3] pry(main)> u = User.for_email 'peter.karman@gsa.gov'
  User Load (27.7ms)  SELECT  "users".* FROM "users" WHERE "users"."email_address" = $1 LIMIT 1  [["email_address", "peter.karman@gsa.gov"]]
=> #<User:0x007ffcda5d7418
 id: 664,
 email_address: "peter.karman@gsa.gov",
 first_name: "Peter",
 last_name: "Karman - XFB",
 created_at: Mon, 14 Sep 2015 17:09:04 UTC +00:00,
 updated_at: Fri, 13 Nov 2015 17:29:05 UTC +00:00,
 client_slug: "ncr",
 active: true>
[4] pry(main)> u.client_slug
=> "ncr"
[5] pry(main)> u.client_model_slug
=> "ncr_work_order"
```

the point is that it is theoretically possible for there to be more than one model in a client namespace. The `client_slug` is just the toplevel namespace. We want the specific model name as a single token (according to the ES tokenization rules).

lib/query/proposal/search_dsl.rb
pkarman: from the hackers dictionary, "munge" is:
> (1) a verb, used in a derogatory sense, meaning to imperfectly transform information

in this case, it is re-writing the user-supplied string value of `created_at` with a stringified range query, based on the presence and parse-ability of the `created_within` field.

If "munge" doesn't work for you, what word would you prefer?

spec/features/search_spec.rb
pkarman: good point. I will work one up.

lib/query/proposal/search_dsl.rb
jessieay: yes I understand the meaning of munge, but I think what is happening here is more precise, as you explained. how about something like `convert_created_at_to_range` ?

app/models/user.rb
jessieay: given the current setup (just two clients) do you think it makes sense to set this via the client slug? I was not aware of future plans or potential plans to add more clients within each namespace.

```

def client_model_slug
  if client_slug == "ncr"
      "ncr_work_order"
  elsif client_slug == "gsa"
    "gsa18f_procurement"
  end
end
```

app/models/gsa18f/procurement.rb
pkarman: fair enough. changed in eba12d9

app/models/user.rb
pkarman: given that one of the business goals at the moment is to scale up to a large number of clients, I would prefer to avoid any kind of hard-coding of client-specific strings.

spec/features/search_spec.rb
pkarman: feature test added in fd29c27. It was back-ported from #899 which has additional tests that address advanced search.

pkarman:
@jessieay I think I've hit all your comments. I pinged #cap about the search-bar-everywhere comment.

app/controllers/application_controller.rb
amberwreed: I think we should continue to have search on "My Requests" and add it to "My Dashboard". For the New Request page and Request Details it seems like it would be a confusing context switch to move from a detail view to a search results page so let's not include search on those pages. @phirefly what do you think?

jessieay:
Aside from Amber's comment re: location of search this looks ready for QA!

app/controllers/application_controller.rb
phirefly: I agree with @amberwreed. Let's keep this on the My Requests + My Dashboard and remove elsewhere. I could definitely see this being globally available if supported by a design that has a global nav, but it doesn't have that treatement atm.

phirefly: Added a parking item with the `Research` tag:
https://trello.com/c/NOqEhnN5

pkarman:
search UI now limited to My Requests and My Dashboard pages, via 8538974

@jessieay ready for QA

jessieay:
Few questions from QA:

 I cannot change the "cardholder name" field -- is this right? If so, why do we have it in the search form?

screen shot 2016-02-02 at 11 59 22 am

Should it be possible to not include an expense type? there is no blank option in the "expense type" dropdown

Also, I think there's a bug in the date range dropdown:

screen shot 2016-02-02 at 12 01 07 pm

pkarman:
@jessieay Card Holder Name is definitely change-able. I think you're seeing the "placeholder" value. Can you not enter arbitrary values?

I will look at the other two bugs.

pkarman:
@jessieay bug fixes pushed to this branch.

jessieay:
@pkarman card holder name was a user error - thought it would be a select! maybe the prompt should be empty like `amount`?

pkarman:
happy to make it empty. I think I wanted to show an example but maybe that's not necessary.

jessieay:
@pkarman I think it would be helpful if there were examples for all. As it is, it appears as a greyed out select to me. Defer to you / Amber!

jessieay:
@pkarman just realized my last comment is not very clear. What I meant to communicate is that either having example input or no example input for **all** fields would be  clearer than as-is. Defer to you on which way we go.

pkarman:
The mock-ups lack any kind of examples, so I've just zapped the placeholder text to make it consistent.

jessieay:
QAing again --form looks great!

just go an error

query url:

https://c2-dev.18f.gov/proposals/query?ncr_work_order%5Bstatus%5D=*&ncr_work_order%5Brequester%5D=&ncr_work_order%5Bclient_data.vendor%5D=&ncr_work_order%5Bclient_data.amount%5D=&ncr_work_order%5Bclient_data.expense_type%5D=*&ncr_work_order%5Bcreated_within%5D=1+month&ncr_work_order%5Bcreated_at%5D=2015-11-11&text=test

logs:

```
February 2nd 2016, 13:19:52.000	INFO	App/0	c2-dev	21:19:52 web.1    |   lib/tabular_data/container.rb:15:in `alter_query'
February 2nd 2016, 13:19:52.000	INFO	App/0	c2-dev	21:19:52 web.1    | 
February 2nd 2016, 13:19:52.000	INFO	App/0	c2-dev	21:19:52 web.1    |   lib/query/proposal/listing.rb:119:in `block in apply_text_filter'
February 2nd 2016, 13:19:52.000	INFO	App/0	c2-dev	21:19:52 web.1    | Completed 500 Internal Server Error in 17ms (ActiveRecord: 5.4ms | Elasticsearch: 5.5ms)
February 2nd 2016, 13:19:52.000	INFO	App/0	c2-dev	21:19:52 web.1    | 
February 2nd 2016, 13:19:52.000	INFO	App/0	c2-dev	21:19:52 web.1    |   lib/query/proposal/listing.rb:43:in `query'
February 2nd 2016, 13:19:52.000	INFO	App/0	c2-dev	21:19:52 web.1    | Started GET "/proposals/query?ncr_work_order%5Bstatus%5D=*&ncr_work_order%5Brequester%5D=&ncr_work_order%5Bclient_data.vendor%5D=&ncr_work_order%5Bclient_data.amount%5D=&ncr_work_order%5Bclient_data.expense_type%5D=*&ncr_work_order%5Bcreated_within%5D=1+month&ncr_work_order%5Bcreated_at%5D=2015-11-11&text=test" for 159.142.71.1 at 2016-02-02 21:19:52 +0000
February 2nd 2016, 13:19:52.000	INFO	App/0	c2-dev	21:19:52 web.1    | Processing by ProposalsController#query as HTML
February 2nd 2016, 13:19:52.000	INFO	App/0	c2-dev	21:19:52 web.1    |   Parameters: {"ncr_work_order"=>{"status"=>"*", "requester"=>"", "client_data.vendor"=>"", "client_data.amount"=>"", "client_data.expense_type"=>"*", "created_within"=>"1 month", "created_at"=>"2015-11-11"}, "text"=>"test"}
February 2nd 2016, 13:19:52.000	INFO	App/0	c2-dev	21:19:52 web.1    | 
February 2nd 2016, 13:19:52.000	INFO	App/0	c2-dev	21:19:52 web.1    | Elasticsearch::Transport::Transport::Errors::BadRequest ([400] {"error":"SearchPhaseExecutionException[Failed to execute phase [query_fetch], all shards failed; shardFailures {[kBMQHCCQRzuZdPnkkNYkHQ][proposals][0]: SearchParseException[[proposals][0]: from[-1],size[-1]: Parse Failure [Failed to parse source [{\"query\":{\"query_string\":{\"query\":\"(test) AND (created_at:[2015-10-11 08:00:00 UTC TO 2015-11-11 08:00:00 UTC])\",\"default_operator\":\"and\"}},\"filter\":{\"bool\":{\"must\":[{\"term\":{\"client_data_type\":\"Ncr::WorkOrder\"}}]}},\"size\":20,\"from\":0}]]]; nested: QueryParsingException[[proposals]
```

pkarman:
@jessieay date parsing fixed in 690ad77

PR #889, opened 2015-12-22
Show "Purchaser" instead of "Approver" on Proposal detail page

jessieay:
* Refer to users as `step_users` rather than `approvers` since some are
  users connected to a proposal via a purchase step

screen shot 2015-12-22 at 2 51 46 pm

https://trello.com/c/HDp7pbS7

### DIFF
--- a/app/controllers/proposals_controller.rb
+++ b/app/controllers/proposals_controller.rb
@@ -50,9 +50,9 @@ class ProposalsController < ApplicationController
   end
 
   def approve
-    approval = proposal.existing_step_for(current_user)
-    approval.update_attributes!(completer: current_user)
-    approval.approve!
+    step = proposal.existing_step_for(current_user)
+    step.update_attributes!(completer: current_user)
+    step.approve!
     flash[:success] = "You have approved #{proposal.public_id}."
     redirect_to proposal
   end
--- a/app/controllers/proposals_controller.rb
+++ b/app/controllers/proposals_controller.rb
@@ -50,7 +50,7 @@ class ProposalsController < ApplicationController
   end
 
   def approve
-    approval = proposal.existing_approval_for(current_user)
+    approval = proposal.existing_step_for(current_user)
     approval.update_attributes!(completer: current_user)
     approval.approve!
     flash[:success] = "You have approved #{proposal.public_id}."
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -31,8 +31,12 @@ class Proposal < ActiveRecord::Base
   acts_as_taggable
 
   has_many :steps
-  has_many :individual_steps, ->{ individual }, class_name: 'Steps::Individual'
-  has_many :approvers, through: :individual_steps, source: :user
+  has_many :individual_steps, ->{ individual }, class_name: "Steps::Individual"
+  has_many :approval_steps, class_name: "Steps::Approval"
+  has_many :purchase_steps, class_name: "Steps::Purchase"
+  has_many :approvers_and_purchasers, through: :individual_steps, source: :user
+  has_many :approvers, through: :approval_steps, source: :user
+  has_many :purchasers, through: :purchase_steps, source: :user
   has_many :completers, through: :individual_steps, source: :completer
   has_many :api_tokens, through: :individual_steps
   has_many :attachments, dependent: :destroy
@@ -77,17 +81,18 @@ class Proposal < ActiveRecord::Base
     approval_delegates.exists?(assignee_id: user.id)
   end
 
-  def existing_approval_for(user)
+  def existing_step_for(user)
     where_clause = <<-SQL
       user_id = :user_id
       OR user_id IN (SELECT assigner_id FROM approval_delegates WHERE assignee_id = :user_id)
       OR user_id IN (SELECT assignee_id FROM approval_delegates WHERE assigner_id = :user_id)
     SQL
+
     steps.where(where_clause, user_id: user.id).first
   end
 
   def subscribers
-    results = approvers + observers + delegates + [requester]
+    results = approvers + purchasers + observers + delegates + [requester]
     results.compact.uniq
   end
 
@@ -137,7 +142,7 @@ class Proposal < ActiveRecord::Base
 
   def add_requester(email)
     user = User.for_email(email)
-    if awaiting_approver?(user)
+    if awaiting_step_user?(user)
       fail "#{email} is an approver on this Proposal -- cannot also be Requester"
     end
     set_requester(user)
@@ -180,7 +185,7 @@ class Proposal < ActiveRecord::Base
   end
 
   # Returns True if the user is an "active" approver and has acted on the proposal
-  def is_active_approver?(user)
+  def is_active_step_user?(user)
     individual_steps.non_pending.exists?(user: user)
   end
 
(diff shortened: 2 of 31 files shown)


jessieay:
also shows up above

screen shot 2015-12-22 at 2 52 23 pm

app/models/proposal.rb
```diff
@@ -31,8 +31,12 @@ class Proposal < ActiveRecord::Base
   acts_as_taggable
 
   has_many :steps
-  has_many :individual_steps, ->{ individual }, class_name: 'Steps::Individual'
-  has_many :approvers, through: :individual_steps, source: :user
+  has_many :individual_steps, ->{ individual }, class_name: "Steps::Individual"
+  has_many :approval_steps, class_name: "Steps::Approval"
+  has_many :purchase_steps, class_name: "Steps::Purchase"
+  has_many :approvers_and_purchasers, through: :individual_steps, source: :user
```
yozlet: If we're using the term `step_users` elsewhere, perhaps we should use it here too? I also worry about the naming of `approvers_and_purchasers` because if/when we add `Steps::Reconcile` then this becomes `approver_and_purchasers_and_reconcilers` (oh my), etc. etc.

jessieay: yeah you are right -- potentially really annoying naming situation down the line.

what do you think about `step_users`? Is that a good name? Makes me think about evil step sisters, but I can't think about anything better...

jessieay: renamed at 493914b0a56c41e319cd43ace5b31ec223ea9169 -- let me know what you think!

app/policies/gsa18f/procurement_policy.rb
```diff
@@ -15,7 +15,7 @@ def can_create!
     def can_cancel!
       not_cancelled! && check(
         (approver? || delegate? || requester? || admin?) && !purchaser?,
-        "Sorry, you are neither the requester, approver or delegate"
+        "Sorry, you are neither the requester, approver, or delegate"
```
pkarman: oxford comma, eh? http://www.grammarly.com/blog/2015/what-is-the-oxford-comma-and-why-do-people-care-so-much-about-it/

app/controllers/proposals_controller.rb
```diff
@@ -50,7 +50,7 @@ def cancel
   end
 
   def approve
-    approval = proposal.existing_approval_for(current_user)
+    approval = proposal.existing_step_for(current_user)
```
pkarman: how do we know that the step in question is an approval type step?

pkarman:
I think re-labeling the step type is great.

I'm a little more ambivalent about the mass renaming of `approval` -> `step` within the code. I understand why the table was renamed and agreed with that. It seems, though, like there are valid reasons for retaining the sense of `approver` within the code when the business use case actually represents "a step of approval" vs "a step of purchase". I.e. I think the renaming is actually more nuanced.

E.g. should the method `existing_approval_for` include a `where` clause that checks the step type, instead of simply being renamed? Or perhaps we go further, and rename `approval_delegates` table to be `step_delegates`? And add a new method called `existing_approval_for` that is similar to `existing_step_for` but includes the step type amongst the qualifiers?

I.e. there are times when in the flow of the code we really want to know if the step user is an approver, as opposed to being a purchaser or some other step type. Was that considered within this renaming?

app/policies/gsa18f/procurement_policy.rb
jessieay: hahah YES! perhaps we should consult a content person on this? :abc:

app/controllers/proposals_controller.rb
jessieay: we do not know if is an approval type, but both Approval and Purchase steps hit this `approve` method. A completed purchase step has a state of `approved` (at least I think this is the case...paging @yozlet )

pkarman: ah! good point. The word "approve" and derivatives are overloaded in this app.

I guess the assumption is that `current_user` would only ever have one pending step of any given flavor at a time?

jessieay:
@pkarman It seems like in most of the places where we are referring to an `approval` in the code, we are actually referring to a `step`. eg: in `app/decorators/proposal_decorator.rb` we did this:

```ruby
step = existing_approval_for(user)
klass = step.class.name.demodulize.downcase.to_sym
scope = [:decorators, :steps, klass]
I18n.t(key, scope: scope)
```

clearly, on the first line we are looking for a `step` rather than an `approval` (we even named the variable `step`) -- this PR makes that clearer by renaming `existing_approval_for` to `existing_step_for`

I think I understand what you mean -- there are cases where we want to know about an approval, not just any old step. 

What I do not see is where in this PR we would want that behavior. Can you be more specific? 

I think there are further changes to be made after this. Eg: the `approve` action in the controller is used by both approval and purchase steps, which is confusing. Ideally we could tackle these piece by piece so we don't have to deal with any massive pull requests.

app/controllers/proposals_controller.rb
pkarman: should the variable also be renamed then?
```
step = proposal.existing_step_for(current_user)
```

pkarman:
How about we only ever use `approve` (the verb) to indicate that the status of the step will change. I.e. as a synonym for _finish_.

Otherwise, we use `step` in place of `approval` (the noun) to indicate what the thing is that is being acted upon (approved/finished).

pkarman:
That means that, ideally, we would not have "approval" steps at all, but maybe "accept" or "signoff" steps?

jessieay:
@pkarman I think it still makes sense to use the concept of `approval` for some steps because those steps are about getting approval. `accept` or `signoff` (as names) could also be related to a purchase, couldn't they? 

What about keeping step names as `purchase` or `approval` and then calling the completed step `completed` ?

pkarman:
@jessieay that's fine. Just the inverse of my suggestion: keep the noun `approval` and change the verb from `approve` to `complete`.

jessieay:
@pkarman do you think we should do that as part of this PR, or separately? It's already pretty sizable but happy to do either way...

pkarman:
fine to do it as a separate PR. The whole naming thing has been stretched over months and many PRs.

jessieay:
@pkarman ok! let me know if there's anything in this PR that should be changed. Will work on other renaming this afternoon!

app/policies/proposal_policy.rb
```diff
@@ -7,17 +7,19 @@ def initialize(user, record)
   end
 
   def can_approve!
```
pkarman: maybe rename to `can_complete!` ?

jessieay: renamed at 349747ef0503069688d5fe8af6efadb4fe61608c :cherries:

pkarman:
@jessieay just this https://github.com/18F/C2/pull/889#discussion_r48369493 and otherwise looks great.

app/controllers/proposals_controller.rb
jessieay: renamed at 46924c3519716eb766e98e33599a11d4499f029f :cat:

jessieay:
@pkarman updated

pkarman:
looks good on c2-dev
screen shot 2015-12-24 at 9 55 12 am

PR #890, opened 2015-12-22
Remove comments from config files and alpha order

jessieay:
* Comments make config harder to read
* Alpha order makes things easier to find


jessieay:
hmmm @pkarman I just re-read your comment and it seems like maybe you wanted comment deletion and re-ordering to happen separately. Let me know if this is still difficult to read and I can undo the re-ordering portion.

PR #892, opened 2015-12-22
Fix path-checking spec by using a waiting matcher

yozlet:
This "Discard Changes" spec keeps breaking when run locally because testing `current_path` doesn't use Capybara's waiting mechanisms to ensure that `click_link` is done. But good news! [There's now a waiting matcher for current_path](https://github.com/jnicklas/capybara/pull/1567), so this fix uses that.


jessieay:
THANK YOU for fixing this. Was totally avoiding it and hoping it would just go away.

PR #893, opened 2015-12-23
Move `current_approver` logic out of Ncr::WorkOrder

jessieay:
* We already have the concept of `currently_awaiting_approvers` in
  StepManager
* Part of https://trello.com/c/smBumRFX

### DIFF
--- a/app/decorators/ncr/work_order_decorator.rb
+++ b/app/decorators/ncr/work_order_decorator.rb
@@ -20,7 +20,7 @@ module Ncr
     end
 
     def pending_approver_email_address
-      approver_email_address(proposal.currently_awaiting_approvers.first)
+      approver_email_address(proposal.currently_awaiting_step_users.first)
     end
 
     def approver_email_address(approver)
--- a/app/decorators/ncr/work_order_decorator.rb
+++ b/app/decorators/ncr/work_order_decorator.rb
@@ -2,26 +2,26 @@ module Ncr
   class WorkOrderDecorator < Draper::Decorator
     delegate_all
 
-    EMERGENCY_APPROVER_EMAIL = 'Emergency - Verbal Approval'
-    NO_APPROVER_FOUND = 'No Approver Found'
+    EMERGENCY_APPROVER_EMAIL = "Emergency - Verbal Approval"
+    NO_APPROVER_FOUND = "No Approver Found"
 
     def current_approver_email_address
-      approver_email_address(current_approver)
+      if proposal.approved?
+        final_approver_email_address
+      else
+        pending_approver_email_address
+      end
     end
+
+    private
 
     def final_approver_email_address
       approver_email_address(final_approver)
     end
 
-    def status_aware_approver_email_address
-      if proposal.approved?
-        final_approver_email_address
-      else
-        current_approver_email_address
-      end
+    def pending_approver_email_address
+      approver_email_address(proposal.currently_awaiting_approvers.first)
     end
-
-    private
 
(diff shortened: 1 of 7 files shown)


app/decorators/ncr/work_order_decorator.rb
```diff
@@ -2,27 +2,27 @@ module Ncr
...
       end
     end
 
     private
 
+    def final_approver_email_address
+      approver_email_address(final_approver)
+    end
+
+    def pending_approver_email_address
```
pkarman: is the use of `pending` here meant to correlate with the step status? Does `pending` == `current` ?

pkarman:
I'm +1 for the moving of the method to the new class.

I'm a little confused by the method renaming. Instead of `current` now the view uses `final` which seems like a bigger change than simply moving methods from one class to another. Is this renaming intended to reflect a business rule change? Or part of clarifying something internally?

app/decorators/ncr/work_order_decorator.rb
jessieay: pending approver is the current approver *if* a request is pending. otherwise, it is the "final" approver

jessieay:
@pkarman did you mean to say "instead of `final` the view now uses `current`?

re: "Is this renaming intended to reflect a business rule change? Or part of clarifying something internally?"

previously, the method `current_approver` in `WorkOrder` did this logic:

- if a proposal is PENDING
-> return the user for the first actionable step
- otherwise, if there is an approving official
-> return approving official
- if it is an emergency (there would never be any approvals for an emergency)
->  return nil
- ELSE
-> return the first system approver
(system approvers are all budget mailboxes for NCR requests)

this method was only being used in the decorator, within the `current_approver_email_address` method.

the `current_approver_email_address` method was only being used by another method in the decorator, `status_aware_approver_email_address`, which returned the `current_approver_email_address` for non-approved requests. 

So there was a bit of duplicate logic there -- `status_aware_approver_email_address` only returns `current_user_approver_address` (which uses `current_approver`) for pending requests, and then `current_approver` checks on the status of the request yet again. 

Also, in the case of an emergency request, it would be marked as `approved`, and the `approver_email_address` method in the decorator takes care of that case, so we don't need to worry about handling that logic elsewhere (like the `current_approver` method did).

So there should be no behavior change here. Previously, places where we used `final_approver_email_address` were places that were for approved requests.  The `current_approver_email_address` method returns the final approver email address for approved requests, so the same info will be returned.

pkarman:
> @pkarman did you mean to say "instead of final the view now uses current?

Yes, I did. Thanks for the catch.

Funny. I didn't realize till just now reading your narrative that I wrote all the code we're now discussing. As Gandalf said, I have no memory of this place.

Fortunately, Github remembers for me. The decorator logic was part of these PRs: #713 #681 

In particular, the changes relative to `lib/ncr/reporter.rb` are worrisome to me. IIRC #713 exists mostly to fix from `current` to `status_aware` because `current` was not correct. I suspect our test coverage is not complete enough to reflect that there may be a regression introduced within the changes you have here, which appear to me to be switching it back.

pkarman:
@jessieay I have added tests against master in #897 that should, when also merged to this PR, reveal whether any of my fears are founded.

pkarman:
@jessieay tests in #897 were merged to master. If you re-base, we'll see if my fears were well-founded or not.

jessieay:
@pkarman updated. had to rename one method that was renamed since this PR was first opened (c587bfd) but otherwise it is all green

PR #894, opened 2015-12-23
Get rid of old communicart references

jessieay:
* We are C2!!


pkarman:
cc/ @phirefly 

+1 to renaming. Not sure how/if partner needs to sign off on this.

jessieay:
This is just stuff that was forgotten. The rest of the app was renamed long
ago.

app/views/help/email_filtering.md
```diff
@@ -4,7 +4,7 @@ Yes. You’ll need to set up a filter in Gmail.
 
 ## First, select the C2 request messages to filter.
 
-1. Enter the following into the search bar at the top of your inbox: `from:communicart.sender@gsa.gov`
+1. Enter the following into the search bar at the top of your inbox: `from:c2.sender@gsa.gov`
```
pkarman: is there such an email address as `c2.sender@gsa.gov` ?

pkarman:
Was it really forgotten? This isn't just internal code references but all the public-facing text.

I remember talking with @afeld about this some months ago; seems like it wasn't just inertia keeping us from renaming it publicly.

afeld:
Correct – it was that users are probably more familiar with that name than C2, so we'd want to make sure they know the changeover is happening.

screen_shot_2015-12-23_at_12_49_05_pm

app/views/help/email_filtering.md
phirefly: There is no email for this. I can set it up. We will need to also login/create users on servers when this gets pushed.

jessieay: ah yes I can return it to the old one. Did a global find an replace and thought I'd reverted all instances of `communicart.sender` but I see that I missed this one.

jessieay: updated at 5e203b55a2f8e750e7a1655be3057e74c692a19a

jessieay:
We had a card for this here https://www.pivotaltracker.com/story/show/103027122

so I assumed this decision was a done deal. @phirefly what do you think we should do re: communicating with clients re: @afeld and @pkarman comments?

phirefly:
Thanks! I will put a request in for this new gsa email and create a story to remove this when it's ready.

I have created these gmail addresses:
`c2.approver`
`c2.ofm.approver`
`c2.budget.approver`

We are aligned with CAP folks in moving away from "Communicart". This is a reasonable first step. TY

jessieay:
@phirefly cool, thanks!

I kept the email as `communicart.sender` in here - will open a separate PR to change the email addresses

PR #898, opened 2015-12-28
rubygems.org giving SSL errors -- try this mirror instead

pkarman:
also see https://github.com/rubygems/rubygems.org/issues/644#issuecomment-33130622 for hardcoded alternate mirror list


pkarman:
verified with production deployment. see discussion here https://18f.slack.com/archives/ruby/p1451318939001011

PR #899, opened 2015-12-30
Saved searches and reports

pkarman:
Story: https://trello.com/c/AEEDqLVV

This PR implements a new table `reports` and UI for saving a search as a `Report`, and downloading a search resultset as a CSV file.

PR is dependent on #888 so that PR should be merged first.


rememberlenny:
@pkarman Do we need all green from Hakiri before merging?

pkarman:
@rememberlenny yes I will try and track it down.

pkarman:
@rememberlenny all :green_heart: now

jessieay:
@pkarman this PR is quite large -- would it be possible to break it into a few discrete chunks for reviewing? The most difficult part with it being this size is comparing test coverage with code. I like to scroll back and forth but with this many changes it gets hard to keep track of it all.

pkarman:
@jessieay this includes everything from #871 and #888 so let's review/merge those first and then this one should "shrink" considerably.

jessieay:
@pkarman ah yes makes sense. didn't realize that was the case.

pkarman:
@jessieay et al, this is re-based and ready for review.

app/assets/javascripts/search.js
```diff
@@ -18,9 +20,71 @@ $(document).ready(function() {
...
+  // defined inline on HTML page
+  if (typeof C2searchQuery != "undefined") {
+    $("#save-search-query").text(C2searchQuery.humanized);
+  }
+  $("#save-search form").on("submit", function(e) {
+    e.preventDefault();
+  });
+  $("#save-search-button").click(function() {
+    var btn = $(this);
+    var form = $("#save-search form");
```
jessieay: I like the explicit naming of `savedSearchName` below -- perhaps this var would be more clear if it were named `savedSearchForm` ?

app/assets/javascripts/search.js
```diff
@@ -18,9 +20,71 @@ $(document).ready(function() {
...
+
+    // validation ok -- fire the XHR
+    form.find('input').prop("disabled", true);
+    btn.prop("disabled", true);
+    $.post("/reports.json", {
+      query: JSON.stringify(C2searchQuery),
+      name: savedSearchName.val()
+    })
+    .fail(function(payload) {
+      form.append($('<div class="form-alert alert alert-danger">Something went wrong! Please try again or contact your administrator.</div>'));
```
jessieay: link to feedback form instead of instructing to contact an administrator?

app/assets/stylesheets/webapp/webviews.scss
```diff
@@ -294,6 +294,14 @@ body.webapp {
     text-decoration: underline;
     color: #0C75E1;
   }
+  #pages {
```
jessieay: this is unrelated to this PR, but I've always found it weird that we have a `webapp` dir in our assets or a file named `webviews` since this whole app is a web app at the moment. maybe something to rename in the near future. created a card for this: https://trello.com/c/i0COT88J

app/controllers/proposals_controller.rb
```diff
@@ -4,7 +4,7 @@ class ProposalsController < ApplicationController
   skip_before_action :authenticate_user!, only: [:approve]
   skip_before_action :check_disabled_client, only: [:approve]
   # TODO use Policy for all actions
-  before_action ->{authorize proposal}, only: [:show, :cancel, :cancel_form, :history]
+  before_action -> { authorize proposal }, only: [:show, :cancel, :cancel_form, :history]
```
jessieay: :heart_eyes:

app/controllers/proposals_controller.rb
```diff
@@ -34,13 +34,9 @@ def cancel_form
 
   def cancel
     if params[:reason_input].present?
-      comments = "Request cancelled with comments: " + params[:reason_input]
-      proposal.cancel!
-      proposal.comments.create!(comment_text: comments, user: current_user)
-
+      cancel_proposal
```
jessieay: `cancel_proposal_and_send_cancellation_emails` ?

app/decorators/proposal_decorator.rb
```diff
@@ -54,4 +54,12 @@ def step_text_for_user(key, user)
     scope = [:decorators, :steps, klass]
     I18n.t(key, scope: scope)
   end
+
+  def self.csv_headers
```
jessieay: these methods could be good candidates for the Fields objects created in https://github.com/18F/C2/pull/920 once that is merged.

app/views/proposals/download.csv.haml
```diff
@@ -0,0 +1,4 @@
+= CSV.generate_line [ProposalDecorator.csv_headers, current_user.client_model.csv_headers].flatten
+- @proposals_data.rows.each do |proposal|
+  != CSV.generate_line proposal.decorate.as_csv
```
jessieay: woahh what does `!=` do in haml? (not very googleable, unfortunately)

app/views/proposals/query.html.haml
```diff
@@ -2,16 +2,57 @@
...
+      #pages
+        = paginate @proposals_data.es_response, current_page: (params[:page]||1).to_i
   - if @proposals_data.rows.empty?
-    %p.empty-list-label
+    .empty-list-label
       No results
+
+:javascript
+  // global var referenced by search.js
+  var C2searchQuery = #{@search_query.to_json}
```
jessieay: ` c2SearchQuery` ?

config/routes.rb
```diff
@@ -48,6 +49,9 @@
     resources :observations, only: [:create, :destroy]
   end
 
+  resources :reports, only: [:index, :show, :create, :destroy] do
```
jessieay: think we don't need this `do` and `end` unless there's a nested route we forgot here?

spec/models/report_spec.rb
```diff
@@ -0,0 +1,36 @@
+describe Report do
+  it "#client_query" do
+    report = create(:report)
+    expect(report.client_query.to_s).to eq "client_data.amount:<123"
```
jessieay: these values feel a little bit mysterious because they come from the Report factory's default vals. perhaps we could set them explicitly in the factory instantiation so it is clearer what the method has to do with object attrs?

app/views/proposals/download.csv.haml
pkarman: http://haml.info/docs/yardoc/file.REFERENCE.html#unescaping_html

app/views/proposals/query.html.haml
pkarman: I initial-capped it to indicate it is a global var. Happy to use whatever syntax best communicates that. Some folks prepend with a `g` -- I was using `C2`.

pkarman:
@jessieay thanks for the review. I think I've hit all your comments.

jessieay:
QAing now

Looks like icon isn't working correctly on CF

screen shot 2016-02-03 at 10 40 16 am

jessieay:
CSV download appears to only grab page 1 -- is that expected?

also, it has extra newlines -- looks nice in CSV format, but spacing is kind of annoying when formatted to chart (have to delete newlines to get accurate record count). this also means the CSV only has 10 records because there are 10 newlines (download first page of results, which is 20, but guessing newlines are being counted as records somehow)

screen shot 2016-02-03 at 10 45 00 am

pkarman:
re: icon missing, I wonder if its a browser caching issue? I see the icons.

screen shot 2016-02-03 at 1 35 09 pm

pkarman:
@jessieay the CSV should not be limited nor should there be extra newlines. Fixing now.

jessieay:
@pkarman I see it on that page too but I do not see it at the root path on dev. do you see the icon when you are at https://c2-dev.18f.gov/?

pkarman:
ah the home page. Looks like that page does not share the application layout. Fixed in 6b6c067

CSV bugs also fixed.

pkarman:
@jessieay I think I've hit all your QA comments.

jessieay:
Ok cool, round 2 of QA done. Only real bug I am seeing is that I created a search 101 results but the CSV download only has 93 rows including the header. (a search with 4 results did return a CSV with all 4 results) Another search with 93 results had 84 items in the CSV. Not sure if this is a problem with Kaminari pagination / record counting or with the CSV download feature.

Otherwise, looks great!

Here are some notes on things we can improve for future iterations but don't need to happen Right Now:

- When viewing an existing report, "Save as Report" link still shows on righthand, which is kinda confusing
- Reports page should have some content that prompts the user to "search to create a report" or something when it is blank. Before I created my first report, I expected there to be either a report or a link to create a report there.
- Having the grey box that shows the SQL equivalent of a search is helpful for people who know SQL but might be confusing for others. Maybe keeping the advanced search fields populated in the field would be clearer for allowing user to understand the parameters of a search. Maybe put fields on page when a user is viewing search results rather than modal (although the modal is nice for creating a new search)

pkarman:
@jessieay can you point me at the specific searches that gave those unexpected results?

jessieay:
@pkarman sure thing!

https://c2-dev.18f.gov/proposals/query?report=1&text=test

101 results

screen shot 2016-02-04 at 10 08 04 am

93 rows (92 rows of data) in exported CSV

pkarman:
@jessieay good QA catch!

The problem was bad data in the dev environment: the missing records had no `requester_id` defined, which meant they were included in the ES search results, but the SQL query that fetched them from the db silently skipped them because it depended on a join against the `users` table, which failed because there was no requester_id defined.

So I deleted all the bad data (part of the `vacuum:old_proposals` rake task) and re-built the index. Now the search results match the CSV.

jessieay:
@pkarman glad we got to the root of the problem! merging

PR #901, opened 2016-01-22
upgrade Nokogiri

afeld:
Fixes security vulnerability:

https://github.com/sparklemotion/nokogiri/blob/master/CHANGELOG.rdoc#1672--2016-01-20

PR #902, opened 2016-01-22
update gems

afeld:
Builds on #901.


Gemfile
```diff
@@ -1,8 +1,8 @@
-source "http://mirror.ops.rhcloud.com/mirror/ruby/"
+source "https://rubygems.org"
 ruby "2.2.3" # this should match `.ruby-version` and docs/setup.md
 
 gem "active_model_serializers"
-gem "activeadmin", github: "activeadmin"
+gem "activeadmin", github: "activeadmin", ref: "d787029e"
```
afeld: Wasn't sure if there was a reason for using the latest from GitHub, but since the latest from `master` isn't necessarily stable, I was nervous about upgrading it. We should probably add a comment here about why we are using this instead of a released gem, but wasn't sure for the reason.

pkarman:
@afeld activeadmin master is being used because they haven't released a gem in so long, and there were some issues fixed in master. I'm happy to peg it to a specific rev.

pkarman:
@afeld this PR has been superseded yes?

afeld:
A few of the security-related updates were already made in the meantime (#905 and #906), but probably worth doing an across-the-board upgrade sometime soon.

pkarman:
superseded by #911

PR #907, opened 2016-01-26
Purchase delegate bugfix

annalee:
Fixes the bug wherein purchase delegates weren't getting the "mark as purchased" button, per https://trello.com/c/ymdHIPMF .

(and no I can't spell 'assignee' either).

### DIFF
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -40,7 +40,7 @@ class Proposal < ActiveRecord::Base
   has_many :completers, through: :individual_steps, source: :completer
   has_many :api_tokens, through: :individual_steps
   has_many :attachments, dependent: :destroy
-  has_many :approval_delegates, through: :approvers, source: :outgoing_delegations
+  has_many :approval_delegates, through: :step_users, source: :outgoing_delegations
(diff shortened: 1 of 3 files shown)


yozlet:
Note that this is a temporary fix in lieu of a larger renaming (see https://trello.com/c/KeQbIIvQ)

app/models/proposal.rb
```diff
@@ -40,7 +40,7 @@ class Proposal < ActiveRecord::Base
   has_many :completers, through: :individual_steps, source: :completer
   has_many :api_tokens, through: :individual_steps
   has_many :attachments, dependent: :destroy
-  has_many :approval_delegates, through: :approvers, source: :outgoing_delegations
+  has_many :approval_delegates, through: :step_users, source: :outgoing_delegations
```
pkarman: ah yes! that looks exactly right.

pkarman:
this looks right to me.

it should be merged into both `stable-v1` branch and master, assuming it should be deployed soon.

phirefly:
:shipit: Merging into master and will cherry pick into stable branch separately

PR #911, opened 2016-02-01
bundle update

pkarman:
supersedes #902


jessieay:
some ppl like to update gems in separate commits for git bisect help in future. this is a tool that automates the process https://github.com/lpender/bummr

I don't mind doing all at once, though, just thought I'd throw that out there since I know you're a git bisect fan

pkarman:
that is a very cool looking gem. I will try it, and since we have test failures, we can hope it will help diagnose.

pkarman:
bummr is cool. new PR coming.

PR #912, opened 2016-02-02
NCR Work Order associated with NCR Organization

jessieay:
* Use foreign key instead of string
* https://trello.com/c/HRGz6lNP/58-make-ncr-organization-use-a-real-foreign-key

### DIFF
--- a/app/services/proposal_update_recorder.rb
+++ b/app/services/proposal_update_recorder.rb
@@ -24,10 +24,10 @@ class ProposalUpdateRecorder
   end
 
   def update_comment_format(key)
-    if key !~ /id/
-      "#{bullet}*#{property_name(key)}* was changed " + former_value(key) + "to #{new_value(key)}"
-    else
+    if key =~ /id/
       "#{bullet}*#{association_name(key)}* was changed " + former_association_value(key) + "to #{new_association_value(key)}"
+    else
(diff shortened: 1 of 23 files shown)


jessieay:
this still has some failing specs but would be great if you could look at it @pkarman -- there is some real ugliness happening in `app/services/proposal_update_recorder.rb`

app/services/proposal_update_recorder.rb
```diff
@@ -24,7 +24,11 @@ def changed_attributes
   end
 
   def update_comment_format(key)
-    "#{bullet}*#{property_name(key)}* was changed " + former_value(key) + "to #{new_value(key)}"
+    if key !~ /id/
```
pkarman: maybe switch the order in order to make this a positive assertion instead of a negative one? i.e. `if key =~ /id/` first.

pkarman:
@jessieay I haven't looked at the failing tests at all, but the code looks fine to me. Nothing surprising.

jessieay:
@pkarman this is ready for real review now - thanks for the feedback on v0!

pkarman:
+1

I will QA

pkarman:
tested on c2-dev:

* created new work order https://c2-dev.18f.gov/proposals/269
* edited work order to add NCR Org
* edited work order again to change NCR Org
* verified NCR Org was "sticky" across form reloads

pkarman:
screen shot 2016-02-03 at 9 12 45 am

PR #914, opened 2016-02-02
add document describing setup for a new client

pkarman:
Trello: https://trello.com/c/i4Koni9u

The example code is based on the `Test::ClientRequest` code in `spec/`. NOTE this example code hasn't been tested. I would expect that anyone adding a new client would amend this document based on real-world experience. This is intended as a starting place.

### DIFF
--- a/doc/client_setup.md
+++ b/doc/client_setup.md
@@ -1,0 +1,199 @@
+## Setting up a new C2 client
+
+This document describes the details of setting up a new C2 client.
+
+### Overview
+
+A *client* represents a group of Users within the C2 system that share
+a data model. Existing examples include **GSA18F** and **NCR**.
+
+Within a *client* users may be assigned different roles. See the [Authorization](roles.md)
+document for details on roles.
+
+A *client* namespace is assigned to each User via a `client_slug` model value. The
+convention is that the `client_slug` is a downcased string matching the client model
+namespace. For example `ncr` for **NCR**.
+
+A User without an assigned `client_slug` may not create, edit, view or search Proposals,
+since the User will have no authorization context. The User may, however, login to the system.
+
+### Models
+
+The first thing to identify when adding a new client is the data schema for the client-specific
+extension to the Proposal. This extension is called *client_data*.
+
+Here is an example db migration for a new client called **Foo** with a client_data schema
+called **SpendingRequest**:
+
+```ruby
+class AddFooSpendingRequests < ActiveRecord::Migration
+  def change
+    create_table :foo_spending_requests do |t|
+      t.decimal  :amount
+      t.string   :project_title
+      t.datetime :created_at
+      t.datetime :updated_at
+    end
+  end
+end
+```
+
+Here is an example of the corresponding ActiveRecord model class:
+
+```ruby
+module Foo
+  def self.table_name_prefix
+    "foo_"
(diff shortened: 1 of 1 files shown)


doc/client_setup.md
```diff
@@ -0,0 +1,199 @@
...
+a data model. Existing examples include **GSA18F** and **NCR**.
+
+Within a *client* users may be assigned different roles. See the [Authorization](roles.md)
+document for details on roles.
+
+A *client* namespace is assigned to each User via a `client_slug` model value. The
+convention is that the `client_slug` is a downcased string matching the client model
+namespace. For example `ncr` for **NCR**.
+
+A User without an assigned `client_slug` may not create, edit, view or search Proposals,
```
jessieay: this is not *technically* true at the moment. When my client slug is set to `gsa18f`, I can still create an NCR work order if I know the URL for doing so. But the link for creating one is not surfaced in the header.

jessieay: same goes for users without a client slug at all. They won't see a link to create a work order or procurement, but it is possible to do so.

doc/client_setup.md
```diff
@@ -0,0 +1,199 @@
...
+called **SpendingRequest**:
+
+```ruby
+class AddFooSpendingRequests < ActiveRecord::Migration
+  def change
+    create_table :foo_spending_requests do |t|
+      t.decimal  :amount
+      t.string   :project_title
+      t.datetime :created_at
+      t.datetime :updated_at
```
jessieay: super small nit but these two columns can be and are usually represented via `t.timestamps null: false`

doc/client_setup.md
```diff
@@ -0,0 +1,199 @@
...
+
+```ruby
+FactoryGirl.define do
+  factory :foo_spending_request, class: Foo::SpendingRequest do
+    amount 123
+    project_title "I am a test request"
+    association :proposal, client_slug: "foo"
+
+    trait :with_approvers do
+      association :proposal, :with_serial_approvers, client_slug: "foo"
```
jessieay: should we get into describing approver flows?

jessieay:
This document is very clear and helpful.

One thing I think we should add is some information about how to deal with approvers. Who receives emails / approves requests is an important part of any purchase request.

doc/client_setup.md
pkarman: @jessieay you might be seeing something different than "normal" users because you are an admin. These tests demonstrate that "normal" users are limited based on `client_slug`.

https://github.com/18F/C2/blob/master/spec/features/client_slug_authz_spec.rb

doc/client_setup.md
pkarman: agreed. fixed in 7623706

doc/client_setup.md
pkarman: Maybe? Probably? That seems out of scope for this doc though. You can set up a new client w/o needing to know much about approver flows, though they are important.

doc/client_setup.md
jessieay: oh ok, good to know

jessieay:
Perhaps this should get a set of @phirefly eyes before merging?

phirefly:
This looks great! TY
:shipit:

PR #915, opened 2016-02-02
if final step is_a Purchase, use different text in email

pkarman:
Trello: https://trello.com/c/6cqZ3aHm


jessieay:
this looks great! do we need to merge differently because this is 18F-specific?

pkarman:
@jessieay we can cherry-pick into `stable-v1` if we need to deploy this to master before NCR is greenlit.

PR #917, opened 2016-02-02
try to avoid NR template error

pkarman:
Trello: https://trello.com/c/BqUkShEV


jessieay:
any way to QA this?

pkarman:
not really. This wasn't a reproducible bug, IME. I'm really baffled by it, and can only assume that the stack trace is somehow reporting the wrong file line number.

PR #921, opened 2016-02-03
Lazily evaluate logs for expired record cleaner

jessieay:
* Will only log if log level is set to `info` or lower
* The default Rails log level is debug in all environments (lowest)
* Log level is set to `warn` in test env so output not shown
* http://guides.rubyonrails.org/debugging_rails_applications.html#the-logger#impact-of-logs-on-performance

### DIFF
--- a/config/environment.rb
+++ b/config/environment.rb
@@ -5,5 +5,5 @@ require File.expand_path("../application", __FILE__)
 C2::Application.initialize!
 
 C2::Application.configure do
-  config.logger = Logger.new(STDERR)
+  config.logger = Logger.new(STDOUT)
 end
--- a/config/environment.rb
+++ b/config/environment.rb
@@ -3,3 +3,7 @@ require File.expand_path("../application", __FILE__)
 
 # Initialize the Rails application.
 C2::Application.initialize!
+
+C2::Application.configure do
+  config.logger = Logger.new(STDERR)
+end
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -8,6 +8,7 @@ C2::Application.configure do
   config.cache_classes = true
   config.consider_all_requests_local = true
   config.eager_load = false
+  config.log_level = :warn
   config.middleware.use RackSessionAccess::Middleware
   config.serve_static_files = true
   config.static_cache_control = "public, max-age=3600"
--- a/lib/expired_record_cleaner.rb
+++ b/lib/expired_record_cleaner.rb
@@ -36,23 +36,23 @@ class ExpiredRecordCleaner
   end
 
   def handle_no_requester(proposal)
-    Rails.logger.info { puts "#{proposal.id} <= no Requester defined" }
+    Rails.logger.info { "#{proposal.id} <= no Requester defined" }
 
     if @ok_to_act
       proposal.destroy
     else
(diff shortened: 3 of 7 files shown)


jessieay:
haha check out those code climate errors -- I kept it as a `puts` since that is what we had, but happy to just use the Rails logger. I am not sure why this is a rubocop rule. Thoughts?

pkarman:
fwiw: the reason I used `puts` originally is that this library is used via a Rake task in an interactive terminal, and Rails.logger writes to a file. There are clever ways to avoid this (e.g. http://macbury.ninja/2014/8/better-loging-in-rake-tasks) but puts is simple and Just Works.

config/environment.rb
```diff
@@ -3,3 +3,7 @@
 
 # Initialize the Rails application.
 C2::Application.initialize!
+
+C2::Application.configure do
+  config.logger = Logger.new(STDERR)
```
pkarman: how does this effect existing logging in cloud.gov envs?

jessieay:
What I was hoping to achieve with this Pr is to move away from what I am currently seeing, which is a bunch of puts statements when I run the test suite:

screen shot 2016-02-03 at 10 19 37 am

It might make more sense to just wrap these logging statements in a `rails.env` if statement, but thought I'd try this method instead since I learned about log levels in Rails just yesterday. what do you think?

pkarman:
ah ok I see your motivation.

Yes, Rail logger levels are great, and I keep those in mind when I `.warn` vs `.debug`.

For the 2 cases you are trying to quiet, I don't think the Rails.logger would help. I would suggest a `Rails.env.test?` check for the `lib/expired_record_cleaner.rb` case, and checking `if ENV['ES_DEBUG']` for the ES helper case.

config/environment.rb
jessieay: it seems that all logs would go to STDERR. This does not seem to unusual. Heroku automatically redirects all Rails logs to STDOUT http://stackoverflow.com/a/7035237/1019369

jessieay:
@pkarman The changes here *should* help because I moved the logging messages into blocks, so they will be lazily evaluated based on the log level of the rails app: http://guides.rubyonrails.org/debugging_rails_applications.html#impact-of-logs-on-performance

What took my a while to grok is that the lower the log level of a Rails app, the more messages that will show. Debug is lowest. We have the level set to `info` on prod so anything tagged with info or above will show up in prod. 

This PR sets the log level of test env to `warn` so these messages will not be evaluated (because they are `info` and that is a lower level than `warn`

pkarman:
ok I will QA

pkarman:
So the behavior here is what I expected: when running a rake task now appear in the log file, instead of in the terminal. You can try this yourself by running:
```
rake vacuum:old_proposals
```
and tailing the `log/development.log` file in a different terminal window. I would expect you to see zero output from the command in the same window where it is running, which IME, gives the impression that either (a) nothing happened or (b) everything finished successfully. If an error message appears, it will be in a log file, not in the terminal window.

What this means is that folks running rake tasks will not see output in the same window as where they have executed the command, which means they are far less likely to see the output at all unless they also remember to tail the log file at the same time. That feels sub optimal to me. Might as well not log it at all.

jessieay:
@pkarman OK I will close this and reopen a PR with changes based on env rather than logging the output.

PR #922, opened 2016-02-03
add elasticsearch:true tag to specs that require it

pkarman:
since specs run in a random order, it's hard to know if we've added the necessary `elasticsearch: true` tag to all relevant specs. Let's leave this PR open and add to it as CI reveals more that need it.

### DIFF
--- a/spec/features/return_to_spec.rb
+++ b/spec/features/return_to_spec.rb
@@ -1,9 +1,9 @@
-describe "the return_to url option" do
+describe "the return_to url option", elasticsearch: true do
   include ReturnToHelper
 
   it 'defaults to the proposals listing' do
     login_as(create(:user))
-    visit '/proposals/query?text=test'
+    visit query_proposals_path(text: "test")
     expect(page).to have_content('Back to main portal')
     click_on('Back to main portal')
     expect(current_path).to eq('/proposals')


jessieay:
I think this is slightly irrelevant, but since the es spec helper refactor yesterday all of my specs are now passing locally except for 1. this one is consistent failing:

```

     Elasticsearch::Transport::Transport::Errors::InternalServerError:
       [500] {"error":{"root_cause":[{"type":"null_pointer_exception","reason":null}],"type":"search_phase_execution_exception","reason":"all shards failed","phase":"query_fetch","grouped":true,"failed_shards":[{"shard":0,"index":"proposals","node":"LYcZWYcmTri4jr7hqrKBIQ","reason":{"type":"null_pointer_exception","reason":null}}]},"status":500}
     # ./spec/lib/query/proposal/search_spec.rb:55
``

pkarman:
@jessieay I have not seen that particular failure. I've started running that test in a `watch` loop just now and will look for any errors. Is it possible that there is something amiss with your particular elasticsearch installation?

pkarman:
I've not seen any other regular failure on travis except for the specs altered in this PR, so I'm removing the [WIP] and this can be merged.

pkarman:
@jessieay I do not get that particular error (maybe it is specific to version 2.1 and I have 1.7) but I do very occasionally get this error:
screen shot 2016-02-04 at 10 55 41 am

jessieay:
How about that, I finally figured out how to switch to an older version of elasticsearch and it is no longer failing (http://effectif.com/mac-os-x/installing-specific-version-of-homebrew-formula was v helpful)

PR #924, opened 2016-02-04
Add flag for specs that require test model

jessieay:
* Make running individual specs faster because we aren't creating those
  tables before every test run

### DIFF
--- a/spec/models/test/client_request_spec.rb
+++ b/spec/models/test/client_request_spec.rb
@@ -1,4 +1,4 @@
-describe Test::ClientRequest do
+describe Test::ClientRequest, test_client_reuest: true do
   it_behaves_like "client data"
 
   describe "#editable?" do
@@ -7,5 +7,4 @@ describe Test::ClientRequest do
       expect(client_request).to be_editable
     end
   end
-
 end
--- a/spec/factories/gsa18f/procurements.rb
+++ b/spec/factories/gsa18f/procurements.rb
@@ -9,6 +9,16 @@ FactoryGirl.define do
     urgency Gsa18f::Procurement::URGENCY[10]
     association :proposal, client_slug: "gsa18f"
 
+    transient do
+      requester nil
+    end
+
+    after(:create) do |procurement, evaluator|
+      if evaluator.requester
+        procurement.proposal.update(requester: evaluator.requester)
+      end
+    end
+
     trait :with_steps do
       after(:create) { |procurement| procurement.add_steps }
     end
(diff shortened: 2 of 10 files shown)


pkarman:
I understand where you're going with this change. Have you measured before/after to see how much faster tests run?

Here's what I got:
`Test.setup_models took 0.00766500 seconds`
`Test.teardown_models took 0.00583000 seconds`

Those methods run only once per suite, not before/after each spec test.

My concern is that if I have to remember to tag each spec that requires the `Test::ClientRequest` model in order to use that model, I will often forget, resulting in at least one more run of the test in order to remember. If one goal of the existence of that test model is to encourage its use, rather than using the NCR or GSA18F models, why prejudice its use? We don't have to explicitly ask that the NCR or GSA18F models be created, they just are.

SQL read/write is going to be the biggest timedrag on our tests. Have you considered something like http://philippe.bourgau.net/simplest-way-to-speed-up-rspec-with-in-memory-sqlite-db/ ?

spec/models/test/client_request_spec.rb
```diff
@@ -1,4 +1,4 @@
-describe Test::ClientRequest do
+describe Test::ClientRequest, test_client_reuest: true do
```
yozlet: Typo?

jessieay: indeed! hmm I wonder if that means we don't even need it here.

jessieay:
@pkarman What I am hoping to work towards for speeding up our test suite is to remove as much as possible from our `before :suite` block.

I started with this because it seemed straightforward and I've come into a few cases now where I run an individual spec totally unrelated to this test client model table and then my test DB gets out of sync and I need to drop it before I can run specs again. 

You are right that there are not large performance improvements for this particular change.

> My concern is that if I have to remember to tag each spec that requires the Test::ClientRequest model in order to use that model, I will often forget, resulting in at least one more run of the test in order to remember. If one goal of the existence of that test model is to encourage its use, rather than using the NCR or GSA18F models, why prejudice its use?

I am with you on that. I feel like the existence of this table is in itself an orange flag that we need to change something in how to DB is set up. Having a test class is helpful but ultimately any class that is *only* for test cases is going to be limited in its usefulness. If I remember correctly, we wanted to wait until we get a 3rd client using C2 before extracting common logic out of the client model classes. What I expect will happen then is that we will do a refactor that genericizes client model classes in general so when we instantiate a generic client model it will be a reflection of how the app actually works.

Until then, if you think we should not merge this I am fine with that. I will keep tearing down and re setting up my test DB from time to time :)

pkarman:
@jessieay agreed on needing to solve how to test-clients-without-using-clients -- the test models are just one approach.

PR #925, opened 2016-02-04
myusa failure sometimes redirects to a non-existent endpoint

pkarman:
Trello: https://trello.com/c/yp6uuxcF

Note that I was not able to duplicate the exact scenario reported in the Trello card, though I have hit that 404 page before via that same URL. This PR at least gets rid of the 404, and should (hopefully) smooth out the UX till we can figure out why it is getting triggered.

### DIFF
--- a/app/views/auth/failure.html.haml
+++ b/app/views/auth/failure.html.haml
@@ -4,4 +4,4 @@
 
   %h4
     There was a problem signing you in. Please
-    = link_to 'Try signing in with MyUSA again', auth_path
+    = link_to 'try signing in with MyUSA again.', auth_path
--- a/app/views/auth/failure.html.haml
+++ b/app/views/auth/failure.html.haml
@@ -1,0 +1,7 @@
+.inset
+  %h1 
+    Sign in problem
+
+  %h4
+    There was a problem signing you in. Please
+    = link_to 'Try signing in with MyUSA again', auth_path
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -11,6 +11,7 @@ C2::Application.routes.draw do
   post "/feedback" => "feedback#create"
 
   match "/auth/:provider/callback" => "auth#oauth_callback", via: [:get]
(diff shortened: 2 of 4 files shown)


app/views/auth/failure.html.haml
```diff
@@ -0,0 +1,7 @@
+.inset
+  %h1 
+    Sign in problem
+
+  %h4
+    There was a problem signing you in. Please
+    = link_to 'Try signing in with MyUSA again', auth_path
```
yozlet: Perhaps lower-case the T in "Try" and add a period at the end of the sentence? </pedant>

yozlet:
Other than my grammar pickiness, :+1:

pkarman:
completely agree about the grammar pickiness. I was struggling to get HAML to allow the period outside the anchor tag and then left it off completely. English grammar pickiness trumps HTML semantic pickiness today.

config/routes.rb
```diff
@@ -11,6 +11,7 @@
   post "/feedback" => "feedback#create"
 
   match "/auth/:provider/callback" => "auth#oauth_callback", via: [:get]
+  get "/auth/failure" => "auth#failure"
```
jessieay: would it make sense to mock oauth in a way that we direct a user to this endpoint on login?

pkarman: I would be happy to mock this up, if I knew what series of events would lead to hitting this page. Right now my mock would like like:

```ruby
post "/auth/myusa"
# ignore response
get "/auth/failure"
```

which would not really tell us much more than that the route works, not what triggers its.

jessieay: yeah I am think it is OK for us not to know what causes it. Just thinking it would be nice to have some sort of test failure in the case that someone either deleted or updated this route incorrectly.

pkarman:
@jessieay valid point. Test added in 9cdd76e

jessieay:
@pkarman cool looks awesome. specs failures are unrelated to this PR and fixed in master via #922. 
merging

PR #926, opened 2016-02-04
Move database cleaner setup to its own file

jessieay:
* Easier to read / tweak
* We want this in all specs, not just specs the include `rails_helper`
  (even though all of our specs require it now)

### DIFF
--- a/spec/support/database_cleaner.rb
+++ b/spec/support/database_cleaner.rb
@@ -1,0 +1,31 @@
+RSpec.configure do |config|
+  config.before(:suite) do
+    DatabaseCleaner.clean_with(:truncation)
+    Test.setup_models
+    Rails.application.load_seed
+  end
+
+  config.before(:each) do
+    DatabaseCleaner.strategy = :transaction
+  end
+
+  config.before(:each, js: true) do
+    DatabaseCleaner.strategy = :truncation
+  end
+
+  config.before(:each) do
+    DatabaseCleaner.start
+  end
+
+  config.after(:each) do
+    ActionMailer::Base.deliveries.clear
+    DatabaseCleaner.clean
+    if Capybara.current_driver != :rack_test
+      Rails.application.load_seed
+    end
+  end
+
+  config.after(:suite) do
+    Test.teardown_models
+  end
+end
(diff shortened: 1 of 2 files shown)


spec/support/database_cleaner.rb
```diff
@@ -0,0 +1,31 @@
...
+    Test.setup_models
+    Rails.application.load_seed
+  end
+
+  config.before(:each) do
+    DatabaseCleaner.strategy = :transaction
+  end
+
+  config.before(:each, js: true) do
+    DatabaseCleaner.strategy = :truncation
```
jessieay: do we want to use this strategy for all capybara specs? I did some research on db cleaner and then got in a little over my head http://stackoverflow.com/questions/11419536/postgresql-truncation-speed/11423886#11423886

pkarman: I think it's one of those cases where it is better to be conservative and always do it, rather than be bitten later by trying too hard to finely parse whether you need it or not. The reason we truncate is that capybara is operating in a different process and transactions can't be assured across processes.

pkarman:
+1 looks great. Unless you want to continue discussing truncation-vs-transaction stuff, I am happy to merge.

spec/support/database_cleaner.rb
jessieay: Agreed.

does that mean I should wrap this in a `if Capybara.current_driver == :rack_test` rather than `js: true` ?

pkarman: I thought `js: true` meant that Capybara was always going to be used? I.e. what about the current syntax is not working?

jessieay: Yes it does mean capybara is always used. 

But there are specs that use capybara where js is not true.  

Honestly, I copied some of the setup from another Rails project you might be familiar with: https://github.com/18F/e-manifest/blob/master/spec/support/database_cleaner.rb

Seems to work there to have different strategies for js vs non-js so I think this should be fine.

jessieay:
fine to merge for me. Do we need to QA if this is a testing-specific PR?

spec/support/database_cleaner.rb
pkarman: Not only am I familiar with that project, but the code there came from this project! :)

https://github.com/18F/e-manifest/commit/c86e5d934d2d0bd47e0ec9fbedb67c29cbb0b1ec#diff-93830fa29d616f7c87903d08b5b1b29a

pkarman:
I don't bother QAing things that only affect the tests themselves.

PR #927, opened 2016-02-05
Wrap es output in check for `ES_DEBUG` var

jessieay:
* To prevent unnecessary output during spec runs

### DIFF
--- a/lib/expired_record_cleaner.rb
+++ b/lib/expired_record_cleaner.rb
@@ -61,7 +61,7 @@ class ExpiredRecordCleaner
   end
 
   def notify_no_action_taken_if_non_test_env(proposal)
-    if !Rails.env.test?
+    unless Rails.env.test?
       STDERR.puts "set OK_TO_ACT=true to clean up #{proposal.id}"
     end
   end
--- a/lib/expired_record_cleaner.rb
+++ b/lib/expired_record_cleaner.rb
@@ -37,7 +37,7 @@ class ExpiredRecordCleaner
   end
 
   def handle_no_requester(proposal)
-    if @verbose && ENV["RAILS_ENV"] != "test"
+    if @verbose && !Rails.env.test?
       STDERR.puts "#{proposal.id} <= no Requester defined"
     end
     if @ok_to_act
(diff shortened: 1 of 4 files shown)


jessieay:
specs failure would be fixed vai #922

lib/expired_record_cleaner.rb
```diff
@@ -37,24 +37,30 @@ def notify_proposal_requester(proposal)
   end
 
   def handle_no_requester(proposal)
-    if @verbose
+    if @verbose && ENV["RAILS_ENV"] != "test"
```
pkarman: Why not `Rails.env.test?` instead of checking the env var explicitly?

spec/lib/query/proposal/search_spec.rb
```diff
@@ -83,7 +83,7 @@
       user = create(:user, client_slug: "test")
 
       proposal1 = create(:proposal, id: 199, requester: user)
-      test_client_request1 = create(:test_client_request, proposal: proposal1)
+      create(:test_client_request, proposal: proposal1)
```
pkarman: why throw away the return value?

spec/support/es_spec_helper.rb
```diff
@@ -28,25 +28,31 @@ def es_server_running?
...
         STDOUT.puts errors.pretty_inspect
       end
     end
-    puts "Refreshing index for class #{klass}"
+
+    output_if_debug_true { "Refreshing index for class #{klass}" }
     klass.__elasticsearch__.refresh_index!
   end
 
+  def output_if_debug_true
```
pkarman: nice!

spec/lib/query/proposal/search_spec.rb
jessieay: unused var. (I have a plugin that highlights these for me)

you think we should re-add and prepend with underscore to show that they are known unused vars?

lib/expired_record_cleaner.rb
jessieay: I am not sure why. I guess I was thinking we might want to run the specs with the env var changed, but `Rails.env` wraps the env var so it is basically the same thing. Will change.

jessieay: updated at d804538

spec/lib/query/proposal/search_spec.rb
pkarman: I don't have that plugin so it hasn't bothered me.

I guess I wrote enough C that the idea of ignoring the return value (because memory leak) just got ingrained. Since they aren't used, I don't care if you leave it as you've got it.

jessieay: I think unused vars can be helpful in some cases but I don't think they are particularly enlightening in the particular context. let's keep as-is! :hamburger:

PR #928, opened 2016-02-05
Extract Query objects from Reporter class

jessieay:
* Part of https://trello.com/c/kBQy3z6F


jessieay:
argh this code climate error is due to duplication and I am not sure if / how best to fix.

pkarman:
I like the code refactor here to split out the queries into their own classes.

As far as going further to rewrite in raw SQL, I've commented on that Trello card that I don't consider it worth the effort.

jessieay:
@pkarman agreed. replied to your comment!

pkarman:
@jessieay test failures?

I would ignore that code duplication error. It's not a duplicate. Methinks codeclimate algorithm needs some tweeking.

jessieay:
@pkarman woops, should pass now.

Agreed on duplication!

PR #929, opened 2016-02-05
Add event, controller/action, client-side, and email tracking

rememberlenny:
## New Gems
Adding two gems: ```ahoy_matey``` and ```ahoy_email```. 

## Database changes
Adding table for ```visits```, ```ahoy_events```, and ```ahoy_messages```. Adding column on ```proposals``` and ```comments``` for visit tracking.

```visits``` and ```ahoy_events``` are used for tracking frontend and backend activity in relation to the logged on user. The events can be used to make details reports for error handling and observe user behavior.

```ahoy_messages``` is used for email tracking. All emails sent out will have their links proxied by the ```ahoy_email``` gem. The proxied links will pass through the webapp before routing to destination, to track clicks. Emails will also have a 1x1 pixel added at ```</body>``` to track opens.

```visit_id``` column is added to ```proposals``` and ```comments``` to associate the visit of a user, when new instances are created. For example, the visit behavior leading up to a proposal being created can be tracked and observed.

# How
*Magic*

---

- [Ahoy Email repository](https://github.com/ankane/ahoy_email)

## Admin

Added three fields in ActiveAdmin for Tracking.

### Note

```blazer``` gem was also added in this set. Blazer will provide a UI to parse the tracking data. It is not currently needed on production, so I mounted it behind the development environment.

### CODE REVIEW

File: db/migrate/20160204205859_add_visit_id_to_comment_and_proposal.rb [hunk 1]
```diff
@@ -0,0 +1,11 @@
+class AddVisitIdToCommentAndProposal < ActiveRecord::Migration
+  def up
+    add_column :proposals, :visit_id, :integer
+    add_column :comments, :visit_id, :integer
```
pkarman: can we also `add_foreign_key` here for some referential integrity?

pkarman:
@rememberlenny this looks great. The only thing missing is tests. Seems like putting something in `spec/features` and/or `spec/controllers` would be appropriate.

A couple other notes:

* we are already tracking email stats via Mandrill. We should confirm that this doesn't interfere.

* Not sure how/if all the demographic stuff would be captured for each visit, and whether any of that qualifies as PII and so would trigger any ATO/security alarms. @phirefly should weigh in.

jessieay:
Neat-o. Perhaps add a note about this in the docs as well? https://github.com/18F/C2/blob/master/doc/production.md

rememberlenny:
@pkarman The email tracking wont interfere. This will give us the ability to trigger rails functions on email related activity. For example, when email is opened or clicked, the activity feed can update accordingly. The emails will still be tracked in Mandrill.

app/views/layouts/_analytics.html.erb
```diff
@@ -14,4 +14,4 @@
     ga('set', 'forceSSL', true);
     ga('send', 'pageview');
   </script>
-<% end %>
+<% end %>
```
jessieay: hey this means there's no newline at EOF!  are you perhaps using sublime? http://docs.sublimetext.info/en/latest/reference/settings.html

db/migrate/20160205184644_add_visit_id_to_comment_and_proposal.rb
```diff
@@ -0,0 +1,15 @@
...
+    add_foreign_key :comments, :visits
+  end
+
+  def down
+    remove_foreign_key :proposals, :visits
+    remove_foreign_key :comments, :visits
+    drop_column :proposals, :visit_id
+    drop_column :comments, :visit_id
+  end
+end
```
jessieay: new newline at EOF again

config/blazer.yml
```diff
@@ -0,0 +1,44 @@
+# see https://github.com/ankane/blazer for more info
+
+data_sources:
+  main:
+    url: <%= ENV["BLAZER_DATABASE_URL"] %>
+
+    # statement timeout, in seconds
```
jessieay: hey there is a lot of commented code in this gem setup (in this file and others in this PR) - can we remove? easier to see what is being set up without all of the surrounding commented code

rememberlenny: Ill do it!

db/migrate/20160205184644_add_visit_id_to_comment_and_proposal.rb
rememberlenny: Fixed

rememberlenny:
@pkarman For the tracking, there isnt any explicit demographic data associated to the user. There is a lat/lon data point, but otherwise it only stores browser information.

rememberlenny:
Closing for now, but keep branch for reference.

PR #932, opened 2016-02-05
Elasticsearch rails ha update

pkarman:


lib/tasks/search.rake
```diff
@@ -0,0 +1,24 @@
+require 'elasticsearch/rails/ha/tasks'
+
+namespace :search do
+  desc 're-index all records, in parallel'
+  task index: :environment do
```
jessieay: should we be using this task elsewhere in the codebase?

jessieay:
code climate is mad at you :fire

lib/tasks/search.rake
pkarman: It's possible that the /admin button to re-index should use the stage+promote tasks in order to provide zero-downtime index rebuilds. But that can be an optimization later.

pkarman:
codeclimate, cruel mistress, you are :green_heart: again!

cc/ @jessieay

spec/lib/roles_conversion_spec.rb
```diff
@@ -1,8 +0,0 @@
-describe RolesConversion do
```
jessieay: did you mean to delete this?

pkarman: yes. see https://18f.slack.com/archives/cap/p1454714455001197

rememberlenny:
Merged

PR #935, opened 2016-02-08
move session expiration from 2.weeks to 8.hours

pkarman:
Trello: https://trello.com/c/0cNHsYk1

8 hours is roughly once a day. If we hear user feedback that this is too short, we can adjust (though IMO 1x daily is a decent compromise between usability and security).


jessieay:
looks like this needs a rebase. Failure related to `RolesConveresion` thing again

jessieay:
this makes sense to me. I am not 100% sure that the linked error from Trello is due to ENT password reset. But if it is, then this is a :+1: solution

pkarman:
@jessieay I am nearly 99% sure the error has nothing to do with ENT password reset. But this change is good regardless.

PR #936, opened 2016-02-08

rememberlenny:

## New Gems

## Database changes

```visits``` and ```ahoy_events``` are used for tracking front-end and backend activity in relation to the logged on user. The events can be used to make details reports for error handling and observe user behavior.

```visit_id``` column is added to ```proposals```, ```reports``` and ```comments``` to associate the visit of a user, when new instances are created. For example, the visit behavior leading up to a proposal being created can be tracked and observed.

# How
*Magic*

---

## Admin

### CODE REVIEW

**Comment by rememberlenny** [comment]
Testing ahoy without the "visitable" model change.

**Comment by rememberlenny** [comment]
Alright!!!!!! GREEN

File: Gemfile [hunk 1]
```diff
@@ -5,6 +5,8 @@ gem "active_model_serializers"
 gem "activeadmin", git: "https://github.com/activeadmin/activeadmin.git"
 gem "acts_as_list"
 gem "acts-as-taggable-on", "~> 3.4"
+# gem "ahoy_email"
```
jessieay: should we remove this?

spec/controllers/ahoy/events_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
+describe Ahoy::EventsController, :type => :controller do
+  routes { Ahoy::Engine.routes }
+
+  it "creates event with other attr" do
+    time = (Time.now - 30).utc.change(:usec => 0)
+    post :create, :id => "11111111-2222-3333-4444-555555555555",
+      :name => "$submit",
```
jessieay: if this were in non-spec code, Rubocop would be complaining about the usage of old hash syntax. not a biggie but an FYI / if you wanna change

jessieay:
The PR description here is *great* -- can we add something like that to the project docs so people who are looking at this project know we are using Ahoy?

jessieay:
cool! this looks good to merge to me. any comments from @pkarman first? I was not part of the convo about why we are adding this so want to make sure I am not missing any acceptance criteria

doc/analytics.md
```diff
@@ -0,0 +1,75 @@
+# Analytics
+
+We use event and visit tracking to better recognize our user's needs. On the backend, we record our controllers and actions. On the client side, we track page's visited, elements clicked, and provide functions for JavaScript triggered events. Our primary tracking mechanism is based on the ```ahoy_matey``` gem. 
+
+For more information about ```Ahoy``` go to [the repo page](https://github.com/ankane/ahoy).
```
pkarman: just a markdown observation: inline verbatim markup only requires single backticks `like this` -- the triple ticks are only for block-style verbatim sections. Not a showstopper.

pkarman:
+1 

@jessieay the impetus here was to start some gathering some in-app UX data beyond what GA can give us.

pkarman:
I will QA

pkarman:
QA feedback:

* deployed ok to c2-dev
* events are getting created for `/admin` actions which seem like they will only take up db and screen space (i.e. we won't ever take action on that kind of internal use, right?)
* there are Create/Edit/Delete actions visible as links/buttons for ahoy events. They should be read-only yes? Esp since clicking any of those actions generates an error. The actions can be filtered away via activeadmin config.
* are we concerned about all the db space this kind of logging is going to consume? Should we have a data retention policy (e.g. drop everything, once a month, that is older than 180 days).

screen shot 2016-02-10 at 10 39 34 am
screen shot 2016-02-10 at 10 39 13 am
screen shot 2016-02-10 at 10 38 28 am

db/migrate/20160208173646_add_the_visit_id_to_models.rb
```diff
@@ -0,0 +1,19 @@
...
+    add_foreign_key :proposals, :visits
+    add_foreign_key :comments, :visits
+    add_foreign_key :reports, :visits
+  end
+
+  def down
+    remove_foreign_key :proposals, :visits
+    remove_foreign_key :comments, :visits
+    remove_foreign_key :reports, :visits
+    drop_column :proposals, :visit_id
```
pkarman: no such method as `drop_column` -- you want `remove_column` instead.

jessieay: good catch! we should make sure we are running and rolling back each migration while they are in developing. ensures we have migrations with proper `up` and `down` content

pkarman: Agreed. I only noticed this after trying to `rake db:rollback` on c2-dev after QA was complete.

pkarman:
@rememberlenny just wanted to be sure you saw this has been handed back to you with some QA feedback.

db/migrate/20160208173646_add_the_visit_id_to_models.rb
rememberlenny: @pkarman Got it!

rememberlenny:
@pkarman Updated the migration

pkarman:
@rememberlenny you might need to manually restart these CI failures.

jessieay:
@rememberlenny you want to update this to make code climate happy?

jessieay:
@rememberlenny yay green! were all of pkarman's QA feedback comments (https://github.com/18F/C2/pull/936#issuecomment-182471500) addressed?

rememberlenny:
@jessieay : ( No! Ill address them tomorrow

db/migrate/20160208173646_add_the_visit_id_to_models.rb
pkarman: @rememberlenny this still needs fixing `s/drop_/remove_/g`

PR #937, opened 2016-02-09
[Refactor] Set approving official as NCR relation instead of attribute

jessieay:
* This enables us to more easily do validations and assign steps for NCR
  Work Orders
* This will make it easier to set up steps as belonging to a user or a
  system approver
* Step 0 of https://trello.com/c/hHfzNgSu

### DIFF
--- a/app/controllers/ncr/work_orders_controller.rb
+++ b/app/controllers/ncr/work_orders_controller.rb
@@ -4,7 +4,7 @@ module Ncr
     MAX_UPLOADS_ON_NEW = 10
 
     def new
-      @client_data_instance.approving_official = suggested_approver_email
+      work_order.approving_official = suggested_approver_email
       super
     end
 
--- a/app/controllers/ncr/work_orders_controller.rb
+++ b/app/controllers/ncr/work_orders_controller.rb
@@ -4,7 +4,7 @@ module Ncr
     MAX_UPLOADS_ON_NEW = 10
 
     def new
-      work_order.approving_official_email = suggested_approver_email
+      @client_data_instance.approving_official = suggested_approver_email
       super
     end
 
@@ -14,8 +14,6 @@ module Ncr
     end
 
     def edit
-      @client_data_instance.approving_official_email = @client_data_instance.approvers.first.try(:email_address)
-
       if proposal.approved?
         flash.now[:warning] = "You are about to modify a fully approved request. Changes will be logged and sent to approvers, and this request may require re-approval, depending on the change."
       end
@@ -37,6 +35,11 @@ module Ncr
       @client_data_instance
     end
 
+    def suggested_approver_email
+      last_proposal = current_user.last_requested_proposal
+      last_proposal.try(:client_data).try(:approving_official)
+    end
+
     def record_changes
       ProposalUpdateRecorder.new(work_order).run
     end
@@ -49,17 +52,8 @@ module Ncr
       updater.run
     end
 
-    def attribute_changes?
-      super || work_order.approver_changed?
-    end
-
     def model_class
       Ncr::WorkOrder
-    end
-
-    def suggested_approver_email
-      last_proposal = current_user.last_requested_proposal
-      last_proposal.try(:approvers).try(:first).try(:email_address) || ''
     end
 
     def permitted_params
@@ -76,7 +70,6 @@ module Ncr
       Ncr::WorkOrderFields.new.relevant(params[:ncr_work_order][:expense_type])
     end
 
-    # @pre: work_order.approving_official_email is set
     def add_steps
       super
       if errors.empty?
--- a/spec/controllers/ncr/work_orders_controller_spec.rb
+++ b/spec/controllers/ncr/work_orders_controller_spec.rb
@@ -16,7 +16,7 @@ describe Ncr::WorkOrdersController do
           code: "Work Order",
           project_title: "Title",
           description: "Desc",
-          approving_official_email: "bob@example.com"
+          approving_official_id: approving_official.id
         }
       }
 
@@ -26,8 +26,8 @@ describe Ncr::WorkOrdersController do
       ncr = Ncr::WorkOrder.order(:id).last
 
       expect(ncr.code).to eq 'Work Order'
-      expect(ncr.approvers.first.email_address).to eq 'bob@example.com'
-      expect(email_recipients).to eq(['bob@example.com', ncr.requester.email_address].sort)
+      expect(ncr.approvers.first).to eq approving_official
+      expect(email_recipients).to eq([approving_official.email_address, ncr.requester.email_address].sort)
     end
   end
 
@@ -61,26 +61,10 @@ describe Ncr::WorkOrdersController do
   end
 
   describe '#update' do
-    let (:work_order) { create(:ncr_work_order, :with_approvers) }
+    let (:work_order) { create(:ncr_work_order) }
     let (:requester) { work_order.proposal.requester }
     before do
       login_as(requester)
-    end
-
-    it "uses existing first approver when there is a blank approver email" do
-      post :update, {
-        id: work_order.id,
-        ncr_work_order: {
-          approving_official_email: '',
-          expense_type: 'BA61',
-          project_title: 'new name'
-        }
-      }
-      expect(flash[:success]).to be_present
-      expect(flash[:error]).not_to be_present
-      work_order.reload
-      expect(work_order.project_title).to eq('new name')
-      expect(work_order.approving_official_email).to eq(work_order.approvers.first.email_address)
     end
 
     it 'does not modify the work order when there is a bad edit' do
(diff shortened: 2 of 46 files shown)


app/controllers/ncr/work_orders_controller.rb
```diff
@@ -4,7 +4,6 @@ class WorkOrdersController < ClientDataController
     MAX_UPLOADS_ON_NEW = 10
 
     def new
-      work_order.approving_official_email = suggested_approver_email
```
jessieay: adding this line back in unfortunately does not fix the regression noted in the PR description:

> we are no longer auto-suggesting the approver from the last proposal

I am not sure why, but wanted to debug separately since this PR was getting larger. Note that the `edit` action still auto populates the existing approving official email in the form.

spec/controllers/ncr/work_orders_controller_spec.rb
```diff
@@ -61,35 +61,19 @@
...
 
   describe '#update' do
-    let (:work_order) { create(:ncr_work_order, :with_approvers) }
+    let (:work_order) { create(:ncr_work_order) }
     let (:requester) { work_order.proposal.requester }
     before do
       login_as(requester)
     end
 
-    it "uses existing first approver when there is a blank approver email" do
```
jessieay: this spec is no longer relevant because we are now called `validates :approving_official, presence: true`, so there will be a validation error of a work order is updated with no approving official FK

app/controllers/ncr/work_orders_controller.rb
```diff
@@ -4,7 +4,7 @@ class WorkOrdersController < ClientDataController
...
+      @client_data_instance.approving_official = suggested_approver_email
```
pkarman: if you're going to use `@client_data_instance` here instead of the `work_order` method, should we change it everywhere? Inconsistent use may confuse as to whether it was intentional or not.

pkarman:
Code looks fine.

I think I see the overall approach here: use real User objects rather than email addresses, and make the WorkOrder approver an actual foreign key, rather than a fictional attribute that existed only to make the form play nicely with the actual underlying data model.

Are you concerned at all that the relevant `Step.user_id` will get out of sync with the `WorkOrder.approving_official_id` ? I.e. AFAIK the existing implementation was a grand hack to present the experience of a simple association while maintaining a highly normalized model on the backend. Now the grand hack is removed in exchange for the actual simple association, at the expense of some normalization.

E.g. if a WorkOrder is created with an `approving_official_id` and a matching Step is created pointing at the same User, and then somehow the `approving_official_id` is edited but the corresponding Step is not, could we end up with mismatched data pointing at 2 different Users?

app/controllers/ncr/work_orders_controller.rb
jessieay: ah yes I changed this when debugging something and forgot to change back. I am happy to change back or change everywhere - which do you think is clearer?

pkarman: I think `work_order` is clearest.

jessieay:
@pkarman yes, that is a risk of this setup. 

The reason for this change was that I started down the path we discussed before of assigning steps to an "assignee" that can be either a User or a SystemApprover (or some other class like that). As I started that refactor, I kept running into this issue where we passed email addresses all over the place and the found or created a user based on that email address. 

In the context of adding a polymorphic relationship, this became a lot more complicated because there was no way of knowing if the email address was for a user or another type of object. This PR moves from passing around email addresses to passing around objects, which mitigates that complication.

Re: your point about data denormalization. 

That *is* a new risk. 

But it is already a risk for any other changed attribute on a work order

eg: if a work order expense type is changed and we do not call `setup_approvals_and_observers`, the approvers for that work order will not reflect the approvers that it should have. Similarly, if an approving official is changed, we need to call `setup_approvals_and_observers` to make sure the steps are assigned correctly.

There is a new slightly higher risk of data de-normalization, but given that we need to update approvals and observers whenever a work order is updated anyway, I don't think that the risk is unusually high.

pkarman:
@jessieay your point about `setup_approvals_and_observers` already being an accepted risk is well made. I'm +1 on this change, if for no other reason than I've wasted hours working on/around the Grand Hack myself.

pkarman:
error during QA deploy:
```
2016-02-11T12:54:59.72-0600 [App/0]      ERR StandardError: An error has occurred, this and all later migrations canceled:
2016-02-11T12:54:59.72-0600 [App/0]      ERR undefined method `individual_steps' for nil:NilClass/home/vcap/app/db/migrate/20160208182623_add_approving_official_email_to_work_orders.rb:10:in `block in up'
```

which is apparently due to bad data (work_order with null proposal_id).

jessieay:
hmmm @pkarman do you think we should remove bad data or rewrite migration to deal with this case? I wouldn't expect us to have and client data sans proposal, wonder how that got in there...

pkarman:
@jessieay I think both: separate PR to clean up bad data though.

This is one of the side effects of assuming all db interactions happen via code and a schema with polymorphic references: you can't put a FK constraint on `proposals` to verify that `client_data_id` and `client_data_type` point at a valid record, or vice versa. I'm sure there was a code bug and/or manual db intervention at some point that allowed the bad data to exist.

jessieay:
@pkarman ok will work on that now. 

Agreed with validation. I think we can validate presence of the parent (proposal), which I can add, but that doesn't mean we won't end up with a proposals sans client data. :cry:

jessieay:
@pkarman updated migration at 8c3243f so it should not fail pre-Proposal cleaner

pkarman:
I will re-QA

pkarman:
I created https://c2-dev.18f.gov/proposals/270 and then tried re-setting the approver via dropdown several times. Screenshots attached.

* The change history doesn't seem to track the `to` value
* Occasionally the switch didn't seem to match (top list of approvers didn't match the selected approver in the dropdown). I couldn't seem to repeat that though.

screen shot 2016-02-11 at 3 17 49 pm
screen shot 2016-02-11 at 3 17 14 pm
screen shot 2016-02-11 at 3 16 54 pm
screen shot 2016-02-11 at 3 16 34 pm
screen shot 2016-02-11 at 3 15 45 pm

jessieay:
@pkarman thanks for the thorough QA! will investigate findings and get back to you

jessieay:
@pkarman update comment issue fixed via a050576e9ed497ca038edbedf1c2501202be63dc

jessieay:
@pkarman was not able to reproduce the approving official issue (not seeing correct value in form on edit) but I did make some changes in 415e35d that might fix. 

Basically, I originally got selectize working in the form via JSON values auto-populating but then some tests without javascript turned on started failing. So i added some code to make the form also work *without* selectize. wondering if having select options set via non-js and options set via selectize was causing inconsistencies you were seeing

app/controllers/ncr/work_orders_controller.rb
jessieay: fixed 2bd0d25

pkarman:
Alas there is still some wonkiness with syncing changes to the "approving official" via the dropdown and the steps assigned.

I created https://c2-dev.18f.gov/proposals/271 as before and re-assigned approving official several times with a variety of responses. Sometimes the step was missing (as when assigned to jessica.young@gsa.gov). Sometimes the step was assigned to someone different (step says jessica.young+18fpurchaser@gsa.gov, approving official says jessie.a.young@gmail.com).

Screenshots below.

screen shot 2016-02-12 at 9 46 51 am
screen shot 2016-02-12 at 9 47 10 am
screen shot 2016-02-12 at 9 47 30 am
screen shot 2016-02-12 at 9 47 42 am
screen shot 2016-02-12 at 9 47 55 am
screen shot 2016-02-12 at 9 48 04 am
screen shot 2016-02-12 at 9 48 22 am
screen shot 2016-02-12 at 9 48 30 am
screen shot 2016-02-12 at 9 48 38 am
screen shot 2016-02-12 at 9 48 47 am
screen shot 2016-02-12 at 9 49 04 am
screen shot 2016-02-12 at 9 49 13 am
screen shot 2016-02-12 at 9 49 19 am
screen shot 2016-02-12 at 9 49 29 am
screen shot 2016-02-12 at 9 49 44 am
screen shot 2016-02-12 at 9 49 51 am
screen shot 2016-02-12 at 9 50 00 am
screen shot 2016-02-12 at 9 50 12 am
screen shot 2016-02-12 at 9 50 23 am
screen shot 2016-02-12 at 9 50 36 am
screen shot 2016-02-12 at 9 50 46 am

jessieay:
ok @pkarman code updated and regression spec added in https://github.com/18F/C2/commit/60e1c6cfed14d624e943cfd9b1c736eceff65679 for issue where a step for approving official was not showing.

jessieay:
code climate failure is a false negative.

I've tried to repro the "wrong approving official showing in list" issue but have had no luck.

pkarman:
@jessieay should I re-QA as-is?

jessieay:
@pkarman I guess? I am not sure what to do about the non-reproducible issue. Wondering if this fix is related but not sure because I never saw it. :(

pkarman:
re-tested and I cannot duplicate the error from before so your fix must have caught that too.

PR #938, opened 2016-02-10
Upgrade to AWS 2

jessieay:
* Newest version of Paperclip are compatible with AWS 2
* Per upgrade guide, need to specify region
* Source: https://github.com/thoughtbot/paperclip/blob/76f825a8b616fabf8bfba433ac3107e1b1a079ae/UPGRADING
* We are already using `us-east-1` region
* http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region
* Use paperclip shoulda matchers
* http://www.rubydoc.info/github/thoughtbot/paperclip/Paperclip/Shoulda/Matchers
* https://trello.com/c/DyY3O3QM


config/initializers/paperclip.rb
```diff
@@ -3,8 +3,9 @@
     bucket: ENV["S3_BUCKET_NAME"],
     s3_credentials: {
       access_key_id: ENV["S3_ACCESS_KEY_ID"],
-      secret_access_key: ENV["S3_SECRET_ACCESS_KEY"]
+      secret_access_key: ENV["S3_SECRET_ACCESS_KEY"],
     },
+    s3_region: "us-east-1",
```
pkarman: should this be an env var like the other `S3_*` values?

pkarman:
@jessieay looks great. just one question about env var vs hardcoded string.

config/initializers/paperclip.rb
jessieay: good idea. done in 80da33c and jut set env var on dev, staging, and prod.

pkarman:
:+1: -- needs QA on c2-dev

pkarman:
:+1: QA, attachments still work:

screen shot 2016-02-11 at 1 08 05 pm

PR #940, opened 2016-02-11
Move app-specific code out of lib dir

jessieay:
* Lib should be for code not specific to app (eg: does not use classes
  in `app`
* http://blog.lunarlogic.io/2013/declutter-lib-directory/
* Put all query classes in `app/queries` instead of splitting into two
 dirs
* Put all service classes in `app/services`
* Move classes for rake tasks into `db/chores` dir


pkarman:
I'm fine with the file re-org. When in Rails, do as the Railers do.

The test failures maybe are false positives? Maybe just a manual restart?

Guessing the code climate stuff is just because CC treating it as a new file. Fine to ignore those IMO.

jessieay:
@pkarman re-running specs now. Did not see failure locally so we shall see!

Fixed that last couple of code climate problems that were easily solvable. I am not sure what to do about Code Climate not liking methods that contain long arrays. Did some reading up on it and there is no easy solution. Could put an ignore flag in code but I would rather just have a red code climate if that's cool with you.

Thanks for the review!

jessieay:
ok @pkarman this is ready for re-review. Code climate is failing on 3 methods that are too long because they contain arrays of attributes.

spec/queries/fielded_search_query.rb
```diff
@@ -1,29 +1,29 @@
-describe Query::Proposal::FieldedSearch do
+describe Proposal::FieldedSearchQuery do
```
pkarman: should this be `ProposalFieldedSearchQuery` (no hierarchy) ?

pkarman:
@jessieay just a question about class name. Otherwise, I'm happy to merge with no QA since it is all class name changes.

spec/queries/fielded_search_query.rb
jessieay: it should! this got missed because there is no `_spec` at the end of the file -- my fault! will rename. great catch!

jessieay: moved in 941bdd2

PR #941, opened 2016-02-11
Add class to remove proposals without client data

jessieay:
* To prevent migration issue https://github.com/18F/C2/pull/937#issuecomment-183009980

### DIFF
--- a/db/chores/proposal_cleaner.rb
+++ b/db/chores/proposal_cleaner.rb
@@ -1,0 +1,5 @@
+class ProposalCleaner
+  def run
+    Proposal.where(client_data: nil).destroy_all
+  end
+end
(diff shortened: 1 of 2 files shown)


jessieay:
I originally was going to have this class remove client data without a proposal, but we have a validation for that already so there should not be client data without a proposal in the db

pkarman:
@jessieay the validation is in code and therefore doesn't touch existing data, unless attempting to re-save that data. We have data in c2-dev that have orphaned client data records. These ncr_work_order records, e.g.:

```
[14, 15, 1, 2, 19, 33, 13, 39, 21, 24, 22, 25, 26, 27, 28, 29, 31, 32, 34, 35, 41, 48, 50, 81, 62, 70, 71, 73, 74, 76, 77, 79, 80, 82, 88, 84, 85, 83, 86, 87, 78]
```

jessieay:
@pkarman good point. added code to remove orphan client data at 30d05b8

pkarman:
:+1:

PR #942, opened 2016-02-12
Update archive link spec

jessieay:
* Was not appended with `spec` so was not running
* Stub constant to to so can create fewer records for spec


app/views/proposals/_archive_link.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="col-sm-2">
-  <%- unless @approved_data.rows.size < @CLOSED_PROPOSAL_LIMIT %>
-    <%= link_to "View the archive", url_for(:controller => 'proposals', :action => 'archive') %>
-  <%- end %>
+  <% if @approved_data.rows.size > @closed_proposal_limit %>
```
pkarman: should this be `if @approved_data.rows.size >= @closed_proposal_limit` instead?

jessieay: we are limiting the results in the controller to  `@closed_proposal_limit` so I think greater than is all we need

previous logic would have showed archive link if there were 10 records, but there would be nothing to more to see in that case

pkarman: failing test?

pkarman: @jessieay the previous commit (72277b495152fe948b253cba94ca14ace815cf40) intentionally biased in favor of showing the link, even if there was nothing behind it. I can't seem to get into pivotal tracker to find the original card. I'm just noticing that this change basically just reverts the previous change.

jessieay: oh interesting, I assume it was not intentional. Do you remember why we did that?

jessieay: failing test is due to constant stubbing not working as I thought it was. working on it!

pkarman: My memory is limited to the commit message, alas.

jessieay: fixed via 8d6f886

pkarman:
@jessieay failures might need manual restart?

jessieay:
@pkarman done!

pkarman:
@jessieay merge conflict alas

jessieay:
@pkarman updated~

PR #943, opened 2016-02-12
add es_execute_with_retries method

pkarman:
Trello: https://trello.com/c/8nuPwMap

ES server may fail to sync immediately in CI environment, causing spurious errors. Can't always test for this precisely during feature tests, but in all others, wrap in a try/catch and re-try if ES-specific error is caught.


jessieay:
this looks great! out of curiosity, why retry 3 times rather than 2 or 4?

pkarman:
3 was what the example I based it on used. makes sense to me. 1 failure is the status quo, 2 is reasonable edge case, and 3 consecutive failures means Things Are Really Broken.

PR #944, opened 2016-02-12
Streamline env var stubbing in specs

jessieay:
* Refactor


pkarman:
@jessieay might need manual restart for failures?

jessieay:
@pkarman done!

PR #945, opened 2016-02-16
use special tier1 approver if NCR org code matches

pkarman:
Trello: https://trello.com/c/FJLK6vYZ


db/chores/roles_conversion.rb
```diff
@@ -1,5 +1,6 @@
 class RolesConversion
   def ncr_budget_approvers
+    ba61_tier1_budgetteam_approver
```
jessieay: should this be `ba61_tier1_budget_team_approver` for better readability?

spec/services/ncr/approval_manager_spec.rb
```diff
@@ -175,6 +175,26 @@
...
+          org_code = "P11#{org_letter}XXXX"
+          ncr_org = create(:ncr_organization, code: org_code)
+          ba60_work_order = create(:ba60_ncr_work_order, ncr_organization: ncr_org)
+          ba61_work_order = create(:ba61_ncr_work_order, ncr_organization: ncr_org)
+          ba80_work_order = create(:ba80_ncr_work_order, ncr_organization: ncr_org)
+          ba60_work_order.setup_approvals_and_observers
+          ba61_work_order.setup_approvals_and_observers
+          ba80_work_order.setup_approvals_and_observers
+
+          expect(ba60_work_order.budget_approvals.first.user_id).to eq(Ncr::Mailboxes.ba61_tier1_budgetteam.id)
```
jessieay: could simplify as:

```
expect(ba60_work_order.budget_approvals.first.user).to eq(Ncr::Mailboxes.ba61_tier1_budgetteam)
```

app/services/ncr/approval_manager.rb
```diff
@@ -62,7 +62,11 @@ def notify_removed_approvers(original_approvers)
     def ba_6x_approvers
       results = []
 
-      unless work_order.for_whsc_organization?
+      if work_order.for_whsc_organization?
+        # no tier 1
+      elsif work_order.ba_6x_tier1_team?
+        results << Ncr::Mailboxes.ba61_tier1_budgetteam
+      else
```
jessieay: maybe this could be re-written as:

```ruby
if work_order.ba_6x_tier1_team?
   results << Ncr::Mailboxes.ba61_tier1_budgetteam
elsif !work_order.for_whsc_organization?
  results << Ncr::Mailboxes.ba61_tier1_budget
end
```

I am not sure if the logic holds if rewritten this way. But having an `if` condition with a result of `nil` (as it is written currently) does seem unusual to me.

app/models/ncr/organization.rb
```diff
@@ -23,5 +23,9 @@ def ool?
     def whsc?
       code == WHSC_CODE
     end
+
+    def ba_6x_tier1_team?
```
jessieay: maybe we should have a unit test for this and then just test one case in `spec/services/ncr/approval_manager_spec.rb` ?

spec/services/ncr/approval_manager_spec.rb
pkarman: yes I expected that to work too, and I started with that. But there is some wonkiness is comparing the AR objects there and I got an error (it was not comparing them using the User pk).

db/chores/roles_conversion.rb
pkarman: good point.

app/services/ncr/approval_manager.rb
pkarman: I wanted to make it clear there are 3 distinct cases:

1. the WO has a whsc_organization (so no tier1 approver at all)
1. the WO has a ba_6x special organization (special tier1 approver)
1. neither of the first 2 are true (default tier1 approver)

I can rewrite it as an if/elsif but I was worried that it the business logic would be less clear later.

app/models/ncr/organization.rb
pkarman: done in 44b9339

I'd rather leave the explicit cases in the `spec/services/ncr/approval_manager_spec.rb` so that it is clear that it does not affect `BA80` expense types and applies equally to `BA60` and `BA61` (since the approver only has `BA61` in the name, per the trello card).

pkarman:
@jessieay thanks for the review. I think I've responded to all your comments with either a change or more questions.

app/models/ncr/organization.rb
jessieay: sound good!

jessieay:
mergin'! :ship:

PR #948, opened 2016-02-17
align GSA18F email with NCR email template

pkarman:
Trello: https://trello.com/c/46RtuI2C

This has the added benefit of making it easier to refactor all emails now that the client-specific templates are virtually identical.

### DIFF
--- a/app/views/gsa18f/_proposal_mail.html.haml
+++ b/app/views/gsa18f/_proposal_mail.html.haml
@@ -2,7 +2,8 @@
   %table.w-container.main-container{ width: "800" }
     %tr
       %td
-        %h1.c2_header= "#{title} : #{proposal.name}"
+        %h1.c2_header
+          = link_to proposal.name, proposal_url(proposal),  { target: "C2" }
         .c2_description
           %p
             Requested by:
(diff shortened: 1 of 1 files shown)


app/views/gsa18f/_proposal_mail.html.haml
```diff
@@ -2,7 +2,8 @@
   %table.w-container.main-container{ width: "800" }
     %tr
       %td
-        %h1.c2_header= "#{title} : #{proposal.name}"
+        %h1.c2_header
+          = link_to proposal.name, proposal_url(proposal),  { target: "C2" }
```
jessieay: I just removed the `target` from the URL in https://github.com/18F/C2/pull/949 because it didn't seem to be doing anything - do you know why we have it in this link?

jessieay:
looks great! will QA now

app/views/gsa18f/_proposal_mail.html.haml
pkarman: no idea. I suspect it was there in order to avoid creating new tabs/windows for each click on an email.

jessieay: isn't that usually done with `target: '_blank'` ?

jessieay:
looks good on dev - mergin'! we can address `target` issue L8R

app/views/gsa18f/_proposal_mail.html.haml
pkarman: `target: '_blank'` would create a new window/tab for each click. By naming it, you can re-use the same window/tab over and over again (maybe most helpful on mobile?).

jessieay: ah interesting! did not know that

PR #949, opened 2016-02-17
Begin moving email copy to i18n

jessieay:
* Use default Rails email preview instead of `mail_view`

https://trello.com/c/xvL5TN7d
https://trello.com/c/LEWukGeB


jessieay:
@pkarman moved copy from a single email to I18n as a proof of concept. If you think this looks good, we can merge and continue to move rest of emails over.

What I found while doing this was that we didn't have previews for many emails. We should create previews for emails as we go forward so we can easily view them as we update styles.

pkarman:
:heart: it

when tests finish I will merge.

PR #953, opened 2016-02-17
Move even even more emails to I18n

jessieay:
to be merged after https://github.com/18F/C2/pull/952


jessieay:
merge https://github.com/18F/C2/pull/952 first since this branches off of that

PR #954, opened 2016-02-18
Make partial rendering in emails less fancy

jessieay:
* `client_partial` method used complicated logic to pick a
  partial for a user and did not error if there was no partial (eek!)
* Easier to find partials if their paths are set explicitly


pkarman:
@jessieay I really like this, +1

merge conflict, alas

jessieay:
@pkarman updated!

PR #955, opened 2016-02-18
[WIP] Rewrite emails using ink

jessieay:
I am currently rewriting the emails using Zurb's INK framework:

http://foundation.zurb.com/emails/docs.html

The process is something like this: 

http://www.sitepoint.com/responsive-emails-rails-ink/

Current design status:

screen shot 2016-02-18 at 2 42 02 pm

So far, my evaluation of INK is:

1. Decent docs
2. Give you a nice starting point template
3. Conventions are a little but confusing. (eg: grid system)
4. Grid not compatible with gmail mobile - sad trombone
5. Might be difficult to re-use elements since the styles must be inlined before sending the email

Before moving further down this path, I am going to look into using straight HTML / CSS with Roadie like we currently do.


jessieay:
screenshot from a gmail test on dev:

screen shot 2016-02-18 at 3 30 46 pm

jessieay:
Hey @gemfarmer - wow looks so much better! 

Comment on implementation - have you considered using the USWDS gem? That would remove the need to have a lot of this boilerplate code in version control (and it has a dependency on bourbon and neat so we wouldn't need those, either). 

The released version of the gem has some isssues with Rails / asset pipeline, but I have a fork that works: https://github.com/18F/e-manifest/blob/master/Gemfile#L39

to see how @jeremiak moved e-manifest to USWDS gem, see: https://github.com/18F/e-manifest/pull/55

If you think it best to do that another time, that's cool too.

gemfarmer:
@jessieay I was not aware of the gem, but honestly didn't really use the USWDS code much. I think just the one stylesheet, and I just copied that code directly. More than anything I used the color palette and icons from the [18F Branding page](https://pages-staging.18f.gov/brand/visual-style/). I thought that this was appropriate because it is an internal app. 

If we want to shift the site to more closely resemble the USWDS then, then I think we should add it, but maybe not for this first style iteration??

PR #958, opened 2016-02-20
Rewrite emails from scratch

jessieay:
This pull request re-writes one email: the proposal created confirmation email to the requester. I don't have the numeric logos yet so that and the approver roles are not yet included in the approver list.

Today I went down the path of rewriting emails from scratch since the grid provided by Ink does not work in gmail mobile and even the desktop version of the email I created with ink had formatting weirdness (https://github.com/18F/C2/pull/955) which was the whole point of using it in the first place.

https://trello.com/c/lf0XzPWo/120-email-conversion-requester-new-request-confirmation

For images, I created an s-3 bucket to host images so we aren't hitting our app when people load emails. Could send images as inline attachments, but that seems like a lot of overheard for images that are going to be the same for every email. We should move these images to a c2 S3 bucket in the short term but this is a good short term solution.

Based on the screenshots below, I think we should consider making the email text smaller so more of it can be seen at once.

Screenshot of the email in the Rails preview (web):

screen shot 2016-02-19 at 5 14 31 pm

Screenshots of this email in desktop gmail:

Above fold:

screen shot 2016-02-19 at 5 06 17 pm

Below fold:

screen shot 2016-02-19 at 5 06 22 pm

And in mobile gmail (ios):

Above fold:

img_3292

Below fold:

img_3293


pkarman:
@jessieay seems like a decent start on the CSS front.

I don't understand the images-via-s3 thing. Why should we be avoiding serving static assets from our site?

jessieay:
@pkarman re: images - if we serve the assets from our site, our servers will be hit every time someone loads an email. Not a big deal at this scale, but easily solved by uploading an image to S3. Also prevents need to deal with asset pipeline.

PR #959, opened 2016-02-22
Rewrite emails from scratch

jessieay:
This is a PR off of the email rewrite master branch.


jessieay:
email screenshots located in https://github.com/18F/C2/pull/958

pkarman:
@jessieay my concern with hosting some images via S3 is that (a) they aren't versioned and (b) we're spreading the project assets across multiple places, which can make it harder to maintain.

Particularly at this early stage of development I would prefer to see all the related files in one place, to make it easier to keep track of.

jessieay:
@pkarman ok updated at baf84b0

PR #960, opened 2016-02-22
do not limit text fields to 255 chars

pkarman:
Trello: https://trello.com/c/xC3qGgZL/138-gsa18f-procurements-limited-to-255-chars

### DIFF
--- a/db/migrate/20160222185247_convert_gsa18_f_to_text.rb
+++ b/db/migrate/20160222185247_convert_gsa18_f_to_text.rb
@@ -1,7 +1,13 @@
 class ConvertGsa18FToText < ActiveRecord::Migration
-  def change
+  def up
     change_column :gsa18f_procurements, :office, :text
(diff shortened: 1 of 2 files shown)


jessieay:
should we do this for `ncr_work_orders` while we're here?

pkarman:
maybe? we just haven't had any bug reports about that client type because all the fields (except `project_title`) are populated via dropdown menus, not text boxes. I.e. there aren't free-form inputs on the NCR form.

pkarman:
@jessieay since this represents an actual bug, as opposed to a hypothetical one, let's move forward with this fix and change the NCR case if/when it is a problem.

jessieay:
@pkarman sounds good. only update I'd make before merging is to make the migration reversible.

pkarman:
@jessieay good call. I guess the `change` method doesn't handle `change_column` very well.

Fixed in 516bbc5

jessieay:
@pkarman :+1:

PR #961, opened 2016-02-22
Add new attachment email

jessieay:
https://trello.com/c/80dedGV8/134-email-conversion-attachment-added

HTML:

screen shot 2016-02-22 at 12 57 46 pm

Text only:

screen shot 2016-02-22 at 12 57 54 pm


jessieay:
Note: the HTML email is not entirely "to spec" -- getting it there will take more time so I am going to do a first pass where I get all of the emails generally over to the new styles  and then we can get everything in tip top shape later.

pkarman:
:+1:  w/r/t this styles approach.

code looks fine. when tests pass I can merge.

jessieay:
@pkarman updated to be non-broken

PR #964, opened 2016-02-23
Move attachment email to its own mailer

jessieay:
* Slowly but surely moving everything out of Mailer


pkarman:
@jessieay merge conflicts, alas

jessieay:
@pkarman updated!

lib/mail_previews/mailer_preview.rb
```diff
@@ -33,7 +29,12 @@ def proposal
     Proposal.last
   end
 
+<<<<<<< HEAD
```
pkarman: looks like a merge conflict got committed @jessieay ?

jessieay: oy! fixed in feac16b - thanks for the careful re-review

PR #966, opened 2016-02-23
c2-prod-elasticsearch is now provisioned

pkarman:
prep for production release of master tonight


jessieay:
woo hoo! should we merge or do you want to wait until you release?

pkarman:
fine to merge now -- the ES instance is up so this would only bind to it.

PR #972, opened 2016-02-24
Refactor Dispatcher classes

jessieay:
* Make them easier to read and understand


baccigalupi:
Win for easier reasonableness, as in I can reason-about this!

app/models/dispatcher.rb
```diff
@@ -99,10 +96,16 @@ def on_comment_created(comment)
     end
   end
 
-  def on_proposal_update(_proposal)
+  def on_proposal_update(modifier)
+    if proposal.client_slug == "ncr"
```
pkarman: this client-specific logic in a non-client-specific class smells a little funny to me. Should we be delegating? or subclassing?

if we're going to keep the client-specific code, maybe we should make it a little less brittle and do something like:

```
if proposal.ncr?
```

?

app/models/dispatcher.rb
```diff
@@ -110,28 +113,26 @@ def on_approver_removal(proposal, removed_approvers)
...
   end
 
-  def requires_approval_notice?(_approval)
-    true
-  end
-
-  def send_notification_email(approval)
-    Mailer.actions_for_approver(approval).deliver_later
+  def requires_approval_notice?(approval)
+    if proposal.client_slug == "ncr"
```
pkarman: same comment about client-specific code

app/models/ncr_dispatcher.rb
```diff
@@ -1,67 +0,0 @@
-class NcrDispatcher < LinearDispatcher
```
pkarman: ah I see you are getting rid of the subclass.

My fear is that if we have 100 clients, our Dispatcher is a long list of `if proposal.client_slug == whatever` chains.

app/models/dispatcher.rb
jessieay: we were delegating  before and it was getting very confusing ->

`Dispatcher` would initialize `NcrDispatcher` or `LinearDispatcher`, `NcrDispatcher` inherited from `LinearDispatcher`, which in turn inherited from `Dispatcher`. This PR is an attempt to make the branching much more explicit. I expect that we will have separate dispatchers at some point in the future, but consolidating makes the logic easier to read for now and then we can break out the differences from here.

Why is `proposal.ncr?` less brittle than looking at the `client_slug` attribute? I am open to it, just wondering.

app/models/ncr_dispatcher.rb
jessieay: I agree that would be problematic. But what we have now was already confusing so this PR is my attempt at clarifying the logic. Ideas on a better way of dealing with the branching?

app/models/dispatcher.rb
pkarman: if I fat-finger while coding and type this:

```
if proposal.client_slug == "mcr"
```

and then wonder why my tests are failing (or worse! not failing), that's more brittle than mistyping `proposal.mcr?` which will raise a method not found error (catastrophic failure being a Good Thing).

i.e. if we're going to hard-code client-specific things into client-agnostic classes, why not make the client-specific code into a method that is harder to mistype.

jessieay: ok that makes sense. do you think that checking `ncr?` here is sufficient, or should I make a larger change to clarify branching?

app/models/ncr_dispatcher.rb
pkarman: I agree that this PR clarifies a lot and on the whole I like it.

I'm wondering if we could just create a `NcrDispatcher` subclass that only overrides the methods where you have `ncr?` checks. Then it would be up to the calling code to call the client-specific `NcrDispatcher` rather than `Distpatcher`.

jessieay:
@pkarman good suggestions , let me know what you think of my changes in c545e54

pkarman:
@jessieay -- `DispatchFinder` -- brilliant. I love it. All the client-specific routing in one easy-to-read place.

PR #975, opened 2016-02-24
Remove unused `proposal_observer_email` method

jessieay:
* Was moved to ObserverMailer previously


pkarman:
@jessieay merge conflict alas

jessieay:
@pkarman updated!

pkarman:
@jessieay our ES retry approach seems to not be working. I'll investigate.

Otherwise this looks great.

PR #981, opened 2016-02-26
Single -> Double quotes

jessieay:
* Greplace is fun


pkarman:
@jessieay some test failures.

jessieay:
@pkarman argh missed one. Fixed in 962e439

PR #984, opened 2016-03-01
Create proposal notification email for step user

jessieay:
* https://trello.com/c/60fsK5PN/147-email-conversion-approver-new-request-for-approval

for approval step:

screen shot 2016-02-29 at 3 47 56 pm

screen shot 2016-02-29 at 3 48 01 pm

for purchase step:

screen shot 2016-02-29 at 4 17 54 pm

screen shot 2016-02-29 at 4 17 48 pm


app/helpers/mailer_helper.rb
```diff
@@ -56,6 +46,14 @@ def add_reason(text, reason)
     end
   end
 
+  def proposal_attributes(proposal)
```
pkarman: this is the ol' "what happens when we have a dozen clients" comment.

could we delegate this to the client_data class and require it to have a `proposal_attributes` method that is called? That way the client_data class knows how to bootstrap its own client-specific proposal attributes.

jessieay: good point - we already have a `fields_for_display` method on each client data and that method is delegated from `proposal` (we use it in the web views) so we can take advantage of that here

jessieay:
@pkarman updated at 33bf647

PR #987, opened 2016-03-01
Welcome mailer

pkarman:
Trello: https://trello.com/c/HvrtBUy8

### DIFF
--- a/app/controllers/auth_controller.rb
+++ b/app/controllers/auth_controller.rb
@@ -31,12 +31,14 @@ class AuthController < ApplicationController
   def do_user_auth(auth)
     sign_out
     user = User.from_oauth_hash(auth)
-    send_welcome_mail(user)
+    unless ENV["NO_WELCOME_EMAIL"]
+      send_welcome_mail(user)
+    end
     sign_in(user)
   end
 
   def send_welcome_mail(user)
-    if (Time.current - user.created_at) < 10.seconds
+    if (Time.current - user.created_at) < 2.seconds
       WelcomeMailer.welcome_notification(user).deliver_later
     end
   end
--- a/app/controllers/auth_controller.rb
+++ b/app/controllers/auth_controller.rb
@@ -31,6 +31,13 @@ class AuthController < ApplicationController
   def do_user_auth(auth)
     sign_out
     user = User.from_oauth_hash(auth)
+    send_welcome_mail(user)
     sign_in(user)
+  end
+
+  def send_welcome_mail(user)
+    if (Time.current - user.created_at) < 10.seconds
+      WelcomeMailer.welcome_notification(user).deliver_later
+    end
   end
 end
--- a/app/views/welcome_mailer/welcome_notification.text.erb
+++ b/app/views/welcome_mailer/welcome_notification.text.erb
@@ -1,0 +1,10 @@
+<%= t("mailer.welcome_mailer.welcome_notification.header") %>
+
+<%= t("mailer.welcome_mailer.welcome_notification.para1") %>
+
+<%= t("mailer.welcome_mailer.welcome_notification.para2", help_url: help_url('') ) %>
+
+<%= t("mailer.welcome_mailer.welcome_notification.para3", feedback_url: feedback_url ) %>
+
+<%= t("mailer.welcome_mailer.welcome_notification.signature") %>
+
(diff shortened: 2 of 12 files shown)


jessieay:
@pkarman circle failing

spec/requests/set_user_after_login_spec.rb
```diff
@@ -17,6 +17,14 @@
     expect(new_user.last_name).to eq("Jetsonian")
   end
 
+  it "sends welcome email to a new user" do
```
jessieay: should we perhaps wrong a spec that ensures we don't send a welcome email to a user on login when the user is not new?

pkarman: good call. added in e578670

pkarman:
@jessieay ok I think I've tracked down all the affected tests. All the `deliveries` counts were off-by-one after adding the welcome email.

app/controllers/auth_controller.rb
```diff
@@ -31,6 +31,15 @@ def fetch_return_to_path
   def do_user_auth(auth)
     sign_out
     user = User.from_oauth_hash(auth)
+    unless ENV["NO_WELCOME_EMAIL"]
```
jessieay: why do we have an env var for this? not sure why we'd want to suppress welcome emails but not any others

pkarman: for all the tests. Every time a user gets created via FactoryGirl and then `login_as()` is called in a spec, a welcome email is getting generated, which was creating lots of test errors because `deliveries` was incrementing where not expected.

jessieay: ahhh ok makes sense - might be good to fix all those tests at some point rather than not sending the emails by default (the tests should probably be re-written to not actually git the auth controller for login purposes) but this seems like a good solution for now

app/controllers/auth_controller.rb
```diff
@@ -31,6 +31,15 @@ def fetch_return_to_path
...
+      send_welcome_mail(user)
+    end
     sign_in(user)
   end
+
+  def send_welcome_mail(user)
+    if (Time.current - user.created_at) < 2.seconds
+      WelcomeMailer.welcome_notification(user).deliver_later
```
jessieay: thoughts on sending this via `Dispatcher` instead? that is what we use to shell out the rest of our emails

app/views/welcome_mailer/welcome_notification.text.erb
```diff
@@ -0,0 +1,10 @@
+<%= t("mailer.welcome_mailer.welcome_notification.header") %>
+
+<%= t("mailer.welcome_mailer.welcome_notification.para1") %>
+
+<%= t("mailer.welcome_mailer.welcome_notification.para2", help_url: help_url('') ) %>
+
+<%= t("mailer.welcome_mailer.welcome_notification.para3", feedback_url: feedback_url ) %>
```
jessieay: these urls are still in `a` tags so we should probably include these links separately for a non-HTML version

jessieay:
a few final comments, otherwise looks ready for :ship:

app/controllers/auth_controller.rb
pkarman: there wasn't anything client-specific about a "welcome to the product" message so my initial thought was to skip the extra level of indirection. I'm happy to make it a Dispatcher method if you think consistency is most important.

app/views/welcome_mailer/welcome_notification.text.erb
pkarman: Since these were long strings I put a `*_html` version in the locale file with <a> tags.

app/controllers/auth_controller.rb
pkarman: ah now I remember. Dispatcher.new requires a proposal. A welcome has no proposal.

jessieay: I think it's fine to keep it as-is. I do like having all emails triggered via dispatcher for grepping reason but I agree with you that this isn't client specific so not needed. Plus easy to move later if need be

app/views/welcome_mailer/welcome_notification.text.erb
jessieay: missed that!

PR #993, opened 2016-03-01
ES retry

pkarman:
example: https://circleci.com/gh/18F/C2/4010


pkarman:
ignore the security warnings. already fixed in master.

PR #994, opened 2016-03-01

rememberlenny:

## New Gems

## Database changes

```visits``` and ```ahoy_events``` are used for tracking front-end and backend activity in relation to the logged on user. The events can be used to make details reports for error handling and observe user behavior.

```visit_id``` column is added to ```proposals```, ```reports``` and ```comments``` to associate the visit of a user, when new instances are created. For example, the visit behavior leading up to a proposal being created can be tracked and observed.

# How
*Magic*

---

## Admin

### CODE REVIEW

File: db/migrate/20160208173646_add_the_visit_id_to_models.rb [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+    add_foreign_key :proposals, :visits
+    add_foreign_key :comments, :visits
+    add_foreign_key :reports, :visits
+  end
+
+  def down
+    remove_foreign_key :proposals, :visits
+    remove_foreign_key :comments, :visits
+    remove_foreign_key :reports, :visits
+    drop_column :proposals, :visit_id
```
pkarman: @rememberlenny this is the wrong method. use `remove_column` instead.

PR #997, opened 2016-03-01
Create email templates for emails to step users after step complete

jessieay:
* one template for steps that need to be re-reviewed, one for steps
  that do not
* re-reviewed: https://trello.com/c/jWfjbfxD
* not re-reviewed: https://trello.com/c/jWfjbfxD

Needs re review:

screen shot 2016-03-01 at 1 38 03 pm

screen shot 2016-03-01 at 1 37 33 pm

Does not need re-review:

screen shot 2016-03-01 at 1 29 14 pm

screen shot 2016-03-01 at 1 29 08 pm

### DIFF
--- a/app/mailers/proposal_mailer.rb
+++ b/app/mailers/proposal_mailer.rb
@@ -38,6 +38,36 @@ class ProposalMailer < ApplicationMailer
     )
   end
 
+  def proposal_updated_step_complete_needs_re_review(step, modifier = nil)
+    @step = step.decorate
+    unless @step.api_token
+      @step.create_api_token
+    end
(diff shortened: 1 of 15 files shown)


pkarman:
@jessieay looks ok.

under what conditions would the `modifier` ever legitimately be a `NullUser` ?

jessieay:
@pkarman if someone modified it via Rails console there would be no `current_user`, I believe.

jessieay:
@pkarman if the `modifier` argument were non-optional I'd assume we would always have one. I think it is optional because of the rails console case.

pkarman:
@jessieay :+1:

PR #999, opened 2016-03-01
Share logic in mailers

jessieay:
* Define `subject` in ApplicationMailer
* Use new layout as deafult layout


pkarman:
@jessieay merge conflict alas

jessieay:
@pkarman updated

PR #1000, opened 2016-03-02

rememberlenny:

## New Gems

## Database changes

```visits``` and ```ahoy_events``` are used for tracking front-end and backend activity in relation to the logged on user. The events can be used to make details reports for error handling and observe user behavior.

```visit_id``` column is added to ```proposals```, ```reports``` and ```comments``` to associate the visit of a user, when new instances are created. For example, the visit behavior leading up to a proposal being created can be tracked and observed.

# How
*Magic*

---

## Admin

### CODE REVIEW

**Comment by pkarman** [comment]
I will QA

File: db/migrate/20160208173646_add_the_visit_id_to_models.rb [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+    add_foreign_key :proposals, :visits
+    add_foreign_key :comments, :visits
+    add_foreign_key :reports, :visits
+  end
+
+  def down
+    remove_foreign_key :proposals, :visits
+    remove_foreign_key :comments, :visits
+    remove_foreign_key :reports, :visits
+    drop_column :proposals, :visit_id
```
pkarman: @rememberlenny this seems to need updating (again) -- should be `remove_column`

pkarman:
QA:

* deployed to c2-dev
* previous issues described here: https://github.com/18F/C2/pull/936#issuecomment-182471500
* `/admin` events are still being logged. buttons have been fixed.
* data retention question not addressed.

@rememberlenny what are your thoughts on how long this data should be retained, and whether `/admin` events should be logged or not?

rememberlenny:
@pkarman Thanks for taking a look at this!

For logs, I'd like to keep all the data if possible. If the table size does seem like an issue, we could always push the events onto another database/rails instance. For immediate, Im open to doing a 60 day retention policy. What do you think?

I think the admin events being logged is good. What do you think?

pkarman:
I'm fine to keep everything (for now) and log admin events. Just wanted to make sure those were intentional.

PR #1004, opened 2016-03-02
Observer added or removed

pkarman:
trello: https://trello.com/c/LDQG6dFR
trello: https://trello.com/c/9wJjCoK0

The `ActivityMailer` stuff is a placeholder for https://trello.com/c/4jRKhfhA so that dispatcher tests continue to work.


app/views/observer_mailer/observer_added_notification.html.haml
```diff
@@ -0,0 +1 @@
+<!-- TODO activity template -->
```
pkarman: this is where we would use the standard "activity happened" template with some local `message`

app/controllers/observations_controller.rb
```diff
@@ -17,6 +17,7 @@ def destroy
     else
       redirect_path = proposal_path(proposal)
     end
+    Dispatcher.on_observer_removed(observation)
```
jessieay: `DispatchFinder` ?

app/mailers/observer_mailer.rb
```diff
@@ -1,28 +1,43 @@
 class ObserverMailer < ApplicationMailer
-  def on_observer_added(observation, reason)
+  def observer_added_confirmation(observation, reason)
```
jessieay: NAMING alert: I've been naming / re-naming these emails kind of willy nilly but one emerging pattern is that emails that are triggered by a user and go to that same user (eg: You cancelled your own request!) are considered `confirmation`s while actions triggered by another user are called `notification`s. If we think that is a good pattern, I think both of these should be `notification`s. Thoughts? 

There is another school of thought that those are both redundant words in the context of an email and should just be removed from all email names.

app/mailers/welcome_mailer.rb
```diff
@@ -1,7 +1,7 @@
 class WelcomeMailer < ApplicationMailer
   def welcome_notification(user)
     mail(
-      to: email_with_name(user.email_address, user.full_name),
+      to: email_to_user(user),
```
jessieay: I've been wondering if there are potential benefits of sending emails 'to' the formatted email / name.

app/views/activity_mailer/activity_notification.text.erb
```diff
@@ -0,0 +1,6 @@
+<%= @activity %>
+
+<%= t("mailer.proposal_link_text",
+      proposal_url: proposal_url(@proposal)) %>
+
+<%= t("mailer.footer", feedback_url: feedback_url) %>
```
jessieay: if the `activity_mailer` has a layout, should we include the footer in that? Realize there may also be conflicts with work @rememberlenny is working on...

app/mailers/observer_mailer.rb
pkarman: ah ok. I was trying to follow your convention but did not understand the correlation between `to` and the name. Agreed this one should change.

I think its fine to have that convention; I have tried to avoid `email` in all mailer method names for the redundancy reasons.

pkarman: since the `observer_removed_*` action can be performed by either the observer themselves or some other user, should we leave that as `confirmation` or change?

app/mailers/welcome_mailer.rb
pkarman: I consider it friendlier to use the recipient's name, so just a UX benefit. Same with `from`.

app/views/activity_mailer/activity_notification.text.erb
pkarman: do we have a layout for text emails? I hadn't noticed one.

jessieay: oh hah I didn't even see this was text. I haven't been including the feedback link in text emails -- we should do that! added a card https://trello.com/c/Un6KSc7U

app/mailers/welcome_mailer.rb
jessieay: :+1:

jessieay:
woo hoo! two more email cards bit the dust

app/mailers/observer_mailer.rb
jessieay: as-is works for me!

PR #1007, opened 2016-03-03
Approving official download

pkarman:
trello: https://trello.com/c/UctuAOdM

This change should stringify associations in CSV download, rather than just printing the foreign key id.

### DIFF
--- a/app/models/concerns/client_data_mixin.rb
+++ b/app/models/concerns/client_data_mixin.rb
@@ -77,10 +77,10 @@ module ClientDataMixin
       field_values = []
       self.class.column_names.sort.each do |attribute|
         field_values << if self.class.foreign_key_to_method_map[attribute]
-          send(self.class.foreign_key_to_method_map[attribute])
-        else
-          send(attribute)
-        end
+                          send(self.class.foreign_key_to_method_map[attribute])
+                        else
+                          send(attribute)
+                        end
       end
       field_values
     end
--- a/app/models/concerns/client_data_mixin.rb
+++ b/app/models/concerns/client_data_mixin.rb
@@ -76,10 +76,10 @@ module ClientDataMixin
     def csv_fields
       field_values = []
       self.class.column_names.sort.each do |attribute|
-        if self.class.foreign_key_to_method_map[attribute]
-          field_values << send(self.class.foreign_key_to_method_map[attribute])
+        field_values << if self.class.foreign_key_to_method_map[attribute]
+          send(self.class.foreign_key_to_method_map[attribute])
         else
-          field_values << send(attribute)
+          send(attribute)
         end
       end
       field_values
--- a/app/models/concerns/client_data_mixin.rb
+++ b/app/models/concerns/client_data_mixin.rb
@@ -73,13 +73,25 @@ module ClientDataMixin
       @_fk_map ||= Hash[reflect_on_all_associations(:belongs_to).map { |a| [a.foreign_key, a.name] }]
     end
 
+    def is_association_column?(column_name)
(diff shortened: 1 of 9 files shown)


app/models/concerns/client_data_mixin.rb
```diff
@@ -69,8 +69,20 @@ def self.csv_headers
       column_names.sort.map { |attribute| human_attribute_name(attribute) }
     end
 
+    def self.foreign_key_to_method_map
+      @_fk_map ||= Hash[reflect_on_all_associations(:belongs_to).map { |a| [a.foreign_key, a.name] }]
+    end
+
     def csv_fields
```
jessieay: test for this new behavior?

pkarman: test added in 069a1d2

app/models/concerns/client_data_mixin.rb
```diff
@@ -69,8 +69,20 @@ def self.csv_headers
...
-      self.class.column_names.sort.map { |attribute| send(attribute) }
+      field_values = []
+      self.class.column_names.sort.each do |attribute|
```
jessieay: would it be possible to throw some of this logic into a well-named method? I think the test clarifies behavior but this is still a little bit complicated for me to read and understand

spec/controllers/proposals_controller_spec.rb
```diff
@@ -194,6 +194,7 @@
       es_execute_with_retries 3 do
         get :download, text: "Work Order", format: "csv"
         expect(response.body).to include "Work Order 29"
+        expect(response.body).to include proposals.last.client_data.approving_official.display_name
```
jessieay: I feel like I'd want to see an email address here -- names are not unique and difficult if someone wants to, say, follow up with XYZ person about something based on a CSV download. Or did we hear otherwise?

spec/support/fixtures/test_models.rb
```diff
@@ -4,6 +4,8 @@ def self.table_name_prefix
   end
 
   class ClientRequest < ActiveRecord::Base
+    belongs_to :approving_official, class_name: User
```
jessieay: not all client requests have an approving official (eg: 18F does not). do we still want to include the relation here?

spec/controllers/proposals_controller_spec.rb
pkarman: just so we're talking about the same thing: `user.display_name` returns `Jane Smith <jane@example.com>` if the first/last are defined, otherwise returns just `<jane@example.com>`.

spec/support/fixtures/test_models.rb
pkarman: this class `ClientRequest` is only for testing and I added the association so that we didn't have to test `NCR::WorkOrder` explicitly. I.e. this is an example of an association.

spec/controllers/proposals_controller_spec.rb
jessieay: oh! cool. didn't realize that. was thinking about the `full_name` method. I feel like we should use `display_name` everywhere although I realize not everyone agrees with that

pkarman:
@jessieay rubocop fixes :green_heart: now

spec/support/fixtures/test_models.rb
jessieay: make sense, I think it is fine to keep as-is!

jessieay:
sweet, wil QA now

jessieay:
download looking fly on dev

screen shot 2016-03-04 at 10 31 57 am

:rainbow: 

mergin'

PR #1008, opened 2016-03-03
Update Proposals#update logic

jessieay:
* After going through emails with Amber yesterday, we decided to
* simplify the email notifications after a proposal is updated

* IF a proposal is updated and DOES NOT need to be re-reviewed, we send
  observers, approvers who have approved, and requester an email that
  says "no action required"
* In that same situation, any pending approvers are notified of change
  and told that the request was updated since they were last notified

* IF a proposal is a updated and DOES  need to be re-reviewed, we send
  the observers and requester an email notifying them that the request
  is being re-routed.
* In that same situation, the new current approver is notified of the
  proposal as if it is a new proposal
* This is potentially confusing if that new current approver approved
  the request before it was updated, but it seems like an edge case and
  the over head required to keep track of old approvers vs new and who
  approved what seemed not worth it for the payoff
12369aa  

Email 8: Requester / Observer, modified post-approval (re-approval required)
https://trello.com/c/O9bI84jL

Email 9: Observer / Requester /Approver, modified request, no action required
https://trello.com/c/dnhMDryn

Email 10: Approver, modified request for an awaiting approval
https://trello.com/c/nbI2mxDU

Email 11 (removed) Email conversion: Approver, Observer, modified request need reapproval
https://trello.com/c/jWfjbfxD


pkarman:
code looks fine, the narrative explanation makes sense.

I'm trusting that the UX changes implicit in these workflows will be thoroughly tested for all the emails we're touching, so merging.

PR #1010, opened 2016-03-04
Myusa doc update

yozlet:
At last. https://trello.com/c/p9dKEhQb/32-expand-myusa-setup-instructions

I think this covers the problems that were had in the past.


pkarman:
lovely

PR #1011, opened 2016-03-04
Prevent Roadie from inlining media queries

rememberlenny:
Isolate stylesheet with media queries and set up the roadie ignore

## Before
screen shot 2016-03-04 at 11 16 08 am

## After
screen shot 2016-03-04 at 11 16 05 am


jessieay:
nice! so if roadie doesn't inline these will they be recognized by gmail? assuming them *not* being inlined means they will be in a stylesheet, which I thought email clients ignored. Educate me!

rememberlenny:
@jessieay 🐋

So this is the part that makes emails complicated.

Most web pages are written with the assumption that all the content is loaded, then the CSS determines the priority for styling. In emails, you have to design your CSS/class/markup base on the assumption that you can have some parts of your code NOT run. 

When it comes to responsive layouts, you have to design for the care where desktop is default (without the media queries) and any media queries are added, only if its possible to run inline styles. In the case where you have things like Gmail, where the style tags are stripped out, you need to account for this. 

The problem with Roadie was that it was inlining the media query styles that would run on mobile, which broke the default desktop layout. By having roadie ignore certain styles, we control the designed layout to render the desktop mode in all cases, and then serve the inline style tag with media queries to be run where possible.

To make things more complicated, Outlook is very outdated. Specifically, there are a lot of CSS3 properties and even CSS2 properties it doesnt render. As a result, outlook has some legacy hacks where you can use comment flags to tell Outlook NOT to render certain content.

For what its worth, we are putting these things into consideration in our templates : )

jessieay:
@rememberlenny thank you for the detailed explanation.

I am still not sure I understand the media query issue

> When it comes to responsive layouts, you have to design for the care where desktop is default (without the media queries) and any media queries are added, only if its possible to run inline styles. In the case where you have things like Gmail, where the style tags are stripped out, you need to account for

Can you explain this a little further?

I am going to merge because your comment made it sound like you are ready for this to be merged. In the future, un-assign yourself from PRs when they are ready so we know it is time for re-review. Thanks!

PR #1013, opened 2016-03-04
search for associated records

pkarman:
Now that NCR uses foreign keys rather than strings for related organization, searches for org code fail. This PR fixes that.

trello: https://trello.com/c/a6SjKfiH

Also included:

* zero downtime for rebuilding the Proposals index
* re-index a single Proposal via the /admin UI

### DIFF
--- a/app/admin/proposal.rb
+++ b/app/admin/proposal.rb
@@ -11,4 +11,14 @@ ActiveAdmin.register Proposal do
     column :requester
     actions
   end
+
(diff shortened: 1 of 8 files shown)


jessieay:
code looks good! will QA a bit later today

pkarman:
QA note: you will need to re-index via /admin to see the change.

jessieay:
looks good on dev! merging

PR #1012, opened 2016-03-04
Create Cancellation Notification Email (read before merge)

rememberlenny:
# Note

This contains https://github.com/18F/C2/pull/1011

# Test

[http://localhost:5000/rails/mailers/cancellation_mailer/cancellation_notification](http://localhost:5000/rails/mailers/cancellation_mailer/cancellation_notification)

# Visual 

screen shot 2016-03-04 at 11 56 19 am

PR #1014, opened 2016-03-04
Setup New Proposal email (read before merge)

rememberlenny:
# Note

This comes after https://github.com/18F/C2/pull/1011 and https://github.com/18F/C2/pull/1012

# Design note

Authorization list is wider than the design. The design breaks w/ outlook, if the name listed is a long email address. The area is now wider to cater to a email address.

o
chromegmailnew-vertical-allowed-1366 2
ol2013-vertical-allowed-1366 2


app/helpers/application_helper.rb
```diff
@@ -23,4 +23,5 @@ def display_profile_warning?
   def display_search_ui?
     current_user && current_user.client_model && !client_disabled?
   end
+
```
jessieay: looks like you added newlines after blocks in a few files - I generally like newlines before/after multi-line blocks *except* when the end the block is directly nested inside another block (eg: two `end`s next to each other) what do you think?

jessieay:
I'm having a little trouble figuring out if some of these changes are included because they are part of commits that this PR depends on or if they are part of the PR itself. GitHub should auto-update to only show the commits for this PR since the two it depends on are merged, but I've noticed recently that it doesn't always work that way (Perhaps because we are merging into a non-master branch?)

Which Trello card does this PR represent? If you can let me know then I think it will be easier to figure it out. Also re-rebasing against `EMAIL-REWRITE-MASTER` would be cool

Thanks! this all looks awesome

rememberlenny:
@jessieay Email is from this ticket: https://trello.com/c/lf0XzPWo/120-1-email-conversion-requester-new-request-confirmation

jessieay:
thanks for the update, @rememberlenny -- mergin!

PR #1015, opened 2016-03-04
Added email template for approval role removed

rememberlenny:
# Note

This comes after https://github.com/18F/C2/pull/1011 and https://github.com/18F/C2/pull/1012 and https://github.com/18F/C2/pull/1014

screen shot 2016-03-04 at 4 11 17 pm

Adding email ticket: https://trello.com/c/uWQFAloH/135-20-email-conversion-removal-as-approver


rememberlenny:
@jessieay Adding email ticket: https://trello.com/c/uWQFAloH/135-20-email-conversion-removal-as-approver

jessieay:
:tada:

config/locales/mailers/en.yml
```diff
@@ -34,6 +34,10 @@ en:
         header: "A new request submitted by %{requester_name} needs your %{step_type_noun}"
         cta: "Add comment"
     proposal_mailer:
+      generic_approval_name:
```
jessieay: would it be possible to make these dynamic now? We could do this via a decorator for each user based on their role. 

Could also keep it mocked out but if we do we need to create a separate card so we don't forget to switch these out. These roles are accurate for some NCR requests but not all (eg: not all will have 3 approvers, not all approval chains will have these roles)

jessieay:
@rememberlenny mind re-rebasing? seems like some of those proposal confirmation changes are in here :question:

PR #1016, opened 2016-03-04
Fix org-reassigner migration to not run AR validations

yozlet:
If you were running this migration in a world where the NCR work order
approving official is not yet a relation, the validation will fail, but we
want the update to happen anyway.

Fixes https://trello.com/c/XSbWulQS/166-update-migration-to-not-fail

Paired with @jessieay


jessieay:
@yozlet this created some failing specs for that reassigner class

yozlet:
Whoops, sorry! Fixed.

jessieay:
looks great, thanks for updating @yozlet !

PR #1019, opened 2016-03-07
Report scheduler

pkarman:
trello: https://trello.com/c/S4PVVXP5/165-backend-mail-out-custom-report-daily-weekly-monthly

This is the backend/model support only, no UI as part of this card.

### DIFF
--- a/app/admin/report.rb
+++ b/app/admin/report.rb
@@ -12,4 +12,23 @@ ActiveAdmin.register Report do
     end
     f.actions
   end
+
+  show do
+    attributes_table do
+      row :name
+      row :query
+      row :shared
+      row :user
+      row :created_at
+      row :updated_at
+    end
+
+    panel "Scheduled Subscriptions" do
+      table_for report.subscriptions.order("created_at DESC") do |tbl|
+        tbl.column("Name") { |scheduled_report| link_to scheduled_report.name, admin_scheduled_report_path(scheduled_report) }
+        tbl.column("Owner") { |scheduled_report| scheduled_report.user }
+        tbl.column("Freqency") { |scheduled_report| scheduled_report.frequency }
+      end
+    end
+  end
 end
--- a/app/admin/scheduled_report.rb
+++ b/app/admin/scheduled_report.rb
@@ -6,7 +6,7 @@ ActiveAdmin.register ScheduledReport do
       f.input :name
       f.input :report
       f.input :user
-      f.input :frequency, as: :select, collection: ["daily", "weekly", "monthly"]
+      f.input :frequency, as: :select, collection: ScheduledReport.frequencies.keys
     end
     panel "Notes" do
       "Weekly reports are sent on Mondays. Monthly reports are sent on the first day of the month."
--- a/app/admin/scheduled_report.rb
+++ b/app/admin/scheduled_report.rb
@@ -1,0 +1,13 @@
+ActiveAdmin.register ScheduledReport do
+  permit_params :name, :report_id, :frequency, :user_id
+
+  form do |f|
+    f.inputs "Scheduled Report" do
+      f.input :name
+      f.input :report
+      f.input :user
+      f.input :frequency, as: :select, collection: ["daily", "weekly", "monthly"]
+    end   
+    f.actions
+  end
+end
--- a/app/admin/scheduled_report.rb
+++ b/app/admin/scheduled_report.rb
@@ -8,6 +8,9 @@ ActiveAdmin.register ScheduledReport do
       f.input :user
       f.input :frequency, as: :select, collection: ["daily", "weekly", "monthly"]
     end
+    panel "Notes" do
+      "Weekly reports are sent on Mondays. Monthly reports are sent on the first day of the month."
+    end
     f.actions
   end
 end
--- a/app/admin/scheduled_report.rb
+++ b/app/admin/scheduled_report.rb
@@ -7,7 +7,7 @@ ActiveAdmin.register ScheduledReport do
       f.input :report
       f.input :user
       f.input :frequency, as: :select, collection: ["daily", "weekly", "monthly"]
-    end   
+    end
     f.actions
   end
 end
--- a/app/services/scheduled_reporter.rb
+++ b/app/services/scheduled_reporter.rb
@@ -8,7 +8,7 @@ class ScheduledReporter
     @check_time = check_time
   end
 
-  def check
+  def run
     if check_time.day == 1
       send_monthlies
     elsif check_time.monday?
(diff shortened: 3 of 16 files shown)


pkarman:
these codeclimate errors all look like false positives to me.

app/admin/report.rb
```diff
@@ -12,4 +12,23 @@
...
+      row :shared
+      row :user
+      row :created_at
+      row :updated_at
+    end
+
+    panel "Scheduled Subscriptions" do
+      table_for report.subscriptions.order("created_at DESC") do |tbl|
+        tbl.column("Name") { |scheduled_report| link_to scheduled_report.name, admin_scheduled_report_path(scheduled_report) }
+        tbl.column("Owner") { |scheduled_report| scheduled_report.user }
```
jessieay: I think what code climate is suggesting is this:

`tbl.column("Owner")(&:user)`

but I do think what we have here is easier to read so fine with either

app/admin/scheduled_report.rb
```diff
@@ -0,0 +1,13 @@
+ActiveAdmin.register ScheduledReport do
+  permit_params :name, :report_id, :frequency, :user_id
+
+  form do |f|
+    f.inputs "Scheduled Report" do
+      f.input :name
+      f.input :report
+      f.input :user
+      f.input :frequency, as: :select, collection: ["daily", "weekly", "monthly"]
```
jessieay: should we clarify in the UI when each of these types of reports is sent out?

spec/mailers/report_mailer_spec.rb
```diff
@@ -44,4 +45,31 @@
...
+    it "mails report on schedule", :elasticsearch do
+      deliveries.clear
+      owner = create(:user)
+      report = create(:report, query: { text: "something" }.to_json, user: owner)
+      scheduled_report = create(:scheduled_report, frequency: "daily", user: owner, report: report)
+
+      proposals = 3.times.map do |i|
+        tcr = create(:test_client_request, project_title: "something #{i}")
+        tcr.proposal.update(requester: owner)
+        tcr.proposal.reindex
```
jessieay: do we need to call `reindex` if we are refreshing the search index below?

spec/models/report_spec.rb
```diff
@@ -36,4 +36,29 @@
     user = create(:user, client_slug: "test")
     expect(Report.sql_for_user(user)).to include "WHERE client_slug='#{user.client_slug}'"
   end
+
+  it "#run", :elasticsearch do
```
jessieay: I think we've been including method names in the `describe` block in most specs and then using `it` to describe what the method does -- do you think we should move away from that here?

db/migrate/20160307161402_add_scheduled_reports.rb
```diff
@@ -0,0 +1,13 @@
+class AddScheduledReports < ActiveRecord::Migration
+  def change
+    create_table :scheduled_reports do |t|
+      t.string :name, null: false
+      t.string :frequency, null: false
```
jessieay: did you consider using AR enums for this? would give us all the boolean methods (`weekly?` etc) for free

app/services/scheduled_reporter.rb
```diff
@@ -0,0 +1,56 @@
+class ScheduledReporter
+  attr_reader :check_time
+
+  def initialize(check_time)
```
jessieay: interested in why you chose to pass in time rather than look at current time within this class - assume we will always be using current time and could mock time with Timecop in specs, but there are probably other reasons for doing this that I am not thinking of.

jessieay:
agreed on code climate -- all seems minor / false.

app/admin/scheduled_report.rb
pkarman: good call. added in 426ca09

spec/mailers/report_mailer_spec.rb
pkarman: `reindex` pushes content to the ES server. 

`refresh` forces the ES server to re-open all its searcher objects to make recent updates accessible. https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-refresh.html

spec/models/report_spec.rb
pkarman: fixed in 88a1165

db/migrate/20160307161402_add_scheduled_reports.rb
pkarman: I did consider an enum-style integer definition, or even a `char(1)` (since it takes less storage).

I did want to eventually support full crontab-style frequency values though, for arbitrary schedules, and that would require a string. I.e. "send this to me on MWF" or "every other week" or "every day at noon".

app/services/scheduled_reporter.rb
pkarman: I didn't assume we would always use current time. I.e., the clockwork job misfires and we need to manually run all reports for a given day. Also makes testing more explicit (no need for Timecop).

pkarman:
@jessieay thanks for the review. I think I responded to everything.

db/migrate/20160307161402_add_scheduled_reports.rb
jessieay: wouldn't it be possible to do that with AR enum? the enum vals would be long but it seems like working with strings that act like enums would be slightly hairier than working with enums.

app/services/scheduled_reporter.rb
jessieay: makes sense!

jessieay:
@pkarman one final question https://github.com/18F/C2/pull/1019#discussion_r55398445

otherwise looks :heart: :heart_eyes: :ship:

db/migrate/20160307161402_add_scheduled_reports.rb
pkarman: perhaps I am not understanding how AR enums work.

I expect that an enum is an array of possible integer values, that are mapped to string constants in the AR class. If I wanted to save an arbitrary crontab string like `0 1 3 * 2` how would I support that with an enum?

jessieay: yah your understanding is correct I guess I was imagining that for the cases you mentioned above you'd have preset values:

```
enum frequency: { daily: 0, weekly: , monthly: 2, every_other_week: 3, monday_wednesday_friday: 4 }
```

this isn't ideal, but I guess I am not sure how storing these values as a string is any better?

pkarman:
@jessieay ok I'll convert this to an enum for the 3 known values, and if users ever need more manual control over schedule frequency, it should not be hard to migrate the db at that point.

pkarman:
@jessieay enum in f616a44

jessieay:
looks good on dev!

PR #1021, opened 2016-03-07
make sure we have email address via myusa

pkarman:
trello: https://trello.com/c/lkYMVM4H/184-unchecking-email-share-in-myusa-authorization-results-in-horrible-rails-error

I am not sure how to test this because (a) it seems to require mocking MyUSA more thoroughly than we do and (b) I could not duplicate the error locally.

### DIFF
--- a/app/controllers/auth_controller.rb
+++ b/app/controllers/auth_controller.rb
@@ -9,7 +9,7 @@ class AuthController < ApplicationController
       try_user_auth(auth, return_to_path)
     rescue EmailRequired => error
       Rails.logger.error(error)
-      render :failure, status: 400
+      render :failure, status: 200
     end
   end
 
--- a/app/controllers/auth_controller.rb
+++ b/app/controllers/auth_controller.rb
@@ -5,10 +5,14 @@ class AuthController < ApplicationController
   def oauth_callback
     auth = request.env["omniauth.auth"]
     return_to_path = fetch_return_to_path
-    do_user_auth(auth)
-    session[:token] = auth.credentials.token
-    flash[:success] = "You successfully signed in"
-    redirect_to return_to_path || proposals_path
+    begin
+      do_user_auth(auth)
+      session[:token] = auth.credentials.token
+      flash[:success] = "You successfully signed in"
+      redirect_to return_to_path || proposals_path
+    rescue => _error
+      render :failure
+    end
   end
 
   def failure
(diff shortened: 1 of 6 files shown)


app/controllers/auth_controller.rb
```diff
@@ -5,10 +5,14 @@ class AuthController < ApplicationController
...
-    do_user_auth(auth)
-    session[:token] = auth.credentials.token
-    flash[:success] = "You successfully signed in"
-    redirect_to return_to_path || proposals_path
+    begin
+      do_user_auth(auth)
+      session[:token] = auth.credentials.token
+      flash[:success] = "You successfully signed in"
+      redirect_to return_to_path || proposals_path
+    rescue => _error
```
jessieay: is it possible to rescue a particular error rather than any error? Ideally we do not rescue all errors so we can surface errors we care about and rescue those we want to ignore.

jessieay:
Another way of solving this would be to check for presence of the email string in the `User.for_email` method - if there is no email passed in, then we return a helpful error message?

app/controllers/auth_controller.rb
pkarman: since I could not duplicate the error, I wasn't sure what error to capture.

I did just change in 9132202 to at least log the error.

pkarman:
@jessieay thanks for review.

I added tests for `User.for_email` and the oauth callback for when email is missing, with a custom Exception class. Hopefully that at least captures that specific case.

app/controllers/auth_controller.rb
jessieay: oh ok! I could duplicate the error- when I get the myusa screen and un select "email" on the permissions, I get the same error as the screenshot in the trello card

jessieay:
@pkarman given the instructions to reproduce ( https://github.com/18F/C2/pull/1021#discussion_r55394101) do you want to take another pass at this? From your comment above I am guessing that you wanted to log the error because you couldn't reproduce but I have been able to reproduce so it seems we *do* know the source of this error...

pkarman:
@jessieay yes, I followed those instructions, and just re-tried. I still cannot reproduce the error locally, or in dev.

I also can't reproduce the myusa bug with getting re-prompted every time, so there must be something in how I am set up that is working differently.

If you are able to reproduce, can you confirm that this PR actually fixes the problem?

jessieay:
@pkarman good idea, will do that. My usa is down but will QA when it is back

jessieay:
weird! just went through flow on branch and got different error:

screen shot 2016-03-08 at 11 19 05 am

screen shot 2016-03-08 at 11 19 11 am

pkarman:
@jessieay odd. Maybe c8d3e59 will help?

jessieay:
update works - mergin' :ship:

PR #1022, opened 2016-03-07
Adding emergency requested email

rememberlenny:
https://trello.com/c/vj5xTBYA/122-6-email-conversion-requester-emergency-request

--

# note 

I struggled with this for a while, but there is no way to make a two column desktop and one column mobile layout inside the panel element, without making it unstable for email browsers.

emer-gmail

emer-ios

emer-out


jessieay:
this looks excellent - woo hoo!

PR #1023, opened 2016-03-08
Email/approved by one

rememberlenny:
https://trello.com/c/VQLjzOh1/170-2-email-conversion-requester-approved-by-one-approver

## Note: Need to decide on which visual element communicates the approved/current/past the best

screen shot 2016-03-07 at 11 40 36 pm


jessieay:
I think I like the top image - without check mark, hard to tell what was done (although text does clarify this). Eager to hear what others think

app/mailers/step_mailer.rb
```diff
@@ -3,6 +3,15 @@ def step_reply_received(step)
     @proposal = step.proposal.decorate
     assign_threading_headers(@proposal)
     @step = step
+    last_approved_user = nil
+    
+    # Iterate through steps to find the LAST 'approved' person
```
jessieay: If the logic of the subsequent code is hard to parse, maybe put it in a well-named private method? Eg `last_step_user` (remember, these people are not always approvers! :smile: )

app/mailers/step_mailer.rb
```diff
@@ -3,6 +3,15 @@ def step_reply_received(step)
     @proposal = step.proposal.decorate
     assign_threading_headers(@proposal)
     @step = step
+    last_approved_user = nil
```
jessieay: let's call this `last_step_user` since the step is not always an approval step!

app/mailers/step_mailer.rb
```diff
@@ -3,6 +3,15 @@ def step_reply_received(step)
...
+    
+    # Iterate through steps to find the LAST 'approved' person
+    @proposal.individual_steps.each_with_index do |step, index|
+      if step.status == 'approved'
+        last_approved_user = User.find step.user_id
+      end
+    end
```
jessieay: this loop could be re-written as:

```ruby
@proposal.individual_steps.map do |step|
  if step.status == 'approved'
    step.user
  end
end.last
```

jessieay: or better yet:

```ruby
last_completed_step = @proposal.individual_steps.select { |step| step.status == "approved" }.last
last_completed_step_user = last_completed_step.user

```

app/views/mail_shared/_email_reply.html.haml
```diff
@@ -1,16 +0,0 @@
-- # pass in show_approval_actions bool and an approval instance
```
jessieay: yaaay so happy we are getting rid of these old partials :clap:

app/views/mail_shared/call_to_action/_email_reply.html.haml
```diff
@@ -0,0 +1,9 @@
+- # pass in show_approval_actions bool and an approval instance
```
jessieay: do we need this comment?

app/views/mail_shared/call_to_action/_email_reply.html.haml
```diff
@@ -0,0 +1,9 @@
...
+- if show_step_actions
+  = render partial: "mail_shared/call_to_action/button", locals: { link_path: generate_approve_url(step),
```
jessieay: I find it a bit easier to read if the `locals` arg is put on the next line and then all locals are on the same line. What do you think?

eg:

now: 

```ruby
render partial: "mail_shared/call_to_action/button", locals: { link_path: generate_approve_url(step),
                       link_text: step.action_name}
```
vs:

```ruby
render partial: "mail_shared/call_to_action/button",
  locals: { link_path: generate_approve_url(step), link_text: step.action_name}
```

app/mailers/step_mailer.rb
rememberlenny: I love these iterations

jessieay: ruby :heart:

jessieay: additional thought: could include this logic in `app/models/concerns/step_manager.rb` (which we use in `Proposal`) so we could have a unit test for this

PR #1027, opened 2016-03-08
Search UI fixes

pkarman:
trello: https://trello.com/c/nDSxYIGp/20-as-a-user-i-have-the-ability-to-perform-an-advanced-search-based-on-attributes-of-my-request-details

screen shot 2016-03-08 at 12 16 30 pm

### DIFF
--- a/app/views/shared/_search_ui.html.haml
+++ b/app/views/shared/_search_ui.html.haml
@@ -11,26 +11,26 @@
               %form.adv-search.form-horizontal{role: "form", action: query_proposals_path, method: "get"}
                 %fieldset
                   .form-group
-                    %label{for: "#{current_user.client_model_slug}[status]"}
+                    %label.control-label{for: "#{current_user.client_model_slug}[status]"}
                       Search in
-                    %select.form-control{name: "#{current_user.client_model_slug}[status]"}
+                    %select.short.form-control{name: "#{current_user.client_model_slug}[status]"}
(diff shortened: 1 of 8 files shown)


app/views/shared/_search_ui.html.haml
```diff
@@ -6,59 +6,9 @@
...
-                    .col-xs-5
-                      %label{for: "#{current_user.client_model_slug}[created_at]"}
-                        of
-                      %input.form-control{ type: "date",
-                        name: "#{current_user.client_model_slug}[created_at]", 
-                        value: @adv_search.value_for("created_at"),
-                      }
-                  %button#adv-search-button.btn.btn-primary{type: "submit"}
-                    Search
+              = render partial: "shared/search_form"
```
jessieay: niiice this makes this view much easier to read / parse

jessieay:
code looks good - QA instructions?

pkarman:
most of this is just UI tweeks, so only QA would be to make sure it looks good to you.

the only new feature is the "includes attachments" checkbox. You'll need to re-index to see that feature in action. Checking it will include proposals that have 1 or more attachments.

jessieay:
`rebuild_index` is broken on master so I will need to re-QA this when that is fixed (since the attachment search can only be tested with a rebuilt index)

jessieay:
@pkarman can you rebase this against master so I can test it with the updates in `rebuild_index` that are on `master`

PR #1028, opened 2016-03-08
Render attachment preview in email

jessieay:
* https://trello.com/c/e4FmHzbx/191-fix-the-email-attachments-and-create-condition-for-file-image-or-file-link-based-on-uploads

screen shot 2016-03-08 at 11 28 08 am

### DIFF
--- a/app/decorators/attachment_decorator.rb
+++ b/app/decorators/attachment_decorator.rb
@@ -5,9 +5,9 @@ class AttachmentDecorator < Draper::Decorator
 
   def file_preview
     if file.content_type =~ /\Aimage/
-      image_tag(file.url, alt: "", class: "image-with-border")
+      link_to image_tag(file.url, alt: "", class: "image-with-border"), file.url 
     else
-      link_to(file.original_filename, file.url)
+      return '<br><table class="button"><tr><td>' + link_to('Click to view ' + file.original_filename, file.url) + '</td></tr></table>'
     end
   end
 end
--- a/app/decorators/attachment_decorator.rb
+++ b/app/decorators/attachment_decorator.rb
@@ -1,0 +1,13 @@
+class AttachmentDecorator < Draper::Decorator
+  include ActionView::Helpers::AssetTagHelper
+  include ActionView::Helpers::UrlHelper
+  delegate_all
+
+  def file_preview
+    if file.content_type =~ /\Aimage/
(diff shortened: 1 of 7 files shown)


jessieay:
via Lenny: outlook doesnt listen to “width” css statements, so if possible, can the image be resized to be 400px max width?

jessieay:
@rememberlenny 

400px width still looks funky

screen shot 2016-03-08 at 12 23 34 pm

also not sure the PDF Url is obvious:

screen shot 2016-03-08 at 12 22 45 pm

design advice?

jessieay:
@rememberlenny code is updated with correct logic. design still slightly off (see screenshots in my last comment)

rememberlenny:
@jessieay Can I pick it up from here?

app/decorators/attachment_decorator.rb
```diff
@@ -0,0 +1,13 @@
+class AttachmentDecorator < Draper::Decorator
+  include ActionView::Helpers::AssetTagHelper
+  include ActionView::Helpers::UrlHelper
+  delegate_all
+
+  def file_preview
+    if file.content_type =~ /\Aimage/
+      link_to image_tag(file.url, alt: "", class: "image-with-border"), file.url 
+    else
+      return '<br><table class="button"><tr><td>' + link_to('Click to view ' + file.original_filename, file.url) + '</td></tr></table>'
```
jessieay: don't need explicit `return` statement here

jessieay: we are still linking to the file rather than the request -- is that what you wanted?

jessieay: let's use a translation rather than text :cherries:

jessieay:
@rememberlenny can you put screenshots of updating designs in here? screenshot in Slack seems to have updated copy so would be good to have latest in here

jessieay:
@rememberlenny updated code so that image preview and PDF button link to proposal for the attachment.

final thing to address here (and I am not sure how to) is that the button turns blue on hover but text does not turn white:

screen shot 2016-03-08 at 2 51 47 pm

the buttons below in the email are working correctly w/r/t color on hover so figure it is something small

rememberlenny:
@jessieay Fixed up

screen shot 2016-03-09 at 11 47 08 am

PR #1029, opened 2016-03-09
A finished proposal is `completed`, not `approved`

jessieay:
* Update status of proposal and steps to be `complete`
* Rename `cancelled` status to `canceled`
* Keep `approve` path so links in old emails work, but add `complete`
  path in routes


jessieay:
great renaming tour of 2015-2016 continues :guitar: cc @yozlet

jessieay:
@pkarman I did a bunch of local QA for this to make sure everything still works - everything, including link in email, seems to work a-ok. but this should probably receive slightly more rigorous QA than usual since it is such a large / impactful change

yozlet:
Yay, thank you @jessieay!

db/schema.rb
```diff
@@ -212,7 +212,7 @@
     t.integer  "client_data_id"
     t.string   "client_data_type", limit: 255
     t.integer  "requester_id"
-    t.string   "public_id",        limit: 255
```
pkarman: I think this keeps happening because I am running my migrations against a copy of the prod db, which has the explicit `limit` declared.

jessieay: ah yes will get checkout

jessieay: or should I? I am not sure which of us should change...

pkarman: imo, the prod db is canonical. My guess is that we have different versions of psql in cloud.gov vs local dev and that is throwing off the syntax slightly.

My suggestion: compare your local psql schema against this:
```
capc2-prod=# \d proposals
                                        Table "public.proposals"
      Column      |            Type             |                       Modifiers                        
------------------+-----------------------------+--------------------------------------------------------
 id               | integer                     | not null default nextval('proposals_id_seq'::regclass)
 status           | character varying(255)      | 
 created_at       | timestamp without time zone | 
 updated_at       | timestamp without time zone | 
 client_data_id   | integer                     | 
 client_data_type | character varying(255)      | 
 requester_id     | integer                     | 
 public_id        | character varying(255)      | 
 visit_id         | uuid                        | 
Indexes:
    "proposals_pkey" PRIMARY KEY, btree (id)
    "index_proposals_on_client_data_id_and_client_data_type" btree (client_data_id, client_data_type)
Foreign-key constraints:
    "fk_rails_c557d6413c" FOREIGN KEY (visit_id) REFERENCES visits(id)
    "requester_id_fkey" FOREIGN KEY (requester_id) REFERENCES users(id)
Referenced by:
    TABLE "attachments" CONSTRAINT "proposal_id_fkey" FOREIGN KEY (proposal_id) REFERENCES proposals(id)
    TABLE "comments" CONSTRAINT "proposal_id_fkey" FOREIGN KEY (proposal_id) REFERENCES proposals(id)
    TABLE "proposal_roles" CONSTRAINT "proposal_id_fkey" FOREIGN KEY (proposal_id) REFERENCES proposals(id)
    TABLE "steps" CONSTRAINT "proposal_id_fkey" FOREIGN KEY (proposal_id) REFERENCES proposals(id) ON DELETE CASCADE
```

if they are identical, then there is something else going on. Otherwise, run the appropriate `alter` sql commands to get your local db in sync with the schema above, and then run `rake db:schema:dump` and `git diff db/schema.rb` and see if anything shows up. If nothing differs, then we have some other problem.

It's not a big deal; just wanted to note somewhere that I keep seeing that noise in the PRs for schema.db and why it is happening.

pkarman:
I will QA

pkarman:
QA results:

* https://trello.com/c/AWeO7Cy2 (possibly unrelated)
* clicking `Approve` button via web UI throws 500 error:
```
2016-03-10T11:17:37.86-0600 [App/0]      OUT 17:17:37 web.1    | Workflow::NoTransitionAllowed (There is no event complete defined for the completed state):
2016-03-10T11:17:37.86-0600 [App/0]      OUT 17:17:37 web.1    |   app/controllers/proposals_controller.rb:54:in `complete'
```

I also note that single-click Approve from an email does not work. I suspect this has nothing to do with the renaming but noting here for completeness.

pkarman:
also, for backcompat with existing emails-in-the-wild, probably need to whitelist the `approve` action for use with the auth token:

```diff
diff --git a/app/controllers/proposals_controller.rb b/app/controllers/proposals_controller.rb
index 7cc8e04..ffb07df 100644
--- a/app/controllers/proposals_controller.rb
+++ b/app/controllers/proposals_controller.rb
@@ -1,12 +1,12 @@
 class ProposalsController < ApplicationController
   include TokenAuth
 
-  skip_before_action :authenticate_user!, only: [:complete]
-  skip_before_action :check_disabled_client, only: [:complete]
+  skip_before_action :authenticate_user!, only: [:approve, :complete]
+  skip_before_action :check_disabled_client, only: [:approve, :complete]
   # TODO use Policy for all actions
   before_action -> { authorize proposal }, only: [:show, :cancel, :cancel_form, :history]
-  before_action :needs_token_on_get, only: :complete
-  before_action :validate_access, only: :complete
+  before_action :needs_token_on_get, only: [:approve, :complete]
+  before_action :validate_access, only: [:approve, :complete]
   helper_method :display_status
   add_template_helper ProposalsHelper
   rescue_from Pundit::NotAuthorizedError, with: :auth_errors
```

jessieay:
@pkarman thanks for QA! Good call on whitelisting.

I *was* able to get link in email to work (clicked locally via letter_opener) -- what was the error you saw?

jessieay:
@pkarman what did you do to trigger the error snippet in https://github.com/18F/C2/pull/1029#issuecomment-194964180 ? I just used Rails console to find an approval step with a status of `actionable` and called `complete!` and it worked so it seems the even of `complete!` works in that case but not sure what setup you had there.

jessieay:
@pkarman updated controller at e924705 for backcompat with existing emails

pkarman:
I got that error via `cf logs c2-dev` when I clicked on the `Approve` (https://c2-dev.18f.gov/proposals/287/complete) button.

jessieay:
Having a hard time reproducing on c2-dev-jessiay

was able to complete step via email link and web ui link (was doing both as the first approve for an NCR request)

logs from web UI approve action:

```
2016-03-10T11:05:49.02-0800 [App/0]      OUT 19:05:49 web.1    | Started POST "/proposals/6/complete?method=POST" for 159.142.71.2 at 2016-03-10 19:05:49 +0000
2016-03-10T11:05:49.02-0800 [App/0]      OUT 19:05:49 web.1    | Processing by ProposalsController#complete as HTML
2016-03-10T11:05:49.02-0800 [App/0]      OUT 19:05:49 web.1    |   Parameters: {"utf8"=>"✓", "authenticity_token"=>"lcTBAtZYLSnQX2nL8Dlh0RsFUl5un9IDoxxQAiIgeCRrcbnsNvmacPB17800RtIMFhyfaU757QjkYIzhNK6haQ==", "version"=>"1457636669", "commit"=>"Approve", "method"=>"POST", "id"=>"6"}
```

logs from email link approve action:

```
2016-03-10T11:02:28.43-0800 [App/0]      OUT 19:02:28 web.1    | Started GET "/proposals/5/complete?cch=tuU7gUbVbYLQrjCZ2HxeaQy8&scheme=https&version=1457636518" for 159.142.71.2 at 2016-03-10 19:02:28 +0000
2016-03-10T11:02:28.44-0800 [App/0]      OUT 19:02:28 web.1    | Processing by ProposalsController#complete as HTML
2016-03-10T11:02:28.44-0800 [App/0]      OUT 19:02:28 web.1    |   Parameters: {"cch"=>"tuU7gUbVbYLQrjCZ2HxeaQy8", "scheme"=>"https", "version"=>"1457636518", "id"=>"5"}
2016-03-10T11:02:28.58-0800 [App/0]      OUT 19:02:28 web.1    | [ActiveJob] Enqueued ActionMailer::DeliveryJob (Job ID: bce6c165-5d89-44b4-84b7-84d44f92b6da) to DelayedJob(mailers) with arguments: "Mailer", "actions_for_approver", "deliver_now", gid://c2/Steps::Approval/19
```

any other ideas?

jessieay:
@pkarman I see -- if we look for an existing step that is *not* complete, that might get is into a weird case where we get `nil` for `proposal.existing_or_delegated_step_for(current_user)` and then try to call `update_attributes!` on that. so in the case above that you described, there would still be an error.

what about ensuring that delegates are not system approvers? seems like that is a bit out of scope for this PR but I think it makes sense to not allow users to be system approvers and delegates for system approvers...

pkarman:
I don't want to wade into the business rules of what kind of relationships users can have with one another. There is just enough rope to hang yourself with our current delegate system, and I seem to have uncovered an edge case. 

A more straightforward approach in the case of a `complete` action is to verify that the step you want to act upon actually needs to be completed. So this in the controller makes more sense to me:

```ruby
  def complete
    step = proposal.existing_or_delegated_actionable_step_for(current_user)
    step.update_attributes!(completer: current_user)
    step.approve!
    flash[:success] = "You have approved #{proposal.public_id}."
    redirect_to proposal
  end
```

where `existing_or_delegated_actionable_step_for` looks like

```ruby
  def existing_or_delegated_actionable_step_for(user)
    where_clause = <<-SQL
      status = 'actionable'
      user_id = :user_id
      OR user_id IN (SELECT assigner_id FROM user_delegates WHERE assignee_id = :user_id)
      OR user_id IN (SELECT assignee_id FROM user_delegates WHERE assigner_id = :user_id)
    SQL
    steps.where(where_clause, user_id: user.id).first
  end
```

jessieay:
@pkarman agree with your logic. in the code above, `step.update_attributes!(completer: current_user)` will return an `undefind method` error since `step` will be `nil` if the step in question is not actionable. Do we want that? I know historically we had an error msg somewhere that told the person a step was already approved but cannot find that code now

pkarman:
I think there should be a sanity check after fetching the step and respond to the user with a 400 error and nice message about "you can't approve this proposal".

the current code is a bit naive to expect that someone might try and approve a step only once or "out of order" or (as in this case) where the current_user might be involved somehow in multiple steps.

pkarman:
yes let's handle it separately. I will merge.

PR #1030, opened 2016-03-09
Add spec for `.rebuild_index` in Searchable module

jessieay:


jessieay:
@pkarman failure here is same one I am seeing locally

spec/models/concerns/searchable_spec.rb
```diff
@@ -0,0 +1,7 @@
+describe Searchable do
+  describe ".rebuild_index" do
+    it "should rebuild the index" do
```
pkarman: should probably `create(:proposal)` here at least once so there is something to index.

pkarman:
elasticsearch-rails-ha (1.0.7) will fix issue when no Proposals are present in the db.

however, if you really want to test more real-world, try creating some proposals before indexing.

jessieay:
@pkarman argh updated gem but still seeing error locally:

``` 
1) Searchable.rebuild_index should rebuild the index
     Failure/Error: expect(Proposal.rebuild_index).not_to raise
     RuntimeError:
     # ./spec/models/concerns/searchable_spec.rb:5:in `block (3 levels) in <top (required)>'
```

can revisit tomorrow or happy to have you check it out too since you are the search :crown:

pkarman:
@jessieay I *think* this latest error was just the way the rspec was written. See da1e577

jessieay:
@pkarman smart! Looks great will QA on dev after my cofffee break

jessieay:
rebuild job now runs on dev! yeahhhhhh

PR #1033, opened 2016-03-09
🎉 StepMailer#proposal_notification

rememberlenny:
https://trello.com/c/60fsK5PN/147-4-email-conversion-approver-new-request-for-approval

fireshot capture 6 - c2 - http___localhost_5000_rails_mailers_step_mailer_proposal_notification_par

### DIFF
--- a/app/mailers/proposal_mailer.rb
+++ b/app/mailers/proposal_mailer.rb
@@ -35,9 +35,11 @@ class ProposalMailer < ApplicationMailer
     )
   end
 
-  def proposal_updated_no_action_required(user, proposal, modifier = nil)
+  def proposal_updated_no_action_required(user = User.last, proposal = Proposal.last, modifier = nil)
     @proposal = proposal.decorate
     @modifier = modifier || NullUser.new
+    # Example ideal data to consume
+    @changes = [{verb: 'modified', result: 'this to that'},{verb: 'added', result: 'that'},{verb: 'modified', result: 'this to that'}]
     assign_threading_headers(@proposal)
 
     mail(
--- a/app/mailers/proposal_mailer.rb
+++ b/app/mailers/proposal_mailer.rb
@@ -35,7 +35,7 @@ class ProposalMailer < ApplicationMailer
     )
   end
 
-  def proposal_updated_no_action_required(user = User.last, proposal = Proposal.last, modifier = nil)
+  def proposal_updated_no_action_required(user, proposal, modifier = nil)
     @proposal = proposal.decorate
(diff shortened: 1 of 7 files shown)


jessieay:
@rememberlenny can you comment in the related Trello cards on the elements here that need to be updated after this is merged ? eg: `step_status_html` locale is static, needs to be dynamic.

app/mailers/proposal_mailer.rb
```diff
@@ -35,9 +35,11 @@ def proposal_complete(proposal)
     )
   end
 
-  def proposal_updated_no_action_required(user, proposal, modifier = nil)
+  def proposal_updated_no_action_required(user = User.last, proposal = Proposal.last, modifier = nil)
```
jessieay: we don't want to set default values for this. Since it is the actual email, we want to make sure we are always passing in the `user` and `proposal` . we have a default value of `nil` for modifier because sometimes will update a proposal via `rails console` in which case there will be no recorded modifier.

app/views/step_mailer/proposal_notification.text.erb
```diff
@@ -2,6 +2,8 @@
     requester_name: @proposal.requester.full_name,
     step_type_noun: @step.noun) %>
 
+<%= t("mailer.step_mailer.proposal_notification.step_status_html") %>
```
jessieay: if this locale has `html` in it, we want to use something else for the text-only version of the email

jessieay: if it doesn't, we should rename and remove `_html` from the name!

app/views/proposal_mailer/proposal_updated_no_action_required.html.haml
```diff
@@ -1,25 +1,92 @@
...
+    locals: { subheader: panel_preface }
+
+  = render partial: "mail_shared/panel/header_name", 
+                  locals: { name: @proposal.name,
+                            subtext: @proposal.public_id }
+
+  = render partial: "mail_shared/panel/proposal_attributes",
+    locals: { proposal: @proposal }
+
+  - if !@proposal.attachments.nil?
```
jessieay: seems like all of this logic is exactly the same as the attachment section in `app/views/step_mailer/proposal_notification.html.haml` - put in a shared partial?

app/mailers/proposal_mailer.rb
rememberlenny: Shoot- that was for testing :) I'll fix!

--
Leonard Bogdonoff
949-322-8496 | @rememberlenny

PR #1036, opened 2016-03-10
Add comment text to Proposal update emails

jessieay:
* Add specs for `Ncr::WorkOrderUpdater`, too

screen shot 2016-03-09 at 5 40 56 pm

https://trello.com/c/nbI2mxDU/150-10-email-conversion-approver-modified-request-for-an-awaiting-approval

### DIFF
--- a/app/controllers/client_data_controller.rb
+++ b/app/controllers/client_data_controller.rb
@@ -44,9 +44,9 @@ class ClientDataController < ApplicationController
 
   def update_or_notify_of_no_changes
     if attribute_changes?
-      record_changes
+      comment = record_changes
       @client_data_instance.save
-      setup_and_email_approvers
+      setup_and_email_approvers(comment)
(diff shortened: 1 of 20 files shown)


jessieay:
Note that this design is slightly different from the mocks. I spoke with Amber about this. Since our current update comments have this format, we decided to stick with it. If we want to change how update text looks in email, we should update the `comment.comment_text` data to look the same so we can be consistent.

jessieay:
updates to comment only version: 

screen shot 2016-03-09 at 5 37 32 pm

rememberlenny:
@jessieay Tested in browser runner. 

Looks good for first pass.

iphone6s-vertical-allowed-1366
chromegmailnew-vertical-allowed-1366
ol2013-vertical-allowed-1366

jessieay:
This should be green now (needed a small update in specs 1a605aa0

@rememberlenny can you review ?

PR #1037, opened 2016-03-10
Setup tracking GUI for admin

rememberlenny:
https://trello.com/c/nLQACF4D/197-create-gui-to-review-tracking-data-from-ahoy-on-production

Blazer was recently accepted at PR that allows us to customize the authentication. Now that we can authenticate the user's admin status, we can expose Blazer to the production environment.

[FTR, Blazer is awesome](https://github.com/ankane/blazer)

screen shot 2016-03-09 at 11 59 00 pm

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -2,5 +2,9 @@ dependencies:
   pre:
     - cp .env.example .env
     - bash ./build-elasticsearch.sh
+    - sudo apt-get update; sudo apt-get install libicu52
+    - curl --output /home/ubuntu/bin/phantomjs-2.0.1-linux-x86_64-dynamic https://s3.amazonaws.com/circle-support-bucket/phantomjs/phantomjs-2.0.1-linux-x86_64-dynamic
+    - chmod a+x /home/ubuntu/bin/phantomjs-2.0.1-linux-x86_64-dynamic
+    - sudo ln -s --force /home/ubuntu/bin/phantomjs-2.0.1-linux-x86_64-dynamic 
   cache_directories:
     - elasticsearch
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -63,8 +63,8 @@ C2::Application.routes.draw do
   end
 
   mount Peek::Railtie => "/peek"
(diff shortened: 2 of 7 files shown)


pkarman:
@rememberlenny looks like failing tests.

rememberlenny:
@pkarman got it to go green!

rememberlenny:
@pkarman Thank you!

pkarman:
screen shot 2016-03-10 at 12 34 57 pm

pkarman:
@rememberlenny you might need to add a `config/initializers/blazer.rb` that sets `BLAZER_DATABASE_URL` to `DATABASE_URL` -- since the latter is defined explicitly at cloud.gov.

rememberlenny:
@pkarman Updated!

pkarman:
@rememberlenny not sure that phantomjs change actually helps -- I just manually restarted circleci after another `DeadClient` error.

pkarman:
@rememberlenny I still get `BLAZER_DATABASE_URL required` error here https://c2-dev.18f.gov/blazer/

`c2-dev` is available to test with, if you need.

circle.yml
```diff
@@ -2,5 +2,9 @@ dependencies:
   pre:
     - cp .env.example .env
     - bash ./build-elasticsearch.sh
+    - sudo apt-get update; sudo apt-get install libicu52
```
pkarman: I don't think upgrading phantomjs is helping, frankly. I still see DeadClient errors occasionally.

another approach e.g. https://github.com/18F/C2/pull/1043

config/routes.rb
```diff
@@ -63,8 +63,8 @@
   end
 
   mount Peek::Railtie => "/peek"
+  mount Blazer::Engine, at: "blazer"
   if Rails.env.development?
     mount LetterOpenerWeb::Engine => "letter_opener"
-    mount Blazer::Engine, at: "blazer"
```
pkarman: let's leave this as a dev-only feature, and copy prod db to local env when needed. that keeps blazer security issues to a minimum.

PR #1038, opened 2016-03-10
add css class fix for login alert

pkarman:
I know I saw a support ticket for this, could not find a trello card.

just happened to me on dev. simple CSS bug.

screen shot 2016-03-10 at 10 26 36 am


jessieay:
niiiiice :thought_balloon: :+1:

PR #1041, opened 2016-03-10
allow delegates to edit ncr work orders

pkarman:


jessieay:
@pkarman spec failure at `rspec ./spec/features/ncr/work_orders/edit_spec.rb:2` (simple text failure -- we should put this in a locale to prevent this issue in future)

pkarman:
thanks @jessieay -- db01062 adds a start for using locales. I will create a card for moving all policies to use them.

PR #1042, opened 2016-03-10
Validate step is actionable before completing it

pkarman:
trello: https://trello.com/c/K3Qppj5S/209-validate-step-before-completing-it

### DIFF
--- a/app/controllers/proposals_controller.rb
+++ b/app/controllers/proposals_controller.rb
@@ -5,8 +5,8 @@ class ProposalsController < ApplicationController
   skip_before_action :check_disabled_client, only: [:approve, :complete]
   # TODO use Policy for all actions
   before_action -> { authorize proposal }, only: [:show, :cancel, :cancel_form, :history]
-  before_action :needs_token_on_get, only: :complete
-  before_action :validate_access, only: :complete
+  before_action :needs_token_on_get, only: [:approve, :complete]
+  before_action :validate_access, only: [:approve, :complete]
   helper_method :display_status
   add_template_helper ProposalsHelper
   rescue_from Pundit::NotAuthorizedError, with: :auth_errors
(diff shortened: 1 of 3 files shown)


app/controllers/proposals_controller.rb
```diff
@@ -49,10 +49,12 @@ def approve
...
   def complete
-    step = proposal.existing_or_delegated_step_for(current_user)
-    step.update_attributes!(completer: current_user)
-    step.complete!
-    flash[:success] = "You have approved #{proposal.public_id}."
+    step = proposal.existing_or_delegated_actionable_step_for(current_user)
+    if step
+      complete_step(step)
+    else
+      flash[:alert] = "There are no actionable steps available for you."
```
jessieay: should we have a feature spec that confirms this flash shows on the page?

pkarman: Unfortunately (or not) there's currently no way to access that code path, because we've also changed the step WHERE clause to include `status=actionable`.

Every spec I attempted to construct would instead hit this special exception check:

https://github.com/18F/C2/blob/master/app/controllers/concerns/token_auth.rb#L35

which means the authz check failed via Policy, before it ever attempted to fetch the step in this controller method.

jessieay: doesn't that mean that we should never git this case in the controller in real life? my instinct is that if we can't construct a test to mimic this situation, we can't mimic IRL either? or am I missing something?

pkarman: I guess I think of it as a belt-and-suspenders, defensive programming thing. If e.g.

https://github.com/18F/C2/blob/master/app/controllers/concerns/token_auth.rb#L35

ever changes (and it was a hack, to be sure), then it's quite possible we find IRL situations that exercise the code path where `step` is nil. Just from a code reading POV, when I look at the "before" version I get a little antsy when I ask myself "what happens if that method fails to return a step?" because then there's an instant 500 error for the user.

pkarman: maybe another way to say it: I try not to assume the happy path.

jessieay: makes sense :+1:

PR #1043, opened 2016-03-10
Turn off parallel circleci

pkarman:
This can be merged as long as we're comfortable with the slightly slower run time. No testing/measuring has been done on how this affects run time.

This is to try and address the periodic failures at CircleCI.


jessieay:
works for me. anything to end the intermittent failures!

PR #1047, opened 2016-03-10
Merge master into email branch

jessieay:


rememberlenny:
😶

jessieay:
@pkarman :green_heart: at long last - ready for a glance over

spec/models/incoming_mail_handler_spec.rb
```diff
@@ -1,9 +1,8 @@
 describe "Handles incoming email" do
```
pkarman: the changes in this file smell funny to me. did master really change so much here?

jessieay: the changes in this file are not purely from the master merge. I had an error in `Dispatcher` after merge where we were looking at `approved?` and then I realized we were still calling the `actions_for_approver` mailer, which should have been replaced with `StepMailer.proposal_notification` so I ended up changing it all at https://github.com/18F/C2/commit/e7a1c5e1fa0077190e7f1410e72dd1c82004c055 -- if you think it better to do that separately I can PR that commit on its own. 

you are right to think this smells funny

pkarman:
@jessieay I basically eye-balled every thing that changed in `spec/*` under the assumption that the changes there are the most important to get correct.

the one thing that raised flags for me was when there was a lot of pink/red indicating removed code in master. My expectation is that master is mostly additive at this point, so I would not expect to see a lot of red, except for spelling changes, etc.

spec/models/incoming_mail_handler_spec.rb
pkarman: ok that makes some sense.

I'm fine with it being part of this PR as long as we have some rationale around why that code changed. Could you maybe annotate the commit just a little so that future us is clear about why all the tests got removed? It seems like a lot more deletion than just replacing `actions_for_approver` with `StepMailer.proposal_notification`

jessieay: I removed the commits for this for now -- will update separately since it was not strictly required for merge and this puppy is already quite large :dog:

PR #1049, opened 2016-03-11
Use `attachment_items` partial everywhere

jessieay:
* Remove old/unused `attachments` partial


pkarman:
@jessieay merge conflicts alas

jessieay:
@pkarman updated!

PR #1050, opened 2016-03-11
Remove ActivityMailer / Update Fiscal Cancelation mailer

jessieay:
* Update fiscal cancelation email to use new styles (no card for that one)
* Remove activity email. Was being used for observer emails but those
  need to be created separately anyway.

fiscal cancelation email:

screen shot 2016-03-10 at 5 19 04 pm


app/models/dispatcher.rb
```diff
@@ -16,16 +16,9 @@ def deliver_new_proposal_emails
       StepMailer.proposal_notification(step).deliver_later
     end
 
-    email_observers("proposal created")
```
pkarman: so we are no longer email observers when proposals are created or completed?

pkarman:
@jessieay the `ActivityMailer` was a generic stopgap -- what is replacing it? Or are we just no longer notifying observers the same way?

jessieay:
Yeah I opened cards for each of these issues. It seems that we do need an
email to observers after a proposal is finished. I labeled that card with
"(25)" since it is the 25th email. I am not as sure about emailing after a
proposal is created because they also get an email when they are added as
an observer so seems like a dupe. I was combing through and checking for
emails without the new new styles and these plus the fiscal cancelation
emails were the last if the bunch. Can keep until we have new designs for
observer email but thought this was one way to go about finalizing
transition

pkarman:
it's fine with me to merge -- conflict atm

jessieay:
@pkarman updated!

PR #1052, opened 2016-03-11
Set up Email 1 w/ fixes across all emails

rememberlenny:
https://trello.com/c/6LxCjA2T/210-1-round-3

# Note

Fix a centering issue across templates on mobile

screen shot 2016-03-11 at 2 39 25 pm
screen shot 2016-03-11 at 2 39 18 pm

### DIFF
--- a/app/models/gsa18f/procurement.rb
+++ b/app/models/gsa18f/procurement.rb
@@ -64,6 +64,10 @@ module Gsa18f
       (cost_per_unit * quantity) || 0.0
     end
 
+    def amount
+      total_price
+    end
+
     # may be replaced with paper-trail or similar at some point
     def version
       updated_at.to_i
--- a/app/models/gsa18f/procurement.rb
+++ b/app/models/gsa18f/procurement.rb
@@ -64,10 +64,6 @@ module Gsa18f
       (cost_per_unit * quantity) || 0.0
     end
 
-    def amount
-      total_price
-    end
-
     # may be replaced with paper-trail or similar at some point
     def version
       updated_at.to_i
--- a/app/models/ncr/work_order.rb
+++ b/app/models/ncr/work_order.rb
@@ -134,7 +134,7 @@ module Ncr
     def total_price
       amount || 0.0
     end
-
+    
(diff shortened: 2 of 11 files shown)


jessieay:
@rememberlenny I don't see the approval chain stuff in this email, but it is in the screenshot in the trello card..

app/models/ncr/work_order.rb
```diff
@@ -134,7 +134,7 @@ def not_ba60?
     def total_price
       amount || 0.0
     end
-
+
```
jessieay: :heart_eyes:

pkarman: heh I didn't realize we were already normalizing on `total_price`

let's pick one or the other. creating new clients will make this a PITA to keep sync'd.

app/models/gsa18f/procurement.rb
```diff
@@ -64,6 +64,10 @@ def total_price
       (cost_per_unit * quantity) || 0.0
     end
 
+    def amount
```
jessieay: I might suggest also adding the `amount` method to the `ProposalDecorator` and then calling `@proposal.amount` in the view. you can define `amount` in the decorator like this:

```ruby
def amount
  client_data.try(:amount) || ""
end
```

then you will also need to define `amount` on `WorkOrder` -- the benefit of this is that when a proposal doesn't have any client data (which won't happen in real life but does happen in tests) you won't deal with a `nil` error

jessieay: also see https://github.com/18F/C2/pull/1052/files#r55882651 -- maybe use `total_price` instead?

rememberlenny:
@jessieay Re: approval chain, it was a local issue. Things are alright!

jessieay:
@rememberlenny made some changes here d82d4ec and committed to your branch -- spec problems should be fixed now (issue was in tests that created a proposal without client data...always nice to make things less dependent on each other so created a method on the proposal decorator that returns an empty string when there is no client data)

rememberlenny:
@jessieay Thanks!

PR #1056, opened 2016-03-11
harden ES service detection to provide nicer UX

pkarman:
trello: https://trello.com/c/uFwAucpL/185-search-service-responds-with-502

Creates a custom exception class for when ES service goes AWOL.


app/controllers/proposals_controller.rb
```diff
@@ -61,7 +61,12 @@ def complete
   def query
     check_search_params
     query_listing = listing
-    @proposals_data = query_listing.query
+    begin
```
yozlet: Should we have this in `#download` as well (and possibly the report generation) or is that overkill?

yozlet:
Other than the question above, :+1:

app/controllers/proposals_controller.rb
pkarman: good idea, @yozlet -- added in 2219a96

jessieay:
Looks awesome - mergin' :ship:

PR #1068, opened 2016-03-15
Include next step type when proposal is "pending"

jessieay:
* If next step is approval, status is "pending approval"
* If next step is purchase, status is "pending purchase"

Looks like this in emails:

screen shot 2016-03-15 at 2 58 16 pm

https://trello.com/c/wSUB35wL/237-update-status-in-emails


rememberlenny:
Looks good! Thanks for adding those tests

PR #1069, opened 2016-03-15
Move email styles to stylesheet

jessieay:
* Instead of view partial
* No visible change here, just reorganizing files


rememberlenny:
@jessieay This wont work. I think because you are testing in browsers, you can't tell that the media queries are going to be broken.

https://github.com/18F/C2/blob/jy-move-styles-to-css-file/app/assets/stylesheets/email.css#L1161-L1163

We could do this, but would need to break the styles in to two partials. One with media queries, and the other without. That way, we can do the roadie ignore.

jessieay:
ah ok! will update. thanks for the tip

jessieay:
@rememberlenny 

updated code.

this is what it looks like when you inspect element:

screen shot 2016-03-15 at 3 48 58 pm

how does that naming sound to you? I think the ignored one is for fullscreen, right? or would a different word be better for describing it?

rememberlenny:
@jessieay Its actually reversed! The mobile stylesheets are mobile-only. `email-mobile` would be a better name.

jessieay:
@rememberlenny d'oh! clearly CSS challenged over here. Renamed at 0fd3735

PR #1071, opened 2016-03-16
Pass canceler to cancelation mailer methods

jessieay:
* Previously, only the requester could cancel a proposal
* Now approvers can also cancel a proposal
* So we must pass canceler to mailers to know who gets what email

https://trello.com/c/tsQSwMvz

### DIFF
--- a/app/controllers/proposals_controller.rb
+++ b/app/controllers/proposals_controller.rb
@@ -86,7 +86,7 @@ class ProposalsController < ApplicationController
     comments = "Request canceled with comments: " + params[:reason_input]
     proposal.cancel!
     proposal.comments.create!(comment_text: comments, user: current_user)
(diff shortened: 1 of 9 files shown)


rememberlenny:
Looks good! I dont fully understand the Dispatcher yet, so could we talk about this before I merge it? @jessieay

rememberlenny:
Merged, but still want to understand how this works

jessieay:
@rememberlenny yes let's chat!

In short: `Dispatcher` is the class that sends all of our emails. We have an `NcrDispatcher` as well, which inherits from `Dispatcher`. This allows `NCR`-specific email logic to override the default logic (which it does right now). I didn't create `Dispatcher`, but I am guessing the original intent was to have a class that decide *who* should receive each email, then calls the appropriate mailer.

PR #1073, opened 2016-03-16
Merge `master` into `EMAIL` master

jessieay:


jessieay:
this is so that when we QA today we are testing how the code will be in `master` with `EMAIL` merged into it

jessieay:
going to merge this since it is just a PR which updates merge conflicts with master

PR #1076, opened 2016-03-16
Changes from Amber's design review

rememberlenny:


app/assets/stylesheets/email.css
```diff
@@ -0,0 +1,1198 @@
+table.header {
```
jessieay: remove this? or move any new styles to the view partial in `app/views/mail_shared` since that is what we are including for CSS Now

app/views/comment_mailer/comment_added_notification.html.haml
```diff
@@ -9,7 +9,7 @@
 - cta_subheader = t("mailer.updated_subheader_html",
   public_id: @proposal.public_id,
   requester_name: @proposal.requester.full_name,
-  proposal_status: @proposal.status)
+  proposal_status: @proposal.status.capitalize)
```
jessieay: we may want to use `display_status` here rather than `status` (`display_status` is a decorated method that says `pending approval` or `pending purchase` depending on the next step if a proposal is pending)

app/views/comment_mailer/comment_added_notification.text.erb
```diff
@@ -4,7 +4,7 @@
 <%= t("mailer.updated_subheader",
   public_id: @proposal.public_id,
   requester_name: @proposal.requester.full_name,
-  proposal_status: @proposal.status) %>
+  proposal_status: @proposal.status.capitalize) %>
```
jessieay: same comment re: `display_status`

app/views/mail_shared/panel/_proposal_attributes.html.haml
```diff
@@ -4,6 +4,8 @@
   subtext: proposal.public_id }
 
 - proposal.fields_for_display.each_slice(2) do |fields|
+  - field_1_data = !property_to_s(fields[0][1]).empty? ? property_to_s(fields[0][1]) : '—'
```
jessieay: we may want to call `blank?` rather than `empty?` here since you cannot call `empty?` on `nil`

eg:

```
[1] pry(main)> nil.empty?
NoMethodError: undefined method `empty?' for nil:NilClass
from (pry):1:in `<main>'
[3] pry(main)> nil.blank?
=> true
```

app/views/observer_mailer/proposal_complete.html.haml
```diff
@@ -2,7 +2,7 @@
 - top_head = t("mailer.observer_mailer.proposal_complete.header")
 - proposal_link_text = t("mailer.view_request_cta")
 - cta_subheader = t("mailer.proposal_status_html",
-  proposal_status: @proposal.detailed_status)
+  proposal_status: @proposal.detailed_status.capitalize)
```
jessieay: maybe we should call `capitalize` in the `detailed_status` method itself so we don't have to do it each time we call the method?

app/views/proposal_mailer/emergency_proposal_created_confirmation.html.haml
```diff
@@ -4,7 +4,15 @@
 - cta_subheader = t("mailer.proposal_mailer.emergency_proposal_created_confirmation.subheader")
 - proposal_link_text = t("mailer.view_request_cta")
 - cta_subheader_foot = t("mailer.proposal_status_html",
-  proposal_status: @proposal.detailed_status)
+  proposal_status: @proposal.detailed_status.capitalize)
+
+
+%table.container
+  = render partial: "mail_shared/call_to_action/subheader",
+    locals: { subheader: "Emergency Request",
```
jessieay: can we put ` "Emergency Request"` into a locale?

app/views/mail_shared/panel/_proposal_attributes.html.haml
```diff
@@ -4,6 +4,8 @@
...
+  - field_2_data = !property_to_s(fields[1][1]).empty? ? property_to_s(fields[1][1]) : '—'
```
jessieay: `fields[1]` will sometimes be `nil`, which is why we have the check on L 22. I can help put this into a helper, but I think we could define something in `MailerHelper` that takes care of this and would prevent the `nil` errors we're getting in specs:

```ruby
# MailerHelper

def property_display_value(field)
 if property_to_s(field[1])
    property_to_s(field[1])
  else
    "-"
  end
end
```

and then here where we used to have

`property_to_s(fields[0][1])` and `property_to_s(fields[1][1])`

we can call

`property_display_value(fields[0])` and `property_display_value(fields[1])`

app/views/observer_mailer/observer_added_notification.html.haml
```diff
@@ -1,6 +1,8 @@
 - content_for :header_icon, "emails/icon-page-circle.png"
-- top_head = observer_text(@observation)
-- cta_subheader = @reason
+- top_head = t("mailer.observer_mailer.observer_added_notification.header" )
+- cta_subheader = t("mailer.observer_mailer.observer_added_notification.subheader", 
+                    amount: number_to_currency(@proposal.total_price),
+                    status: @proposal.status)
```
jessieay: `display_status` ?

jessieay:
A few comments about moving stuff around, otherwise looks ready to :ship:  !

app/views/comment_mailer/comment_added_notification.html.haml
rememberlenny: @jessieay On this template specifically, tests fail for display_status not existing

jessieay:
Hey @rememberlenny saw an email notification for a comment about `display_status` not existing -- that is because `@proposal` in that case is not decorated. We should be decorating it for all mailers. Look for `.decorate` being called in the mailer.

Thoughts on moving `.capitalize` logic into the `display_status` method itself?

app/helpers/mailer_helper.rb
```diff
@@ -16,6 +16,14 @@ def observer_text(observation, reason = nil)
     text + "."
   end
 
+  def property_display_value(field)
```
jessieay: I added a very similar method in https://github.com/18F/C2/pull/1079/files as part of the proposal attributes refactor so you may want to remove so we don't have merge conflicts.

app/decorators/proposal_decorator.rb
```diff
@@ -12,9 +12,9 @@ def detailed_status
 
   def display_status
     if object.pending?
-      "pending approval"
+      "Pending approval"
     else
-      object.status
+      object.status.capitalize
```
jessieay: :+1:

lib/mail_previews/cancelation_mailer_preview.rb
```diff
@@ -2,7 +2,7 @@ class CancelationMailerPreview < ActionMailer::Preview
   def cancelation_notification
     CancelationMailer.cancelation_notification(
       recipient_email: email,
-      cancler: user,
+      canceler: user,
```
jessieay: :skull:

jessieay:
spec is failing because `observer_text(@observation)` was removed from observer added email. 

if we not longer want to use that text (which includes adder name),then let's remove the method and remove the tests in `observer_mailer_spec` that are looking that that text (but keep test that looks for `reason` since we are still including that)

PR #1077, opened 2016-03-17
Attachment mailer updates

jessieay:
* Link to `files` anchor when user clicks attachment in email
* Do not send attachment email to person who added attachment

https://trello.com/c/oCt6aVdo
https://trello.com/c/v24zGByh

### DIFF
--- a/app/decorators/attachment_decorator.rb
+++ b/app/decorators/attachment_decorator.rb
@@ -2,11 +2,17 @@ class AttachmentDecorator < Draper::Decorator
   include Rails.application.routes.url_helpers
   include ActionView::Helpers::AssetTagHelper
   include ActionView::Helpers::UrlHelper
+  default_url_options[:host] = ::Rails.application.routes.default_url_options[:host]
+
   delegate_all
 
   def file_preview
     if file.content_type =~ /\Aimage/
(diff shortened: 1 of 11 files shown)


jessieay:
this is still a WIP

jessieay:
this is now ready for review

rememberlenny:
Looks good!

app/decorators/attachment_decorator.rb
```diff
@@ -2,11 +2,17 @@ class AttachmentDecorator < Draper::Decorator
   include Rails.application.routes.url_helpers
   include ActionView::Helpers::AssetTagHelper
   include ActionView::Helpers::UrlHelper
+  default_url_options[:host] = ::Rails.application.routes.default_url_options[:host]
```
rememberlenny: What exactly does this do?

jessieay: oh shoot I left this in accidentally. was trying to use some named routes in here and was looking at this https://github.com/drapergem/draper/wiki/Using-Rails-Path-Helpers-in-Draper-Decorators-with-Grape but ended up sticking with `object.url` instead of going for `proposal_attachment_url` since I looked in the controller and realized that just redirects to `object.url` anyway :)

not bad that it was left in but unnecessary so we can remove soon

PR #1079, opened 2016-03-17
Update display attributes in emails

jessieay:
* Put them in a specific order for NCR / 18F

https://trello.com/c/8hxppCQO/242-email-details-display-revision

NCR:

screen shot 2016-03-17 at 1 08 04 pm

18F:

screen shot 2016-03-17 at 1 16 42 pm


app/helpers/value_helper.rb
```diff
@@ -24,11 +24,16 @@ def decimal?(val)
...
+    elsif val.is_a?(ActiveSupport::TimeWithZone)
+      I18n.l(val, format: :date)
+    elsif val == true
+      "Yes"
+    elsif val == false
+      "No"
     else
-      val.to_s
+      val
     end
```
afeld: A `case` statement may be easier here.

app/models/gsa18f/procurement_fields.rb
```diff
@@ -18,6 +18,24 @@ def display
...
+        ["Quantity", procurement.quantity],
+        ["Urgency", procurement.urgency_string],
+        ["Cost Per Unit", procurement.cost_per_unit],
+        ["Office", procurement.office],
+        ["Total Price", procurement.total_price],
+        ["Justification", procurement.justification],
+        ["Link to Product", procurement.link_to_product],
+        ["Additional Info", procurement.additional_info]
+      ]
+    end
```
afeld: Not a blocker, but worth investigating if these can be done via i18n.

app/models/gsa18f/procurement_fields.rb
```diff
@@ -18,6 +18,24 @@ def display
...
+
+    def top_email_field
+    end
```
afeld: This is meant to be empty?

app/models/gsa18f/procurement_fields.rb
afeld: Also, seems like this may not belong in a model...? Ditto to both comments below.

app/models/proposal.rb
```diff
@@ -239,6 +239,20 @@ def fields_for_display
...
+      client_data.public_send(:fields_for_email_display)
+    else
+      []
+    end
+  end
+
+  def top_email_field
+    if client_data
+      client_data.public_send(:top_email_field)
+    end
```
afeld: I would be explicit about returning `nil` otherwise, if that's the intention.

afeld:
Just some style things. Looks good otherwise!

app/models/gsa18f/procurement_fields.rb
jessieay: this model is used for picking out which fields belong in views. It's kind of a weird duck. 

Was extracted from some methods that used to be in the `Gsa18f::Procurement` and `Ncr::WorkOrder` models themselves.

Where would you put it?

Agreed on i18n. We are moving all email stuff there right now. Will do that for this.

app/models/gsa18f/procurement_fields.rb
jessieay: yes, we want to place `description` at top for NCR and nothing at top for 18F so we are defining the method for duck-typing purposes.

app/models/gsa18f/procurement_fields.rb
afeld: I'd do it in a presenter, I suppose.

jessieay: Moved to decorators at 1761162 -- I think it's a nice cleanup, thanks for the suggestions ;)

app/helpers/value_helper.rb
jessieay: its funny, I never think to use case statements... guess I usually aim to have branching statements small enough that they wouldn't make sense...but I am interested: what about them is simpler? I dont fully understand why people use case statements vs `if` and `elsif`

rememberlenny:
Reviewed and look good. Thanks @afeld for pointers too

app/helpers/value_helper.rb
afeld: Well, it wouldn't work for the `decimal?` call, but for the rest, just saves you writing `val == ` or `val.is_a?` over and over.

```ruby
case val
when ActiveSupport::TimeWithZone
  I18n.l(val, format: :date)
when true
  "Yes"
when false
  "No"
else
  val
end
```

Not a big deal!

PR #1082, opened 2016-03-18
Update logic around `proposal_complete` emails

jessieay:
* Was not sending to NCR requester / observers because the former
  "pending" status was cached (fixed by reloading proposal)
* Was not sending to 18F for same reason.
* Was sending `step_complete` email to 18F for same reason too
* One tweak: sending email to all observers rather than
  `active_observers`

https://trello.com/c/PMcncLYA
https://trello.com/c/ri0POaQW

### DIFF
--- a/spec/features/gsa18f/procurements/approve_spec.rb
+++ b/spec/features/gsa18f/procurements/approve_spec.rb
@@ -1,5 +1,18 @@
 feature "Approve a Gsa18F procurement" do
   context "when signed in as the approver" do
+    context "last step is completed" do
+      it "sends one email to the requester" do
+        procurement.individual_steps.first.complete!
+        deliveries.clear
+
+        login_as(purchaser)
+        visit proposal_path(proposal)
+        click_on("Mark as Purchased")
+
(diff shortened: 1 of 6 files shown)


spec/features/gsa18f/procurements/approve_spec.rb
```diff
@@ -1,5 +1,18 @@
...
+    context "last step is completed" do
+      it "sends one email to the requester" do
+        procurement.individual_steps.first.complete!
+        deliveries.clear
+
+        login_as(purchaser)
+        visit proposal_path(proposal)
+        click_on("Mark as Purchased")
+
+        expect(deliveries.length).to eq(1)
```
afeld: Would it be useful to ensure it sent the message to the right person?

jessieay: good idea!

afeld:
Not sure if I understand all the context, but the code looks fine!

rememberlenny:
Entire way of testing emails is new and very educational. 

Looks good. Merging

PR #1083, opened 2016-03-18
attachment button (when attachment is PDF) styling is a bit off

rememberlenny:
https://trello.com/c/iENa45O8/250-attachment-button-when-attachment-is-pdf-styling-is-a-bit-off

screen shot 2016-03-18 at 7 30 08 pm


jessieay:
I just noticed when QA-ing on dev that the button (when attachment is PDF) is not working. Can you QA locally and ensure that the link now works when the attachment is a PDF? Easiest thing to do is just attach a PDF to a proposal locally and then the preview will show the email for the last attachment...

PR #1086, opened 2016-03-21
Send email to requester when final step is completed

jessieay:
* Was not sending in production because `proposal` was not reloaded
* Wrote feature specs for each use case to confirm behavior
* Always good to actually send emails :)

https://trello.com/c/CP1x6c5e


pkarman:
@jessieay assuming that `NO_WELCOME_EMAIL` is going to change in #1087 this looks fine.

PR #1087, opened 2016-03-21
Show "false" proposal values as "No" in email

jessieay:
* Were showing up as "-"
* Because we were checking for `.present?`
* Change to check for `.nil?`
* Add specs YEAH testing

https://trello.com/c/oo2LPdqD


pkarman:
so many ways to say no, so little time.

PR #1089, opened 2016-03-21
only fire ahoy tracking in non-test environments

pkarman:


jessieay:
re-running false negative CI :whale:

PR #1092, opened 2016-03-21
Update method name

jessieay:
* Updated incorrectly in merge
* https://github.com/18F/C2/commit/a340de63d51c47467dadd5bbf3d6d9d9d31a9881

### DIFF
--- a/spec/support/es_spec_helper.rb
+++ b/spec/support/es_spec_helper.rb
@@ -64,8 +64,8 @@ module EsSpecHelper
   def es_execute_with_retries(retries = 3, &block)
     begin
       retries -= 1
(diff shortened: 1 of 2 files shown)


spec/support/es_spec_helper.rb
```diff
@@ -64,8 +64,8 @@ def create_es_index(klass)
   def es_execute_with_retries(retries = 3, &block)
     begin
       retries -= 1
-      response = block.call
-    rescue SearchUnavailable => _error
+      block.call
+    rescue SearchUnavailable => error
```
jessieay: changed bc we are using the `error` var below

PR #1093, opened 2016-03-21
Final changes for Amber

rememberlenny:
- Change size of feedback footer text
- Remove the "actionable" state image, but keep for future
- Make the blue label text on the attribute table smaller
- Change the button padding on call-to-action when text is long


jessieay:
@rememberlenny spec failure in `spec/helpers/mailer_helper_spec.rb`-- you're gonna want to change the expectation given new behvaior

app/helpers/mailer_helper.rb
```diff
@@ -30,12 +30,10 @@ def add_reason(text, reason)
   end
 
   def step_status_icon(step)
-    if step.status == "actionable"
-      "emails/numbers/icon-number-" + (step.position - 1).to_s + ".png"
```
jessieay: since we are not longer using these numbered icons can we remove them from source control?

rememberlenny: @jessieay Amber said we might use them later, so she suggested we keep them.

jessieay: ok sounds good - generally I am opposed to keeping things in source control that are unused bc that can end up piling up, so let's try and remember to remove these if we aren't using them in a few weeks

rememberlenny: @jessieay Im going to delete for now. We can revert them back in when we need them.

jessieay:
this is wonderful. mergin'

PR #1097, opened 2016-03-22
ClockTasks are not Rake tasks, so load them like any other class

pkarman:


jessieay:
are we sure this is the right solution? Dolores bot also uses clockwork (setup inspired by C2) and the clockwork task appears to be working with this config https://github.com/18F/dolores-landingham-bot/blob/master/config/clock.rb

pkarman:
I agree that it seems suspicious that this suddenly stopped working last month. My best guess is that the `lib/tasks` files stopped being autoloaded.

I have tested this on c2-prod-ssh and it seemed to resolve the problem. Only way to really know (ime) is to deploy and make sure clockwork does its thing.

jessieay:
I think I know the root of the problem! 
When I moved around all that code, I moved the `ClockTasks` class: https://github.com/18F/C2/commit/e2d614da3742671c163552ea1e4e86b115ea0952#diff-d10485443437eed0f42399a05865ddb0 (`lib/clock_tasks.rb` → `lib/tasks/clock_tasks.rb`) -- previous path must have been autoloaded but new one is not. 

This explains why Dolores bot config work -- that class is in `app/services`, which is def autoloaded.

Although that change was on 2/10, we probably didn't deploy it for a few weeks which is why the problem started at the end of Feb.

This is all to say: this change makes total sense and looks good to go. :)

PR #1098, opened 2016-03-22
Building updates per support ticket request #152

phirefly:


pkarman:
:+1:  -- waiting for CI to restart before merging

phirefly:
It occurs to me that if a request that used the previously used versions of this building, one wouldn't be able to modify the request without problems. Thoughts?

pkarman:
don't our JSON fields allow arbitrary input? I know they do for vendor.

phirefly:
I will test this....

PR #1099, opened 2016-03-22
update PhantomJS, workaround DeadClient CI failures

pkarman:


jessieay:
If we need to use `let` statements to fix the problem, I am on board. But I am not clear on why `let` would behave differently than a memoized method.

As the docs say
> Use let to define a memoized helper method. The value will be cached
across multiple calls in the same example but not across examples.

That sounds like what we are doing with private methods. What is the difference?

pkarman:
the problem is caching across multiple runs of the same example.

here's a small example:

```
RSpec.configure do |config|
  config.around(:each) do |spec|
    spec.run
    spec.run
  end
end

describe "example of let vs private memoization" do
  def foo
    @_foo ||= "bar"
  end

  let(:color) { "green" }

  it "reinits let() each time run is called" do
    expect(color).to eq "green"
  end

  it "has private var initialized to nil" do
    expect(@_foo).to eq nil
    expect(foo).to eq "bar"
    expect(@_foo).to eq "bar"
  end
end
```

jessieay:
Oh interesting I see I didn't realize `let` cached differently.

would you be open to inlining some of these rather than using `let` -- it sounds like that would fix this issue and would not require us to use `let` -- If `let` were the only way I would be into it, but given that we can inline these and have the same result I feel that is a good path! While inlining will initially mean more code, it will be good motivation to have our objects less dependent on each other! Thoughts?

pkarman:
there are 115 instances of `let()` in our test suite at the moment. when you say "some of these" which ones are you meaning?

`grep -r "let\(" spec | wc`

jessieay:
I think in general we've been moving away from `let` and `before` statements. We've had separate threads about this but I think there was general agreement that this was a good way to go because:

- makes for more readable specs; if you have a 200+ line spec file with `let` statements at the top, it's hard to know where the vars in your spec are coming from and what exactly they are.
- makes it easier to understand dependencies if you can see them all encapsulated in a spec. While `let` statements are lazily evaluated so they are not created until referenced, `before` statements *do* run before each spec whether they are really needed or not.
- generally, moving away from the RSpec DSL makes for more readable / maintainable specs. For example, it's easier to move methods into private / shared files for re-use whereas `let` and `before` statements cannot be shared.

I don't think it makes sense to remove all `let` and `before` statements in one sweep, but while we are updating specs I think it would be :100: if we didn't move toward `let`, which is what we've been slowly moving away from. 

I understand the issues that private methods have been causing us; seems that inlining this data creation in each spec might be a happy medium. Thoughts?

pkarman:
how about we change all of these:

```
let(:foo) { create(:bar) }
```

to these

```
def foo; create(:bar); end
```

? that's effectively the private method pattern, w/o the caching.

pkarman:
I retract that suggestion. We have plenty of instances where `let()` is called multiple times in a given spec.

Ok, I will start inlining.

jessieay:
:+1: looks great, will merge as soon as CI passes. HOORAY this is huge I am so happy

PR #1102, opened 2016-03-22
Inline images for emails

jessieay:
* Prevents email formatting from breaking when web app assets are down

https://trello.com/c/pHou61Gd


jessieay:
@pkarman after a few iterations, this is ready for review. (tested on prod to make sure). 

Interesting learning: inline attachments do not render properly (at least in gmail) without a file extension.

PR #1103, opened 2016-03-22
AG dashboard & summary updates

yozlet:
For [this Trello card](https://trello.com/c/A7EoU4o1/106-acquisition-gateway-summary-dashboard-updates), and also includes a fix for [the "No approved count" bug](https://trello.com/c/i80CJ99O).

### DIFF
--- a/app/controllers/summary_controller.rb
+++ b/app/controllers/summary_controller.rb
@@ -1,5 +1,6 @@
 class SummaryController < ApplicationController
-  before_action :authorize
+  before_action :block_non_admins
+  before_action :block_client_admins_without_client_slugs
 
   def index
     @summaries = titleized_client_namespaces.map do |cn|
@@ -9,14 +10,16 @@ class SummaryController < ApplicationController
 
   private
 
-  def authorize
-    if !current_user.any_admin? || (needs_client_slug? && !client_slug?)
+  def block_non_admins
+    unless current_user.any_admin?
       render "authorization_error", status: 403
     end
   end
 
-  def needs_client_slug?
-    current_user.client_admin?
+  def block_client_admins_without_client_slugs
+    if current_user.client_admin? && !client_slug?
+      render "authorization_error", status: 403
+    end
   end
 
   def client_slug?
--- a/app/controllers/summary_controller.rb
+++ b/app/controllers/summary_controller.rb
@@ -2,17 +2,15 @@ class SummaryController < ApplicationController
   before_action :authorize
 
   def index
-    if current_user.gateway_admin?
-      client_namespaces = Proposal.client_slugs.map(&:titleize)
-    else
-      client_namespaces = [current_user.client_slug.titleize]
-    end
+    client_namespaces = if current_user.gateway_admin?
+                          Proposal.client_slugs
+                        else
+                          [current_user.client_slug]
+                        end
+                        .map(&:titleize)
 
     @summaries = client_namespaces.map do |cn|
-      ClientSummarizer.new(
-        client_namespace: cn,
-        fiscal_year: params[:fiscal_year]
-      ).run
+      get_client_summary(cn, params[:fiscal_year])
     end
   end
 
@@ -30,5 +28,12 @@ class SummaryController < ApplicationController
 
   def client_slug?
     current_user.client_slug.present?
+  end
+
+  def get_client_summary(client_namespace, fiscal_year)
+    ClientSummarizer.new(
+      client_namespace: client_namespace,
+      fiscal_year: fiscal_year
+    ).run
   end
 end
--- a/app/controllers/summary_controller.rb
+++ b/app/controllers/summary_controller.rb
@@ -2,14 +2,7 @@ class SummaryController < ApplicationController
   before_action :authorize
 
   def index
-    client_namespaces = if current_user.gateway_admin?
-                          Proposal.client_slugs
-                        else
-                          [current_user.client_slug]
-                        end
-                        .map(&:titleize)
-
-    @summaries = client_namespaces.map do |cn|
+    @summaries = titleized_client_namespaces.map do |cn|
       get_client_summary(cn, params[:fiscal_year])
     end
   end
@@ -17,7 +10,7 @@ class SummaryController < ApplicationController
   private
 
   def authorize
-    if !adminish_role? || (!client_slug? && !current_user.gateway_admin?)
+    if !adminish_role? || (needs_client_slug? && !client_slug?)
       render "authorization_error", status: 403
     end
   end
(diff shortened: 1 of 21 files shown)


yozlet:
Note: The CodeClimate issues are legacy. (I'm going to try and get a de-dupe in, but I vote that this shouldn't depend on it)

yozlet:
Also, CircleCI seems to be failing for unrelated reasons :frowning:

app/controllers/summary_controller.rb
```diff
@@ -2,27 +2,38 @@ class SummaryController < ApplicationController
   before_action :authorize
 
   def index
-    client_namespace = current_user.client_slug.titleize
-    @client_summarizer = ClientSummarizer.new(
-      client_namespace: client_namespace,
-      fiscal_year: params[:fiscal_year]
-    )
-    @summary = @client_summarizer.run
+    client_namespaces = if current_user.gateway_admin?
```
jessieay: 1 thoughts:

1. should regular admins also be able to see all proposals in the summary view? 

2. It seems like this logic should go into its own method -- perhaps a helper method we can test?

app/controllers/summary_controller.rb
```diff
@@ -2,27 +2,38 @@ class SummaryController < ApplicationController
...
+    @summaries = client_namespaces.map do |cn|
+      get_client_summary(cn, params[:fiscal_year])
+    end
   end
 
   private
 
   def authorize
-    if !adminish_role? || !client_slug?
+    if !adminish_role? || (!client_slug? && !current_user.gateway_admin?)
```
jessieay: I think `adminish_role?` takes care of the `current_user.gateway_admin?` case so we don't need to add that here

app/presenters/client_summary.rb
```diff
@@ -25,21 +25,35 @@ def subtotal(status)
 
   def total
     total = 0
-    @subtotals.values.each { |i| total += i }
+    @subtotals.each { |k, i| total += i unless k == :canceled }
```
jessieay: can we use var names that are more descriptive than `k` and `i`?

app/views/layouts/_header_nav.html.erb
```diff
@@ -25,6 +25,11 @@
         <%= image_tag "img-cart.png", width: "18px", height: "16px" %>
       </li>
       <%= render client_specific_partial(current_user, "header_links") %>
+      <%- if current_user.gateway_admin? %>
```
jessieay: should regular admins be able to see this too?

db/seeds.rb
```diff
@@ -10,3 +10,4 @@
 # the relevant ENV vars after it has run at least once.
 RolesConversion.new.ncr_budget_approvers
 RolesConversion.new.gsa18f_approvers
+RolesConversion.new.gateway_roles
```
jessieay: does this need to be in seeds, or can we create in selectively for the specs that require this role?

spec/models/user_spec.rb
```diff
@@ -91,8 +91,16 @@
...
+    it 'is true if the user has the gateway_admin role' do
+      admin = create(:user, :gateway_admin)
+
+      expect(admin).to be_gateway_admin
+    end
+  end
+
   describe '#admin?' do
-    it 'is true of the user has the admin role' do
+    it 'is true if the user has the admin role' do
```
jessieay: :smile:

spec/presenters/client_summary_spec.rb
```diff
@@ -1,21 +1,26 @@
 describe ClientSummary do
-  it "adds status" do
+  it "adds non-canceled status" do
```
jessieay: what about `it "shows non-canceled proposals" do` -- does that fit the spec expectation more clearly?

spec/queries/ncr/dashboard_query_spec.rb
```diff
@@ -3,10 +3,14 @@
     it "returns scoped NCR records" do
       user = create(:user)
       proposal = create(:proposal, requester: user)
+      # canceled proposals should be excluded from the totals
```
jessieay: perhaps we can add this as a `context ` block rather than a comment?

app/views/layouts/_header_nav.html.erb
yozlet: good point, they should.

app/controllers/summary_controller.rb
yozlet: 1. The card calls for gateway admins to have that power, but now that you mention it, it makes sense for regular admins (but not client admins) to have it too. I'll tweak it.

2. OK!

app/controllers/summary_controller.rb
yozlet: Not quite - the previous logic says that admins need to have an adminish role _and_ a client slug, whereas in the new logic, gateway admins don't need a client slug.

However, there's probably a clearer way to write this. I'll give it a go.

db/seeds.rb
yozlet: I assumed it should be in seeds because it's now an important role in the app, and so dev instances should have it. But we also do a `Role.find_or_create_by(name: 'gateway_admin')` in the `User` factory (for the `:gateway_admin` trait).

jessieay: benefit of *not* including in seeds is that we run them before each test, so each new record added to seeds is slowing us down quite a bit

yozlet:
I've taken `gateway_admin` creation out of `seeds.rb`, but I've left it in `RolesConversion` given that it feels like a useful list of base roles. But I'm happy to take it out of there too.

app/controllers/summary_controller.rb
```diff
@@ -2,27 +2,40 @@ class SummaryController < ApplicationController
...
+    @summaries = titleized_client_namespaces.map do |cn|
+      get_client_summary(cn, params[:fiscal_year])
+    end
   end
 
   private
 
   def authorize
-    if !adminish_role? || !client_slug?
+    if !current_user.any_admin? || (needs_client_slug? && !client_slug?)
```
jessieay: nice! much better. one question:

for users with `client_admin` role, we are checking that they do not have a client slug. But then in ` titleized_client_namespaces` we return the `client_slug` for the user. What do we want client admins to see and is this accomplishing that?

jessieay:
one last comment on authorization logic for summaries. Otherwise looking perfecto!

app/controllers/summary_controller.rb
yozlet: `client_admin` is only meant to see a summary of their own `client_slug` (i.e. the client for which they are an admin). Other admins are meant to see summaries of _all_ clients. So `titleize_client_namespaces` returns an array containing all the clients for which which this user is allowed to see summaries.

PR #1104, opened 2016-03-23
add 2 columns to CSV for completion date and duration

pkarman:
trello: https://trello.com/c/3ZCgmZAR


app/decorators/proposal_decorator.rb
```diff
@@ -9,6 +9,38 @@ def total_approvers
...
+
+  def final_step_label
+    if total_approvers > 0
+      "Final #{object.individual_steps.last.decorate.label} Completed"
+    else
+      "Final Step Completed"
+    end
+  end
+
+  def self.final_step_label(proposal = nil)
```
jessieay: why is this a class method rather than an instance method on the decorator? We are only using it one one record at a time based on what I am seeing here so would expect an instance method.

spec/decorators/step_decorator_spec.rb
```diff
@@ -1,4 +1,22 @@
 describe StepDecorator do
+  describe "#label" do
+    context "when the step is an approval" do
+      it "fetches the correct role text" do
+        user = build_stubbed(:user)
+        step = Steps::Approval.new(user: user).decorate
+        expect(step.label).to eq "Approval"
```
jessieay: should we use the `I18n` val in these expectations?

app/decorators/proposal_decorator.rb
```diff
@@ -9,6 +9,38 @@ def total_approvers
...
+  def total_completion_days
+    if completed?
+      (final_completed_date.to_date - created_at.to_date).to_i
+    else
+      ""
+    end
+  end
+
+  def final_step_label
+    if total_approvers > 0
```
jessieay: rename this to `total_step_users` while we are here ? :heart_decoration:

jessieay:
just a few small comments from me -- looking good! back to you @pkarman

app/decorators/proposal_decorator.rb
pkarman: there are 2 methods: one an instance method and the other a class method. The class method is necessary because there might not be a Proposal instance available (imagine a search with zero results).

pkarman: the dilemma is that the CSV header row is generated regardless of whether we have any body rows.

spec/decorators/step_decorator_spec.rb
pkarman: good question. I was following the existing specs which did not use the `I18n` value.

I will change all of them.

app/decorators/proposal_decorator.rb
jessieay: in that case maybe it makes sense to have a helper method to handle this? I think I understand the need to have something other than an instance method, but I find the class method confusing because usually a class method is called on a collection rather than a single object or no objects.

app/decorators/proposal_decorator.rb
pkarman: :100:

jessieay:
@pkarman one final comment for you consideration: https://github.com/18F/C2/pull/1104#discussion_r57216388

pkarman:
@jessieay fair point about class method expectations. refactored to eliminate that method.

jessieay:
@pkarman :cookie: :cool:

jessieay:
these circle failure are legit (related to new implementation)

pkarman:
:green_heart: again

PR #1108, opened 2016-03-23
Reference approval chain icons via http

jessieay:
* When attach all images as inline, unused images show as attachments
* In future we can inline but need to be more sophisticated about
  logic, shouldn't just inline all images


yozlet:
:+1:

PR #1110, opened 2016-03-23
Use step user in `StepMailer`

jessieay:
* Was defaulting to last completed step
* Should be the same thing, but this is more straightforward


yozlet:
:+1:

PR #1112, opened 2016-03-24
switch from cf-blue-green to autopilot for zero downtime deploys

pkarman:
tested successfully on c2-dev and c2-staging


jessieay:
Should we update The Docs to reflect this change? https://github.com/18F/C2/blob/master/doc/production.md#deploying

pkarman:
@jessieay good catch! fixed in e80cc1a

pkarman:
@jessieay test failure is just the Ghost in the Machine

jessieay:
:+1:

PR #1113, opened 2016-03-24
proposal.final_completed_date requires at least one step

pkarman:
trello: https://trello.com/c/Mlt0yGBC/291-reporting-email-bug

### DIFF
--- a/app/decorators/proposal_decorator.rb
+++ b/app/decorators/proposal_decorator.rb
@@ -34,7 +34,7 @@ class ProposalDecorator < Draper::Decorator
   end
 
   def total_completion_days
-    if completed?
+    if completed? && total_step_users > 0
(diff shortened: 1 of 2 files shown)


app/decorators/proposal_decorator.rb
```diff
@@ -34,7 +34,7 @@ def final_completed_date
   end
 
   def total_completion_days
-    if completed?
+    if completed? && total_step_users > 0
```
jessieay: should we add a test for this to prevent regressions?

pkarman: as always, a good idea. e6812f6

jessieay: :heart_eyes_cat:

PR #1114, opened 2016-03-24
rework search UI

pkarman:
trello: https://trello.com/c/LvyMo4E6/86-as-a-user-i-can-easily-find-advanced-search

screen shot 2016-03-24 at 11 16 12 am

screen shot 2016-03-24 at 11 15 32 am

screen shot 2016-03-24 at 11 15 09 am


pkarman:
codeclimate error about "similar" code is accurate but irrelevant.

pkarman:
tests are passing! :green_heart:

app/assets/javascripts/search.js
```diff
@@ -1,31 +1,145 @@
 $(document).ready(function() {
 
-  /* open the Adv Search UI immediately if param set */
+  /*************************************************************************************************/
+  /* ** Search Form ** */
```
jessieay: oops! just left some JS comments on https://github.com/18F/C2/pull/1117 bc I didn't realize it was a branch off of this. Sorry! reference that for my js feedback :green_apple:

spec/features/search_spec.rb
```diff
@@ -35,10 +35,10 @@
 
     visit proposals_path
     fill_in "text", with: proposals.first.name
-    dropdown_button = find("button.dropdown-toggle")
-    dropdown_button.trigger("click") # open dropdown
+    adv_options = find("a.adv-options")
+    adv_options.trigger("click") # open dropdown
     fill_in "test_client_request[client_data.amount]", with: proposals.first.client_data.amount
-    dropdown_button.trigger("click") # close dropdown
+    find("a.closer").trigger("click") # close dropdown
```
jessieay: put in a private method named `close_dropdown` instead?

pkarman:
@jessieay tests passing, and I think I hit all your comments from this PR and #1117

jessieay:
looks like some or all of the updates from #1099 got pulled into this one -- was that intentional? make diff much noisier than I remember it being when I last looked

pkarman:
yes #1099 was merged to here to verify it fixed the problem on other open PRs: https://18f.slack.com/archives/cap/p1458919683002033

jessieay:
my merge of other PR caused conflicts here - sorry!

jessieay:
niiiice code looks good.  I will QA Now

jessieay:
QA feedback:

- advanced search options hidden on Proposals#index

screen shot 2016-03-25 at 3 26 17 pm

- `reset search` link doesn't appear to do anything. It does trigger the spinny wheel (perhaps refreshing results?) but I would expect it to clear the inputs on the search form.

pkarman:
both issues addressed.

pkarman:
This is deployed on https://c2-dev-pkarman.18f.gov/ for QA purposes

cc @yozlet @rememberlenny @ericronne @phirefly

rememberlenny:
Looks good. One small issue, the "Date range" filter can't be unselected.

get out

pkarman:
thanks @rememberlenny -- fixed and redeployed. please test again.

rememberlenny:
@pkarman Found another small one.

If "AMOUNT" field takes a alpha-character string, then it returns with error. Can you turn the input field to accept numbers only?

amoun

yozlet:
This may be more of a problem with the underlying search, but: I created a proposal today, then searched for Date Within 1 month of 03/30/16, and no results were found. The generated search string is `Created at:[2016-02-29T07:00:00Z TO 2016-03-30T07:00:00Z]`, so it's searching up to the _start_ of the target day, rather than the end. But also, I expect "Date within 1 month of..." to mean 1 month on either side, rather than just one month before. (GMail works this way too: searching for "Date within 1 week of 2016/03/09" produces `after:2016/3/2 before:2016/3/17`)

yozlet:
Also, the old plural problem: "1 results"

pkarman:
@yozlet that is a good point about the date math. I'm not sure (a) what is most useful to users and (b) how best to represent that in the UI. Since this PR doesn't represent any function change to the date math, can we open a new ticket for that issue to do some UX research?

pkarman:
@yozlet plural issue fixed in 6339836

pkarman:
@rememberlenny good catch on the bad `Amount` field value.

I'm reluctant to limit the kind characters allowed because in theory a value like `[1 TO 1000]` is a valid query (range of values).

Instead I have added a more granular search exception class `SearchBadQuery` with a more helpful error message. The ES exception messages are notoriously difficult to parse (though apparently are much better in ES 2.x) so the generic message is as close as I can do for now.

See cc78093

pkarman:
opened https://trello.com/c/4CTs5vfP/306-search-within-date-range-ux-is-math-correct to track the UX issue.

@rememberlenny @yozlet if this is :+1: then please merge.

PR #1116, opened 2016-03-24
Ensure client admins see all summaries if they have other admin roles

yozlet:
Fixes bug where a regular or gateway admin was only seeing one summary because they were also a client admin.


pkarman:
@yozlet looks good. ignoring the Ghost in the CI machine.

PR #1117, opened 2016-03-24
use relative created_at date in search/reports

pkarman:
trello: https://trello.com/c/HkOpuTs8/226-as-a-user-i-would-like-my-timeboxed-report-to-always-be-relative-to-today

NOTE this is based on #1114 so please merge that first.


app/assets/javascripts/search.js
```diff
@@ -1,31 +1,145 @@
...
+
+    if (previewCountTimer) {
+      clearTimeout(previewCountTimer);
+    }
+
+    previewCountTimer = setTimeout(function() {
+      $.get(url, function(resp) {
+        countEl.html(resp.total);
+      }).fail(function(xhr, err, msg) {
+        console.log('fail!', msg);
```
jessieay: remove this console log?

app/assets/javascripts/search.js
```diff
@@ -1,31 +1,145 @@
...
+      if (!advOptionsVisible()) {
+        $("#adv-options").fadeIn(FADE_SPEED);
+      }
+      $(".m-search-ui .input-group-addon.magnifier").hide();
+      $("#search-button").fadeIn(FADE_SPEED);
+    }
+  };
+
+  var showAdvOptions = function() {
+    $("fieldset.adv").fadeIn(FADE_SPEED);
```
jessieay: rather than referencing these selectors in various functions, should we assign them to a var so we can easily update our js if markup changes? Also, thoughts on adding selectors with `js` prefix so we know they are there for js?

app/assets/javascripts/search.js
```diff
@@ -1,31 +1,145 @@
...
   });
-  $('form.adv-search').on("submit", function(e) {
-    var searchForm = $(this);
-    var termsInput = $('.search-terms');
-    if (termsInput.val().length) {
-      var textInput = $('<input type="hidden" name="text">');
-      textInput.val(termsInput.val());
-      searchForm.append(textInput);
+
+  // fetch search total for preview count
```
jessieay: should we put this chunk of logic into a function so we don't need the comment?

app/assets/javascripts/search.js
```diff
@@ -1,31 +1,145 @@
...
+        console.log('fail!', msg);
+      });
+    }, 1000); // TODO experiment with this delay
+  };
+
+  // if any adv search form inputs change, fetch new preview total
+  // the 'keyup' listener handles text input immediately (change waits for focus change)
+  $('form.search :input').keyup(function(e) {
+    var el = $(e.target);
+    //console.log('adv search keyup: ', el[0].name);
```
jessieay: remove commented console log?

app/assets/javascripts/search.js
```diff
@@ -35,6 +149,7 @@ $(document).ready(function() {
     clickOnEnter(e, ".m-search-ui button.search");
   });
 
+  /******************************************************************************************************/
   /* *** setup Save as Report *** */
```
jessieay: this JS file is getting pretty long! any ideas on how we could make it more modular / break it up?

spec/controllers/proposals_controller_spec.rb
```diff
@@ -173,6 +173,24 @@
           expect(query[2].id).to be(single.id)
         end
       end
+
+      it "returns JSON for preview count" do
+        login_as(user)
+        proposals = 3.times.map do |i|
+          wo = create(:test_client_request, project_title: "Work Order #{i}")
+          wo.proposal.update(requester: user)
+          wo.proposal.reindex
```
jessieay: do we need to call `reindex` here if we are refreshing entire index on L 185 anyway?

spec/controllers/proposals_controller_spec.rb
```diff
@@ -173,6 +173,24 @@
           expect(query[2].id).to be(single.id)
         end
       end
+
+      it "returns JSON for preview count" do
+        login_as(user)
+        proposals = 3.times.map do |i|
```
jessieay: this can be shortened to:

```ruby
create_list(:test_client_request, requester: user)
```

perhaps we can sequence the project title in the factory definition attr if it's important for those to be unique?

spec/queries/proposal_search_dsl_spec.rb
```diff
@@ -64,23 +64,43 @@
...
-      "(foo OR Bar) AND (created_at:[#{(now.utc - 6.months).iso8601} TO #{now.utc.iso8601}])"
-    )
+  describe "parses date ranges" do
+    it "when created_at is present" do
+      now = Time.zone.now
+      user = create(:user, client_slug: "test")
+      dsl = ProposalSearchDsl.new(
+        params: {
+          test_client_request: {
+            created_at: now.to_s,
```
jessieay: maybe we should set `created_at` to a different value so the result of this spec is more different from the result in the following spec?

app/assets/javascripts/search.js
pkarman: good point. split now into 2 files.

spec/controllers/proposals_controller_spec.rb
pkarman: refresh is not the same as reindex.

reindex pushes content to the ES server.

refresh tells the ES to commit changes to disk and re-open its searchers to make the reindex changes visible. It's like calling `COMMIT` on a db transaction.

spec/controllers/proposals_controller_spec.rb
pkarman: thanks, shortened. can't quite get rid of the block though, because we need to reindex each of the 3 proposals.

spec/queries/proposal_search_dsl_spec.rb
pkarman: sure 3831be2

pkarman:
This is deployed on https://c2-dev-pkarman.18f.gov/ for QA purposes

cc @yozlet @rememberlenny @ericronne @phirefly

yozlet:
Tested. Works as intended, though I think it'd be a little more helpful if today's date was pre-filled in the form.

pkarman:
@yozlet I wanted the default placeholder text to be `now` but it seems that the field does not support it.

Just found this CSS trick and will try that http://stackoverflow.com/questions/18146350/how-do-i-simulate-placeholder-functionality-on-input-date-field

I avoided defaulting to today's date because that re-introduces the same problem this PR is trying to solve: hard-coding a date into a report, instead of "date window".

pkarman:
@yozlet some sleight-of-hand implemented to more clearly show the default is `now` in 133816d

yozlet:
Nifty! :+1: :shipit:

pkarman:
@yozlet @rememberlenny looks like this can be merged after #1114

PR #1119, opened 2016-03-24
Fix `urgency` display for 18F procurements

jessieay:
* Was showing as integer rather than string
* Move logic from model -> decorator

screen shot 2016-03-24 at 3 25 59 pm

https://trello.com/c/YEbcYtjo/255-18f-urgency-showing-as-integer-not-string


jessieay:
I don't think I'm gonna be able to make code climate happy on this one

pkarman:
ghost in the CI machine keeps failing, merging in good faith.

PR #1120, opened 2016-03-25
create User from /admin UI

pkarman:
card: https://trello.com/c/3Iazk964/296-ability-to-create-users-via-admin-console

### DIFF
--- a/app/admin/user.rb
+++ b/app/admin/user.rb
@@ -17,7 +17,7 @@ ActiveAdmin.register User do
       roles = role_ids_to_roles
       User.transaction do
         user.save!
-        UserRole.create(roles.map{ |role| { user: user, role: role } })
+        UserRole.create(roles.map { |role| { user: user, role: role } })
       end
       redirect_to admin_user_path(user)
     end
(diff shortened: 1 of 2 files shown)


app/admin/user.rb
```diff
@@ -11,6 +11,36 @@
   # https://github.com/activeadmin/activeadmin/blob/master/docs/2-resource-customization.md#setting-up-strong-parameters
   permit_params :active, :first_name, :last_name, :email_address, :client_slug, role_ids: []
 
+  controller do
+    def create
+      user = User.new(user_params)
+      roles = role_ids_to_roles
+      User.transaction do
```
jessieay: niiiiice I need to use transactions more

jessieay:
small question / request while we're here: I noticed when updated users on dev and staging that it isn't possible to select no roles. One or more must always be selected. Would that be hard to fix?

pkarman:
that's not a Rails issue, afaik. If you command-click on a multiselect, you can deselect all roles, which should allow you to save a user with zero roles. I just verified with https://c2-dev.18f.gov/admin/users/146

jessieay:
@pkarman ok cool! I can already envision my future self searching for this thread to remember how to deselect roles. ;)

PR #1122, opened 2016-03-25
Update seeds

jessieay:
* Calling `dev:prime[EMAIL]` will create a user for email and make it an
  admin
* This task clears out DB so there are no problems with uniqueness
* Creates all records necessary to preview emails

https://trello.com/c/voBTTwVx

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -4,7 +4,7 @@
 #
 MYUSA_KEY: consumer_public_key
 MYUSA_SECRET: consumer_secret_key
-DEFAULT_URL_HOST: "http://localhost:3000"
+DEFAULT_URL_HOST: "localhost:3000"
 
 # optional
 #
--- a/.env.example
+++ b/.env.example
@@ -4,6 +4,7 @@
 #
 MYUSA_KEY: consumer_public_key
 MYUSA_SECRET: consumer_secret_key
+DEFAULT_URL_HOST: your_local_server_address
 
 # optional
 #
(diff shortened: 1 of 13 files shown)


.env.example
```diff
@@ -4,6 +4,7 @@
 #
 MYUSA_KEY: consumer_public_key
 MYUSA_SECRET: consumer_secret_key
+DEFAULT_URL_HOST: your_local_server_address
```
pkarman: is this really necessary to be set? should it not default to localhost?

jessieay: we do need this for urls to work in emails -- see `config/initializers/mailer.rb` -- this is a newish change

pkarman: then maybe it should default to something working? `localhost` ?

jessieay: sounds good! question: are you using foreman or rails server? I am using foreman bc of our multiple processes but I see a lot of localhost:3000 in setup so maybe I should just default to that...

pkarman: I use both, but mostly `rails s`

pkarman: fwiw, I think default host is just hostname sans port.

jessieay: I have it set to `http://localhost:5000` locally and that is working for me . http://guides.rubyonrails.org/action_mailer_basics.html#generating-urls-in-action-mailer-views

jessieay: updated at 31ebada

.env.example
```diff
@@ -4,7 +4,7 @@
 #
 MYUSA_KEY: consumer_public_key
 MYUSA_SECRET: consumer_secret_key
-DEFAULT_URL_HOST: your_local_server_address
+DEFAULT_URL_HOST: "http://localhost:3000"
```
pkarman: I think this should just be `localhost`

pkarman:
@jessieay when I check out how `DEFAULT_URL_HOST` is used at cloud.gov, it is like this:

`DEFAULT_URL_HOST: cap.18f.gov`

which suggests it should just be `localhost` (or maybe `localhost:3000`) but definitely no protocol.

jessieay:
@pkarman updated at fc4463b

PR #1123, opened 2016-03-25
missed this inlining of let()

pkarman:


jessieay:
woah I don't know how CI was passing - weird! Must have seen in passing in cached version or something

PR #1130, opened 2016-03-30
add test to validate step assignment reverts on restart

pkarman:
trello: https://trello.com/c/Wib4cFDJ/298-as-a-mailbox-user-i-should-be-assigned-a-request-that-has-been-put-into-re-approval-status

this is just a test to validate the story in the card. the code itself is already in master.

### DIFF
--- a/spec/models/steps/individual_spec.rb
+++ b/spec/models/steps/individual_spec.rb
@@ -40,12 +40,14 @@ describe Steps::Individual do
     end
 
     it "resets completed_by and completed_at" do
-      user = create(:user)
-      step = create(:approval_step)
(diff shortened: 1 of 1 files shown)


yozlet:
Looks good, though it makes me wonder if we need a better name than `completed_by` (given that it returns a user even when the step isn't complete)

phirefly:
@pkarman: Is it possible to have an integration/browser spec for this? i.e. when re-approval is triggerd, the email should show that the affected steps are 1) in pending status and 2) displaying the correct (mailbox) email address

pkarman:
@phirefly I will open a new PR with feature test.

PR #1132, opened 2016-03-31
Rename NCR code column

pkarman:
trello: https://trello.com/c/TfqNfraq/272-rename-code-field-in-database

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160308231131) do
+ActiveRecord::Schema.define(version: 20160331143230) do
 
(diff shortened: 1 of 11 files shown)


db/schema.rb
```diff
@@ -212,7 +212,7 @@
     t.integer  "client_data_id"
     t.string   "client_data_type", limit: 255
     t.integer  "requester_id"
-    t.string   "public_id"
+    t.string   "public_id",        limit: 255
```
yozlet: Should this change be here?

pkarman: it keeps toggling back and forth because the production db schema has that constraint defined and I work off a copy of the production db.

I know @jessieay would be glad to see this noise removed. I will create a new PR to make that limit in the code.

yozlet:
:+1:

PR #1133, opened 2016-03-31
add feature test demonstrating mailbox system user assigned on approval restart

pkarman:
trello: https://trello.com/c/Wib4cFDJ/298-as-a-mailbox-user-i-should-be-assigned-a-request-that-has-been-put-into-re-approval-status

### DIFF
--- a/spec/features/ncr/work_orders/post_approval_modification_spec.rb
+++ b/spec/features/ncr/work_orders/post_approval_modification_spec.rb
@@ -25,7 +25,7 @@ feature "post-approval modification" do
     fill_in 'Amount', with: work_order.amount + 1
     click_on 'Update'
 
-    expect_budget_approvals_restarted(work_order)
+    expect_budget_approvals_restarted(work_order, page)
 
     login_as(work_order.budget_approvers.first)
     visit "/proposals/#{work_order.proposal.id}"
@@ -70,7 +70,7 @@ feature "post-approval modification" do
     linear_approval_statuses(work_order.proposal)
   end
 
-  def expect_budget_approvals_restarted(work_order)
+  def expect_budget_approvals_restarted(work_order, page)
     work_order.reload
 
     expect(work_order.status).to eq('pending')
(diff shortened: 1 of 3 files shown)


pkarman:
cc @phirefly re https://github.com/18F/C2/pull/1130#issuecomment-203962248

spec/features/ncr/work_orders/post_approval_modification_spec.rb
```diff
@@ -82,7 +82,10 @@ def expect_budget_approvals_restarted(work_order)
     ))
 
     approver = work_order.budget_approvers.first
+    expect(approver.email_address).to eq(Ncr::Mailboxes.ba61_tier1_budget.email_address)
     reapproval_mail = deliveries.find { |mail| mail.to.include?(approver.email_address) }
     expect(reapproval_mail.html_part.body).to include('Approve')
+    approver_page_row = page.find(:css, ".step-row.pending.position-1")
```
yozlet: I recommend looking at the Proposal Page object ( `spec/support/page_objects/proposal_page.rb`) and the tests that use it, since page objects are a better home for CSS selectors and other HTML-dependent test details.

yozlet:
Looks good. I'd like us to move to using page objects for this kind of spec (see comment above) but it doesn't have to be in this PR.

spec/features/ncr/work_orders/post_approval_modification_spec.rb
pkarman: That's very neat. I was looking for something like that but didn't immediately see where it was being used. Any pointers?

yozlet: At the moment the only spec that uses `ProposalPage` is in [`spec/features/linear_approvals_spec.rb`](https://github.com/18F/C2/blob/master/spec/features/linear_approvals_spec.rb#L29) but there's more usage of other page objects elsewhere (just search for `@page`)

pkarman:
thanks @yozlet  -- test refactored in 88206c9

yozlet:
Yay! :+1:

PR #1134, opened 2016-03-31
Policies use locale strings

pkarman:
trello: https://trello.com/c/rXd3KLCd/208-policies-should-use-locales-instead-of-hardcoded-strings


yozlet:
:+1:

PR #1135, opened 2016-03-31
Safari and FF require disabling normal form submit

pkarman:


rememberlenny:
Code does what it says

PR #1136, opened 2016-03-31
Admin user delegates

pkarman:
while verifying https://trello.com/c/MRo5TnY3/207-as-a-delegate-i-can-modify-a-request-that-has-not-yet-been-approved-in-my-current-step

I added these /admin/ features to make it easier to review/manage user delegates.

Also fixed a typo and a missing locale translation.

### DIFF
--- a/app/admin/user.rb
+++ b/app/admin/user.rb
@@ -69,6 +69,20 @@ ActiveAdmin.register User do
       row("Roles") { user.roles.map(&:name).join(", ") }
     end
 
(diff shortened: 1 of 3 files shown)


pkarman:
I am inclined to ignore these codeclimate warnings because the spacing is consistent with how it is done elsewhere and the dupe code is a false positive.

rememberlenny:
Looking per comment that CodeClimate is a known issue.

PR #1140, opened 2016-04-04
User timezone

pkarman:
trello: https://trello.com/c/eDRUTNoN/262-show-user-specified-timezone-in-emails

* creates `users.timezone` column
* adds management via `/profile` page
* if set in db, overrides what is set in `browser.timezone` cookie via https://github.com/kbaum/browser-timezone-rails
* `/admin/users` adds timezone

### DIFF
--- a/app/controllers/profile_controller.rb
+++ b/app/controllers/profile_controller.rb
@@ -3,11 +3,11 @@ class ProfileController < ApplicationController
   end
 
   def update
-    first_name = params[:first_name]
-    last_name = params[:last_name]
+    user_params = params.require(:user)
     user = current_user
-    user.first_name = first_name
-    user.last_name = last_name
+    user.first_name = user_params[:first_name]
+    user.last_name = user_params[:last_name]
+    user.timezone = user_params[:timezone] || "Eastern Time (US &amp; Canada)"
     user.save!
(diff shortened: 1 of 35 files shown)


pkarman:
codeclimate issue can't be helped because we are simply overriding the upstream gem method.

app/controllers/profile_controller.rb
```diff
@@ -3,11 +3,11 @@ def show
...
   def update
-    first_name = params[:first_name]
-    last_name = params[:last_name]
+    user_params = params.require(:user)
     user = current_user
-    user.first_name = first_name
-    user.last_name = last_name
+    user.first_name = user_params[:first_name]
+    user.last_name = user_params[:last_name]
+    user.timezone = user_params[:timezone] || "Eastern Time (US &amp; Canada)"
```
yozlet: Does it matter that this is (slightly) different from the `DEFAULT_TIMEZONE` in `app/models/user.rb`?

pkarman: good catch, @yozlet 

The default value will now be whatever the browser timezone is, and if left empty, will default to `User::DEFAULT_TIMEZONE`.

yozlet:
Looks great!

PR #1141, opened 2016-04-04
add comments via ajax

nickbristow:
Its not the best, but its pretty straightforward. 

Get the text value and current user, and adds a new comment. All the HTML is just copied from the page. 

It does actually submit the comment so its there when you refresh. 

@rememberlenny @yozlet


nickbristow:
The ajax does return a failed response...but it does post

rememberlenny:
@nickbristow Looks good! Can you put the JS in the `redesign/detail.js` file?

nickbristow:
@rememberlenny sure, uh oh failed a test.

rememberlenny:
@nickbristow At this point, dont worry too much about the tests

nickbristow:
@rememberlenny ok, should be good to merge then. Added a hidden div with the current user so the JS could grab it.

rememberlenny:
@nickbristow Getting merge conflicts

nickbristow:
coool

PR #1142, opened 2016-04-04
added fake ajax observer functionality

nickbristow:


nickbristow:
@rememberlenny this should be good

rememberlenny:
@nickbristow Looks great!

PR #1146, opened 2016-04-05
refactor test to avoid let() and before()

pkarman:
seeing DeadClient failure trigger psql constraint errors. trying to avoid those here using same approach as elsewhere.


rememberlenny:
Looks great. I'll be excited if this actually prevents the deadclient errors!

PR #1147, opened 2016-04-06
fix versions.item_type for older data

pkarman:
trello: https://trello.com/c/V4fvtjy7/321-500-error-on-proposal-history

see trello code for full explanation.

### DIFF
--- a/db/migrate/20160406142950_fix_versions_item_type_for_steps.rb
+++ b/db/migrate/20160406142950_fix_versions_item_type_for_steps.rb
@@ -4,6 +4,6 @@ class FixVersionsItemTypeForSteps < ActiveRecord::Migration
   end
 
   def down
-    execute "UPDATE versions SET item_type = 'Approval' WHERE item_type = 'Steps::Approval'"
+    execute "UPDATE versions SET item_type = 'Steps::Approval' WHERE item_type = 'Approval'"
   end
 end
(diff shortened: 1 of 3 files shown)


db/migrate/20160406142950_fix_versions_item_type_for_steps.rb
```diff
@@ -0,0 +1,9 @@
+class FixVersionsItemTypeForSteps < ActiveRecord::Migration
+  def up
+    execute "UPDATE versions SET item_type = 'Steps::Approval' WHERE item_type = 'Approval'"
+  end
+
+  def down
+    execute "UPDATE versions SET item_type = 'Approval' WHERE item_type = 'Steps::Approval'"
```
yozlet: This rollback affects _all_ `Steps::Approval` rows, not just those updated in the migration. Maybe limit with a date clause? Or am I being overly cautious? (Probably.)

pkarman: we don't want any `versions` rows with `item_type='Approval'` because there is no such class as `Approval`. The date is just for archeological purposes.

pkarman: maybe it would be better to not have a `down` at all. Under what conditions would we ever want to roll this change back to a broken state?

pkarman: @yozlet I changed the `down` to be identical to the `up` -- we're only fixing data, not the schema itself, and we don't ever want to roll this change back.

yozlet:
Makes sense to me!

PR #1153, opened 2016-04-07
Reindex on new observer

pkarman:
ticket: https://18f.uservoice.com/admin/tickets/184

Proposals are automatically updated in the search index whenever they, or their associated records, change.

Associated observers were not triggering that automatic reindexing.

This PR adds tests for reindexing, and fixes the reindex-on-observation-change bug.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -92,6 +92,7 @@ group :test do
   gem "shoulda-matchers"
   gem "simplecov"
   gem "site_prism"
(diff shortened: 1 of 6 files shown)


pkarman:
codeclimate warnings are all irrelevant. We *want* class variables, similar to how `Mail::TestMailer` does it.

yozlet:
:+1:

PR #1154, opened 2016-04-07
turn off sidebar filters on /admin/user_delegates because un-needed and slow

pkarman:
filters are useless on the `/admin/user_delegates` index page and just slow down response time.


yozlet:
:+1:

PR #1156, opened 2016-04-11
fixed dev.rake, changed def file > def temp_file

nickbristow:
rake dev:prime was failing due to an issue with the method name "file"

Changed to `temp_file`


jessieay:
:+1:

PR #1157, opened 2016-04-11
Automatically generate text only emails

nickbristow:
Added actionmailer-text gem
 Removed mailer text templates
 Removed text translations that were used in text templates
 Made URL's in translations double quotes, to fix issue with actionmailer-text gem
 Fixed bug in comment mailer preview. Preview need non update comment
 https://trello.com/c/oCg3xSTb/282-auto-generate-plain-text-emails-from-html


jessieay:
we paired on this so I obvi think it looks good :)

@nickbristow can you fix the two small code climate offenses?

jessieay:
there's also one failing test: /spec/mailers/welcome_mailer_spec.rb:5 -- looking for a deleted locale. Should update to have it look for `I18n.t("mailer.welcome_mailer.welcome_notification.para2_html", help_url: help_url(''))`

jessieay:
This is amazing. Thanks for deleting so many things @nickbristow and thanks to @rememberlenny for the original inspiration behind the trello card that inspired this PR

PR #1160, opened 2016-04-11
API v2

pkarman:
trello: https://trello.com/c/zi6YR2jB/195-outline-api-needs-design-for-front-end

API, version 2

* Provides Oauth capabilities native to the C2 (users can create oauth apps and generate keys).
* Can search, fetch, create and update proposals based on the `client_slug` of the user.
* API client repo created in https://github.com/18F/c2-api-client-ruby

Currently deployed on c2-dev-pkarman.18f.gov for QA purposes.


pkarman:
the one codeclimate warning is a red herring, imo.

config/routes.rb
```diff
@@ -19,13 +23,11 @@
   # mandrill-rails
   resource :inbox, controller: "inbox", only: [:show, :create]
 
-  namespace :api do
-    scope :v1 do
-      namespace :ncr do
-        resources :work_orders, only: [:index]
+  if ENV["API_ENABLED"] == "true"
+    namespace :api do
+      scope :v2 do
```
jessieay: I think it makes sense to namespace the version, but do we need to make this v2? afaik nobody uses v1

app/controllers/api/proposals_controller.rb
```diff
@@ -0,0 +1,109 @@
+module Api
+  class ProposalsController < BaseController
```
jessieay: should we namespace the controllers with the version?

spec/support/request_spec_helper.rb
```diff
@@ -1,6 +1,7 @@
 module RequestSpecHelper
   def get_json(url)
     get(url)
+    puts response.pretty_inspect
```
jessieay: I think this was left it accidentally :)

spec/support/request_spec_helper.rb
```diff
@@ -22,4 +23,12 @@ def options(*args)
       copy_session_variables!
     end
   end
+
+  def mock_api_doorkeeper_pass
```
jessieay: looks like this is an exact dupe of the method in `spec/support/controller_spec_helper.rb` -- possible to share logic?

.env.test
```diff
@@ -0,0 +1 @@
+API_ENABLED: true
```
jessieay: oooh we have some other env vars in `rails_helper` - perhaps move them into here?

app/controllers/api/base_controller.rb
```diff
@@ -1,19 +1,27 @@
 module Api
   class BaseController < ApplicationController
-    before_action :fail_if_not_enabled
-    skip_before_action :authenticate_user!
-    skip_before_action :check_disabled_client
+    prepend_before_action :doorkeeper_authorize!
```
jessieay: woah I've never used `prepend_before_action` before -- v cool.

app/controllers/api/base_controller.rb
```diff
@@ -1,19 +1,27 @@
...
+
+    protect_from_forgery with: :null_session, if: proc { |c| c.request.format.json? }
```
jessieay: can we assume all API requests will be json? (aka - do we need this `if` ?)

app/controllers/api/proposals_controller.rb
```diff
@@ -0,0 +1,109 @@
...
+      render json: ProposalSerializer.new(proposal).to_json
+    end
+
+    def index
+      listing = ProposalListingQuery.new(current_user, params)
+      begin
+        proposals = listing.query
+        render json: ProposalSearchSerializer.new(proposals).to_json
+      rescue SearchBadQuery, SearchUnavailable => error
+        render json: { error: error.message }, status: 500
```
jessieay: should we include a `message` key  with a generic value here as well to be consistent with other error json in this controller?

app/controllers/api/proposals_controller.rb
```diff
@@ -0,0 +1,109 @@
...
+
+    def update_client_data_instance
+      @client_data_instance = proposal.client_data
+      client_model_class = @client_data_instance.class
+      @client_data_instance.assign_attributes(client_model_class.permitted_params(params, @client_data_instance))
+    end
+
+    def update_proposal
+      update_client_data_instance
+      if errors.any?
```
jessieay: how about checking for `if errors.empty?` instead and putting the logic currently in `else` within the `if`.

Reasoning: I usually prefer to handle the "if all goes well" case first, return `false` second

app/controllers/client_data_controller.rb
```diff
@@ -25,6 +25,8 @@ def edit
   end
 
   def update
+    @client_data_instance.assign_attributes(filtered_params)
```
jessieay: ooooh cool!

jessieay:
this look amaaaaazing. just a few small comments from me. question: do we want to document this API anywhere? :grimacing:

.env.test
pkarman: good catch. 25278b9

pkarman:
@jessieay great review, thanks.

as far as API versioning is concerned, I went with v2 just because I wasn't sure what might be in the wild, esp w.r.t. forks, etc. And it *is* the second version, even if v1 was never enabled anywhere.

jessieay:
@pkarman coolio, re-running CI (ghost failure ugh!)  will merge once it passes

thoughts on my question re: documentation? I liked our swagger approach on e-manifest, but it did feel like a bit much. Perhaps more than we need here. Thoughts on how we'd like to document this? I've used https://github.com/zipmark/rspec_api_documentation and GitHub wikis in the past.

pkarman:
re: documentation, I agreed that swagger feels like too much at this stage.

I'd be inclined to simply update `doc/api.md` since that already exists and is now out of date.

I'd like @adelevie to weigh in on the current implementation though, before we document. I've a feeling the implementation here might change.

Let's wait to merge this till we get some more feedback and get the docs updated.

jessieay:
@pkarman spec failure here seems legit (`spec/requests/set_user_after_login_spec.rb:27`) -- related to env var

PR #1161, opened 2016-04-11
Clarified how active_observer was being used

nickbristow:
Changed active_observer to only_observers
 Changed activ_observer? to observer_only?
 Move comment listeners method to proposal as subscribers_except_future_step_users
 Use only_observers where appropriate
 https://trello.com/c/3KFzjQCC/261-investigate-use-of-active-observers

### DIFF
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -173,6 +173,11 @@ class Proposal < ActiveRecord::Base
     results.compact.uniq
   end
 
+  def subscribers_except_future_step_users
+    results = currently_awaiting_step_users + individual_steps.completed.map(&:user) + observers + [requester]
+    results.compact.uniq
+  end
+
   def subscribers_except_delegates
     subscribers - delegates
   end
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -46,17 +46,4 @@ class Comment < ActiveRecord::Base
   def add_user_as_observer
     proposal.add_observer(user)
   end
-
-  # All of the users who should be notified when a comment is created
-  # This is basically Proposal.users _minus_ future approvers
-  def listeners
-    users_to_notify = Set.new
-    users_to_notify += proposal.currently_awaiting_step_users
-    users_to_notify += proposal.individual_steps.completed.map(&:user)
-    users_to_notify += proposal.observers
-    users_to_notify << proposal.requester
-    # Creator of comment doesn't need to be notified
-    users_to_notify.delete(user)
-    users_to_notify
-  end
 end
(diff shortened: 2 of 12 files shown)


jessieay:
@pkarman would you mind reviewing this? Nick and I paired on it so it already has two sets of eyes but there are some naming changes we weren't 100% confident in. Would :heart: your input!

app/models/proposal.rb
```diff
@@ -173,6 +173,11 @@ def subscribers
     results.compact.uniq
   end
 
+  def subscribers_except_future_step_users
+    results = currently_awaiting_step_users + individual_steps.completed.map(&:user) + observers + [requester]
```
pkarman: should it be `individual_steps.completed.map(&:completed_by)` instead of `&:user` ? Do we want to send to system mailboxes or the human individuals?

pkarman:
these naming changes make sense to me. +1

Only comment was to clarify business rule intent. I'm not sure of the answer; @phirefly should weigh in.

app/models/proposal.rb
nickbristow: I think the humans is what we want.

jessieay: I was under the impression that we *never* email delegates (always email mailboxes instead)...but now that I think it over more, that is more of an assumption than anything else. 

paging @phirefly for clarification -- do we want to email mailboxes when a comment is created, or if there is a delegate who already approved that proposal, do we want the comment email to go to that delegate?

current behavior is that we email mailboxes only, not delegates

jessieay: @nickbristow if you look at the `listeners` method we deleted / rewrote, the logic in there grabs the "step users", which are in some cases shared mailboxes. Like peter said, if we want to email people, we need to look at "step completers", too. Moving in that direction may make sense, but that would be a change from what we have currently (and, like I said above, I am pretty sure that we never email completers atm....but perhaps that should change)

phirefly:
when a comment is posted, the following should happen
1. send an email to delegate who already approved that proposal
2. send an email to the mailbox. afaik, this isn't currently happening but our support requests show that this is a need. this is an opportunity to update this behavior. Pls ping with questions/feedback.

pkarman:
Based on @phirefly comment, I think what we want is a unique union of `user` and `completer`.

jessieay:
@pkarman @phirefly 

The current behavior for comment emails is that we send them to the step user, which is either the approving official or a mailbox in the world of NCR.

We do not send emails to delegates.

It sounds like we want a union of user and completer like Peter said. That is slightly separate from this PR, which is a refactor of existing behavior. Ok to work on that separately? (happy to work on it next) 

one benefit of working on it separately is we could implement the change for multiple emails, not just comment email. 

card for change in behavior: https://trello.com/c/Jr7CubpI/382-send-emails-to-step-completer-and-user

pkarman:
+1 to merge this as-is then.

PR #1162, opened 2016-04-11
Refactor comment specs

jessieay:
* Move data checking into controller spec (out of feature spec)
* Done while investigating https://trello.com/c/1XMufWy1, which I was
  unable to get to the source of


yozlet:
👍

PR #1165, opened 2016-04-13
Move more step-related logic to StepManager concern

jessieay:
* Remove duplicate specs (were in both Proposal and StepManager spec
  files)
* Move specs to StepManager spec when for method in StepManager
* Use factories in specs rather than initializing objects


jessieay:
@pkarman finally got the :ghost: failures to go away

PR #1167, opened 2016-04-13
Steps would be randomly missing from emails

nickbristow:
Changed class in proposal for
 individual steps from "Steps:Individual" to
 "Step"
 Scope "individual" limits steps to "Approval"
 and "Purchase"
 https://trello.com/c/Tsg7daOb/389-sometimes-steps-don-t-show-up

### DIFF
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -37,11 +37,10 @@ class Proposal < ActiveRecord::Base
   visitable # Used to track user visit associated with processed proposal
 
   has_many :steps
-  has_many :individual_steps, ->{ individual }, class_name: "Steps::Individual"
   has_many :approval_steps, class_name: "Steps::Approval"
   has_many :purchase_steps, class_name: "Steps::Purchase"
-  has_many :completers, through: :individual_steps, source: :completer
-  has_many :api_tokens, through: :individual_steps
+  has_many :completers, through: :steps, source: :completer
+  has_many :api_tokens, through: :steps
   has_many :attachments, dependent: :destroy
   has_many :comments, -> { order(created_at: :desc) }, dependent: :destroy
 
--- a/app/models/step.rb
+++ b/app/models/step.rb
@@ -8,6 +8,7 @@ class Step < ActiveRecord::Base
     state :completed
   end
 
+  has_one :api_token, -> { fresh }, foreign_key: "step_id"
   belongs_to :user
   belongs_to :proposal, touch: true
   belongs_to :completer, class_name: "User"
@@ -18,8 +19,8 @@ class Step < ActiveRecord::Base
 
   validates :proposal, presence: true
   validates :user_id, uniqueness: { scope: :proposal_id }, allow_blank: true
-
-  scope :individual, -> { where(type: ["Steps::Approval", "Steps::Purchase"]).order("position ASC") }
+  scope :individual, -> { where.not(type: ["Steps::Serial", "Steps::Parallel"]).order("position ASC") }
+  scope :with_users, -> { includes :user }
 
   statuses.each do |status|
     scope status, -> { where(status: status) }
--- a/app/models/concerns/client_data_mixin.rb
+++ b/app/models/concerns/client_data_mixin.rb
@@ -10,7 +10,6 @@ module ClientDataMixin
 
     has_one :proposal, as: :client_data
     has_many :steps, through: :proposal
-    has_many :individual_steps, -> { individual }, class_name: "Steps::Individual", through: :proposal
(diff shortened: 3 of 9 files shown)


jessieay:
See trello card for some interesting SQL inconsistencies we were seeing with the old definition!

jessieay:
note: steps would also sometimes randomly be missing from web views. :(

pkarman:
the test failure here seems legit.

nickbristow:
@pkarman looking into that

jessieay:
@pkarman ready for re-review

app/models/proposal.rb
```diff
@@ -37,11 +37,10 @@ class Proposal < ActiveRecord::Base
   visitable # Used to track user visit associated with processed proposal
 
   has_many :steps
-  has_many :individual_steps, ->{ individual }, class_name: "Steps::Individual"
   has_many :approval_steps, class_name: "Steps::Approval"
   has_many :purchase_steps, class_name: "Steps::Purchase"
-  has_many :completers, through: :individual_steps, source: :completer
-  has_many :api_tokens, through: :individual_steps
+  has_many :completers, through: :steps, source: :completer
```
pkarman: does this need a scope to `individual` -- the parent (root) step will never have a completer.

app/models/step.rb
```diff
@@ -18,8 +19,8 @@ class Step < ActiveRecord::Base
 
   validates :proposal, presence: true
   validates :user_id, uniqueness: { scope: :proposal_id }, allow_blank: true
-
-  scope :individual, -> { where(type: ["Steps::Approval", "Steps::Purchase"]).order("position ASC") }
+  scope :individual, -> { where.not(type: ["Steps::Serial", "Steps::Parallel"]).order("position ASC") }
```
pkarman: nice

pkarman:
this looks great.

just one question about the `completers` association. I don't know how nil values are treated.

app/models/proposal.rb
jessieay: yeah you're right -- it won't. as-is it just returns nothing if there is no completer for a step, so it will return the same thing whether we scope to `individual` or not.

benefit of scoping would be to clarify that a root would never have a completer, but I am not sure how helpful that would be.

example output for proposal that has no completers:

```
2] pry(main)> Proposal.last.steps
  Proposal Load (0.9ms)  SELECT  "proposals".* FROM "proposals"  ORDER BY "proposals"."id" DESC LIMIT 1
  Step Load (0.8ms)  SELECT "steps".* FROM "steps" WHERE "steps"."proposal_id" = $1  ORDER BY position ASC  [["proposal_id", 4]]
=> [#<Steps::Serial:0x007fb7d34b5450
  id: 12,
  user_id: nil,
  status: "actionable",
  created_at: Mon, 11 Apr 2016 23:02:45 UTC +00:00,
  updated_at: Mon, 11 Apr 2016 23:02:45 UTC +00:00,
  position: 1,
  proposal_id: 4,
  completed_at: nil,
  type: "Steps::Serial",
  parent_id: nil,
  min_children_needed: nil,
  completer_id: nil>,
 #<Steps::Approval:0x007fb7d4fa4a90
  id: 13,
  user_id: 6,
  status: "actionable",
  created_at: Mon, 11 Apr 2016 23:02:45 UTC +00:00,
  updated_at: Mon, 11 Apr 2016 23:02:46 UTC +00:00,
  position: 2,
  proposal_id: 4,
  completed_at: nil,
  type: "Steps::Approval",
  parent_id: 12,
  min_children_needed: nil,
  completer_id: nil>,
 #<Steps::Purchase:0x007fb7d4ebf5d0
  id: 14,
  user_id: 7,
  status: "pending",
  created_at: Mon, 11 Apr 2016 23:02:45 UTC +00:00,
  updated_at: Mon, 11 Apr 2016 23:02:45 UTC +00:00,
  position: 3,
  proposal_id: 4,
  completed_at: nil,
  type: "Steps::Purchase",
  parent_id: 12,
  min_children_needed: nil,
  completer_id: nil>]
[3] pry(main)> Proposal.last.completers
  Proposal Load (0.4ms)  SELECT  "proposals".* FROM "proposals"  ORDER BY "proposals"."id" DESC LIMIT 1
  User Load (0.5ms)  SELECT "users".* FROM "users" INNER JOIN "steps" ON "users"."id" = "steps"."completer_id" WHERE "steps"."proposal_id" = $1  ORDER BY position ASC  [["proposal_id", 4]]
=> []
```

pkarman:
lovely example. thanks.

PR #1168, opened 2016-04-14
API v2 docs

pkarman:
I'm ambivalent about keeping the client-model-specific attributes documented here. It will then be a maintenance burden.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -113,6 +113,8 @@ Attribute | Type | Note
 Consult the [client data models](https://github.com/18F/C2/blob/master/app/models) themselves
 for details on specific required attributes.
 
+All attributes for `client_data` are present in responses, regardless of whether they have a value assigned.
+
 ## Endpoints
 
 ### `GET /api/v2/proposals`
--- a/doc/api.md
+++ b/doc/api.md
@@ -2,31 +2,62 @@
 
 ***If you are running the server and want to enable the API, set the environment variable `API_ENABLED=true`.***
 
-This API is currently in a very limited alpha, and is not enabled in any of our deployments. Note that this API may change at any time, without warning.
+The API design is evolving. Backwards-incompatible changes will result in a version change.
+
+The lastest version is *v2*.
 
 ## Schemas
 
 * All decimals are strings ([more info](https://github.com/rails-api/active_model_serializers/issues/202))
 * All times are in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format
 
-### Step
+### Proposal
+
+Each proposal object is composed of a few attributes and some child objects.
+
+### Attributes
 
 Attribute | Type | Note
 --- | --- | ---
-`id` | integer |
-`status` | string | Can be `pending`, `actionable`, `approved`, or `canceled`
-`user` | [User](#user) | a.k.a. "the approver"
+`id` | integer | primary identifier
+`status` | string | one of: `pending`, `cancelled`, `complete`
+`created_at` | string | timestamp
+`updated_at` | string | timestamp
+`client_data_type` | string | name of the client data class
 
-### [NCR](overview.md#national-capitol-region-ncr-service-centers) Work Order
+### Requester
+
+Who created the proposal. Refers to a [User](#user) record.
+
+Attribute | Type | Note
+--- | --- | ---
+`id` | integer | primary identifier
+`created_at` | string | timestamp
+`updated_at` | string | timestamp
+
+### Steps
+
+Array of step objects.
+
+Attribute | Type | Note
+--- | --- | ---
+`id` | integer | primary identifier
+`status` | string | Can be `pending`, `actionable`, `approved`, or `canceled`
+`user` | [User](#user) | who needs to complete the step
+
+### Client data
+
+The `client_data` object is the client-specific set of attributes. The attribute names
+will vary based on the client. For example, [NCR](overview.md#national-capitol-region-ncr-service-centers) Work Order:
 
 Attribute | Type | Note
 --- | --- | ---
 `amount` | string (decimal) | The cost of the work order
 `building_number` | string | ([full list](../config/data/ncr.yaml))
-`code` | `null` for BA61, string for BA80 | Identifier for the type of work
+`work_order_code` | `null` for BA61, string for BA80 | Identifier for the type of work
 `description` | string |
 `emergency` | boolean | Whether the work order was pre-approved or not (can only be `true` for BA61)
-`expense_type` | string | `BA61` or `BA80`
+`expense_type` | string | `BA60`, `BA61` or `BA80`
 `id` | integer |
 `name` | string |
 `not_to_exceed` | boolean | If the `amount` is exact, or an upper limit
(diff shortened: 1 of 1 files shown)


doc/api.md
```diff
@@ -2,31 +2,89 @@
 
 ***If you are running the server and want to enable the API, set the environment variable `API_ENABLED=true`.***
 
-This API is currently in a very limited alpha, and is not enabled in any of our deployments. Note that this API may change at any time, without warning.
+The API design is evolving. Backwards-incompatible changes will result in a version change.
```
jessieay: since we essentially deleted v1, should we mention that here and and/or confirm that we will not be removing future versions similarly (don't think it's bad that we removed v1, but in the future I imagine we will want to continue to support deployed older API versions)

doc/api.md
```diff
@@ -35,122 +93,73 @@ Attribute | Type | Note
...
+`offset` | alias for `from`
+`status` | one of `pending`, `cancelled`, `complete`
+`start_date` | timestamp string
+`end_date` | timestamp.string
+`text` | full-text search string. See https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#query-string-syntax
+
+#### Example
+
+```bash
+% curl -H 'Authorization: Bearer your-auth-token' \
```
jessieay: is `Bearer` here a token as well? it's mentioned above that we need to "include your Bearer auth token with every API request"

If it is a token, perhaps we can include it in the same format as your-auth-token. Maybe both all caps or something?

doc/api.md
```diff
@@ -2,31 +2,89 @@
...
+```bash
+% export MY_OAUTH_KEY=your-key-here
+% export MY_OAUTH_SECRET=your-secret-here
+% export MY_CREDS=`echo "$MY_OAUTH_KEY:$MY_OAUTH_SECRET" | base64`
+% curl -i -X POST -H "Authorization: Basic $MY_CREDS" \
+  -d 'grant_type=client_credentials' \
+  https://cap.18f.gov/oauth/token
+```
+
+Take note of the `access_token` string value in the response. You will use it below.
```
jessieay: where do we use it? we have something called `your-auth-token` below, but not `access_token` -- perhaps we should stay consistent?

doc/api.md
```diff
@@ -2,31 +2,89 @@
...
+### Steps
+
+Array of step objects.
+
+Attribute | Type | Note
+--- | --- | ---
+`id` | integer | primary identifier
 `status` | string | Can be `pending`, `actionable`, `approved`, or `canceled`
-`user` | [User](#user) | a.k.a. "the approver"
+`user` | [User](#user) | who needs to complete the step
```
jessieay: looks like we deleted the `User` docs below - should w re-add so this is more helpful?

jessieay:
> I'm ambivalent about keeping the client-model-specific attributes documented here. It will then be a maintenance burden.

Agreed -- I do think it's nice to provide an example response but perhaps we can work on automating that later (rspec-api-documentation makes it pretty easy)

doc/api.md
```diff
@@ -2,31 +2,89 @@
 
 ***If you are running the server and want to enable the API, set the environment variable `API_ENABLED=true`.***
 
-This API is currently in a very limited alpha, and is not enabled in any of our deployments. Note that this API may change at any time, without warning.
+The API design is evolving. Backwards-incompatible changes will result in a version change.
+
+The lastest version is *v2*.
+
+## Clients
```
jessieay: maybe include this in the `Authentication` section to clarify that it's an alternative to auth-ing manually

doc/api.md
```diff
@@ -35,122 +93,73 @@ Attribute | Type | Note
...
-  // ...
-]
+```bash
+% curl -i -X POST -H 'Content-Type: application/json' \
+  -H 'Authorization: Bearer your-auth-token' \
+  --data @proposal.json \
+  https://cap.18f.gov/api/v2/proposals
+```
+
+where `proposal.json` looks like:
```
jessieay: should we include or link to required attributes?

doc/api.md
```diff
@@ -35,122 +93,73 @@ Attribute | Type | Note
...
 All are optional.
 
 Name | Values
 --- | ---
-`limit` | an integer >= 0
-`offset` | an integer >= 0
+`size` | an integer >= 0
+`from` | an integer >= 0
+`limit` | alias for `size`
+`offset` | alias for `from`
```
jessieay: can you provide a bit more explanation on how offset works?

doc/api.md
```diff
@@ -35,122 +93,73 @@ Attribute | Type | Note
...
 
 All are optional.
 
 Name | Values
 --- | ---
-`limit` | an integer >= 0
-`offset` | an integer >= 0
+`size` | an integer >= 0
+`from` | an integer >= 0
+`limit` | alias for `size`
```
jessieay: is there a default limit?

doc/api.md
```diff
@@ -2,31 +2,89 @@
...
+`id` | integer | primary identifier
 `status` | string | Can be `pending`, `actionable`, `approved`, or `canceled`
-`user` | [User](#user) | a.k.a. "the approver"
+`user` | [User](#user) | who needs to complete the step
 
-### [NCR](overview.md#national-capitol-region-ncr-service-centers) Work Order
+### Client data
+
+The `client_data` object is the client-specific set of attributes. The attribute names
+will vary based on the client. For example, [NCR](overview.md#national-capitol-region-ncr-service-centers) Work Order:
```
jessieay: will keys be present even if the values for a specific client data instance are not? eg: if `description` is nil, do we return the `description` key? (I think we should!) And if so, what is the value? Nil / empty string?

doc/api.md
pkarman: `Bearer` is literal. Analogous to `Basic` for basic auth.

doc/api.md
pkarman: `access_token` is what the OAuth spec uses and what doorkeeper calls it. I will change to call is `access token` instead of `auth token`.

doc/api.md
pkarman: I believe the serializer class will always include every client_data attribute, even if nil.

doc/api.md
pkarman: yes. it is defined in `Proposal::MAX_SEARCH_RESULTS`.

doc/api.md
jessieay: cool, including that detail here would be helpful

doc/api.md
pkarman: just like for SQL. If you want the first "page" of results, your `offset` (`from`) is `0`. That's the default.

ES uses `size/from` and SQL uses `limit/offset` -- one good reason to switch to all ES internally so we do not have to worry about both sets of terms.

doc/api.md
jessieay: ah ok I saw

> include your Bearer auth token with every API request

and was thinking we'd see some type of auth token named `bearer` :)

doc/api.md
jessieay: can we include that info here? not sure how comprehensive we want these docs to be, but record limits seems potentially problematic if someone doesn't know there is a default and thinks they are returning all

doc/api.md
pkarman: attributes are client-specific. This is where I worry about the scope of the API docs. If we have 20 clients, that's a lot of docs to keep in sync.

Maybe we should have an API endpoint that returns the meta about the model? Like swagger, but built-in so the model just interrogates itself.

`GET /api/v2/about/client_data/ncr_work_order`

```
{
  // list of attributes, types, etc
}
```

jessieay: sure or I was thinking for now we could link to the client data models so someone could see which attrs are required

pkarman:
thanks @jessieay -- edits made.

jessieay:
:heart_eyes_cat:

PR #1169, opened 2016-04-14
Create check for query param to load alternate view

rememberlenny:
On `proposal#show` check for `?detail=view` to load alternate `show_next.html.haml`

### DIFF
--- a/app/controllers/proposals_controller.rb
+++ b/app/controllers/proposals_controller.rb
@@ -14,7 +14,7 @@ class ProposalsController < ApplicationController
     @proposal = proposal.decorate
     cookies[:detail] = params[:detail]
     mode = cookies[:detail]
-    if mode == "view"
+    if mode == "new"
       @client_data_instance ||= proposal.client_data
       render "show_next"
     end
--- a/app/controllers/proposals_controller.rb
+++ b/app/controllers/proposals_controller.rb
@@ -15,7 +15,6 @@ class ProposalsController < ApplicationController
     cookies[:detail] = params[:detail]
     mode = cookies[:detail]
     if mode == "view"
-      @history = HistoryList.new(proposal)
       @client_data_instance ||= proposal.client_data
       render "show_next"
     end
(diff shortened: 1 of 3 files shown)


app/controllers/proposals_controller.rb
```diff
@@ -12,6 +12,13 @@ class ProposalsController < ApplicationController
 
   def show
     @proposal = proposal.decorate
+    cookies[:detail] = params[:detail] ? params[:detail] : nil
+    @mode = cookies[:detail]
```
jessieay: I don't think this needs to be an instance var since we aren't using it in the view

app/controllers/proposals_controller.rb
```diff
@@ -12,6 +12,13 @@ class ProposalsController < ApplicationController
 
   def show
     @proposal = proposal.decorate
+    cookies[:detail] = params[:detail] ? params[:detail] : nil
```
jessieay: this can be simplified to:

```ruby
cookies[:detail] = params[:detail]
```

app/controllers/proposals_controller.rb
```diff
@@ -12,6 +12,13 @@ class ProposalsController < ApplicationController
...
+    @mode = cookies[:detail]
+    if @mode == "view"
```
jessieay: did you want to write a controller test for this?

jessieay:
glad we are breaking this down into smaller chunks! you mentioned in Slack that we'd be adding tests along with each chunk, should we have a test for this, then?

rememberlenny:
@jessieay Added!

pkarman:
+1 looks sane to me.

@jessieay I'll let you merge since you did the initial review.

yozlet:
The string `view` comes from when it was choosing between `view` and `edit` mode. Since we no longer have `edit` mode, perhaps we should change the cookie string to 'new', since that's what it's about (and will be more memorable).

jessieay:
looks good to me too - let me know what you think about yoz's comment!

rememberlenny:
@jessieay Updated per @yozlet suggestions

jessieay:
cool will merge as soon as CI passes

rememberlenny:
Come on CI....

PR #1171, opened 2016-04-14
New details page/add history instance

rememberlenny:
# Important

This is dependent on https://github.com/18F/C2/pull/1169

## What

Parse the history on a proposal and make it accessible as a instance variable to display in show_next.html.haml

## Trello

https://trello.com/c/0OrpiMHD/395-transition-design-new-details-branch-into-master-through-smaller-prs

### DIFF
--- a/app/decorators/step_decorator.rb
+++ b/app/decorators/step_decorator.rb
@@ -18,7 +18,9 @@ class StepDecorator < Draper::Decorator
   end
 
   def action_status
-    "Step #{position - 1}: #{completed} by #{completed_by.full_name}"
+    I18n.t('decorators.steps.summary.action', display_position: object.position - 1,
+                                              completed: completed,
+                                              name: object.completed_by.full_name)
   end
 
   def label
@@ -43,6 +45,10 @@ class StepDecorator < Draper::Decorator
 
   def waiting_text
     get_step_text("status.waiting")
+  end
+
+  def completed
+    get_step_text(:completed)
   end
 
   private
--- a/app/decorators/step_decorator.rb
+++ b/app/decorators/step_decorator.rb
@@ -17,6 +17,10 @@ class StepDecorator < Draper::Decorator
     end
   end
 
+  def action_status
+    "Step #{position - 1}: #{completed} by #{completed_by.full_name}"
+  end
+
   def label
     get_step_text(:label)
   end
--- a/app/decorators/step_decorator.rb
+++ b/app/decorators/step_decorator.rb
@@ -18,7 +18,7 @@ class StepDecorator < Draper::Decorator
   end
 
   def action_status
-    I18n.t('decorators.steps.summary.action', display_position: object.position - 1,
+    I18n.t("decorators.steps.summary.action", display_position: object.position - 1,
                                               completed: completed,
                                               name: object.completed_by.full_name)
   end
--- a/app/presenters/history_event.rb
+++ b/app/presenters/history_event.rb
@@ -18,7 +18,7 @@ class HistoryEvent < SimpleDelegator
     Attachment.find(item_id)
   end
 
-  def decorate
+  def decorate_item
     reify.decorate
   end
 
--- a/app/presenters/history_event.rb
+++ b/app/presenters/history_event.rb
@@ -1,5 +1,6 @@
 require "delegate"
 
+# HistoryEvent wraps & delegates to C2Version
 class HistoryEvent < SimpleDelegator
   include ActiveModel::Conversion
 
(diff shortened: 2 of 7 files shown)


rememberlenny:
Passing this to @yozlet

yozlet:
CodeClimate is suggesting I use delegation in a class that's already a delegator. Silly CodeClimate.

pkarman:
@rememberlenny @yozlet this test failure seems legit

rememberlenny:
@pkarman Yep! @yozlet is on it

app/decorators/step_decorator.rb
```diff
@@ -17,6 +17,10 @@ def display_status
     end
   end
 
+  def action_status
```
jessieay: test for this?

app/presenters/history_event.rb
```diff
@@ -0,0 +1,47 @@
+require "delegate"
+
+# HistoryEvent wraps & delegates to C2Version
```
jessieay: we haven't generally been doing class descriptions in C2. We can discuss if we want to move in that direction, but as of right now it's not something we're doing.

app/presenters/history_event.rb
```diff
@@ -0,0 +1,47 @@
...
+class HistoryEvent < SimpleDelegator
+  include ActiveModel::Conversion
+
+  def event_type
+    model.event
+  end
+
+  def comment_text
+    # For some reason, paper_trail's "reify" doesn't work with comment versions
```
jessieay: zap this?

app/presenters/history_event.rb
```diff
@@ -0,0 +1,47 @@
...
+    Comment.find(item_id).comment_text
+  end
+
+  def attachment
+    # "reify" not working with attachment versions either
```
jessieay: zap this? or try to make it work?

spec/presenters/history_list_spec.rb
```diff
@@ -0,0 +1,20 @@
+describe HistoryList do
+  describe 'contents' do
```
jessieay: usually in the `describe` blocks we include the name of the method we are testing. looks like this one is testing `events` so I'd have this as:

```ruby
describe "#events" do
  it "returns proposal creation first" do
    ...
  end
end
```

spec/presenters/history_list_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    it "starts with proposal creation" do
+      proposal = create(:proposal)
+
+      history = HistoryList.new(proposal)
+      expect(history.events.first.event).to eq("create")
+    end
+
+    describe "filtering" do
```
jessieay: perhaps use `context ` here rather than `describe` since we're still testing `#events`

```ruby
context "events include work order updates" do
  it "excludes client data events" do
    ...
  end
end
```

app/presenters/history_event.rb
pkarman: do you want `reify` or just `item` ?

app/presenters/history_event.rb
```diff
@@ -0,0 +1,47 @@
...
+    Attachment.find(item_id)
```
pkarman: isn't this just `item` ?

app/presenters/history_event.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def to_partial_path
+    "proposals/details/history/" + partial_name
+  end
+
+  private
+
+  def partial_name
+    case item_type
```
jessieay: instead of this case statement we can do:

```ruby
item_type.split("::").last
```

...I think this will work for all cases :)

app/presenters/history_event.rb
yozlet: I added this because there's no other indication that it relates to C2Version, which makes reading it much easier. Or is deducing that from HistoryList enough?

jessieay: I'd say my general philosophy is that the class itself plus related specs should be enough documentation unless something really unusual is going on. If stuff is still confusing, that might mean we need more tests or more clearly written classes. There are def exceptions, though!

app/presenters/history_event.rb
yozlet: Good point :) Thank you!

app/presenters/history_event.rb
yozlet: I didn't realise that was available. Thank you!

app/decorators/step_decorator.rb
yozlet: Added!

pkarman:
@yozlet I'm fine to merge this.

Have you considered the case where a proposal is restarted? I.e. re-approval is triggered. I'm not sure how the filtering out of "created" events will affect how that looks in the timeline. 

cf. https://github.com/18F/C2/commit/182679ab682350ef87698309abd85e5bceaefd29 where I recently added special comments for completion and restart.

PR #1174, opened 2016-04-14
Add new details page base + summary/approvals/action-bar partials+images

rememberlenny:
# Contents

Adding the layout base for the show_next.

Also adding the page partials that dont require large amounts of changes to controller.

## Trello Reference Link
https://trello.com/c/0OrpiMHD 

- Basic Layout Markup
- Summary Card Markup
- Approval Card Markup
- Action Bar Card Markup


rememberlenny:
Tests clear!

jessieay:
seems like `app/assets/images/numbers/icon-number-2-past.png` might have been added accidentally?

jessieay:
seems like `icon-number-3-pending.png` are dupes of the email images -- if so, can we share?

app/views/proposals/detail/_action.html.haml
```diff
@@ -0,0 +1,25 @@
...
+      %li.fl.modify-button.action-default
+        = link_to("Go back", proposals_path, class: "button secondary fr large")
+      
+      - if policy(@proposal).can_complete?
+        %li.fr.modify-button.action-default.action-approve
+          = form_tag(complete_proposal_path(@proposal, method: "POST", remote: true)) do
+            = submit_tag @proposal.step_text_for_user(:execute_button, current_user), class: "form-button button primary large"
+            = hidden_field_tag :version, @proposal.version
+      
+      - if policy(@proposal).can_edit? && !@proposal.completed?
```
jessieay: I think we can combine this line with the next and do 

`if policy(@proposal).can_edit? && policy(@proposal).not_completed? && policy(@proposal).not_canceled?`

app/views/proposals/detail/_summary.html.haml
```diff
@@ -0,0 +1,24 @@
...
+              = proposal.requester.full_name
+            Requested on:
+            %strong
+              %span
+                = proposal.created_at.strftime("%b %d, %Y at %l:%M%P")
+      .medium-3.column.text-center
+        %span.status-wrap.medium-12.small-12{ class: "status-is-" + @proposal.status }
+          Status:
+          %span.status-value
+            = @proposal.status.capitalize
```
jessieay: have you considered using `capitalized_detailed_status` from `ProposalDecorator`? That's what we are using in emails.

spec/controllers/proposals_controller_spec.rb
```diff
@@ -91,16 +91,22 @@
       end
     end
 
-    context 'details beta' do
+  end
+
+  describe '#show + new details' do
+    before do
```
jessieay: I know we have a lot of `before` blocks in this spec file... I prefer to avoid `before` blocks because I feel that they end up adding bloat to specs (you create data you don't need for subsequent specs) and make them harder to read... so if it were me, I would include these steps in the test itself rather than in a `before`

jessieay:
looking good! just a few small comments from me.

would be cool to see screenshots in these PRs.

Also, trello link?

rememberlenny:
Adding Trello card!

https://trello.com/c/0OrpiMHD

rememberlenny:
@jessieay Updated images in emails!

https://github.com/18F/C2/pull/1174#issuecomment-210177378

app/views/proposals/detail/_action.html.haml
rememberlenny: 👍

spec/controllers/proposals_controller_spec.rb
rememberlenny: @jessieay Got it. I think this particular `before` block is valid because its the flag to trigger the cookie. Without the cookie, none of the following specs will be properly working. 

When we remove the cookie flag, it will be easy to remove the content in the before block.

What do you think?

pkarman: I think what @jessieay is after is something more like

```ruby
def setup_proposal_page
  login_as(user)
  get :show, id: proposal.id, detail: "new"
end
```

and then call `setup_proposal_page` within each `it` as necessary.

rememberlenny:
Clearing my name from this!

spec/controllers/proposals_controller_spec.rb
rememberlenny: @pkarman Got it!

jessieay: @rememberlenny you going to move this? fine if you'd rather not, just don't want to merge if you are still working on the code

jessieay:
looking good! just one final question about spec refactor :) :rainbow:

rememberlenny:
Updated spec based on @pkarman suggestion!

PR #1176, opened 2016-04-14
Inline step icons in emails

jessieay:
* To prevent wackiness if we ever remove / move images in app

https://trello.com/c/FGZuU0RY/387-inline-all-icons-in-emails


jessieay:
I am inclined to ignore to code climate warnings ... :traffic_light:

pkarman:
+1 to ignoring codeclimate warnings

jessieay:
@pkarman CI passing now ;)

PR #1180, opened 2016-04-15
switch diff logic in proposal history.

pkarman:
papertrail gem stores "before" version, so switch diff logic
to always compare the version record against what comes next,
not what came before. This allows the current live record
to be included in the history.

trello: https://trello.com/c/TC59Hwop/325-look-into-comment-history-vs-admin-history-disrepancy

before:
history-fix-before

after:
history-fix-after

### DIFF
--- a/lib/c2_version.rb
+++ b/lib/c2_version.rb
@@ -13,16 +13,12 @@ class C2Version < PaperTrail::Version
     end
   end
 
+  def live_version
+    item
+  end
+
   def diff
-    case event
-    when "create"
-      HashDiff.diff({}, attributes)
-    when "update"
-      prev = previous or fail("No previous version for #{self.pretty_inspect}")
-      HashDiff.diff(prev.attributes, attributes)
-    else
-      # not sure what makes the most sense here...
-      nil
-    end
+    next_version = self.next || live_version
+    HashDiff.diff((attributes || {}), next_version.attributes)
   end
 end
--- a/lib/c2_version.rb
+++ b/lib/c2_version.rb
@@ -13,12 +13,8 @@ class C2Version < PaperTrail::Version
     end
   end
 
-  def live_version
-    item
-  end
-
   def diff
-    next_version = self.next || live_version
+    next_version = self.next || item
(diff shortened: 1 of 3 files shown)


lib/c2_version.rb
```diff
@@ -13,16 +13,12 @@ def attributes
...
-    when "create"
-      HashDiff.diff({}, attributes)
-    when "update"
-      prev = previous or fail("No previous version for #{self.pretty_inspect}")
-      HashDiff.diff(prev.attributes, attributes)
-    else
-      # not sure what makes the most sense here...
-      nil
-    end
+    next_version = self.next || live_version
```
jessieay: papertrail has a method called `next` ? yikes, wish they'd picked a different name

lib/c2_version.rb
```diff
@@ -13,16 +13,12 @@ def attributes
     end
   end
 
+  def live_version
```
jessieay: can we put this below the `diff` method? this is in the spirit of:

> dependencies in a given class should always point downward. That is, the first method in a class should depend on methods further down in the class definition, and those methods should depend on methods even further down, and so on.

which is courtesy of Avdi Grimm / Ruby Tapas

lib/c2_version.rb
```diff
@@ -13,16 +13,12 @@ def attributes
...
+    item
+  end
+
   def diff
```
jessieay: yay getting rid of case statements

jessieay: do we have a unit test for this? if not, should we add one? seems like that would be as valuable as the feature spec. I know this logic is mostly from papertrail and we don't want to test papertrail internals, but there was a logic change here that is important so seems like that should be tested and the unit level somehow?

jessieay:
glad we got to the bottom of this! just one minor suggestion from me re: testing

pkarman:
@jessieay added unit test (good call -- it was tested one level up by multiple classes that used it, but not at the unit level).

I just zapped the additional `live_version` method since it didn't really add much (I had created it while figuring out how best to fetch the original record).

jessieay:
:rainbow: :raised_hands:

PR #1181, opened 2016-04-15
fix tests to use locale strings

pkarman:
should fix failing spec on master.


jessieay:
we had failing tests on master ? ee!

PR #1187, opened 2016-04-18
Add pagination everywhere

pkarman:
trello: https://trello.com/c/puZ8ibnu/396-pagination-for-proposals-query-status-foo

This adds pagination to all /proposals list views, and eliminates the /proposals/archive view in favor of a redirect to `status:completed`.


app/controllers/proposals_controller.rb
```diff
@@ -27,11 +27,11 @@ def index
     @pending_data = listing.pending
     @pending_review_data = listing.pending_review
     @completed_data = listing.completed.alter_query { |rel| rel.limit(@closed_proposal_limit) }
-    @canceled_data = listing.canceled
+    @canceled_data = listing.canceled.alter_query { |rel| rel.limit(@closed_proposal_limit) }
```
jessieay: thoughts on putting `alter_query { |rel| rel.limit(@closed_proposal_limit) }` in a method that we can call here? That way we can have a test for this logic without needing a controller spec.

app/serializers/proposal_search_serializer.rb
```diff
@@ -1,9 +1,24 @@
 class ProposalSearchSerializer < ActiveModel::Serializer
   attributes(
```
jessieay: alpha order these attrs?

app/serializers/proposal_search_serializer.rb
```diff
@@ -1,9 +1,24 @@
...
     :total,
+    :size,
+    :from,
+    :current_page,
     :proposals
   )
 
+  def size
+    object.size
```
jessieay: we we need to define these if they are just delegating to the serialized object?

spec/features/archive_link_spec.rb
```diff
@@ -27,7 +27,21 @@
...
-      expect(page).to_not have_content("View the archive")
+      expect(page).to_not have_content("View all completed")
+    end
+  end
+
+  scenario "adds archive link for canceled proposals" do
+    ClimateControl.modify CLOSED_PROPOSAL_LIMIT: "2" do
+      user = create(:user)
+      3.times.map do |i|
+        create(:proposal, requester: user, status: "canceled")
```
jessieay: can be re-written as 

`create_list(:proposal, 3, requester: user, status: "canceled")`

app/queries/proposal_listing_query.rb
```diff
@@ -121,12 +121,20 @@ def apply_text_filter(proposals_data)
 
   def apply_paging_filter(proposals_data)
     unless params[:text] || params[user.client_model_slug.to_sym]
-      limit = params[:size] || params[:limit] || Proposal::MAX_SEARCH_RESULTS
-      offset = params[:from] || 0
+      limit = (params[:size] || params[:limit] || Proposal::MAX_SEARCH_RESULTS).to_i
```
jessieay: why do we accept either `size` or `limit` here? It might be easier to understand code if we only accept one or the other, but maybe there's a good reason why we need to look for either?

pkarman:
thanks for the good review @jessieay 

I think I've hit all your comments.

app/controllers/proposals_controller.rb
```diff
@@ -26,12 +26,12 @@ def index
     @closed_proposal_limit = ENV.fetch("CLOSED_PROPOSAL_LIMIT", 10).to_i
     @pending_data = listing.pending
     @pending_review_data = listing.pending_review
-    @completed_data = listing.completed.alter_query { |rel| rel.limit(@closed_proposal_limit) }
-    @canceled_data = listing.canceled
+    @completed_data = listing.completed.apply_limit(@closed_proposal_limit)
```
jessieay: 😍

jessieay:
looks great! will qa now

app/controllers/proposals_controller.rb
```diff
@@ -26,12 +26,12 @@ def index
     @closed_proposal_limit = ENV.fetch("CLOSED_PROPOSAL_LIMIT", 10).to_i
     @pending_data = listing.pending
     @pending_review_data = listing.pending_review
```
jessieay: should we paginate this?

jessieay:
@pkarman looks like the 'pending' proposals list is not paginated - intentional?

also, thoughts on paginating search?

pkarman:
@jessieay search is already paginated.

pending is not paginated, purposefully. if users have things to do, they don't want to be hidden on a 'next' page.

jessieay:
@pkarman my bad, had 19 results for search and thought it was over 20 but was not seeing pagination. Added 2 more records and saw pagination 👍 

`pending` reasoning makes sense. merging!

PR #1189, opened 2016-04-18
New hashdiff formatting

yozlet:
- HTML format changes to the HashDiffDecorator output (used both in the old proposal history view and new timeline view)
- HashDiffDecorator also looks up field name translations where possible
- Internal change: now uses a hash to represent the change rather than a position-dependent array (though it still needs to translate from an initial array - see `C2VersionDecorator#update_html`


pkarman:
+1 for the named keys instead of ordered array.

I'm curious about the HTML markup changes. I personally find the `code` formatting easier to read because of the fixed-width font. Is the plan to format these changes via CSS in some future PR?

yozlet:
This markup reflects the design in Amber & Eric's sketches. There are no further styling plans as yet, but I put a `span` around the field name just in case. (Personally, I'm torn over which style I prefer.)

PR #1191, opened 2016-04-19
Anonymous user

pkarman:
trello: https://trello.com/c/8CDaBWU7/381-fix-broken-history-view-seed-data

some C2Version records have no `whodunnit` defined because they are created anonymously.

fixes:

screen shot 2016-04-19 at 10 28 35 am

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -33,7 +33,7 @@ class User < ActiveRecord::Base
   DEFAULT_TIMEZONE = "Eastern Time (US & Canada)"
 
   def self.anonymous
-    new(first_name: 'Unknown', last_name: 'Unknown', email_address: 'unknown@example.com')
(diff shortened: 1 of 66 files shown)


app/models/user.rb
```diff
@@ -32,6 +32,10 @@ class User < ActiveRecord::Base
 
   DEFAULT_TIMEZONE = "Eastern Time (US & Canada)"
 
+  def self.anonymous
+    new(first_name: 'Unknown', last_name: 'User', email_address: 'unknown@example.com')
```
jessieay: thoughts on using the null object pattern? https://robots.thoughtbot.com/rails-refactoring-example-introduce-null-object

pkarman:
NullUser it is

jessieay:
looks good!

PR #1192, opened 2016-04-19
Admin tasks no email

pkarman:
trello: https://trello.com/c/OUj6cUMt/317-as-an-admin-i-can-bypass-email-sends-when-updating-the-overall-status-or-specific-step-status-of-a-request-so-that-i-can-perform

Fully completing a Proposal will, by default, send email to all affected subscribers. This adds a second button that completes without notifying.

NOTE that editing the status of a Step directly via the admin UI does **not** generate any email.


jessieay:
looks great!!! V elegant solution

can QA in an hour or so.

jessieay:
looks good on dev!

PR #1193, opened 2016-04-19
Clarify / update admin specs

jessieay:
* Highlight that email not sent on step update

### DIFF
--- a/spec/features/admin_spec.rb
+++ b/spec/features/admin_spec.rb
@@ -8,7 +8,7 @@ describe "admin" do
   end
 
   it "does not allow editing of user delegates" do
-    login_as_admin_user
+    user = login_as_admin_user
     other_user = create(:user)
     user_delegate = create(:user_delegate, assigner: user, assignee: other_user)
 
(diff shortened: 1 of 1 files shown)


spec/features/admin_spec.rb
```diff
@@ -1,14 +1,14 @@
...
+    login_as_admin_user
 
     visit admin_users_path
 
     expect(page).to_not have_content("Delete")
   end
 
   it "does not allow editing of user delegates" do
-    user = login_as_admin_user
+    login_as_admin_user
```
pkarman: don't you need `user` a couple lines later?

spec/features/admin_spec.rb
```diff
@@ -94,7 +94,7 @@
     expect(proposal).to be_completed
   end
 
-  it "triggers actions on Step edit" do
+  it "triggers actions, no email notifications, on Step edit" do
```
pkarman: +1

spec/features/admin_spec.rb
jessieay: you're right. got too greedy. added back at 000c1f2

PR #1198, opened 2016-04-21
be more generous in what we receive (email or User) in mailer method

pkarman:
to fix failing jobs like https://cap.18f.gov/admin/jobs/120940

### DIFF
--- a/spec/mailers/comment_mailer_spec.rb
+++ b/spec/mailers/comment_mailer_spec.rb
@@ -2,6 +2,8 @@ describe CommentMailer do
   include MailerSpecHelper
   include EnvVarSpecHelper
 
+  let(:user) { create(:user) }
+
   describe "#comment_added_notification" do
     it_behaves_like "a proposal email" do
       let(:proposal) { create(:proposal) }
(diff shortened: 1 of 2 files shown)


spec/mailers/comment_mailer_spec.rb
```diff
@@ -2,6 +2,8 @@
   include MailerSpecHelper
   include EnvVarSpecHelper
 
+  let(:user) { create(:user) }
```
jessieay: we only need this `user` in one spec, why add it here rather than in the spec itself?

jessieay: please no `let

pkarman: I wish we could.

however, the rspec retry approach to workaround DeadClient failures has caching problems with our handrolled memoization when we use a `def` method.  see https://github.com/18F/C2/pull/1099#issuecomment-201404576

jessieay: can we inline `user` instead? understand memoization issue for sure!

pkarman: I started doing that, then noticed there is a dependency on `user` in the bunch of `let` statements in the first test.

jessieay: yeah that's annoying -- if it were me I would put a `let` there and then inline `user` for the rest of the tests but I will merge as-is since I know this is related to failing jobs on prod.

PR #1202, opened 2016-04-25
New details page/top notification

rememberlenny:
# What

Add the grey notice bar at the top of the page to explain beta design.

## Trello

https://trello.com/c/l2f5atmK/377-details-grey-bar-for-new-feature


app/controllers/proposals_controller.rb
```diff
@@ -12,7 +12,9 @@ class ProposalsController < ApplicationController
 
   def show
     @proposal = proposal.decorate
-    cookies[:detail] = params[:detail]
+    if !params[:detail].blank?
```
rememberlenny: @jessieay I had to re-add this check because the cookie was getting wiped when the query wasnt set.

app/views/proposals/details/_top_notification.html.haml
```diff
@@ -0,0 +1,6 @@
+.medium-12.top-notification-wrapper
+  .row
+    .medium-12.column
+      = image_tag 'icon-configuration.png'
+      %p
+        This is a beta version of the Request Details page. #{link_to 'Switch to previous design', proposals_revert_design_path} if you prefer.
```
pkarman: instead of a route that deletes a cookie and redirects, did you consider deleting the cookie client-side via JS and then reloading the same page?

rememberlenny: @pkarman I didnt explore that idea originally. I just coded up what that would look like.

Do you think it would be better to do via JavaScript? My only concern would be that the cookie is set in the controller from the first place. I see the single-use route being added as a 'con', but not a show stopper.

Thoughts?

pkarman: I suggested the client-side JS approach only because I thought it would require less code and would not require the one-off route.

The approach you have here is reasonable. My assumption is that this feature is to help in the UI transition and might eventually be excised from the code.

rememberlenny: @pkarman Got it. thanks for merging!

PR #1209, opened 2016-04-26
Fix Steps where completer_id is null

pkarman:
trello: https://trello.com/c/WDBh0nXW/413-fix-completed-steps-where-completer-is-null

includes some history clean up as well, to improve signal-to-noise ratio.

history before:

screen shot 2016-04-26 at 1 13 45 pm

history after:

[nothing, because dates are compared using same timezone and steps are not recreated unless they change]


app/services/ncr/approval_manager.rb
```diff
@@ -50,7 +50,14 @@ def force_approvers(users)
       individuals = users.map do |user|
         proposal.existing_step_for(user) || Steps::Approval.new(user: user)
       end
-      proposal.root_step = Steps::Serial.new(child_steps: individuals)
+      unless proposal.root_step && steps_share_parent?(proposal.root_step, individuals)
```
jessieay: not sure I understand the use case for `steps_share_parent?(proposal.root_step, individuals)` -- perhaps we should have a test for this?

db/migrate/20160426212643_fix_step_user_completer.rb
```diff
@@ -0,0 +1,36 @@
...
+    has_many :incoming_delegations, class_name: "UserDelegate", foreign_key: "assignee_id"
+  end
+
+  class UserDelegate < ActiveRecord::Base
+    belongs_to :assignee, class_name: "User", foreign_key: "assignee_id"
+    belongs_to :assigner, class_name: "User", foreign_key: "assigner_id"
+  end
+
+  def up
+    steps_with_no_completer.each do |step|
```
jessieay: seems like this could live in a one-off rake task rather than migration since we just need to run it once.

app/services/ncr/approval_manager.rb
pkarman: good point. I have added test and renamed method/vars for clarity in 4cbb1fc

The issue is that, previously, a new root step (`Steps::Serial`) was created every time the WorkOrder was modified, regardless of whether the child steps changed at all. This created a lot of noise in the proposal history, with `parent_id changed` type messages.

db/migrate/20160426212643_fix_step_user_completer.rb
pkarman: I suppose it could. It really should have been part of the migration in 04fea6b05913aa543140d64405092ef99d5fb443 (data migration to accompany schema change) but we failed to do that (didn't seem necessary at the time).

Plus, I want it to run once in every env, automatically. Which is what every db migration does.

jessieay:
LGTM! circle is mad at ya

pkarman:
thanks @jessieay -- at last the circle is unbroken, by and by.

PR #1210, opened 2016-04-27
bump up cloud.gov web instance memory use

pkarman:
after 36 hours, production has climbed from 500MB to 624MB usage.


pkarman:
this has been deployed to production, since memory use was at 99%

PR #1213, opened 2016-04-28
New Details Page "Edit mode"

rememberlenny:
# What

Use the `.on` on the `#mode-parent` element to control page's edit/view mode.

## Note

This is a partial PR. The code is functional on its own and can be merged, but doesnt make sense without the following PRs.

## Trello

Not yet.


app/assets/javascripts/details/state/edit_mode.js
```diff
@@ -0,0 +1,57 @@
+"use strict";
+
+var EditStateController = function (el) {
```
nickbristow: Is this used somewhere in the app right now?

rememberlenny: @nickbristow Not now. Since these PRs are coming in piecemeal, the "glue" is missing

rememberlenny: @nickbristow This is where it gets used: https://github.com/18F/C2/blob/new-details-page/edit-mode/app/assets/javascripts/details/app_details.js

PR #1218, opened 2016-04-29
require user logged in to submit feedback form

pkarman:


rememberlenny:
Looks good

afeld:
If requiring login ends up being a problem, could alternatively look into having a captcha or something.

pkarman:
https://trello.com/c/GOYM8mdl/419-feedback-form-requires-authn

PR #1219, opened 2016-04-29
bump ruby version to 2.2.5

pkarman:


rememberlenny:
@pkarman Are there any concerns around this? I take it the tests are failing for the very purpose of the PR.

dogweather:
This is probably not the current issue, but AFAIK one shouldn't use `migrate` to set up the database:

> bundle exec rake db:create db:migrate

It's not guaranteed to work. And it generally doesn't in a mature app because of sync errors between the current code and the old migration state. Instead,

> bundle exec rake db:setup

is the right way to do it. It creates the database schema via `db:schema:load`, which is guaranteed to be correct.

dogweather:
I'm having a whack at it.

dogweather:
That commit fixes the database install. Now seeing the same error further down at the `rspec` invocation.

dogweather:
The core problem is a `rubygems` incompatibility. Running `gem update --system` on the test server resolves it.

pkarman:
This upgrade is for cloud.gov where the default buildpack has now stopped supporting Ruby 2.2.3.

This is actually already deployed on stage and prod, as it was necessary to get out the fix on Friday 4/29 for the feedback spam issue.

dogweather:
Ok, thanks. Do we still need to merge this into master?

pkarman:
definitely, once tests pass. otherwise master cannot be deployed to cloud.gov

rememberlenny:
Going to merge this so we can push to cloud.gov

PR #1225, opened 2016-05-02
Don't Merge: Testing Circle CI configuration

dogweather:
(bump ruby version to 2.2.5)


dogweather:
I've fixed the bundler+rake error by running;

```bash
$ bundle update rake
```

PR #1233, opened 2016-05-06
Make comments ajax

nickbristow:
Comments now appear on page without reloading

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -13,8 +13,7 @@ class CommentsController < ApplicationController
     else
       flash[:error] = comment.errors.full_messages
     end
-
-    redirect_to proposal
+    respond_to_comment
   end
 
   protected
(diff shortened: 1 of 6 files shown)


app/controllers/comments_controller.rb
```diff
@@ -1,6 +1,7 @@
 class CommentsController < ApplicationController
   before_action ->{ authorize proposal, :can_show! }
   rescue_from Pundit::NotAuthorizedError, with: :auth_errors
+  respond_to :js, only: [:create]
```
jessieay: do we need this? I thought that passing a `format.js` option to `respond_to

spec/features/comment_spec.rb
```diff
@@ -10,6 +10,28 @@
     expect(page).to have_content("You successfully added a comment")
   end
 
+  scenario "saves the comment with javascript", js: true do
+    proposal = create_and_visit_proposal
+    visit "/proposals/#{proposal.id}?detail=new" 
+    comment_text = "this is a great comment"
+    js_submit_comment(comment_text, "#add_a_comment")
+    wait_for_ajax
```
rememberlenny: awesome

PR #1235, opened 2016-05-06
Test that should work, doesnt work. Why?

rememberlenny:


rememberlenny:
@nickbristow test works. Issue is code. Im going to break up the other PR

PR #1241, opened 2016-05-07
5️⃣️ Provide ability for tests to use config params [NDJS 5] 

rememberlenny:
# What

Create a condition in the c2 shell file to optionally load config parameters. This provides a way to override the test conditions with a mock dom. Helps for unit tests.

# Note

Merge the NDJS files with smaller numbers first.

### DIFF
--- a/app/assets/javascripts/details/shell/c2.js
+++ b/app/assets/javascripts/details/shell/c2.js
@@ -2,19 +2,42 @@ var C2;
 C2 = (function() {
   
   function C2(){
+    this.config = {
+      editMode:       '#mode-parent',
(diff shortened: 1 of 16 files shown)


app/assets/javascripts/details/shell/c2.js
```diff
@@ -1,8 +1,8 @@
 var C2;
-
 C2 = (function() {
   
-  function C2(){
+  function C2(config){
+    config = config || {};
```
rememberlenny: I think this was the cause of the errors. I was using a ES6 JS syntax for assigning default params and it was failing. @nickbristow

PR #1245, opened 2016-05-10
Update the displayed Request Details form, after saving

rememberlenny:
# What

Create a series of events that pass data from the `ClientDataController#update` controller to the frontend. The `update` method triggers a javascript event that passes the response value to the an event on `DetailsSave`. `DetailsSave` takes the event and fires into the `c2.js` glue. The trigger kicks off the `save` button in the `loading` mode, when the views get updated. Once the views are processed, then the button updates and the view returns to the display mode.

Sort of complicated, but this is the best way to separate responsibility of `data/persistence`, `view manipulation`, and the in betweens.

# Note

Adding a new JS library for making a pleasant button animation after submit.


dogweather:
Huh, weird error. Is it real?

PR #1246, opened 2016-05-11
Fixed two bugs with observer card

nickbristow:
You can now delete yourself as an observer which redirects you like the original page did.
  The dropdown for observers is no longer cut off by the card.


dogweather:
Rebuilding the tests to see if this consistently gives the same error.

rememberlenny:
@nickbristow Branch needs to be caught up

nickbristow:
Whoops this should've been merged before the other PR, so these commits are in master now.

PR #1249, opened 2016-05-12
Added 'completed_at' to active admin ui for Step model

phirefly:


dogweather:
Looks great. :shipit:

PR #1250, opened 2016-05-12
Moved pry-rails to production

dogweather:
Makes it a lot easier to interactively work with the production data.


pkarman:
this would be for an alternate `rails c` when running under `cf-ssh` ? Seems ok to me.

rememberlenny:
Looks good!

PR #1254, opened 2016-05-17
Fix new design when proposal is emergency

rememberlenny:
# What

The emergency requests dont have any approval steps. As a result, the template was failing to display a `proposal.individual_steps.last.created_at` time. The summary partial checks for the `expense_type`, and conditionally uses the `proposal.updated_at` time for emergency requests.


app/views/proposals/details/_approvals.html.haml
```diff
@@ -6,27 +6,31 @@
...
-                  %span.step-data
-                    %span.step-user-name
-                      #{User.find(step.user_id).full_name},
-                    %span.step-user-role #{step_user_title(step)}
-                  - if !step.completed_at.blank? && step.status != "actionable" && step.status != "pending"
-                    %span.step-user-status.step-data
-                      = step.status.capitalize + " "
-                      %span.step-timestamp
-                        = step.completed_at.strftime("%m/%d/%Y %I:%M%p")
+          - if
```
nickbristow: should there be a conditional here?

rememberlenny: nope! @nickbristow

PR #1258, opened 2016-05-18
Added visual notification for added observers

nickbristow:
Notifies user when an observer is added and removed


rememberlenny:
Choosing to merge with codeclimate conflict. We need to update Rubocop rules.

PR #1265, opened 2016-05-20
Show and hide comments

nickbristow:


app/assets/javascripts/details/views/activity_card.js
```diff
@@ -15,6 +15,25 @@ ActivityCardController = (function(){
...
     var self = this;
+    this._setupUpdateEvent();
+    this._setupCommentListToggle();
+  }
+
+  ActivityCardController.prototype._setupCommentListToggle = function(){
+    var self = this;
+
+    this.el.on('click','.status-contract-action, .status-expand-action', function(){
+      self.el.find('.toggle-contracted')
```
rememberlenny: Since these are all doing the same thing, is there anyway to give them all one class? Then you just need to have one thing toggled.

rememberlenny:
Looks good!

PR #1270, opened 2016-05-23
Attachment refinements

nickbristow:
Some minor improvements to the attachments card

Removed header text as per @amberwreed 

added "Remove" when you hover on delete button


rememberlenny:
Looks good!

PR #1272, opened 2016-05-24
[DO NOT MERGE] Show client data updates through HistoryList

yozlet:
Here for @nickbristow to poke at before we pair on getting object updates displayed.


yozlet:
Closing because it's no longer needed - code is now in @nickbristow's `save_card_activity_updates` branch.

PR #1275, opened 2016-05-24
Save card activity updates

yozlet:
Includes formatted diffs of proposal & client_data object data updates.


yozlet:
Killing and recreating, since it includes commits that should already have been merged.

PR #1278, opened 2016-05-25
options passed to date with tooltip that formats time

nickbristow:
Show the date when an activity item wasn't made on the current day


rememberlenny:
Looks good!

PR #1291, opened 2016-05-27
Get js tests solid

dogweather:


app/controllers/proposals_controller.rb
```diff
@@ -11,6 +11,7 @@ class ProposalsController < ApplicationController
   rescue_from Pundit::NotAuthorizedError, with: :auth_errors
 
   def show
+    fail "User #{@current_user} has no client-slug" if @current_user.client_slug.blank?
```
dogweather: What's the best way to check this? Somewhere in the authorization code? It's a little complex.

I added this as a quick diagnostic to eliminate nil errors.

dogweather:
@phirefly @rememberlenny @nickbristow  I added the missing data which was causing the nil errors. But now the tests are mostly having routing errors. I'm thinking the tests might have errors? For sure, it'd be better if they'd use the route helpers to construct the paths.

Here's a typical error now:

```
3.2) Failure/Error: raise ActionController::RoutingError, "No route matches [#{env['REQUEST_METHOD']}] #{env['PATH_INFO'].inspect}"
          
          ActionController::RoutingError:
            No route matches [GET] "/javascripts/details/state/details_request_form_state.js"
```

dogweather:
No one's working on this, and it's pretty out of date at this point.

PR #1298, opened 2016-06-01
Recommend rbenv for developer machines

yozlet:
Trello card: https://trello.com/c/T5K6IQXf/480-revise-docs-to-recommend-rbenv-not-rvm


phirefly:
:shipit:

PR #1300, opened 2016-06-02
redirect to proposals list view after cancel

nickbristow:
After canceling a proposal in the new view, user is redirected to the list view

### DIFF
--- a/app/controllers/proposals_controller.rb
+++ b/app/controllers/proposals_controller.rb
@@ -12,11 +12,8 @@ class ProposalsController < ApplicationController
 
   def show
     @proposal = proposal.decorate
-    unless params[:detail].blank?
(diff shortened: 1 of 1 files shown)


app/controllers/proposals_controller.rb
```diff
@@ -12,11 +12,8 @@ class ProposalsController < ApplicationController
 
   def show
     @proposal = proposal.decorate
-    unless params[:detail].blank?
-      cookies[:detail] = params[:detail]
-    end
-    mode = cookies[:detail]
-    if mode == "new"
+
+    if new_mode
```
rememberlenny: This is way better

rememberlenny:
Looksk good

PR #1302, opened 2016-06-02
Revert design now works even if param is in url

nickbristow:


rememberlenny:
Thanks for doing this!

PR #1304, opened 2016-06-03
Focus is moved to the modal when its activated for keyboard users

nickbristow:
Focus moves to modal when its activated.


rememberlenny:
Looks good!

PR #1306, opened 2016-06-03
Added stubs for when mailer previews to prevent errors

nickbristow:
Was getting errors for mailer previews. Added some stubbing to prevent this.

### DIFF
--- a/lib/mail_previews/mailer_preview_helpers.rb
+++ b/lib/mail_previews/mailer_preview_helpers.rb
@@ -4,11 +4,18 @@ module MailerPreviewHelpers
   end
 
   def completed_proposal
-    Proposal.completed.last
+    Proposal.completed.last || Proposal.last
   end
 
   def completed_step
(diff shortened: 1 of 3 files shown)


lib/mail_previews/mailer_preview_helpers.rb
```diff
@@ -4,11 +4,18 @@ def canceled_proposal
   end
 
   def completed_proposal
-    Proposal.completed.last
+    Proposal.completed.last || Proposal.last
   end
 
   def completed_step
-    Step.where(status: "completed").last
+    last_complete = Step.where(status: "completed").last
```
rememberlenny: Seems like we should figure out what the actual problem is too

rememberlenny:
Merged, but we should figure out what the actual issue is

nickbristow:
Stubbed is not a great word, basically it uses real proposals to show you the emails. If you don't have those proposals in those states, it errored out

PR #1307, opened 2016-06-06
Set cookie for beta users

nickbristow:


nickbristow:
This is for this card
 https://trello.com/c/tMZj9EYK/471-adding-a-beta-users-to-the-system

PR #1308, opened 2016-06-07
Notify user about new design + trigger new design with "beta_mode" and "beta_design". 

rememberlenny:
# What

Create the new design for the notice regarding new design. Notice bar can be clicked to toggle between modes. 

To manage new design, check for `beta_mode`. If `beta_mode` is active, then check for `beta_detail` to show the new detail page. `beta_mode` lets users go back-and-forth between designs, while still having access.

Eventually, `beta_mode` should be accessible via the profile page.


nickbristow:
Yeah, this is exactly what I was thinking.

config/routes.rb
```diff
@@ -42,7 +42,8 @@
       get "cancel_form"
       post "cancel"
       get "history"
-      get "revert_design" => "proposals#revert_design"
+      get "activate_design" => "proposals#activate_detail_design"
```
dogweather: We're going to remove these because we're not going to support that feature.

But also, using GETs to mutate data is a security flaw.

rememberlenny: @dogweather great. Is that clear with Amber?

dogweather: No idea. This was from @phirefly

PR #1309, opened 2016-06-08
Fix ncr_organization/org_code labelling

yozlet:
The "Org Code" label wasn't showing up properly in the new detail page. Also removed the `org_code` key, because AFAIK we're not using it.


rememberlenny:
@yozlet Good to go?

phirefly:
@yozlet Does this mean that `org_code` is in use?
https://github.com/18F/C2/blob/master/app/decorators/ncr/work_order_decorator.rb#L87

I might be misunderstanding the removal of this attribute, but I think this bit of code would call that, wouldn't it?

yozlet:
@phirefly: Argh, yes, you're right. Sorry about that. Will fix.

yozlet:
Closing this in favour of #1327.

PR #1311, opened 2016-06-08
Updated comment spec to use new beta roles

nickbristow:


rememberlenny:
Thanks for updating these

PR #1316, opened 2016-06-08
Dont allow expense type to change in pages

rememberlenny:
# What

Keeping with current behavior, prevent changing expense type once set.


yozlet:
QA'ed, looks good 👍

PR #1317, opened 2016-06-08
Create a pending status on the actionable and pending states of status

rememberlenny:
# What 

https://trello.com/c/iIKdPjz9/501-status-panel-not-displaying-status-of-step

### DIFF
--- a/app/views/proposals/details/_status.html.haml
+++ b/app/views/proposals/details/_status.html.haml
@@ -17,7 +17,7 @@
                 .row.step-row.column.medium-table-row{class: "step-status-" + step.status}
                   .medium-icon-col.text-center.medium-table-cell.step-status-wrapper
                     %span.circle-wrapper{ class: "status-" + step.status }
-                      - if step.status != "completed"
+                      - if !step.completed?
                         = index.to_s
                       - else
                         = "✔"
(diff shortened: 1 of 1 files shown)


app/views/proposals/details/_status.html.haml
```diff
@@ -33,3 +33,6 @@
                         = step.status.capitalize + " "
                         %span.step-timestamp
                           = step.completed_at.strftime("%m/%d/%Y %I:%M%p")
+                    - elsif step.status == "actionable" || step.status == "pending"
```
yozlet: So, you can do `step.actionable?` instead of a string comparison here. But also you can possibly replace this test with `!step.completed?` since it's the only other status, isn't it?

app/views/proposals/details/_status.html.haml
```diff
@@ -33,3 +33,6 @@
...
+                      %span.step-user-status.step-data
+                        = "Pending"
```
yozlet: There are already strings in `locales/decorators/en.yml` for this, but we don't yet have a label for "not completed". The `Step` model uses the word `outstanding` for that, so I suggest adding `outstanding: "Pending"` under `en.decorators.steps.status`

app/views/proposals/details/_status.html.haml
yozlet: (ditto on line 31)

rememberlenny:
@yozlet Thanks! Updated!

yozlet:
👍

rememberlenny:
Thanks!
ᐧ

--
*Leonard Bogdonoff*
rememberlenny@gmail.com
Blog <http://blog.rememberlenny.com> | Portfolio <http://rememberlenny.com/>
 | @rememberlenny <http://twitter.com/rememberlenny>
949-322-8496

PR #1319, opened 2016-06-08
Update notification behavior

rememberlenny:
# What

Updates to notifications after editing.

- Alert notices do not automatically disappear
- Extend time of notices to 7.5 seconds
- Clear all notices when save/cancel is clicked


nickbristow:
verified on dev

PR #1327, opened 2016-06-09
Fix ncr_organization labeling in decorator strings

yozlet:
The "Org Code" label wasn't showing up properly in the redesign. (This is a second attempt at #1309)


yozlet:
OK, sorry, STILL more investigation needed.

PR #1328, opened 2016-06-10
Step counter set by index instead of model

nickbristow:
https://trello.com/c/mKEhP6tK/515-steps-start-at-step-0


nickbristow:
There was a bug where the position didn't line up with the step number, now using the number of the step displayed for the page.

nickbristow:
screen_shot_2016-06-08_at_4 02 14_pm

rememberlenny:
Looks good

PR #1329, opened 2016-06-10
Change the action bar button conditions

rememberlenny:
# What

Make the action bar buttons visible for anyone with can_edit access.

https://trello.com/c/5RjPTQQR/494-action-bar-has-no-buttons-on-completed-and-canceled-requests


nickbristow:
Looks good

PR #1330, opened 2016-06-10
Remove option to change emergency field

rememberlenny:
# What

Remove the input fields for the emergency checkbox and replace with value.

# Reference

https://trello.com/c/CIvM5KYu/517-as-a-user-i-am-not-allowed-to-change-a-ba61-request-to-an-emergency-request-after-submitting-it


nickbristow:
Looks good!

PR #1334, opened 2016-06-13
fixes bug where the same approval shows twice in activity feed

nickbristow:
## What

On the redesign page, the activity feed shows the approval entry twice

## Note

Approvals created two entries in the change log, one updating who was doing the approval, and another that marked the step as complete. This update just filters out the first case since we aren't showing that information in the feed. 

https://trello.com/c/cgIK47ON/519-status-check-marks-not-matching-timeline-steps


rememberlenny:
Looks good!

PR #1335, opened 2016-06-13
Removed code checking for cookie

nickbristow:
## What

Code was stopping notices from showing when you had the detail cookie set to new. 

## Notes

https://trello.com/c/y65WhdCj/493-sign-in-notification-is-displayed-over-page


rememberlenny:
@nickbristow We should be replacing the check for cookie with check for beta+detail mode

nickbristow:
@rememberlenny well the issue with doing that is it turns off notifications on the whole site. 

But it looks like that bar needs to be removed on the new page now that im looking at it

I'll figure out a solution

nickbristow:
@rememberlenny hows that

rememberlenny:
Looks good!

PR #1338, opened 2016-06-13
Enable API on prod

jessieay:
* Micropurchase wants to use Prod API!
* See https://github.com/18F/C2/blob/a935eb3807ec7a859941e590616f883ae4be285c/doc/api.md

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -95,6 +95,7 @@ applications:
   instances: 2
   command: script/server_start
   env:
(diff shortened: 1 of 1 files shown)


rememberlenny:
@jessieay Is this good to go?

rememberlenny:
@phirefly Are we okay with merging this?

phirefly:
:shipit:

PR #1342, opened 2016-06-14
Adding tests to the notification's page

rememberlenny:
# What

Add the tests to the notification's javascript

### DIFF
--- a/spec/javascripts/details/views/notification_bar_spec.js.coffee
+++ b/spec/javascripts/details/views/notification_bar_spec.js.coffee
@@ -15,7 +15,12 @@ describe 'Notification', ->
   
   describe '#_postNotification', ->
     it "create a notification", ->
-  
+      notification = new Notifications(getNotificationContent())  
+      expect(notification.el.find('ul li').length).to.eql(0)
+      notice = notification._prepare({type: "success", content: "This is content", title: ""})
+      notification._postNotification(notice)
+      expect(notification.el.find('ul li').length).to.eql(1)
+
   describe '#_prepareOnLoadNotifications', ->
     it "on page load, create a notification for each meta tag", ->
       notification = new Notifications(getNotificationContent())  
@@ -47,6 +52,7 @@ describe 'Notification', ->
       expect(notification.el.find('ul li').is(':animated')).to.eql(false)
       notification.el.find('ul li .close').trigger('click')
       expect(notification.el.find('ul li').is(':animated')).to.eql(true)
+  
   describe '#_prepare', ->
     it "prepare a notification", ->
     it "handle a notification without a title", ->
--- a/spec/javascripts/details/views/notification_bar_spec.js.coffee
+++ b/spec/javascripts/details/views/notification_bar_spec.js.coffee
@@ -35,9 +35,18 @@ describe 'Notification', ->
       notification._prepareOnLoadNotifications(el)
       expect(notification.el.find('ul li .notification-content').text()).to.eql("Notification text here")
       expect(notification.el.find('ul li.notice-type-success').length).to.eql(1)
+
   describe '#_closeButton', ->
     it "make sure the close button deletes the notification", ->
-  
+      notification = new Notifications(getNotificationContent())  
+      el = $('
+        <meta name="flash-message" type="success" content="Notification text here">  
+      ')
+      notification._prepareOnLoadNotifications(el)
+      expect(notification.el.find('ul li').length).to.eql(1) 
+      expect(notification.el.find('ul li').is(':animated')).to.eql(false)
+      notification.el.find('ul li .close').trigger('click')
+      expect(notification.el.find('ul li').is(':animated')).to.eql(true)
   describe '#_prepare', ->
     it "prepare a notification", ->
     it "handle a notification without a title", ->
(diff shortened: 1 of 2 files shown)


spec/javascripts/details/views/notification_bar_spec.js.coffee
```diff
@@ -13,30 +14,119 @@ describe 'Notification', ->
       expect(notification instanceof Notifications).to.eql(true)
   
   describe '#_postNotification', ->
-    it "create a notification", ->
+    it "create a notification without type", ->
```
yozlet: Prefer naming of individual specs to read like a sentence including the `it`, so "it create**s** a notification without type" (note the S)

spec/javascripts/details/views/notification_bar_spec.js.coffee
```diff
@@ -13,30 +14,119 @@ describe 'Notification', ->
...
+      notification = new Notifications(getNotificationContent())  
+      expect(notification.el.find('ul li').length).to.eql(0)
+      notice = notification._prepare({type: "", content: "This is content", title: ""})
+      notification._postNotification(notice)
+      expect(notification.el.find('ul li').length).to.eql(1)
+      expect(notification.el.find('ul li.notice-type').length).to.eql(0)
+      expect(notification.el.find('ul li.notice-type-primary').length).to.eql(1)
+      expect(notification.el.find('ul li .notification-content').text()).to.eql("This is content")
+    it "create a notification with type", ->
```
yozlet: Line breaks in between examples would help readability here, especially since this file is quite dense.

spec/javascripts/details/views/notification_bar_spec.js.coffee
```diff
@@ -13,30 +14,119 @@ describe 'Notification', ->
...
+      expect($(notice).attr('data-timeout')).to.eql("10000")
   
   describe '#clearAll', ->
     it "remove all four notifications", ->
-  
+      notification = new Notifications(getNotificationContent())  
+      expect(notification.el.find('ul li').length).to.eql(0)
+      notice = notification._prepare({type: "", content: "This is content", title: ""})
+      notification._postNotification(notice)
+      expect(notification.el.find('ul li').length).to.eql(1)
```
yozlet: There are a lot of repetitive assertions in here and you don't need many of them, especially those which test methods other than the one currently described, or test behaviour that's already tested elsewhere. For example, this spec has several assertions checking that `_postNotification(notice)` increases the list `length` by one, but that behaviour is already asserted by a different test, and anyway we're testing `clearAll()` here. So I think you can probably remove many of the `length` assertions through this file

spec/javascripts/details/views/notification_bar_spec.js.coffee
```diff
@@ -13,30 +14,119 @@ describe 'Notification', ->
...
+      expect(notification.el.find('ul li.notice-type-primary').length).to.eql(0)
+    it "create a notification with title", ->
+      notification = new Notifications(getNotificationContent())  
+      expect(notification.el.find('ul li').length).to.eql(0)
+      notice = notification._prepare({type: "success", content: "This is content", title: "This is a title"})
+      notification._postNotification(notice)
+      expect(notification.el.find('ul li .notification-title').text()).to.eql("This is a title")
+
+  describe '#_prepareOnLoadNotifications', ->
+    it "on page load, create a notification for each meta tag", ->
```
yozlet: This example seems to be testing something slightly different than described. From the description I'd expect one `_prepareOnLoadNotifications(el)` call with `el` containing two `meta` tags, but you've got two calls with identical single elements. Also, the naming of `_prepareOnLoadNotifications()` suggests to me that it should be preparing the entire set of notifications, rather than adding to them.

yozlet:
Yay loads of tests!

spec/javascripts/details/views/notification_bar_spec.js.coffee
```diff
@@ -8,35 +9,138 @@
...
+    it "creates a notification with title", ->
+      notification = new Notifications(getNotificationContent())  
+      expect(notification.el.find('ul li').length).to.eql(0)
+      notice = notification._prepare({type: "success", content: "This is content", title: "This is a title"})
+      notification._postNotification(notice)
+      expect(notification.el.find('ul li .notification-title').text()).to.eql("This is a title")
+
+  describe '#_prepareOnLoadNotifications', ->
+    
+    it "creates a notification for each meta tag on page load", ->
```
yozlet: (re-adding a comment which was outdated by a different commit...)

This example seems to be testing something slightly different than described. From the description I'd expect one `_prepareOnLoadNotifications(el)` call with `el` containing two `meta` tags, but you've got two calls with identical single elements. Also, the naming of `_prepareOnLoadNotifications()` suggests to me that it should be preparing the entire set of notifications, rather than adding to them.

yozlet: OK, this looks better now, but the `each` in the description implies it should be testing with more than one item. Maybe add two `meta` tags to `el` and check length of `2`?

yozlet:
Looks much better now. One remaining note, then it's good to go.

rememberlenny:
Checking now @yozlet

yozlet:
👍 :shipit:

PR #1343, opened 2016-06-14
Add `completed_at` field to `steps` output in API

jessieay:
* This is so micropurchase can see when the purchase step is completed
* Also removed unneeded serializers and updated docs


rememberlenny:
Looks good! Thank you Jessie!!

PR #1345, opened 2016-06-15
Downgraded version of 'aws-sdk' gem to adddress 'uninitialized consta…

phirefly:
…nt Paperclip::Storage::S3::AWS' error


Gemfile
```diff
@@ -10,7 +10,7 @@ gem "ahoy_matey"
 gem "ar_outer_joins"
 gem "autoprefixer-rails"
 gem "awesome_print"
-gem "aws-sdk"
+gem "aws-sdk", "< 2.0"
```
jessieay: I think the "pessimistic operator" might make more sense here https://robots.thoughtbot.com/rubys-pessimistic-operator

PR #1349, opened 2016-06-15
Fix how partials are rendered

rememberlenny:
# What

Partials were rendering partials relative to the current_user. User the requester of the proposal to qualify which partial is rendered.

### DIFF
--- a/app/views/gsa18f/_request_details.html.haml
+++ b/app/views/gsa18f/_request_details.html.haml
@@ -6,7 +6,7 @@
       f: f, |
       proposal: @proposal,
       key: key,
-      t_slug: "decorators." + @current_user.client_slug + "/work_order.",
+      t_slug: "decorators." + @proposal.requester.client_slug + "/work_order.",
       client_slug: @current_user.client_slug,
       client_data: @client_data_instance,
       client_id: @client_data_instance.id.to_s,
--- a/app/views/gsa18f/_request_details.html.haml
+++ b/app/views/gsa18f/_request_details.html.haml
@@ -6,7 +6,7 @@
       f: f, |
       proposal: @proposal,
       key: key,
-      t_slug: "decorators." + @proposal.requester.client_slug + "/work_order.",
+      t_slug: "decorators." + @proposal.client_slug + "/work_order.",
       client_slug: @current_user.client_slug,
       client_data: @client_data_instance,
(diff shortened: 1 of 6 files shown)


phirefly:
@rememberlenny Can you throw a test in here?

app/views/gsa18f/_request_details.html.haml
```diff
@@ -6,7 +6,7 @@
       f: f, |
       proposal: @proposal,
       key: key,
-      t_slug: "decorators." + @current_user.client_slug + "/work_order.",
+      t_slug: "decorators." + @proposal.requester.client_slug + "/work_order.",
```
yozlet: You don't need to go via `requester`, you can just use `@proposal.client_slug`

app/views/gsa18f/_request_details.html.haml
```diff
@@ -6,7 +6,7 @@
...
       client_slug: @current_user.client_slug,
```
yozlet: See above.

app/views/proposals/details/_request_details.html.haml
```diff
@@ -7,4 +7,4 @@
             Request Details
         .content-content.column#view-request-details
           .row
-            = render @current_user.client_slug + '/request_details'
+            = render client_specific_partial(@proposal.requester, '/request_details')
```
yozlet: Should this `client_specific_partial()` method be used elsewhere above (like in the change made to `_request_details.html.haml`, when it loads a client-specific partial)? Or do we have something like it already elsewhere?

Also, the current implementation of `client_specific_partial()` takes a `User` as the first argument, which is kind of strange. Can we tweak it just to take a client slug?

rememberlenny: @yozlet Im not using it anymore because it has a weird partial structure. Most of the partials above are either passing `data` or have a partial prefix. The helper doesnt account for those things.

app/views/gsa18f/_request_details.html.haml
rememberlenny: @yozlet Updated!

app/views/gsa18f/_request_details.html.haml
rememberlenny: @yozlet Thanks! Updated

phirefly:
:shipit:

PR #1350, opened 2016-06-16
Update the direct pay display

rememberlenny:
# What

Change how direct pay is displayed on front end. 

https://trello.com/c/HSNjpjbh/532-view-details-direct-pay-should-display-yes-no-instead-of-a-disabled-checkbox


phirefly:
Ship it

PR #1351, opened 2016-06-16
Make attachments editable when completed

rememberlenny:
# What

Change the condition that hides attach file button.

### DIFF
--- a/app/views/proposals/details/_attachment.html.haml
+++ b/app/views/proposals/details/_attachment.html.haml
@@ -14,5 +14,5 @@
           .medium-12.column
             = render partial: "proposals/details/attachment_list",
               locals: { proposal: @proposal }
-            - if !@proposal.completed? && !@proposal.canceled?
+            - if policy(@proposal).can_edit?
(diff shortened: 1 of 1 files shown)


app/views/proposals/details/_attachment.html.haml
```diff
@@ -14,5 +14,5 @@
           .medium-12.column
             = render partial: "proposals/details/attachment_list",
               locals: { proposal: @proposal }
-            - if !@proposal.completed? && !@proposal.canceled?
+            - if policy(@proposal).can_edit?
```
phirefly: Here's the current logic for `can_edit?`: `(admin? || requester!) && not_completed! && not_canceled!`

Currently, there is no role requirement (https://github.com/18F/C2/blob/master/app/views/proposals/show.html.haml#L100-L120), allowing anyone that access to the request to upload documents. We can remove this `if` and simply display the attachment capability for all roles.

Also note that we limit the ability to delete the attachments:
https://github.com/18F/C2/blob/master/app/views/proposals/show.html.haml#L120

phirefly:
:shipit:

PR #1357, opened 2016-06-17
Update total value on gsa18f proposals

rememberlenny:
# What

Update the total_price field on 18f proposals, when making frontend updates.

# How

Pass the data response into the success trigger for the proposal update event. If the object contains the amount and quantity fields, then process the total_price value and update.

# Reference

https://trello.com/c/coe76G0f/565-updating-the-quantity-or-cost-of-a-18f-proposal-item-does-not-update-the-total-price-field-without-refresh


phirefly:
🚢

PR #1358, opened 2016-06-20
Show ethnio popup on revert

nickbristow:
**Ready for code review**

## WHAT
Add ethnio pop-up after a user clicks "Show old design"

## WHY
To collect feedback as to why the user stopped using the new page

### DIFF
--- a/app/views/proposals/details/_top_notification.html.haml
+++ b/app/views/proposals/details/_top_notification.html.haml
@@ -7,6 +7,8 @@
     .top-notification-wrapper
       %p
         - if cookies[:beta_reverted_survey] == "true"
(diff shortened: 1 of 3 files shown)


app/views/proposals/details/_top_notification.html.haml
```diff
@@ -7,6 +7,8 @@
     .top-notification-wrapper
       %p
         - if cookies[:beta_reverted_survey] == "true"
+          - cookies.delete :beta_reverted_survey
+          = javascript_include_tag "//ethn.io/55009.js", async: true
```
adelevie: try to avoid protocol-relative links and instead explicitly use https

phirefly:
:shipit:

PR #1360, opened 2016-06-20
No more ba6x budget approvers

dogweather:
[Trello Card](https://trello.com/c/2PBUqMkE/523-as-a-user-i-only-need-approving-official-sign-off-when-making-a-ba60-or-ba61-request)


app/services/ncr/approval_manager.rb
```diff
@@ -1,4 +1,6 @@
 module Ncr
+  START_OF_NEW_6X_APPROVAL_POLICY = Time.zone.local(2016, 7, 5, 0, 0, 0)
```
phirefly: Can this just be a feature flag to check if it's on? This will give us flexibility on launch date (which could be turned on July 1st) and allow us to easily revert back in case the policy is reverted.

PR #1366, opened 2016-06-22
Disable email with spec metadata

yozlet:
A slightly easier way of enabling email sending, continuing the work @dogweather did in #1361.


dogweather:
I found that `NcrDispatcherSpec` has great examples of doing faster mail tests through mocking. https://github.com/18F/C2/blob/master/spec/models/ncr_dispatcher_spec.rb

yozlet:
👍

PR #1372, opened 2016-06-28
Update the cancel notification message

rememberlenny:
# What

Update the cancel button notification message.

# Reference

https://trello.com/c/ixQcz0yp/529-reset-modified-values-after-canceling


phirefly:
:shipit:

PR #1375, opened 2016-06-29
New details page/prevent duplicate notification

rememberlenny:
# What

Prevent potential for multiple notifications to pop up, if the content is the same as last rendered notification.

Change cancel notification text when there is no modifications made to field.

# Reference 

https://trello.com/c/WGKNAzEP/496-when-saving-a-change-to-a-request-the-notification-bar-repeats-multiple-times


nickbristow:
👍 merged

PR #1377, opened 2016-06-29
Update link to Reports

phirefly:
@amberwreed to give thumbsup + merge


phirefly:
@amberwreed Here's the text link change we discussed

PR #1379, opened 2016-06-30
Add some docs for Proposals#update action in API

jessieay:
* Yay docs!

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -174,7 +174,7 @@ Create a new [Proposal](#proposal). The root key is the client model slug, e.g. 
 ```bash
 % curl -i -X POST -H 'Content-Type: application/json' \
   -H 'Authorization: Bearer your-access-token' \
-  --data @proposal.json \
+  --data proposal.json \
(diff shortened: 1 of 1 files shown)


doc/api.md
```diff
@@ -174,7 +174,7 @@ Create a new [Proposal](#proposal). The root key is the client model slug, e.g.
 ```bash
 % curl -i -X POST -H 'Content-Type: application/json' \
   -H 'Authorization: Bearer your-access-token' \
-  --data @proposal.json \
+  --data proposal.json \
```
pkarman: my understanding was that `curl` required the `@` prefix to `--data` argument in order to designate it as a file name rather than a literal string.

http://stackoverflow.com/questions/15912924/how-to-send-file-contents-as-body-entity-using-curl

maybe newer versions of `curl` are smarter about that?

jessieay: yes this is very possible! In the end I was unable to actually test this `curl` because of my problems with the auth token authorization. So was a guess.

PR #1380, opened 2016-06-30
As an NCR USER, I can search by Org Code and Building Number, and run reports based on those attributes

dogweather:
[Trello card](https://trello.com/c/Fx9zE3cD/524-as-an-ncr-user-i-can-search-by-org-code-and-building-number-and-run-reports-based-on-those-attributes)


yozlet:
This currently works: there are SELECT boxes for the two fields using new helper methods, the helper methods have specs, there are feature tests that check that the fields are there. _However,_ we don't yet have tests that ensure that these particular field searches work as intended (though we've tested them and they seem to) and, more importantly, they are giant ugly default select boxes when we really should be using Selectize. But we can save both of those for a later PR if we need to get this out now.

dogweather:
@phirefly To recap, this is ready to merge, given the caveats in @yozlet 's comment above.

phirefly:
:shipit:

phirefly:
Creating a follow-up story for tidying this up
https://trello.com/c/KeQNWen2

@dogweather @yozlet, feel free to add details

PR #1381, opened 2016-07-01
Can now close ethnio screener consistently

nickbristow:
bug where the ethnio screener (when user reverts to the old design) couldn't be closed. 

This resolves that issue.


phirefly:
🚢

PR #1384, opened 2016-07-01
Fix OAuth applications authorize method

pkarman:
API docs claim this works. Now it does, with test to prove it.

see https://trello.com/c/5p8smp6E/596-oauth-authorize-action-broken


jessieay:
awesome! thanks a million @pkarman !

spec/features/api/v2/oauth_applications_spec.rb
```diff
@@ -0,0 +1,22 @@
+describe "OAuth applications" do
```
jessieay: may make sense to change the route of this spec file since it doesn't occur in the api (maybe `spec/features/oauth_applications_spec.rb` ?)

jessieay:
LGTM! One comment on feature spec location for future findability :)

pkarman:
good call @jessieay -- renamed in c3afb03

jessieay:
@pkarman these failures seem legit but unrelated to this PR -- any ideas?

pkarman:
agreed, they do seem unrelated. No idea what they're about. I just rebased from master, just in case.

pkarman:
I have lots of rspec failures locally, none for this particular feature test.

pkarman:
use #1388 instead

PR #1385, opened 2016-07-05
First pass at new script

dogweather:
[Trello card](https://trello.com/c/2PBUqMkE/523-as-a-user-i-only-need-approving-official-sign-off-when-making-a-ba60-or-ba61-request)


phirefly:
Ran through ssh. Closing.

PR #1386, opened 2016-07-06
Fix tests after ba6x approval changes

yozlet:
The new BA6x approval policy (no more budget approval steps) unfortunately broke several tests in a way that wasn't caught before the policy became active. This PR fixes those specs by:

 * removing specs testing BA6x budget approval behaviour
 * changing the default `ncr_work_order` factory to use the `BA80` expense type (which has a budget approval step) instead of the `BA61` expense type (which no longer does)
 * due to the above, expecting only one budget approval step instead of two

Also, it removes the "tier 1 only" report from the report mailer, since tier 1 no longer exists.


rememberlenny:
@yozlet Looks good. Feel free to merge as needed.

PR #1396, opened 2016-07-08
Update the details title

rememberlenny:
# What

The details view `title` field was not being updated on the frontend. During refactoring, the selectors targeting the title broke.

There was no test to detect this would break.


phirefly:
🚀

PR #1397, opened 2016-07-11
Create notices around the controller's JS related interactions

rememberlenny:
# What

Fix notices from `proposal#show` view were appearing in other pages.

# How

Add the `.now` attribute on notices.


phirefly:
:shipit:

phirefly:
@rememberlenny I just realized -- can you add specs for the new presenter?

PR #1401, opened 2016-07-15
Print design

micahtaylor:
First pass on print styles.


micahtaylor:
print design

phirefly:
:shipit:!

PR #1406, opened 2016-07-18
Turn on JS tests on circleCI

rememberlenny:
# What

Turn on the CircleCI JavaScript tests.

They were disabled and shouldnt have been.

### DIFF
--- a/spec/javascripts/details/shell/c2_spec.js.coffee
+++ b/spec/javascripts/details/shell/c2_spec.js.coffee
@@ -15,6 +15,8 @@
 #= require details/views/summary_bar
 #= require details/views/modal_card
 #= require details/views/notifications
+#= require details/views/view_helper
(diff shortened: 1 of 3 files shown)


spec/javascripts/details/shell/c2_spec.js.coffee
```diff
@@ -15,6 +15,8 @@
 #= require details/views/summary_bar
 #= require details/views/modal_card
 #= require details/views/notifications
+#= require details/views/view_helper
```
phirefly: I think I'm missing why this is needed for this

rememberlenny: @phirefly They are dependencies for the file. Without them, this test fails.

phirefly:
:shipit:

PR #1408, opened 2016-07-19
Use JS to submit the comments

rememberlenny:
# What

Replace the comments `form` with JavaScript functions. 

# Why 

You can't have nested `form` tags.

# How

Store the values on the old `form` tag, and save them as JavaScript object params.


nickbristow:
👍

PR #1409, opened 2016-07-20
removed "Observer added/removed" from notification in redesign

nickbristow:
https://trello.com/c/aq3wTFN1/562-content-change-for-notificaitons


rememberlenny:
Looks gooooood!

PR #1415, opened 2016-07-27
updated home page styles to be less jarring on load

nickbristow:
## WHAT
The home page loaded with a jump, as the header was displayed. This fixes this by adding the custom styles in at the start and setting a minimum height temporarily to keep it from jumping

## WHY 

https://trello.com/c/od2gcZME/438-fix-visual-glitch-when-loading-c2


rememberlenny:
Looks good!

PR #1416, opened 2016-07-27
Use a single form for the entire request details page

rememberlenny:
# What 

Remove the two separate forms for summary bar and request details card. Put everything under a single form.


nickbristow:
👍

PR #1417, opened 2016-07-27
Create validation flow and action bar

rememberlenny:
# What

The front-end now sends the form params to the backend for validation. After the validation check, the front-end does a form `submit`. This allows any errors from validation to be properly handled on the front-end, without accidentally updating the `dirrty` fields or getting inconsistent notifications

screen shot 2016-07-15 at 8 31 08 pm

### DIFF
--- a/app/assets/javascripts/details/shell/action_bar_bridge.js
+++ b/app/assets/javascripts/details/shell/action_bar_bridge.js
@@ -1,4 +1,5 @@
 var ActionBarBridge;
+
 ActionBarBridge = (function() {
 
   function ActionBarBridge(c2, config){
--- a/app/assets/javascripts/details/shell/action_bar_bridge.js
+++ b/app/assets/javascripts/details/shell/action_bar_bridge.js
@@ -1,0 +1,116 @@
+var ActionBarBridge;
+ActionBarBridge = (function() {
+
+  function ActionBarBridge(c2, config){
+    this.c2 = c2;
+    config = config || {};
+    this.config = {
+      actionBar:      '#action-bar-wrapper',
+      detailsSave:    '#request-details-card',
+      notifications:  '#action-bar-status',
+      editMode:       '#mode-parent',
+      modalCard:      '#modal-wrapper',
+      updateView:     '#mode-parent'
+    }
+    this._overrideTestConfig(config);
+    this._blastOff();
+  }
+
+  ActionBarBridge.prototype._blastOff = function(){
+    var config = this.config;
+
+
+    // Data
+    this.detailsSave = this.c2.detailsSave;
+    this.updateView = this.c2.updateView;
+
+    // State
+    this.editMode = this.c2.editMode;
+
+    // Views
+    this.modals = this.c2.modals;
+    this.actionBar = this.c2.actionBar;
+    this.notification = this.c2.notification;
+    this._setupEvents();
+  }
+
+  ActionBarBridge.prototype._overrideTestConfig = function(config){
+    this.config = config;
+  }
+
+  ActionBarBridge.prototype._setupEvents = function(){
+    this._setupActionBar();
+    this._setupSaveModal();
+    this._setupFormSubmitModal();
+  }
+
+  /* Action Bar */
+
+  ActionBarBridge.prototype._setupActionBar = function(){
+    var self = this;
+    this.actionBar.el.on("action-bar-clicked:cancel", function(){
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/details/shell/action_bar_bridge.js
```diff
@@ -0,0 +1,116 @@
+var ActionBarBridge;
```
dogweather: This creates a global `ActionBarBridge` variable, doesn't it? Should this instead simply assign `window.ActionBarBridge` to the result of the IIFE?

app/assets/javascripts/details/shell/action_bar_bridge.js
```diff
@@ -0,0 +1,116 @@
...
+      notifications:  '#action-bar-status',
+      editMode:       '#mode-parent',
+      modalCard:      '#modal-wrapper',
+      updateView:     '#mode-parent'
+    }
+    this._overrideTestConfig(config);
+    this._blastOff();
+  }
+
+  ActionBarBridge.prototype._blastOff = function(){
```
dogweather: The name `_blastOff` isn't too clear about what this does — would you mind renaming it?

app/assets/javascripts/details/shell/action_bar_bridge.js
```diff
@@ -0,0 +1,116 @@
+var ActionBarBridge;
+ActionBarBridge = (function() {
+
+  function ActionBarBridge(c2, config){
```
dogweather: Could you add some docs; what this class's responsibilities are, and why we have it?

app/assets/javascripts/details/shell/action_bar_bridge.js
rememberlenny: @dogweather This is the consistent name for the `init` method. Its being used in 4 other files, so I think its okay. 

@nickbristow @phirefly thoughts?

app/assets/javascripts/details/shell/action_bar_bridge.js
rememberlenny: @dogweather This is the naming convention we are using. Its consistent with the `coffeescript` output for creating constructors.

app/assets/javascripts/details/shell/action_bar_bridge.js
rememberlenny: @dogweather Yep! Will do.

app/assets/javascripts/details/shell/action_bar_bridge.js
dogweather: if it's the `init` method, could we just call it `init` then? Smells of cowboy coding, IMO ;-)

app/assets/javascripts/details/shell/action_bar_bridge.js
dogweather: @rememberlenny think of my point of view: a skilled developer who's unfamiliar with the framework you've created. If this was angular, there'd be docs online that'd tell me the purpose of the class and point to examples how to use it. But here, we don't have that. So for the sake of someone new to the code, what does the ActionBarBridge do?

app/assets/javascripts/details/shell/action_bar_bridge.js
rememberlenny: @dogweather Its a standard from the modular `backbone.js` library `human.js`.

Link: https://github.com/HenrikJoreteg/humanjs

I didnt come up with the syntax. 

Theres a whole book outlining the writing style and naming pattern: http://read.humanjavascript.com/

app/assets/javascripts/details/shell/action_bar_bridge.js
rememberlenny: @dogweather I agree with the window/global overkill. I was doing this intentionally to make these files testable while running the beta testing. The scope can be removed, but it helps with debugging.

These files will eventually be in the asset pipeline. When we make that switch, the plan is to change how the files are invoked.

nickbristow:
@dogweather did you want to weigh in on this again before we merge?

dogweather:
@nickbristow in my opinion, its not ready to be merged because of the issues i raised.  but if it works for you, thats cool.

nickbristow:
Merging, we'll open a new PR to address some of the JS issues effecting the wider project.

dogweather:
👍  @rememberlenny  @nickbristow Sounds good, thanks!

PR #1418, opened 2016-07-28
Show current expense type when editing

nickbristow:
When modifying a current proposal, instead of showing disabled radio buttons
just show the current expense type.
Card: https://trello.com/c/JXHxpEpD/610-only-display-expense-type-of-request-remove-radio-button-and-other-expense-types


rememberlenny:
Looks good

PR #1419, opened 2016-07-28
show description field label

nickbristow:
https://trello.com/c/w2PoepWN/608-move-text-below-description-to-be-under-description-or-in-the-form-field


rememberlenny:
Looks good

PR #1423, opened 2016-07-29
Tweaking selectors and markup to reflect single form changes.

micahtaylor:


rememberlenny:
Looks good! Merging after green @micahtaylor

PR #1429, opened 2016-08-01
update aws-sdk version

nickbristow:
https://trello.com/c/P2edJBtK/542-investigate-why-aws-sdk-gem-version-cannot-be-upgraded-to-2-0

There was an issue where a `uninitialized constant` error would occur. 

This error is due to a namespace change from 1.0 -> 2.0+ versions of the gem.
<2 = AWS namespace
>2 = Aws namespace 

I've made the update and the name space errors seem to be gone already. 

I've deployed to c2-dev with no issues.


rememberlenny:
Merged. Thanks for the explanation.

PR #1432, opened 2016-08-02
Feature/env feature flag

nickbristow:
Based on discussions with @phirefly and @rememberlenny 

Moved checks for features off of roles and into env files.


rememberlenny:
Looks good!

PR #1440, opened 2016-08-08
upcased action bar

nickbristow:
https://trello.com/c/Zs56zRTq/655-buttons-on-action-bar-should-be-all-caps


rememberlenny:
Looks good

PR #1469, opened 2016-08-17
Show New features beta toggle at top of list view

nickbristow:
https://trello.com/c/fA0YU6Zs/700-add-active-beta-notification-to-top-of-list-view
 
 Shows the "new version of this page" at the top
    of the list view.
      Only shows when
        1) The env[BETA_LIST_VIEW] is set to true
        2) The user is a beta user

  Updated the route logic to a single route to toggle
    active beta user status

  Added new method to make clear how to check if beta
    features can be activated.

my_requests_-_c2

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,0 +1,8 @@
+class UsersController < ApplicationController
+  def update
+    if params[:user] && params[:user][:update_beta_active] && current_user
+      current_user.toggle_active_beta
+    end
+    redirect_to(:back)
+  end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -139,6 +139,11 @@ class User < ActiveRecord::Base
     in_beta_program? && active_beta_user? && feature
   end
 
+  def can_see_beta?(feature = nil)
+    feature = feature ? ENV[feature] == "true" : true
+    in_beta_program? && feature
+  end
+
   def in_beta_program?
     role? ROLE_BETA_USER
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -140,7 +140,7 @@ class User < ActiveRecord::Base
   end
 
   def can_see_beta?(beta_feature = nil)
-    feature_enabled = beta_feature == nil || ENV[beta_feature] == "true"
+    feature_enabled = beta_feature.nil? || ENV[beta_feature] == "true"
     in_beta_program? && feature_enabled
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -134,13 +134,14 @@ class User < ActiveRecord::Base
     role? ROLE_ADMIN
   end
 
-  def should_see_beta?(feature = nil)
-    feature = feature ? ENV[feature] == "true" : true
-    in_beta_program? && active_beta_user? && feature
+  def should_see_beta?(feature)
+    can_see_beta?(feature) && active_beta_user?
   end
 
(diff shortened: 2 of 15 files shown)


app/controllers/users_controller.rb
```diff
@@ -0,0 +1,16 @@
+class UsersController < ApplicationController
+  def toggle_active_beta
+    if current_user.active_beta_user?
+      current_user.remove_role(ROLE_BETA_ACTIVE)
+    else
+      current_user.add_role(ROLE_BETA_ACTIVE)
+    end
+  end
```
dogweather: Cool. We could refactor this into `User` in a future PR i think.

app/models/user.rb
```diff
@@ -139,6 +139,11 @@ def should_see_beta?(feature = nil)
     in_beta_program? && active_beta_user? && feature
   end
 
+  def can_see_beta?(feature = nil)
+    feature = feature ? ENV[feature] == "true" : true
```
dogweather: This is hard for me to follow — what type is `feature` expected to be? Like, a String with a feature name? And then line 143 changes its type from a String to a Boolean?

app/views/proposals/details/_top_notification.html.haml
```diff
@@ -11,6 +12,8 @@
           = javascript_include_tag "details/force_ethnio"
           = javascript_include_tag "https://www.ethn.io/55009.js", async: true
           
-          = t("proposals.details.notification.beta.reverted", link: activate_design_proposal_path(proposal) ).html_safe
+          = t("proposals.details.notification.beta.reverted")
+          = link_to "click here", user_path(current_user, user: {update_beta_active: true}), method: :patch
         - else
-          = t("proposals.details.notification.beta.new", link: activate_design_proposal_path(proposal) ).html_safe
+          = t("proposals.details.notification.beta.new")
+          = link_to "click here", user_path(current_user, user: {update_beta_active: true}), method: :patch
```
dogweather: Great - glad to see `patch` working here.

config/routes.rb
```diff
@@ -20,6 +20,8 @@
   match "/auth/:provider/callback" => "auth#oauth_callback", via: [:get]
   get "/auth/failure" => "auth#failure"
   post "/logout" => "auth#logout"
+  resources :users, only: [:update]
```
dogweather: Sweet! +100 for the better security and standards-compliance this change gives us.

app/models/user.rb
```diff
@@ -139,6 +139,11 @@ def should_see_beta?(feature = nil)
     in_beta_program? && active_beta_user? && feature
   end
 
+  def can_see_beta?(beta_feature = nil)
+    feature_enabled = beta_feature.nil? || ENV[beta_feature] == "true"
+    in_beta_program? && feature_enabled
+  end
```
dogweather: Thanks! Much clearer. But should this be **not** nil? I.e.,

```ruby
  !beta_feature.nil? || ENV[beta_feature] == "true"
```

dogweather: I'm wondering, When would this be called with `nil`? What does that mean?

app/models/user.rb
```diff
@@ -147,6 +152,14 @@ def active_beta_user?
     role? ROLE_BETA_ACTIVE
   end
 
+  def toggle_active_beta
+    if active_beta_user?
+      remove_role(ROLE_BETA_ACTIVE)
+    else
+      add_role(ROLE_BETA_ACTIVE)
+    end
+  end
```
dogweather: Cool!

PR #1473, opened 2016-08-24
Added new wiki images

phirefly:


phirefly:
Talked through adding these files with @amberwreed. This replaces https://github.com/18F/C2/pull/1472

PR #1479, opened 2016-08-26
Fixes an ungraceful error that occurs when the user

nickbristow:
dosen't reload the page after being away for awhile

To recreate, login to a page that triggers an update
  new proposal for example,
  go to chrome settings, and clear recent history including
  cache and cookies.
  Return to the page and attempt a patch action,

  You will now be returned to the home page of the app instead
  of crashing to an error message.

### DIFF
--- a/app/controllers/client_data_controller.rb
+++ b/app/controllers/client_data_controller.rb
@@ -2,6 +2,7 @@
 # * model_class
 # * permitted_params
 class ClientDataController < ApplicationController
+  include TokenAuth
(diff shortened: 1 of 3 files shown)


nickbristow:
https://trello.com/c/SwPnhsbT/725-bug-if-a-user-s-session-expires-and-then-tries-to-perform-a-patch-action-submit-proposal-update-proposal-etc-an-error-is-raised

rememberlenny:
Looks good! Nice catch.

PR #1484, opened 2016-08-30
Setup sidebar active states for new request/reports

rememberlenny:
# What

When new request or reports is active, change the sidebar styling on the links

screen shot 2016-08-30 at 2 03 33 pm

screen shot 2016-08-30 at 2 06 02 pm


nickbristow:
👍

PR #1489, opened 2016-08-31
Add tests to new request page

rememberlenny:
# What

Check to make sure the templates work with proper env vars.


nickbristow:
👍

PR #1490, opened 2016-08-31
Load correct new form on unvalidated submissions

rememberlenny:
screen shot 2016-08-31 at 5 54 03 pm

# What

On beta, when form submission is rejected due to validation, load new form.


micahtaylor:
Code looks good and passes visual QA as well.

PR #1492, opened 2016-09-01
Create the condition for non-list_view_conditions to work

rememberlenny:
# What

Add helper to check for details view beta. Keep assets in the details page if list view is deactivated.

### DIFF
--- a/app/views/proposals/show_next.html.haml
+++ b/app/views/proposals/show_next.html.haml
@@ -1,3 +1,6 @@
+- if !list_view_conditions && detail_view_conditions
+  = render partial: 'layouts/scripts'
+
 - content_for :top_notification do
   = render partial: "proposals/details/top_notification", locals: { proposal: @proposal }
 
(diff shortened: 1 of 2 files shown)


app/views/proposals/show_next.html.haml
```diff
@@ -1,3 +1,6 @@
+- if !list_view_conditions && detail_view_conditions
```
phirefly: Wouldn't this break in our dev environments then?

phirefly: ^^ meaning if we have list view and detail view turned on in our dev/staging

rememberlenny: @phirefly Then this kicks in: https://github.com/18F/C2/blob/master/app/views/layouts/application.html.haml#L17-L18

phirefly:
:shipit:

phirefly:
Verified through manual QA

PR #1504, opened 2016-09-09
Listview: Adding transitions to page load

rememberlenny:
# What

- Remove the bug report text in "new request"
- Add transitions to how listview is loaded. Remove the FOUC and add user feedback.

loading

sidebar


micahtaylor:
Looks good!

PR #1505, opened 2016-09-09
Remove duplicate events

nickbristow:
https://trello.com/c/3Ef39EmD/762-step-1-approval-appearing-twice-in-the-timeline

Simple fix to remove duplicate renders from appearing in the timeline


rememberlenny:
Looks Gooooood!

PR #1510, opened 2016-09-13
Fix the pending state on the list view

rememberlenny:
# What

The filter for list view sidebar buttons uses the `status` column to display/hide. The conditions for hiding were cased on `ncr` status attributes. The conditions are now updated for the `gsa18f` status attributes for pending.

screen shot 2016-09-13 at 12 30 30 am

# What

The Pending option for one user was not displaying. This is because the text in the pending area was changed.

# Reference

https://trello.com/c/iJdSnPAP/786-redesign-pending-navigation-indicates-a-collection-of-requests-but-does-not-contain-any-in-the-body


nickbristow:
👍

PR #1516, opened 2016-09-14
Update new request form to be more 508 compliant

nickbristow:
Adds explicit labeling, aria attributes, field sets.


rememberlenny:
Looks good for when tests pass.

PR #1523, opened 2016-09-19
[can merge]Add MVP gsa 18f training form to c2

nickbristow:
Add new form for 18f employees to request training

This is the barebones for a new form under 18f. 

It includes 
* migrations for a new table
* new model
* new form 
* new fields partials for detail page
* new controller
* new policy file
* new decorator
* new routes


rememberlenny:
Looks good! Thanks for handling this!

PR #1528, opened 2016-09-23
Add Pegasys Document Number to gsa18f

rememberlenny:
# What

Add the Pegasys Document Number field in the GSA 18f.

Notes: Not added when request is being created. Only appears after request is already paid.

# Reference 

https://trello.com/c/sjwe7RNt/808-as-an-18f-ops-purchaser-i-can-add-a-pegasys-document-number-field-after-marking-a-request-as-purchased-so-that-i-can-easily-atta


rememberlenny:
Code climate complaint is not valid. Please ignore.

nickbristow:
👍

PR #1529, opened 2016-09-23
Update Events form to match mockups

nickbristow:
Add new fields, 
Renamed fields to match mockups
https://trello.com/c/wz8iIjdg/822-update-18f-event-fields-to-match-mockup


rememberlenny:
Looks good!

PR #1531, opened 2016-09-26
Added radio select to event type

nickbristow:
WHAT: Add radio select to event type
WHY: https://trello.com/c/G4L6WbB4/823-18f-event-type-of-event-radio-buttons

c2

18f_events_-_c2


rememberlenny:
Clarified questions over slack. Looks good!

PR #1538, opened 2016-09-28
changed fields to text area

nickbristow:
Purpose, Justification, and instructions are now text areas instead of inputs


nickbristow:
CSS, is limiting the height of the text areas @rememberlenny

PR #1546, opened 2016-09-29
Event Create spec

nickbristow:
Specs for event form.


db/schema.rb
```diff
@@ -349,6 +349,14 @@
 
   add_index "tags", ["name"], name: "index_tags_on_name", unique: true, using: :btree
 
+  create_table "test_client_requests", force: :cascade do |t|
```
rememberlenny: Do you know what this is coming from? It keeps getting deleted/created @nickbristow

rememberlenny:
@nickbristow Added one comment and merging!

PR #1549, opened 2016-09-30
Create active new requests nav class

rememberlenny:
# What

TeamOps: Highlight left nav when on the Event form

# Reference

https://trello.com/c/g4LNsNSG/839-teamops-highlight-left-nav-when-on-the-event-form

# Visual

screen shot 2016-09-30 at 12 34 50 pm


micahtaylor:
Looks good!

PR #1551, opened 2016-09-30
Remove the NSF field and replace with link

rememberlenny:
# What

NSF shouldn't have been a field. Now its just a link.

screen shot 2016-09-30 at 2 49 56 pm

# Referenece

https://trello.com/c/nyOKspcY/840-nsf-field-is-actually-just-a-link-to-be-exposed-please-link-it-to-another-tab


nickbristow:
👍

PR #1552, opened 2016-09-30
Fixed issue with modifying event fields

nickbristow:
Fixes an issue where modifying an event request causes it to fail.


rememberlenny:
Good to go

PR #1558, opened 2016-10-04
Adds form complete to submit button when required form fields are com…

nickbristow:
…pleted


app/assets/javascripts/new_form.js
```diff
@@ -0,0 +1,13 @@
+$(document).ready(function(){
```
rememberlenny: In future, it would be nice to get this in the format of all the other JS we are running. For now, quick fix. @nickbristow

rememberlenny:
Looks good. Left comment @nickbristow

PR #1565, opened 2016-10-07
Fixed some form fields on evnet, made edit spec

nickbristow:
Adds spec to test event spec
fixes helper
fixes some form fields


rememberlenny:
Looks good!

PR #1569, opened 2016-10-11
Teamops/validate attachment

rememberlenny:
# What

Create a validation notice and prevent event submission without attachment.

# Looks

Briefly flash parent div yellow, then revert to normal view. Not pretty, but does the job!

screen shot 2016-10-11 at 5 11 35 pm


phirefly:
🚢

PR #1573, opened 2016-10-12
Loading circle only shows when table has data

nickbristow:
Table loading circle was showing non stop when table had no data. 

Table loading circle only loads now if the table has data


rememberlenny:
Looks good!

rememberlenny:
Merge when green

PR #1579, opened 2016-10-14
Add css styles for error related labels

rememberlenny:
# What

Change the text for the error labels and add styles.

screen shot 2016-10-14 at 4 08 19 pm


micahtaylor:
👍

PR #1582, opened 2016-10-17
Fix "Save the report"

rememberlenny:
# What

Fix the report "Save the report" link.

screen shot 2016-10-17 at 7 10 30 pm


micahtaylor:
Design 👍

PR #1586, opened 2016-10-19
Proposals pages blank/not applicable

rememberlenny:
# What

Create default fields of `n/a` and blank fields with `-`.

screen shot 2016-10-19 at 5 46 16 pm


micahtaylor:
👍

PR #1588, opened 2016-10-25
Index client_slug on proposals and search against them, instead of client_data_type

rememberlenny:
# What

Update the proposal index by adding the client_slug. The client_slug will now be used to determine the scope of search results on a user. Previously the client_data_type was being used and causing issues when users from GSA18f were searching against multiple client_data_types. 

# Proof

screen shot 2016-10-25 at 5 00 45 pm


nickbristow:
👍

PR #1590, opened 2016-10-25
Fix really long titles on summary card

nickbristow:
Titles were being cut off if they took more than one line

c2


rememberlenny:
Merged!

PR #1596, opened 2016-10-28
Ahoy frontend removal

rememberlenny:
# What

Removing Ahoy fronted and unmounting routes.


rememberlenny:
Canceled due to Ahoy gem update

PR #1598, opened 2016-10-28
Add cloud.gov auth

ertzeid:
This includes a lot of Robb's spike work, as you can see. It's essentially using a new outside gem, which does most of the work for us, but should probably still get some tests. Happy to consider this a WIP review and add tests, or to have this merged and then add some end-to-end tests after.

Also wasn't entirely sure whether to take out the MyUSA stuff just yet.


rememberlenny:
@ertzeid Updated for codeclimate stuff. Merging. Would be good to have any tests you think will help. Regarding pulling the MyUSA Auth, I think we should hold off. @phirefly should weigh in.

PR #1600, opened 2016-10-31
Create the proposal services

rememberlenny:
# What

Reduce the size of the `proposals.rb` model by breaking out content into `ProposalServices`.

Based on reducing CodeClimate complexity.

screen shot 2016-10-31 at 4 58 32 pm


nickbristow:
👍

PR #1601, opened 2016-11-01
Clean up cloud auth

ertzeid:
Delete a couple comments erroneously left in and put cloud.gov vs. myUSA login behind an env. var


rememberlenny:
Looks good!

PR #1610, opened 2016-11-17
Implement User Provided Services

rememberlenny:
Use user provided services to access the environment variables needed on app.


Gemfile.lock
```diff
@@ -318,7 +318,7 @@ GEM
     multi_xml (0.5.5)
     multipart-post (2.0.0)
     nenv (0.3.0)
-    newrelic_rpm (3.14.2.312)
+    newrelic_rpm (3.17.1.326)
```
yozlet: Yay, thank you

app/credentials/concerns/user_provided_services.rb
```diff
@@ -0,0 +1,17 @@
+module UserProvidedService
+  def credentials(service_name)
```
yozlet: Surely the arg here is the attribute, since they're all being pulled from the same service?

app/credentials/concerns/user_provided_services.rb
```diff
@@ -0,0 +1,17 @@
...
+    user_provided_service[service_name]
+  end
+
+  def use_env_var?
+    Rails.env.development? || Rails.env.test?
+  end
+
+  def user_provided_service
+    vcap_services["user-provided"][0]["credentials"]
```
yozlet: This assumes that there's only going to be one UPS, and it will hold all of the credentials. The impression I get is that that's not how UPSes should be used - rather, a single UPS per external service, so we'd have one for Mandrill, one for cloud.gov auth, etc. (This would also make the credentials much easier to edit.)

app/credentials/env_credentials.rb
```diff
@@ -0,0 +1,173 @@
+require "concerns/user_provided_services"
+
+class EnvCredentials
+  extend UserProvidedService
+
+  def self.new_relic_app_name
+    if use_env_var?
+      ENV["NEW_RELIC_APP_NAME"]
+    else
+      credentials("NEW_RELIC_APP_NAME")
```
yozlet: See comment above - this assumes that we've got all credentials in a single UPS.

app/credentials/concerns/user_provided_services.rb
```diff
@@ -0,0 +1,17 @@
+module UserProvidedService
+  def credentials(service_name)
+    user_provided_service[service_name]
```
yozlet: It'd be hugely helpful if an exception was thrown for missing required values. AFAIK we didn't have that before, for env vars, but it would make certain setup issues much more discoverable. Perhaps we could do that with two different `credentials` methods, one for required values and one for optional values?

PR #1611, opened 2016-11-17
Fix native browser prompt for unfinished form content

rememberlenny:
# What

Comments and attachments are triggering unfinished form prompt. Fix this by trigging a form submission on the parent form when comment and attachment is complete. This will update the native browser behavior around unfinished form without changing any content on the backend.

Note: No emails or backend activity is triggered when form submit takes place

# Reference

https://trello.com/c/aONpL13i

### DIFF
--- a/app/assets/javascripts/details/views/attachment_card.js
+++ b/app/assets/javascripts/details/views/attachment_card.js
@@ -154,6 +154,8 @@ AttachmentCardController = (function(){
       },
       contentType: false,
       processData: false
+    }).done(function(){
+      $('body').find('form.request-details-form').submit();
     });
   }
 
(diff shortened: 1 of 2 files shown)


rememberlenny:
@nickbristow 

Do you think you can take a look at this for the Comments module? When a comment is submitted and the user refreshes/changes pages, there is a native browser functionality that warns the user of a unfinished form. This is coming up because the form needs to be "submitted". I resolved the issue for attachments, but the view_helper stuff is confusing me. I think you are a bit more familiar with it.

All that needs to happen is: `$('body').find('form.request-details-form').submit();`

phirefly:
@rememberlenny Should we keep this PR open for @nickbristow to add to? I'd be in favor of both changes going in the same PR.

nickbristow:
@rememberlenny @phirefly oh, yes, im working off this branch now.

PR #1619, opened 2016-11-22
Fix issue with NCR not_to_exceed

rememberlenny:
# What

The `not_to_exceed` value was not being updated when user made changes on frontend. Also when values were being reported at `false` during update, they were returning as blank. 

This fixes both of these issues.

screen shot 2016-11-22 at 6 12 29 pm

screen shot 2016-11-22 at 6 12 37 pm


nickbristow:
👍

PR #1640, opened 2017-01-30
Adjusted CSV download on proosal admin to show all fields

nickbristow:
Updates CSV download link on proposals admin page.

### DIFF
--- a/app/admin/proposal.rb
+++ b/app/admin/proposal.rb
@@ -71,4 +71,17 @@ ActiveAdmin.register Proposal do
     flash[:alert] = "Completed!"
     redirect_to admin_proposal_path(resource)
   end
+
+  csv do
+    proposal_attributes = %w(id status created_at updated_at client_data_type public_id visit_id)
+    proposal_attributes.each do |proposal_attr|
+      column(proposal_attr.to_sym) { |proposal| proposal.attributes[proposal_attr] }
+    end
+    column(:requester) { |proposal| proposal.requester.display_name }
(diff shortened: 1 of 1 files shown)


app/admin/proposal.rb
```diff
@@ -71,4 +71,17 @@
...
     redirect_to admin_proposal_path(resource)
   end
+
+  csv do
+    proposal_attributes = %w(id status created_at updated_at client_data_type public_id visit_id)
+    proposal_attributes.each do |proposal_attr|
+      column(proposal_attr.to_sym) { |proposal| proposal.attributes[proposal_attr] }
+    end
+    column(:requester) { |proposal| proposal.requester.display_name }
+    client_data_attributes = %w(expense_type vendor not_to_exceed building_number emergency rwa_number work_order_code project_title description direct_pay cl_number function_code soc_code ncr_organization_id)
```
yozlet: Is this going to break things for 18F proposals, or will these fields just be ignored?

app/admin/proposal.rb
```diff
@@ -71,4 +71,17 @@
...
+    client_data_attributes.each do |data_attr|
+      column(data_attr.to_sym) { |proposal| proposal.client_data.attributes[data_attr] }
+    end
+    column(:approving_offical_name) { |proposal| User.find(proposal.client_data.approving_official_id).display_name }
```
yozlet: We probably want another column for official's email address, just in case they need to disambiguate.

PR #1644, opened 2017-03-23
WIP: Dockerization

adelevie:


yozlet:
@stvnrlly and I started playing with this in order to test some `paperclip` fixes. It's mostly great (thank you @adelevie, I owe you coffee & seltzer) but we hit these issues:

 * `bundle install` seems to install gems somewhere else to where the `bundle exec` process is expecting them. We kept hitting `Could not find rake-11.3.0 in any of the sources` when trying to run `rake`. Doing `docker-compose run web bundle install` helped, though we still hit this problem elsewhere.
* Whatever the shell is in the `web` container, it can't seem to run bash scripts: `./script/bootstrap: 9: ./script/bootstrap: [[: not found`
* The `bundler` version in the `ruby:2.3.5` image is 1.16.0, whereas we've used 1.16.1 to generate the latest `Gemfile.lock`. This causes warnings but doesn't seem to break anything.
* After all the above, we were able to successfully run C2! And _nearly_ log in with the fake auth! But then Rails complains with a CSRF issue, and that's where we ran out of time.

Of the above issues, the first is the one that needs someone with both Ruby & Docker experience to poke at. Anyone?

stvnrlly:
Shout-out to @toolness and @yozlet, who apparently had already found the last piece of the puzzle here [back in 2017](https://github.com/18F/cg-fake-uaa/issues/18).

PR #1651, opened 2017-09-26
Updates to setup docs

ctro:
These helped me get running locally (without using docker).


ctro:
Hi @yozlet here are a couple updates to C2 docs that I found useful.

yozlet:
Thank you @ctro! Sorry for missing this.

Issue #1655, opened 2017-10-26
Add additional logging for ATO


ctro:
To qualify for an ATO we need to add extra logging around many actions.

TOLOG:
- [x] Log successful and unsuccessful logon events
- [x] log account management events : update username, etc
- [x] object access : via SQL logs
- [x] privilege functions : should fall under "all admin actions"
- [x] system events : aka "show me a deployment"
- [x] all administrator activity : "elevate, demote users and show logs"
- [x] authentication checks
- [x] authorization checks
- [x] data deletions
- [x] data access
- [x] data changes
- [x] permission changes

ctro:
@yozlet At some point I'd like to sync up with you about this task to get a feel for where all of these actions take place and must be logged.

PR #1656, opened 2017-11-13
Additional logging for ATO

ctro:
Ref https://github.com/18F/C2/issues/1655

Includes
- Some extra logging calls for things like logins
- Configuration to send database logs to Kibana
- Examples of Kibana queries to surface various logging requirements
- Update to ruby 2.3.5 so we can still use default buildpacks

Note: The documented queries all work with `c2-dev` right now. You can generate events here: https://c2-dev.fr.cloud.gov


manifest.yml
```diff
@@ -33,7 +33,7 @@ applications:
   services:
     - c2-dev-elasticsearch
     - c2-dev-db
-    - c2-dev-ups-s3
+    - c2-dev-s3
```
ctro: I changed this to match what was already existing in the cloud.gov/development env.

PR #1659, opened 2017-12-20
Specify allowed content types for attachments

stvnrlly:
This adds content type checking for uploaded attachments using Paperclip's [security validations](https://github.com/thoughtbot/paperclip#security-validations).

The list of file types is based on the types of files that past users have uploaded, so there shouldn't be any noticeable difference for users.


stvnrlly:
On further thought, I think more specificity would be good.

yozlet:
(though it'd be nice to keep CodeClimate happy)

PR #1661, opened 2017-12-26
Limit attachment size to 30 MB

stvnrlly:
30 MB is around 5 MB larger than the largest two uploaded in the past, but I'd be happy to reduce it for the future.


ethanheppner:
Yep, I don't see why we would need to accommodate attachments in C2 larger than 30 MB.

